PR #4, opened 2016-10-18
Link component: render as <a> if href is present

FreedomBen:


junyper:
@FreedomBen This was intentional because if you're passing '#' as the href, semantically you should probably be using a <button> element. The thinking was it should only render as an <a> if there is a real href. What is the scenario where this isn't working for you?

FreedomBen:
oh shoot, I thought I deleted this.  It was a bug that was fixed in the 0.17 release.  All good!

PR #72, opened 2019-08-05
Spinner title attribute deprecated, update example

buckett:
renderTitle it the newer attribute and should be used instead as the title attribute is being removed in 7.0.0.


buckett:
Just a documentation update.

MatheusRV:
Commit name was divergent name from project default.

matyasf:
This issue is fixed in the next release (8.0), thanks for reporting it!

PR #93, opened 2020-05-04
Generate cache busting files by default.

buckett:
By default the files generated by the webpack configuration will be cached by browsers and when a new version of the application is deployed the changes won’t be detected. This adds a cache busting content hash to the filename to avoid this.

### DIFF
--- a/packages/template-app/template/webpack.config.js.ejs
+++ b/packages/template-app/template/webpack.config.js.ejs
@@ -33,7 +33,7 @@ module.exports = {
   entry: './src/index.js',
   output: {
     path: outputPath,
-    filename: '[name].js'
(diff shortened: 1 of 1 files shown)


buckett:
I can understand not having this in the sample project to keep it simple which makes it easier for people to get onboard.

I just bumped into this as in deploying a couple of projects changes weren't getting picked up.

matyasf:
Thanks for the PR and sorry for the late response. We have integrated this fix into the next major release (8.0) which will go out in approximately a month.

buckett:
Thanks, although I've ended up switching to using https://create-react-app.dev/ and just add the Instructure dependencies as it's out of the box experience fit better with what we need.

PR #95, opened 2020-05-16
Fixes for createSuper being undefined.

### LINKED ISSUE #94: Module not found: Error: Can't resolve '@babel/runtime/helpers/esm/createSuper'
When upgrading from 6.24 to 6.26 I started getting an error of:

    Module not found: Error: Can't resolve '@babel/runtime/helpers/esm/createSuper' 

which was referenced from:
```
[webpack]  @ ./node_modules/@instructure/ui-testable/es/testable.js 6:0-66 53:17-29
[webpack]  @ ./node_modules/@instructure/ui-testable/es/index.js
[webpack]  @ ./node_modules/@instructure/ui-spinner/es/Spinner/index.js
[webpack]  @ ./node_modules/@instructure/ui-spinner/es/index.js
```
When looking at the babel dependencies defined most of them are at 7.9.6, but there are a few 7.8.4:

[babel-runtime-ls.txt](https://github.com/instructure/instructure-ui/files/4637975/babel-runtime-ls.txt)

`ui-testable` isn't listed as one of the packages that requires `@babel/runtime` so I think it's getting the older version which doesn't have `createSuper` available.

buckett:
If the project ui-testable is used in isn’t pulling in a newish version of table then you can end up getting an error of:

Module not found: Error: Can't resolve '@babel/runtime/helpers/esm/createSuper' 

This makes this module explicitly depend on a newer version of @babel/runtime so that this can’t happen.

Fixes #94

### DIFF
--- a/packages/ui-testable/package.json
+++ b/packages/ui-testable/package.json
@@ -26,6 +26,7 @@
     "prop-types": "^15.6.2"
   },
   "dependencies": {
+    "@babel/runtime": "^7.9.2",
     "@instructure/ui-decorator": "^6.26.0"
   },
   "peerDependencies": {


serikjensen:
@buckett thanks for submitting this! we are at the tail end of our version 7.0 release work so it may be a week or so before this actually gets released. We'll keep you posted. thanks again!

PR #257, opened 2020-11-25
WIP(emotion,ui-docs-client): bidirectional polyfill POC

ke1k0:
proof-of-concept polyfill instead of postcss-bidirection

### DIFF
--- a/packages/emotion/src/useStyle/bidirectionalPolyfill.js
+++ b/packages/emotion/src/useStyle/bidirectionalPolyfill.js
@@ -1,0 +1,163 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015 - present Instructure, Inc.
+ *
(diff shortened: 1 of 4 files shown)


packages/emotion/src/useStyle/bidirectionalPolyfill.js
```diff
@@ -0,0 +1,169 @@
...
+  'borderBottomInlineEndRadius',
+  'marginInlineStart',
+  'marginInlineEnd',
+  'offsetInlineStart',
+  'offsetInlineEnd',
+  'insetInlineStart',
+  'insetInlineEnd'
+]
+
+function log(...args) {
```
matyasf, line 56: It might be better to use our own `console` package to log stuff

ke1k0: fixed

HerrTopi (approved):

matyasf (approved):

Issue #158, opened 2020-11-18
Add visual change flag to the commit script


HerrTopi:
### Context
Commit script

### Summary
Add visual change flag to the commit script

### Justification
Users need to know if a commit introduces visual changes

### Additional Details
Indicate this in the docs as well

matyasf:
add a "is visual change?" option to the commit script

PR #258, opened 2020-11-25
Visual change commit

matyasf:
Completes #158

### DIFF
--- a/packages/cz-lerna-changelog/lib/index.js
+++ b/packages/cz-lerna-changelog/lib/index.js
@@ -63,23 +63,28 @@ function makePrompter() {
 
     cz.prompt(autocompleteQuestions(questions))
       .then((answers) => {
-        const { body, footer, ...rest } = answers
+        const { body, testplan, visualChange, footer, ...rest } = answers
 
-        const testplan = answers.testplan
-          ? `\nTEST PLAN:\n${answers.testplan}\n\n`
-          : ''
+        const testplanTxt = testplan ? `\nTEST PLAN:\n${testplan}\n\n` : ''
+
         const issues = footer ? `\n\nCloses: ${footer}\n\n` : ''
+
+        const visualChangeTxt = visualChange
+          ? `\n\nVisual change: ${visualChange}\n\n`
+          : ''
 
         let scope = '*'
 
         if (Array.isArray(answers.scope)) {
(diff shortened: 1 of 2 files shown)


Brailor (approved):

packages/cz-lerna-changelog/lib/index.js
```diff
@@ -63,23 +63,28 @@ function makePrompter() {
...
+
+        const testplanTxt = testplan ? `\nTEST PLAN:\n${testplan}\n\n` : ''
 
-        const testplan = answers.testplan
-          ? `\nTEST PLAN:\n${answers.testplan}\n\n`
-          : ''
         const issues = footer ? `\n\nCloses: ${footer}\n\n` : ''
 
+        const visualChangeTxt = visualChange
+          ? `\n\nVisual change: ${visualChange}\n\n`
```
HerrTopi, line 73: So far every, every major topic (e.g.: TEST PLAN) is with all capital letters, shouldn't we keep this pattern?
```suggestion
          ? `\n\nVISUAL CHANGE: ${visualChange}\n\n`
```

packages/cz-lerna-changelog/lib/index.js
```diff
@@ -63,23 +63,28 @@ function makePrompter() {
...
         }
-
+        // To have part of a commit body appear in the changelog it needs to be after the "BREAKING CHANGE:" text.
+        // see https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-angular
         const message = buildCommit(
           {
             ...rest,
-            body: issues + body + testplan,
+            body: issues + body + testplanTxt,
+            breaking: answers.breaking + visualChangeTxt,
```
HerrTopi, line 87: Since you got every part of `answers` out in line 66, is there any reason why not take out `breaking` as well?
```suggestion
            breaking: breaking + visualChangeTxt,
```

packages/cz-lerna-changelog/lib/index.js
matyasf: good idea, I'll change it

packages/cz-lerna-changelog/lib/index.js
matyasf: I've fixed these, but since they are on another branch now I'll open a new PR

PR #262, opened 2020-11-27
feat(ui-docs-plugin): display the faulty YAML that caused a parsing error

matyasf:
when running the documentation app with `yarn dev` and a source code file had a malformed YAML it caused a generic build error message. This commit prints the wrong YAML string to the console when it causes the build to fail.


ke1k0 (approved):

matyasf:
cherry picked to `next` branch

PR #264, opened 2020-11-27
Update the project to be able to use Node 10+ instead of Node 10.x

matyasf:
Tested with Node 10 and Node 15. `"type": "module"` throws an error when building with Node 12+


packages/ui-docs-plugin/package.json
```diff
@@ -25,7 +25,7 @@
   "dependencies": {
     "@instructure/config-loader": "^7.3.0",
     "@instructure/ui-docs-client": "^7.3.0",
-    "favicons-webpack-plugin": "^1.0.2",
+    "favicons-webpack-plugin": "^4.2.0",
```
matyasf, line 28: This dependency needed to be updated, it had a bug in Node 12+

Brailor:
Doesn't it make sense to merge this into master as well?

matyasf:
> Doesn't it make sense to merge this into master as well?

good idea, but I'd first let us test it on the `V8` branch for a bit, there might be some hidden issues.

HerrTopi (approved):

matyasf:
i cherry picked this to the `next` branch

PR #271, opened 2020-12-03
WIP(emotion,ui-avatar,ui-view): applying theme js changes from v8 to …

ke1k0:
…next branch

Manually applied changes that we already made in v8 branch to the cherrypicked next branch.

### DIFF
--- a/packages/emotion/README.md
+++ b/packages/emotion/README.md
@@ -1,78 +1,87 @@
 ---
-
 category: packages
-
 ---
-
-
 
 ## emotion
 
 The [Emotion design library's](https://emotion.sh/) implementation in Instructure UI.
 
 ### Elements
- ```EmotionThemeProvider``` is a React component, which wraps Emotion's own ```ThemeProvider```.
+
+`EmotionThemeProvider` is a React component, which wraps Emotion's own `ThemeProvider`.
+
 #### Usage
+
 It can be used in two ways.
 Once on the top level, providing the theme for the whole application.
-It accepts a ```theme``` prop, which should be an Instructure UI theme.
-```jsx
-import App from  './App'
-import { EmotionThemeProvider } from  "@instructure/emotion"
-import { canvas, canvasHighContrast } from  "@instructure/ui-themes"
+It accepts a `theme` prop, which should be an Instructure UI theme.
 
-const  RenderApp = () => {
-return (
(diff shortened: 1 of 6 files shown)


packages/emotion/README.md
```diff
@@ -92,180 +101,242 @@ return (
...
+  }
+
+  // return with the css you'd like to apply to the component
+  return {
+    root: {
+      label: 'root',
+      display: 'block',
+      background: componentTheme.background,
+      boxSizing: 'border-box',
+      ...colorStyles[color]
```
matyasf, line 231: Here `color` is undefined

packages/emotion/README.md
```diff
@@ -92,180 +101,242 @@ return (
...
+  // return with the final theme object of the component
+  // with the added overrides
+  return {
+    ...componentVariables,
+    ...themeSpecificStyle[themeName],
+    ...themeOverride
+  }
+}
+export default generateComponentTheme
+```
```
matyasf, line 189: IT would be nice to see here the original `theme.js` too

ke1k0: This "temporary migration guide" section is only made for the 4 of us, since we will be doing all the migrating. :) But I'll add the original theme.js, if you think it will make it clearer for us.

packages/emotion/README.md
ke1k0: fixed! thanks

packages/ui-avatar/package.json
```diff
@@ -28,6 +28,7 @@
     "@instructure/ui-react-utils": "^7.3.1",
     "@instructure/ui-testable": "^7.3.1",
     "@instructure/ui-themeable": "^7.3.1",
+    "@instructure/ui-themes": "^7.3.1",
     "@instructure/ui-view": "^7.3.1",
     "@instructure/emotion": "^7.3.1",
     "classnames": "^2",
```
HerrTopi, line 34: Remove unused dependency please

HerrTopi, line 34:
```suggestion
```

ke1k0: removed, thanks

ke1k0: done, thanks

matyasf (approved):

HerrTopi (approved):

PR #274, opened 2020-12-08
refactor(emotion): applying postcss-bidirection functionality

ke1k0:
Closes: INSTUI-2735

Since postcss is not working with Emotion, we needed to replace the functionality of
postcss-bidirection plugin with a local polyfill.
TEST PLAN:
Making sure ltr and rtl styles generate perfectly.


HerrTopi (approved):

packages/emotion/src/styleUtils/index.js
```diff
@@ -107,26 +115,38 @@ const withStyle = decorator((ComposedComponent, generateStyle) => {
...
   hoistNonReactStatics(WithStyle, ComposedComponent)
   WithStyle.displayName = `WithStyle(${ComposedComponent.displayName})`
 
   return WithStyle
 })
+
+/**
+ * Applies necessary polyfills on the style object
+ * @param {object} styleObject
+ * @returns {object} the modified style object
```
matyasf, line 145: This returns sometimes `undefined` (bidirectional.js line 152), if this is expected it should be mentioned here.

packages/emotion/src/styleUtils/index.js
```diff
@@ -82,7 +83,14 @@ const useTheme = () => {
...
 
-  return makeStyles(theme, componentName, generateStyle, props)(...rest)
+  const styles = calculateStyles(
+    theme,
+    componentName,
+    generateStyle,
+    props
+  )(...rest)
+
+  return polyfillStyles(styles)
```
matyasf, line 93: This returns sometimes undefined (bidirectional.js line 152), if this is expected it should be mentioned in the description.

packages/emotion/src/styleUtils/index.js
ke1k0:
```js
function bidirectionalPolyfill(styles) {
  const stylesCopy = {}
  //..... rest of code, forEach ....
  return stylesCopy
```
It is an empty object by default, so it always gives back an object. In the forEach loop it just skips properties with no value.

packages/emotion/src/styleUtils/index.js
ke1k0: answered above

packages/emotion/src/styleUtils/index.js
matyasf: oh yeah, I missed that part, it looks good :)

matyasf (approved):

ke1k0:
Done in another PR: https://github.com/instructure/instructure-ui/pull/318

PR #279, opened 2020-12-09
INSTUI-2741 Emotion migration - ContextView

Brailor:


packages/ui-view/src/ContextView/__tests__/ContextView.test.js
```diff
@@ -42,74 +40,4 @@ describe('<ContextView />', async () => {
...
-  testArrowPlacement('start', 'end')
-  testArrowPlacement('start top', 'end top')
-  testArrowPlacement('start bottom', 'end bottom')
-  testArrowPlacement('bottom', 'top')
-  testArrowPlacement('bottom start', 'top start')
-  testArrowPlacement('bottom end', 'top end')
-  testArrowPlacement('end', 'start')
-  testArrowPlacement('end top', 'start top')
-  testArrowPlacement('end bottom', 'start bottom')
 })
```
Brailor, line 43: Note: these test cases were deleted because with the current emotion styling solution we simply can't query for css classes like we could do with the previous solution.

HerrTopi (approved):
Seems to be alright. Nice work with that nasty css

packages/ui-view/src/ContextView/index.js
```diff
@@ -21,27 +21,26 @@
...
 
 /**
 ---
 category: components
 ---
 **/
 
-@themeable(theme, styles)
+@withStyle(generateStyles)
+// @themeable(theme, styles)
```
ke1k0, line 43: This could be deleted :)

packages/ui-view/src/ContextView/index.js
```diff
@@ -151,23 +161,18 @@ class ContextView extends Component {
...
-      [styles.contextViewRoot]: true,
-      [styles[`placement--${placement.replace(' ', '-')}`]]: true
-    }
-
     return (
       <View
         {...omitProps(this.props, ContextView.propTypes)}
         style={style}
-        className={classnames(classes)}
+        css={styles.root}
```
ke1k0, line 175: According to our agreement, classes should be named after the component, like "styles.contextView" in this case.

packages/ui-view/src/ContextView/styles.js
```diff
@@ -0,0 +1,283 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+import { mirrorPlacement } from '@instructure/ui-position'
+
+import generateComponentTheme from './theme'
+
+const getPlacementStyle = (placement, theme) => {
```
ke1k0, line 29: It is just a preference/suggestion, but how about writing it as a switch? I find it more readable, but this works fine too :)
```js
switch (placement) {
  case 'end-center':
  case 'end-top':
  case 'end-bottom':
  case 'center-end':
  case 'end':
    return { paddingInlineStart: theme?.arrowSize, paddingInlineEnd: '0' }
  
  case 'start-center':
  ...etc
}
```

Brailor: Yeah, I started of with a switch case, but it started to become too cumbersome so I instead started doing this way which is still hard to read, but IMO better than the huge switch case. (an argument can be made for the switch to be faster, because it doesn't have to do the `includes` part)

packages/ui-view/src/ContextView/index.js
Brailor: Done

packages/ui-view/src/ContextView/index.js
Brailor: Fixed

ke1k0 (approved):

PR #283, opened 2020-12-10
Avatar img header truncatetext

HerrTopi:


packages/ui-heading/src/Heading/styles.js
```diff
@@ -0,0 +1,126 @@
...
+        textAlign: 'start',
+        boxShadow: 'none',
+        display: 'block',
+        '&:focus': { outline: 'none' }
+      },
+      ...levelStyles[level],
+      ...colorStyles[color],
+      ...borderStyles[border]
+    },
+    forwardedStyleProps: {
```
Brailor, line 121: I don't think we need to do this since the View component will override the margin with the inline applied styles.

packages/ui-heading/src/Heading/index.js
```diff
@@ -130,16 +126,10 @@ class Heading extends Component {
...
-          [styles[`level--${level}`]]: true,
-          [styles[`color--${color}`]]: color,
-          [styles[`border--${border}`]]: border !== 'none',
-          [styles.ellipsis]: ellipsis
-        })}
+        css={this.props.styles.heading}
         as={ElementType}
-        margin={margin}
         elementRef={elementRef}
+        margin={this.props.styles?.forwardedStyleProps?.margin}
```
Brailor, line 132: I don't think we need to do this since the View component will override the margin with the inline applied styles.

packages/ui-avatar/src/Avatar/index.js
```diff
@@ -22,6 +21,7 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
+/* eslint-disable react/prop-types */
```
Brailor, line 24: This is not needed anymore.

packages/ui-avatar/src/Avatar/index.js
```diff
@@ -30,6 +30,7 @@ import { View } from '@instructure/ui-view'
 import { ThemeablePropTypes } from '@instructure/ui-themeable'
 import { passthroughProps } from '@instructure/ui-react-utils'
 import { testable } from '@instructure/ui-testable'
+// eslint-disable-next-line import/named
```
Brailor, line 33: Why is this needed?

packages/ui-avatar/src/Avatar/styles.js
```diff
@@ -98,15 +98,15 @@ const generateStyle = (
       ...variantStyles[shape]
     },
     initials: {
-      label: 'initials',
+      label: 'avatar_initials',
```
ke1k0, line 101: The migration guide recommends double underscores: 'avatar__initials' to indicate BEM-like parent-child connection.

packages/ui-truncate-text/src/TruncateText/styles.js
```diff
@@ -22,19 +22,33 @@
...
-  display: block;
+  return {
+    truncateText: {
+      label: 'truncateText',
+      display: 'block',
+      overflow: 'hidden',
+      wordWrap: 'break-word',
+      fontFamily: componentTheme.fontFamily
+    },
+    auto: { label: 'auto_truncateText', height: '100%' },
```
ke1k0, line 45: It should be "truncateText__auto" and "truncateText__spacer", starting with the name of the component.

ke1k0 (approved):

packages/ui-avatar/src/Avatar/styles.js
HerrTopi: Yapp, ty. Starting to get used to all the new rules

packages/ui-truncate-text/src/TruncateText/styles.js
HerrTopi: Fixed

packages/ui-truncate-text/src/TruncateText/index.js
```diff
@@ -148,7 +152,10 @@ class TruncateText extends Component {
   }
 
   componentDidUpdate(prevProps, prevState) {
-    const { children, onUpdate } = this.props
+    // eslint-disable-next-line react/prop-types
```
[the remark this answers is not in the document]
HerrTopi: removed

packages/ui-img/src/Img/styles.js
```diff
@@ -0,0 +1,157 @@
...
+  const containerBackgroundStyle = {
+    backgroundRepeat: 'no-repeat',
+    backgroundPosition: 'center',
+    backgroundSize: constrain
+  }
+
+  const viewBackgroundImageStyle = hasBackground ? `url(${src})` : {}
+  return {
+    overlay: {
+      label: 'overlay',
```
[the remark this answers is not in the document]
HerrTopi: Fixed

packages/ui-heading/src/Heading/styles.js
HerrTopi: Removed

packages/ui-heading/src/Heading/index.js
HerrTopi: Removed

packages/ui-avatar/src/Avatar/index.js
HerrTopi: Removed

packages/ui-avatar/src/Avatar/index.js
HerrTopi: Removed

packages/ui-heading/src/Heading/theme.js
```diff
@@ -22,58 +22,80 @@
...
+  } = theme
 
-    h1FontSize: typography.fontSizeXXLarge,
-    h1FontWeight: typography.fontWeightLight,
-    h1FontFamily: typography.fontFamily,
+  const themeSpecificStyle = {
+    canvas: {
+      primaryColor: brandVariables
+        ? brandVariables['ic-brand-font-color-dark']
+        : ''
```
ke1k0, line 45: Just another thing caught my eye: I remember you made a change on v8 branch about brandVariables, but I don't think it got cherrypicked here, so I'm not sure this would work. In the other components we use `theme['ic-brand-font-color-dark']`. (Btw, I would probably leave it like on the theme object: not sure if brandVariables would break anything for our users.)

HerrTopi: Good catch, thanks a lot

PR #287, opened 2020-12-12
Link emotion migration

matyasf:


packages/ui-link/src/Link/style.js
```diff
@@ -0,0 +1,157 @@
...
+          color: componentTheme.colorInverse
+        },
+        '&:link:focus, &:visited:focus': {
+          outlineColor: componentTheme.focusInverseIconOutlineColor
+        },
+        '&:link:hover, &:link:focus, &:link:active, &:visited:hover, &:visited:focus, &:visited:active': {
+          color: componentTheme.colorInverse
+        }
+      }
+    : {}
```
matyasf, line 81: I'm not sure that this is needed..

HerrTopi: How else would you apply these style rules?

packages/ui-link/src/Link/style.js
```diff
@@ -0,0 +1,157 @@
...
+
+  // If TruncateText is used in Link with icon, align the icon and the text vertically
+  const truncateStyle =
+    props.renderIcon && state.containsTruncateText && state.hasVisibleChildren
+      ? { alignItems: 'center' }
+      : {}
+
+  const linkStyles = {
+    label: 'link',
+    textDecoration: textDecorationStyle,
```
ke1k0, line 106: I don't think it is needed here: in the css file I can see text-decoration only in case of button or a tags, not by default.

packages/ui-link/src/Link/style.js
```diff
@@ -0,0 +1,157 @@
...
+    : componentTheme.textDecorationOutsideText
+  // If Link is a button or link, it should look clickable
+  const buttonOrLinkStyle =
+    state.elementName === 'button' || state.elementName === 'a'
+      ? {
+          cursor: 'pointer',
+          color: inverseStyle
+            ? componentTheme.colorInverse
+            : componentTheme.color,
+          textDecoration: props.isWithinText
```
ke1k0, line 50: You could use "textDecorationStyle" here.

packages/ui-link/src/Link/style.js
ke1k0: My concern is, that it doesn't exactly translates to the rules in the CSS file, but I suppose it needs VRT and manual testing to see for sure. In the css file, many rules might be there to strengthen the selectors.

packages/ui-link/src/Link/index.js
```diff
@@ -148,11 +153,28 @@ class Link extends Component {
...
+
+  componentDidUpdate(prevProps, prevState, snapshot) {
+    this.props.makeStyles(this.makeStyleProps())
+  }
+
+  makeStyleProps = () => {
+    return {
+      containsTruncateText: this.containsTruncateText,
+      hasVisibleChildren: this.hasVisibleChildren,
+      elementName: this.element
```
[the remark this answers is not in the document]
matyasf: elementType sounds good :)

packages/ui-link/src/Link/style.js
matyasf: good catch :)

ke1k0 (approved):

PR #291, opened 2020-12-14
WIP(ui-navigation): migrate AppNav and AppNav.Item to emotion styling ( INSTUI-2763 )

HerrTopi:


packages/ui-navigation/package.json
```diff
@@ -46,6 +46,8 @@
     "@instructure/ui-tooltip": "^7.3.1",
     "@instructure/ui-utils": "^7.3.1",
     "@instructure/ui-view": "^7.3.1",
+    "@instructure/emotion": "^7.3.1",
+    "@instructure/ui-themes": "^7.3.1",
     "classnames": "^2",
```
ke1k0, line 51: Unnecessary classnames

packages/ui-navigation/src/AppNav/Item/styles.js
```diff
@@ -0,0 +1,85 @@
...
+      background: 'transparent',
+      border: 'none',
+      outline: 'none',
+      lineHeight: componentTheme.height,
+      padding: `0 ${componentTheme.padding}`,
+      alignItems: 'flex-start',
+      '&:hover': {
+        textDecoration: 'underline',
+        textDecorationColor: componentTheme.textColor
+      },
```
ke1k0, line 68: This woks fine, so this is only an alternative solution: the original "&:hover .label" css rule can be translated with this syntax too:
```
'&:hover [class$=-item__label]': {
   textDecoration: 'underline'
}
```
We could discuss later which version works best for us. :) Probably better to avoid the selectors when we can, like you did here.

packages/ui-navigation/src/AppNav/Item/theme.js
```diff
@@ -22,8 +22,16 @@
...
+/**
+ * Generates the theme object for the component from the theme and provided additional information
+ * @param  {Object} theme The actual theme object.
+ * @param  {Object} themeOverride User provided overrides of the default theme mapping.
+ * @return {Object} The final theme object with the overrides and component variables
+ */
+const generateComponentTheme = (theme, themeOverride = {}) => {
+  const { colors, spacing, typography } = theme
+
+  const componentVariables = {
```
ke1k0, line 34: Missing optional syntax: eg. typography?.fontFamily

packages/ui-navigation/src/AppNav/index.js
```diff
@@ -186,11 +195,11 @@ class AppNav extends Component {
...
-        className={classnames({
-          [styles.listItem]: true,
-          [styles['listItem--isMenuTrigger']]:
-            isMenuTrigger && this.props.visibleItemsCount > 0
-        })}
+        css={
+          isMenuTrigger
+            ? this.props.styles.listItemWithMenuTrigger
+            : this.props.styles.listItem
+        }
```
ke1k0, line 202: I like what you did here :D

packages/ui-navigation/src/AppNav/index.js
```diff
@@ -251,35 +260,30 @@ class AppNav extends Component {
...
       <View
         {...passthroughProps}
         as="nav"
-        className={classnames({
-          [styles.root]: true,
-          [styles['root--hasRenderedContent']]: hasRenderedContent
-        })}
+        css={[
+          this.props.styles.appNav,
+          hasRenderedContent ? this.props.styles.alignCenter : ''
```
ke1k0, line 269: Do you think the hasRenderedContent (and renderBeforeItems and renderAfterItems) could be used as a getter, so we can pass it to the makeStyles above?

packages/ui-navigation/src/AppNav/theme.js
```diff
@@ -22,8 +22,18 @@
...
+ * @param  {Object} theme The actual theme object.
+ * @param  {Object} themeOverride User provided overrides of the default theme mapping.
+ * @return {Object} The final theme object with the overrides and component variables
+ */
+const generateComponentTheme = (theme, themeOverride = {}) => {
+  const { colors, borders, spacing, typography, key: themeName } = theme
+
+  const themeSpecificStyles = {}
+
+  const componentVariables = {
```
ke1k0, line 36: Missing optional syntax: eg. typography?.fontFamily

packages/ui-navigation/package.json
HerrTopi: It is needed until the navigation component is not migrated

packages/ui-navigation/src/AppNav/Item/styles.js
HerrTopi: As we discussed on daily, we will keep my solution whenever possible. Thank for pointing this out :)

packages/ui-navigation/src/AppNav/index.js
HerrTopi: Oh, It's nice to hear this, thanks :)

packages/ui-navigation/src/AppNav/index.js
HerrTopi: Unfortunately not. we can not call them in componentDidMount , because they do some sort of prop or state changes under the hood and push React into an infinite cycle.

packages/ui-navigation/src/AppNav/Item/theme.js
HerrTopi: Somehow I always forget this :D

packages/ui-navigation/src/AppNav/theme.js
HerrTopi: Not agaaaaaain

matyasf (approved):

packages/ui-navigation/src/AppNav/index.js
ke1k0: Oh, too bad :( Thanks :)

packages/ui-navigation/package.json
```diff
@@ -46,6 +46,8 @@
     "@instructure/ui-tooltip": "^7.3.1",
     "@instructure/ui-utils": "^7.3.1",
     "@instructure/ui-view": "^7.3.1",
+    "@instructure/emotion": "^7.3.1",
+    "@instructure/ui-themes": "^7.3.1",
```
ke1k0, line 50: @matyasf pointed out that this dependency is only needed for tests, so we can move it to "devDependences" (here and all future components, where it is needed for theme.test.js)

ke1k0 (approved):

PR #293, opened 2020-12-14
refactor(ui-rating): refactor to use Emotion

matyasf:


packages/ui-rating/src/Rating/theme.js
```diff
@@ -22,8 +22,17 @@
...
+ */
+const generateComponentTheme = (theme, themeOverride = {}) => {
   return {
-    contentMaxWidth: breakpoints.large
+    contentMaxWidth: theme['breakpoints']['large'],
+    ...themeOverride
   }
 }
+
+export default generateComponentTheme
```
matyasf, line 38: This is some weird stuff, as I can tell the original code was doing nothing

ke1k0: Hm indeed, the code doesn't seem to use it anywhere. I think we should delete it then.

packages/ui-rating/src/RatingIcon/style.js
```diff
@@ -0,0 +1,69 @@
...
+ * @return {Object} The final style object, which will be used in the component
+ */
+const generateStyle = (theme, themeOverride, props, state) => {
+  const componentTheme = generateComponentTheme(theme, themeOverride)
+  const { size } = props
+
+  let fontSize
+  if (size === 'small') fontSize = componentTheme.smallIconFontSize
+  else if (size === 'medium') fontSize = componentTheme.mediumIconFontSize
+  else if (size === 'large') fontSize = componentTheme.largeIconFontSize
```
ke1k0, line 41: Just for the sake of consistency, I'd write it like we do in other components (and also we agreed on not using the single line syntax for if-else):

```suggestion
  const sizeVariants = {
    small: { fontSize: componentTheme.smallIconFontSize },
    medium: { fontSize: componentTheme.mediumIconFontSize },
    large: { fontSize: componentTheme.largeIconFontSize }
  }
```

and then use it like: `...sizeVariants[size]`

packages/ui-rating/src/RatingIcon/index.js
```diff
@@ -95,23 +104,16 @@ class RatingIcon extends Component {
   }
 
   handleTransitionEnter = () => {
-    this.applyTheme()
+    //this.applyTheme() TODO does this need to be handled with emotion?
+    // Likely not good, I see css="[Object object]" in the debugger
```
ke1k0, line 108: In this case I'd rather ask @HerrTopi or @Brailor :)

packages/ui-rating/src/RatingIcon/style.js
```diff
@@ -0,0 +1,69 @@
...
+      lineHeight: 1,
+      '&:first-of-type': {
+        marginLeft: 0
+      },
+      '&:last-of-type': {
+        marginRight: 0
+      }
+    },
+    icon: {
+      label: 'icon',
```
ke1k0, line 58: Please give it the label: "rating__icon" :)

packages/ui-rating/src/Rating/theme.js
matyasf: yea, I double checked, it does nothing, ill delete it

packages/ui-rating/src/RatingIcon/index.js
matyasf: I've tested to dynamically change the number of stars and the rating and it works fine, so it was not needed

packages/ui-rating/src/RatingIcon/style.js
ke1k0: The class ratingIcon is the "root" element, so it should have the label "ratingIcon", named after the component itself. Any other child element of it should have label like "ratingIcon__icon" (in this case), meaning "the icon element in ratingIcon component". Sorry, if it wasn't clear :)

matyasf: done

ke1k0 (approved):

HerrTopi (approved):
Good work :)

PR #294, opened 2020-12-16
WIP(ui-navigation): migrate navigation and navigationItem components to emotion (INSTUI-2763)

HerrTopi:
Closes: INSTUI-2763


packages/ui-navigation/src/Navigation/NavigationItem/styles.js
```diff
@@ -0,0 +1,100 @@
...
+ * @param  {Object} state the state of the component, the style is applied to
+ * @return {Object} The final style object, which will be used in the component
+ */
+const generateStyle = (theme, themeOverride, props, state) => {
+  const { selected } = props
+  const componentTheme = generateComponentTheme(theme, themeOverride)
+
+  return {
+    navigationItem: {
+      label: 'navigationItem',
```
matyasf, line 40: as I know we should use the format `navigation__item`

packages/ui-navigation/src/Navigation/NavigationItem/styles.js
```diff
@@ -0,0 +1,100 @@
...
+          }
+        : {
+            backgroundColor: componentTheme.backgroundColor,
+            color: componentTheme.fontColor,
+            '&:active': {
+              backgroundColor: componentTheme.hoverBackgroundColor
+            },
+            '&:hover': {
+              backgroundColor: componentTheme.hoverBackgroundColor
+            },
```
matyasf, line 79: can this be written as `'&:active, &:hover'`:

packages/ui-navigation/src/Navigation/README.md
```diff
@@ -20,7 +20,7 @@ example: true
         icon={<Img src={placeholderLogo(85, 85)} constrain="cover" />}
         label={<ScreenReaderContent>Home</ScreenReaderContent>}
         href="#"
-        theme={{
+        themeOverride={{
```
matyasf, line 23: good catch, we should aways check readmes and examples.

packages/ui-navigation/src/Navigation/NavigationItem/styles.js
ke1k0: In this case the name of the component is NavigationItem, so the class should be "navigationItem". If there is a child component inside it, like a separator, then that one would have "navigationItem__separator".

matyasf (approved):

packages/ui-navigation/src/Navigation/README.md
ke1k0: It made me think: can we make this work with the old `theme={{}}` syntax? It will be quite annoying to replace all overrides for every component, pretty sure people use it quite often... Don't fix it in this component, just let's talk about it next time :)

ke1k0 (approved):

PR #297, opened 2020-12-16
InstUI-2759 ui-pages emotion refactor

matyasf:


packages/ui-pages/src/Pages/__tests__/Pages.test.js
```diff
@@ -77,7 +77,7 @@ describe('<Pages />', async () => {
       </Pages>
     )
 
-    expect(pageSpy).to.have.been.calledOnce()
+    expect(pageSpy).to.have.been.calledTwice()
```
matyasf, line 80: Its called twice because we call now `this.props.makeStyles()` in `componentDidMount()`

ke1k0: Do you think we should put this as a comment above this test in the code too? Just so the future generations know the reason too :)

ke1k0 (approved):

packages/ui-pages/src/Pages/__tests__/Pages.test.js
matyasf: done

HerrTopi (approved):

PR #300, opened 2020-12-17
InstUI-2746 ui-text emotion migration

matyasf:


packages/ui-text/src/Text/styles.js
```diff
@@ -0,0 +1,165 @@
...
+        color: 'inherit',
+        height: 'auto',
+        width: '100%',
+        lineHeight: 'inherit',
+        textAlign: 'start',
+        boxShadow: 'none',
+        display: 'block',
+        ...textOrInputStyle
+      },
+      ...textOrInputStyle
```
matyasf, line 160: hope this is OK... tests are passing and thing look good in the .design webpage

packages/ui-text/src/Text/styles.js
```diff
@@ -0,0 +1,165 @@
...
+      },
+      'b, strong': {
+        fontWeight: componentTheme.fontWeightBold
+      },
+      p: {
+        display: 'block',
+        padding: 0,
+        margin: componentTheme.paragraphMargin
+      },
+      '& input': {
```
ke1k0, line 142: This is needed for InPlaceEdit, please use styling like this: https://github.com/instructure/instructure-ui/commit/5968508755adf2c9c5c33bc319a3abc522a87866
(You can preview it in InPlaceEdit component, when you click to edit the text and it becomes an input)

packages/ui-text/src/Text/index.js
```diff
@@ -53,24 +45,19 @@ class Text extends Component {
     as: PropTypes.elementType,
     children: PropTypes.node,
     /**
-     * One of: primary, secondary, brand, success, warning, danger, alert, primary-inverse, secondary-inverse
+     * Color of the text
      */
-    color: deprecated.deprecatePropValues(
```
ke1k0, line 58: Since it is a breaking change, please mark it as such in the commit's breaking change block, so it appears in the changelog.

packages/ui-text/src/Text/styles.js
matyasf: done

packages/ui-text/src/Text/index.js
matyasf: done

ke1k0 (approved):

HerrTopi (approved):

PR #301, opened 2020-12-17
FOO-1331 fix packaging for native ESM environments

amireh:
This series of patches makes it possible to import `@instructure` packages found in this repository in environments where ES modules are natively supported, like Node 12.5+ and Webpack 5. Each part includes an explanation of the problem and the proposed solution in its commit message.

Thanks.


amireh:
The last part is optional in that it is not part of the fixes, it only contains the tooling I used to verify the fixes in case you find it helpful too.

packages/ui-themes/package.json
```diff
@@ -5,7 +5,7 @@
   "author": "Instructure, Inc. Engineering and Product Design",
   "type": "module",
   "module": "./es/index.js",
-  "main": "./lib/index.js",
+  "main": "./es/index.js",
```
Brailor, line 8: Just wondering why was this changed to the ES module source?

amireh: There is no `lib/index.js`

Brailor (approved):
Great work, I really appreciated the detailed commit messages, approved from my side.

amireh:
> Great work, I really appreciated the detailed commit messages, approved from my side.

Thank you Brailor

matyasf (approved):
thanks for this big contribution!

PR #303, opened 2020-12-21
InstUI-2747 textinput migration

matyasf:
note: for some reason yarn commit messed up commit names, I need to fix it


packages/ui-text-input/src/TextInput/index.js
```diff
@@ -210,12 +215,29 @@ class TextInput extends Component {
...
+  componentDidMount() {
+    this.props.makeStyles(this.makeStyleProps())
+  }
+
+  componentDidUpdate(prevProps, prevState, snapshot) {
+    this.props.makeStyles(this.makeStyleProps())
+  }
+
+  makeStyleProps = () => {
+    const { interaction } = this
```
ke1k0, line 233: You could get invalid and state from "this" too.

ke1k0 (approved):

PR #304, opened 2020-12-22
INSTUI-2771 - FormField themeable migration

ke1k0:
Closes: INSTUI-2771


packages/ui-docs-client/src/Select/index.js
```diff
@@ -105,6 +105,7 @@ class Select extends Component {
         renderBeforeInput={renderBeforeInput}
         inputValue={inputValue}
         isShowingOptions={isShowing}
+        onInputChange={() => {}}
```
matyasf, line 108: So this is the fix for the error that is thrown now when we load the docs client?

matyasf:
There is something wrong with the commit plugin, it does not include the module name in the commit message :/

packages/ui-docs-client/src/Select/index.js
ke1k0: Well yes, it is a required prop if value is provided, was missing before too... Couldn't figure out why it didn't throw error before :/

ke1k0:
> There is something wrong with the commit plugin, it does not include the module name in the commit message :/

Modified the message, thanks.

Brailor (approved):

packages/ui-form-field/src/FormFieldLabel/styles.css
```diff
@@ -1,27 +0,0 @@
...
-label.root,
-.root {
-  all: initial;
-  display: block;
-}
-
-.root.legend {
-  display: table;
-  width: 100%;
-}
```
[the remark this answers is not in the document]
ke1k0: it was probably some unused, undeleted style, the component never received "legend" class:
```js
    const classes = {
      [styles.root]: true,	
      [styles['has-content']]: hasVisibleChildren(this.props.children)	
    }
```

matyasf (approved):

PR #305, opened 2020-12-22
INSTUI-2796 - badge themeable migration

ke1k0:
Closes: INSTUI-2796


packages/ui-badge/package.json
```diff
@@ -25,20 +25,21 @@
   "dependencies": {
     "@babel/runtime": "^7.9.2",
     "@instructure/console": "^7.3.1",
+    "@instructure/emotion": "^7.3.1",
     "@instructure/ui-color-utils": "^7.3.1",
     "@instructure/ui-position": "^7.3.1",
     "@instructure/ui-react-utils": "^7.3.1",
     "@instructure/ui-testable": "^7.3.1",
     "@instructure/ui-themeable": "^7.3.1",
```
matyasf, line 33: could this dependency be removed?

packages/ui-badge/src/Badge/__tests__/Badge.test.js
```diff
@@ -68,19 +66,6 @@ describe('<Badge />', () => {
...
-        <button type="button">Inbox</button>
-      </Badge>
-    )
-
-    const badge = await BadgeLocator.find()
-
-    expect(await badge.find(`.${styles['positioned--bottom']}`)).to.exist()
-    expect(await badge.find(`.${styles['positioned--start']}`)).to.exist()
-  })
-
```
matyasf, line 83: IMO we should just comment out with a TODO comment tests that are hard to port

packages/ui-badge/src/Badge/styles.js
```diff
@@ -0,0 +1,194 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+import { keyframes } from '@instructure/emotion'
+import generateComponentTheme from './theme'
+
+// keyframes have to be outside of 'generateStyle',
+// since it is causing problems in style recalculation
```
matyasf, line 29: its better here anyway, since it'll run fewer times, so its more performant. Perhaps it could also be made `static`, then we can be sure its running only once.

packages/ui-badge/src/Badge/styles.js
```diff
@@ -0,0 +1,194 @@
...
+      ...(end && {
+        insetInlineEnd: `calc(-1 * ${componentTheme.sizeNotification} / 2)`,
+        insetInlineStart: 'auto'
+      })
+    })
+  }
+
+  const notStandaloneTypeVariant = {
+    count: countPositions,
+    notification: notificationPositions
```
matyasf, line 121: The key here should not be `'&.notification'` and `'.count'`?

matyasf:
There is something wrong with the commit plugin, it does not include the module name in the commit message :/

packages/ui-badge/package.json
ke1k0: `import { ThemeablePropTypes } from '@instructure/ui-themeable'` is still using it in the component (if you meant the ui-themeable dependency :) )

packages/ui-badge/src/Badge/__tests__/Badge.test.js
ke1k0: I removed it, because these are the tests that will be better covered in VRT.

packages/ui-badge/src/Badge/styles.js
ke1k0: count/notification are coming from the `type` property, so we can use it it like any other property variant map

packages/ui-badge/src/Badge/__tests__/Badge.test.js
ke1k0: since I had to modify the commit anyway, I put it back with a comment :)

ke1k0:
> There is something wrong with the commit plugin, it does not include the module name in the commit message :/

nice catch @matyasf :) I changed the commit message, and let's see if the commit script keeps misbehaving

matyasf (approved):

HerrTopi (approved):

PR #307, opened 2021-01-04
InstUI-2748 TextArea Emotion migration

matyasf:


packages/ui-text-area/src/TextArea/styles.js
```diff
@@ -0,0 +1,123 @@
...
+      borderBottomColor: componentTheme.borderBottomColor,
+      borderLeftColor: componentTheme.borderLeftColor,
+      borderRadius: componentTheme.borderRadius,
+      color: componentTheme.color,
+      background: componentTheme.background,
+      whiteSpace: 'pre-wrap',
+      wordWrap: 'break-word',
+      textAlign: 'start',
+      ...fontSizeVariants[size],
+      '&:focus + span': {
```
ke1k0, line 80: Hm, i think using "span" here is pretty risky, the selector is too general. I would rather use the "[class$=-textArea__outline]" syntax, that is specific. Also, the css file used `~` in the selector, which is not equivalent to `+`.  But to be honest, the '+' should work here just as well...

packages/ui-text-area/src/TextArea/styles.js
```diff
@@ -0,0 +1,123 @@
...
+        transform: 'scale(1)',
+        opacity: 1
+      },
+      '&[aria-invalid]': {
+        borderColor: componentTheme.errorBorderColor
+      },
+      '&[aria-invalid]:focus, &[aria-invalid]:focus + span': {
+        borderColor: componentTheme.errorBorderColor
+      },
```
ke1k0, line 89: This can be nested in the previous one. Also, would specify the "span" here too.

packages/ui-text-area/src/TextArea/index.js
```diff
@@ -150,10 +153,12 @@ class TextArea extends Component {
 
   componentDidMount() {
     this.autoGrow()
+    this.props.makeStyles()
   }
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps, prevState, snapshot) {
```
HerrTopi, line 159: `prevProps, prevState, snapshot ` Are these necessary?

packages/ui-text-area/src/TextArea/index.js
```diff
@@ -248,7 +251,7 @@ class TextArea extends Component {
       this._container.style.minHeight = height
     }
 
-    this._height = height
+    this._height = height // string type, e.g. '55px'
```
HerrTopi, line 254: This info should be in the proptypes section imho

HerrTopi (approved):

packages/ui-text-area/src/TextArea/index.js
matyasf: WebStorm gives a warning when these are missing, and it might cause a build error when we convert this to TypeScript. But for now it doesnt change anything

packages/ui-text-area/src/TextArea/styles.js
matyasf: I've pushed an update that adresses this

packages/ui-text-area/src/TextArea/styles.js
matyasf: I've pushed an update that adresses this

ke1k0 (approved):

packages/ui-text-area/src/TextArea/index.js
matyasf: fixed

PR #308, opened 2021-01-05
InstUI-2743 Tray emotion migration

matyasf:


packages/ui-tray/src/Tray/index.js
```diff
@@ -61,7 +58,7 @@ class Tray extends Component {
     /**
      * Placement to determine where the `<Tray />` should display in the viewport
      */
-    placement: PropTypes.oneOf(['top', 'bottom', 'start', 'end']),
+    placement: PropTypes.oneOf(['top', 'bottom', 'start', 'end', 'center']),
```
matyasf, line 61: this was missing in the original code

packages/ui-tray/src/Tray/index.js
```diff
@@ -251,8 +258,6 @@ class Tray extends Component {
 
   handlePortalOpen = (DOMNode) => {
     this.DOMNode = DOMNode
-    // We apply the theme here because now we have a DOM node (provided by Portal)
-    DOMNode && this.applyTheme(DOMNode)
```
matyasf, line 255: seems to work without it, its also tested by the test that I had to rewrite

HerrTopi (approved):
Seems nice and flawless, great work

Brailor (approved):

PR #311, opened 2021-01-06
Fix tests for migrated components

ke1k0:


matyasf:
After re-running tests there are still a few issues that look like a quick fix to me:

in `Media` and `Byline`:
`AssertionError: expected error to have been called with arguments Warning: [Byline] prop 'dir' is not allowed.`

in `Breadcrumb` and `Billboard`:
`ReferenceError: undefinedundefinedundefinedundefinedundefinedundefinedundefinedundefinedundefinedundefinedundefined is not defined`

otherwise nice work!

matyasf (approved):

PR #312, opened 2021-01-06
InstUI-2773 filedrop emotion refactor

matyasf:


packages/ui-file-drop/src/FileDrop/styles.js
```diff
@@ -0,0 +1,132 @@
...
+      zIndex: 1,
+      textAlign: 'center',
+      borderRadius: componentTheme.borderRadius,
+      border: `${componentTheme.borderWidth} ${componentTheme.borderStyle} ${componentTheme.borderColor}`,
+      cursor: 'pointer',
+      '&:hover': {
+        // dragRejected should take precedence!
+        borderColor: dragRejected
+          ? componentTheme.rejectedColor
+          : componentTheme.hoverBorderColor
```
matyasf, line 116: This was accomplished a bit differently in the original, there they had another CSS class for rejected, so it took precedence over `:hover`

ke1k0: Nice solution, but missing a little detail: as I can see in the original, the accepted state takes precedence too. It is not visible, because in the theme the hover and accepted color are set to the same blue color, but the component should handle the difference, so I suggest this:
```js
'&:hover': {
  ...(!dragAccepted && !dragRejected && { borderColor: componentTheme.hoverBorderColor })
},
```

packages/ui-file-drop/src/FileDrop/styles.js
```diff
@@ -0,0 +1,132 @@
...
+    ? {
+        height: '100%'
+      }
+    : {}
+
+  return {
+    fileDrop: {
+      label: 'fileDrop',
+      position: 'relative'
+    },
```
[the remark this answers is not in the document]
matyasf: oh yea, this was the style for the root element, but the `position:relative` is already added in the main file: https://github.com/instructure/instructure-ui/pull/312/files#diff-26afa9aebb62eb371baf31f62e221ac8897ad19747f0a8ef5cf396ec965825deL514

packages/ui-file-drop/src/FileDrop/styles.js
matyasf: nice catch!

ke1k0:

packages/ui-file-drop/src/FileDrop/index.js
```diff
@@ -67,14 +67,14 @@ const IS_MS = isBrowserMS()
 category: components
 ---
 **/
+@withStyle(generateStyle)
 @deprecated('8.0.0', {
   label: 'renderLabel',
   enablePreview: 'shouldEnablePreview',
   allowRepeatFileSelection: 'shouldAllowRepeats',
   allowMultiple: 'shouldAllowMultiple'
 })
```
Brailor, line 76: We should remove these props from the component since they are deprecated.

matyasf: yep, I'm planning to do the removals in a later commit

ke1k0 (approved):

PR #313, opened 2021-01-06
INSTUI-2757 - Position themeable migration

ke1k0:


packages/ui-position/src/Position/index.js
```diff
@@ -327,27 +295,20 @@ class Position extends Component {
...
+    let content = ensureSingleChild(this.props.children)
 
     if (content) {
       content = safeCloneElement(content, {
         ref: (el) => {
           this._content = el
         },
         style: {
+          boxSizing: 'border-box',
+          zIndex: this.props.styles.zIndex,
```
ke1k0, line 307: This is a new breaking change, sadly it is the only way I found to put these 2 css props on the component, emotion doesn't work like this. Also, kind of funny, because if the content block is a View or ContextView, it filters out most of the styles anyway, but the live version works like this too.  Added breaking change comment on the commit.

packages/ui-view/src/View/styles.js
```diff
@@ -382,24 +382,24 @@ const generateStyle = (theme, themeOverride, props, extraArgs = {}) => {
 
   const displayVariants = {
     inline: {
-      label: 'inline',
+      label: 'view--inline',
```
ke1k0, line 385: Modified the labels on View, because they were confusing, sorry :)

Brailor (approved):

HerrTopi (approved):

PR #314, opened 2021-01-08
Numberinput

HerrTopi:


ke1k0:
When deprecating props in the component, please indicate it as a breaking change in the commit, so it appears in the change log :)

packages/ui-number-input/src/NumberInput/styles.js
```diff
@@ -0,0 +1,166 @@
...
+            fontSize: componentTheme.largeFontSize,
+            height: componentTheme.largeHeight
+          })
+    },
+    input: {
+      label: 'numberInput_input',
+      all: 'initial',
+      textAlign: 'start',
+      direction: 'inherit',
+      '&::-ms-clear': { display: 'none' },
```
matyasf, line 146: this is no longer needed in the browsers we support, see https://developer.mozilla.org/en-US/docs/Archive/Web/CSS/::-ms-clear

packages/ui-number-input/src/NumberInput/styles.css
```diff
@@ -1,138 +0,0 @@
...
-  &.large {
-    font-size: var(--largeFontSize);
-    height: var(--largeHeight);
-  }
-}
-
-/* stylelint-disable selector-max-type, no-descending-specificity */
-
-.input,
-input[type].input {
```
matyasf, line 69: I'm just curious, what purpose did this serve? Did it need something to migrate to the new code?

matyasf (approved):

ke1k0:

Brailor (approved):

packages/ui-number-input/src/NumberInput/styles.css
HerrTopi: I think with the restructuring of the styling, this line is implied in the style.

PR #315, opened 2021-01-08
WIP(ui-menu): menu component migration to emotion

HerrTopi:


ke1k0:

packages/ui-menu/src/Menu/MenuItem/styles.js
```diff
@@ -0,0 +1,124 @@
...
+      textAlign: 'start',
+      cursor: 'pointer',
+      userSelect: 'none',
+      fontFamily: componentTheme.fontFamily,
+      fontWeight: componentTheme.fontWeight,
+      lineHeight: componentTheme.lineHeight,
+      fontSize: componentTheme.fontSize,
+      background: componentTheme.background,
+      transition: 'background 0.2s',
+      display: 'block',
```
matyasf, line 86: The original code had a comment about this line, I'd leave it here:

  Changing the following to display: flex; causes a VO
  bug where items with role menuitem, menuitemcheckbox
  and menuitemradio are selected twice with control+
  option+space. So we set the display to block.

packages/ui-menu/src/Menu/MenuItem/styles.js
```diff
@@ -0,0 +1,124 @@
...
+      ...roleStyles,
+      '&:focus,  &:active,  &:hover': {
```
matyasf, line 88: I'm just curious, why have you removed `&[aria-expanded="true"]` from here?

packages/ui-menu/src/Menu/MenuItem/styles.js
```diff
@@ -0,0 +1,124 @@
...
+        background: componentTheme.activeBackground,
+        '[class*="menuItem__label"]': {
+          color: componentTheme.activeLabelColor
+        },
+        '[class*="menuItem__icon"]': {
+          color: componentTheme.activeIconColor
+        }
+      },
+      '&::-moz-focus-inner': {
+        padding: '0',
```
matyasf, line 98: I'd leave the comment here ( `removes extra ff button spacing`)

packages/ui-menu/src/Menu/MenuItem/styles.js
```diff
@@ -0,0 +1,124 @@
...
+        margin: '0',
+        border: '0'
+      },
+      textDecoration: 'none',
```
matyasf, line 102: I'm just curious, this was behind a selector (`a.root, a.root:link,...`) in the .css, is that no longer needed?

packages/ui-menu/src/Menu/MenuItemGroup/index.js
```diff
@@ -245,11 +258,11 @@ class MenuItemGroup extends Component {
   render() {
     const props = omitProps(this.props, MenuItemGroup.propTypes)
     return (
-      <span {...props} className={styles.root} role="presentation">
+      <span {...props} css={this.props.styles.root} role="presentation">
```
matyasf, line 261: this should be `this.props.styles.menuItemGroup`?

matyasf (changes requested):

packages/ui-menu/src/Menu/MenuItem/styles.js
```diff
@@ -0,0 +1,124 @@
...
+ * @param  {Object} props the props of the component, the style is applied to
+ * @param  {Object} state the state of the component, the style is applied to
+ * @return {Object} The final style object, which will be used in the component
+ */
+const generateStyle = (componentTheme, props, state) => {
+  const { type, disabled } = props
+  const { role } = state
+
+  const isMenuitemradioOrMenuitemcheckbox =
+    type === 'checkbox' || type === 'radio'
```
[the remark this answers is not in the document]
HerrTopi: Role is calculated from type in the index file, so it should be type. See: ./index.js Line:200

packages/ui-menu/src/Menu/MenuItem/styles.js
HerrTopi: It seems, that we don't use or set the aria-expanded anywhere for the component, so I've figured, it was mistakenly left there.

packages/ui-menu/src/Menu/MenuItem/styles.js
HerrTopi: It suppose to remove all text decorations from menuItems, and the specificity not needed, since we want to apply that to every case. It seems to do the job.

packages/ui-menu/src/Menu/MenuItemGroup/index.js
HerrTopi: Lol, nice catch

packages/ui-menu/src/Menu/MenuItem/styles.js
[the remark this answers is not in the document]
ke1k0: also, the IDE underlines the name as type: how about `isMenItemRadioOrMenuItemCheckbox` or just simply `isRadioOrCheckbox`?

ke1k0 (approved):

matyasf (approved):
nice work!

PR #326, opened 2021-01-14
fix(ui-modal): fix modal resetting contents when size is changed to/from fullscreen

matyasf:
This behaviour was introduced with INSTUI-1105, where it did not serve any special purpose, so it looks like OK to change it

### DIFF
--- a/packages/ui-modal/src/Modal/index.js
+++ b/packages/ui-modal/src/Modal/index.js
@@ -310,6 +310,7 @@ class Modal extends Component {
       as
     } = this.props
 
+    const isFullScreen = size === 'fullscreen'
     const dialog = (
       <Dialog
         {...passthroughProps(props)}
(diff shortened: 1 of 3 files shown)


packages/ui-modal/src/Modal/index.js
```diff
@@ -336,28 +337,21 @@ class Modal extends Component {
...
+      <Mask
+        placement={this.maskPlacement}
+        fullscreen={constrain === 'window'}
+        theme={
+          isFullScreen ? { borderRadius: '0em', borderWidth: '0em' } : null
+        }
+        className={classnames({
+          [styles.fullscreenLayout]: isFullScreen,
+          [styles[`fullscreenLayout--${constrain}`]]: isFullScreen
+        })}
```
matyasf, line 350: These styles dont seem to change anything now, I just left them there to keep CSS as similar as possible, there might be some edge case where it matters.

packages/ui-modal/src/Modal/index.js
```diff
@@ -336,28 +337,21 @@ class Modal extends Component {
...
-          {dialog}
-        </Mask>
-      )
-    }
+    return (
+      <Mask
+        placement={this.maskPlacement}
+        fullscreen={constrain === 'window'}
+        theme={
+          isFullScreen ? { borderRadius: '0em', borderWidth: '0em' } : null
```
matyasf, line 345: this is to remove the 1px border when set to fullscreen to keep a pixel-by-pixel similarity

ke1k0 (approved):

Brailor (approved):

PR #328, opened 2021-01-14
InstUI-2772: Remove FocusableView, substitute it with View in InPlaceEdit

matyasf:
BREAKING CHANGE:
removed FocusableView. Use View instead, you can set focus with the withFocusOutline property


packages/ui-editable/src/InPlaceEdit/index.js
```diff
@@ -140,9 +159,16 @@ class InPlaceEdit extends Component {
...
     return isEditMode ? (
-      <FocusableView as="span" display="block" focused={showFocusRing}>
+      <View
+        as="span"
+        display="block"
+        withFocusOutline={showFocusRing}
+        position="relative"
+        css={this.props.styles.inPlaceEdit}
+        borderRadius="medium"
+      >
```
matyasf, line 169: Luckily this was the only place InstUI used this component. I've added its CSS for 100% backwards compatibility.

ke1k0: Hm, seems the component looks the same with and without this extra style, but probably it's best to transfer all props, just in case. Good idea :)

packages/ui-editable/src/InPlaceEdit/theme.js
```diff
@@ -22,8 +22,19 @@
...
+ * @param  {Object} theme The actual theme object.
+ * @param  {Object} themeOverride User provided overrides of the default theme mapping.
+ * @return {Object} The final theme object with the overrides and component variables
+ */
+const generateComponentTheme = (theme, themeOverride = {}) => {
+  const { typography } = theme
+
   return {
-    fontFamily: typography.fontFamily
+    fontFamily: typography?.fontFamily,
```
ke1k0, line 35: This one I'm not sure about, because it was the "themeable" prop of FocusView, but not sure we want to make it overridable for InPlaceEdit too. I mean the only usecase of this would be that once you click on the Text to edit, it would change font family...

Edit: I tried to set it with `themeOverride={{fontFamily: "serif"}}` on the component, but it didn't seem to have any effect, since the text child inside had it's own font family anyway. So I'd say this is only confusing to list it as an overridable variable, but has no effect.

packages/ui-focusable/src/Focusable/README.md
```diff
@@ -57,83 +57,3 @@ example: true
 }}
 </Focusable>
 ```
-
-The `FocusableView` component can be used with `Focusable` to render a minimal focus outline when
```
ke1k0, line 61: Running a quick search in the code, I could find a few other usages of `FocusableView` in the code: 
- packages/\_\_docs\_\_/components.js
- packages/ui-focusable/README.md
- rfcs/Button/variants.md

(also in v6 codemods, but that probably should stay)

Can you please have a look if any of this is needed?

packages/ui-editable/src/InPlaceEdit/index.js
matyasf: yea, it might affect some styled text inside this component

packages/ui-editable/src/InPlaceEdit/theme.js
matyasf: good catch, I'll delete it

packages/ui-focusable/src/Focusable/README.md
matyasf: Good catch, I've removed the first 2; the third one is an RFC, this is meant as a historic document that should not be changed

ke1k0 (approved):

Brailor (approved):

PR #338, opened 2021-01-19
INSTUI-2791 - Checkbox themeable migration

ke1k0:


packages/emotion/src/styleUtils/polyFill.js
```diff
@@ -152,7 +152,7 @@ const processStyleProps = (propsObj, dir) =>
             }
           }
 
-          if (isSupportedProps(originalProp) && originalValue) {
+          if (isSupportedProps(originalProp) && originalValue !== 'undefined') {
```
ke1k0, line 155: @Brailor had to make a small tweak, because it was ignoring props with `0` (as number) values, for example `insetInlineStart: 0`.

Brailor (approved):

packages/ui-checkbox/src/Checkbox/styles.js
```diff
@@ -0,0 +1,74 @@
...
+      padding: 0,
+      margin: 0,
+      fontSize: 'inherit',
+      lineHeight: 'inherit',
+      width: 'auto',
+      position: 'absolute',
+      top: 0,
+      insetInlineStart: 0,
+      insetInlineEnd: 'auto',
+      opacity: 0.0001
```
matyasf, line 62: I'd leave the original comment here ("selenium cannot find fully transparent elements"), since this is weird stuff

matyasf (approved):
nice work!

packages/ui-checkbox/src/Checkbox/styles.js
ke1k0: I put it back, thanks!

PR #346, opened 2021-01-21
refactor(emotion,multiple): add annotations to new functions

ke1k0:
Needed to add annotations to our new functions so they appear in the right place on the docs
page.
Fixed errors: page load failed on renderTheme and View with depreacted `background="default"`.


matyasf (approved):
Nice work! Just one thing: We dont really need to do anything with the RFCs, they are feature proposal documents from the past, that might've gotten accepted

Brailor (approved):
Nice!

PR #348, opened 2021-01-25
INSTUI-2754 Refactor RangeInput to use Emotion for theming

matyasf:


packages/ui-range-input/src/RangeInput/styles.js
```diff
@@ -0,0 +1,133 @@
...
+      margin: 0,
+      '&::-webkit-slider-thumb': {
+        ...thumbStyle,
+        marginTop: `calc(-1 * ${componentTheme.handleSize} / 4)`
+      },
+      '&::-moz-range-thumb': thumbStyle,
+      '&:focus, &:active': {
+        outline: 'none',
+        '&::-webkit-slider-thumb': thumbFocusActiveStyle,
+        '&::-moz-range-thumb': thumbFocusActiveStyle
```
matyasf, line 108: for some reason this was not working when I defined in one style ( `'&::-webkit-slider-thumb &::-moz-range-thumb': ...`

packages/ui-range-input/src/RangeInput/styles.js
```diff
@@ -0,0 +1,133 @@
...
+ * Generates the style object from the theme and provided additional information
+ * @param  {Object} componentTheme The theme variable object.
+ * @param  {Object} props the props of the component, the style is applied to
+ * @param  {Object} state the state of the component, the style is applied to
+ * @return {Object} The final style object, which will be used in the component
+ */
+const generateStyle = (componentTheme, props, state) => {
+  const { size } = props
+  const { dir } = state
+  const rtlInputStyle = dir === DIRECTION.rtl ? { direction: 'rtl' } : {}
```
ke1k0, line 40: I am not sure we need to access the dir in the styles.js, other components use the `'[dir="rtl"] &': { ... }` syntax too (SVGIcon, ToggleFacade, Tabs, Tag). It kind of behaves differently too, so we could probably bring up this topic on a standup.

packages/ui-range-input/src/RangeInput/README.md
```diff
@@ -6,7 +6,34 @@ An html5 range input/slider component.
...
+class Example extends React.Component {
+  state = { size: "small" }
+  handleModalSizeChange = (event, value) => {
+    this.setState({ size: value })
+  }
+  render() {
+    return (
+      <div>
+        <RangeInput label="Grading range" defaultValue={30} max={100} min={0} size={this.state.size} />
+        <FormFieldGroup rowSpacing="medium">
```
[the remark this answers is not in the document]
matyasf: good catch

packages/ui-range-input/src/RangeInput/styles.js
matyasf: nice catch :)

packages/ui-range-input/src/RangeInput/styles.js
matyasf: figured out: its because the old code used the `postcss-input-range` plugin

ke1k0 (approved):

Brailor (approved):

PR #349, opened 2021-01-25
INSTUI-2792 - Calendar themeable migration

ke1k0:
Closes: INSTUI-2792


packages/ui-calendar/src/Calendar/index.js
```diff
@@ -160,13 +174,13 @@ class Calendar extends Component {
...
-      [styles.withNavigationButtons]: prevButton || nextButton
-    })
+    const style = [
+      styles.navigation,
+      ...(prevButton || nextButton ? [styles.navigationWithButtons] : [])
+    ]
 
     return (
-      <div className={classes}>
+      <div css={style}>
```
matyasf, line 183: I did not knew we could add multiple classes this way :) Are you sure this evaluates correctly, dont you need brackets? e.g. 
```
...((prevButton || nextButton) ? [styles.navigationWithButtons] : [])
```

matyasf (approved):
nice work :)

Brailor (approved):

packages/ui-calendar/src/Calendar/index.js
ke1k0: I tested it, works just fine. If the second style is added, then the label of the second one will be displayed (in this case css-xxxx-calendar__navigation--withButtons), which is what I'd expect here, since the second one is a "modified version", and it also nicely merges the properties. Works without brackets too.

PR #351, opened 2021-01-26
INSTUI-2798 - migrate template-component from themeable to emotion

ke1k0:
Closes: INSTUI-2798

Component template now uses emotion instead of themeable.


packages/template-component/package.json
```diff
@@ -15,16 +15,17 @@
...
   "devDependencies": {
     "@instructure/ui-babel-preset": "^7.3.1",
     "@instructure/ui-color-utils": "^7.3.1",
     "@instructure/ui-test-locator": "^7.3.1",
-    "@instructure/ui-test-utils": "^7.3.1"
+    "@instructure/ui-test-utils": "^7.3.1",
+    "@instructure/ui-themes": "^7.3.1"
   },
   "peerDependencies": {
     "react": "^16.8"
```
matyasf, line 31: We should also add   "sideEffects": false" here too.

matyasf (approved):
nice work

packages/template-component/package.json
ke1k0: done, thanks :)

Brailor (approved):

PR #352, opened 2021-01-26
INSTUI-2789 - DateInput themeable migration

ke1k0:


packages/ui-date-input/src/DateInput/index.js
```diff
@@ -257,6 +254,14 @@ class DateInput extends Component {
     children: null
   }
 
+  componentDidMount() {
+    this.props.makeStyles()
+  }
+
+  componentDidUpdate(prevProps, prevState, snapshot) {
```
HerrTopi, line 261: Parameters not needed 
```suggestion
  componentDidUpdate() {
```

HerrTopi (approved):

Brailor (approved):

packages/ui-date-input/src/DateInput/index.js
ke1k0: WebStorm complains if they are missing... that is why Matyi and me always add them :D

PR #356, opened 2021-01-27
Release (INSTUI-2827)

HerrTopi:


Brailor:
Slack notification for the status of the release is missing.

HerrTopi:
> Step for the slack notification for the result of the release is missing.

I guess it should be triggered from github itself.

packages/ui-scripts/lib/publish.js
```diff
@@ -25,43 +25,40 @@ const { getPackageJSON } = require('@instructure/pkg-utils')
...
-  createNPMRCFile(config)
-  checkWorkingDirectory()
+  //create NPM RC and try to auth in to npm
+  createNPMRCFile(packageConfig)
 
-  let versionToRelease, tag
+  const isRelease = isReleaseCommit(currentVersion)
+  const infoMessage = isRelease
+    ? `📦  Currently on release commit for ${currentVersion} of ${packageName}.`
+    : `📦  Not on a release commit--publishing a pre-release...`
```
matyasf, line 54: We should call this nightly build.

.github/workflows/release.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Set up git identity
+        run: git config --global user.name "instructure-ui-ci" && git config --global user.email "instructure-ui-ci@instructure.com"
+      - name: Remove __build__ folder
+        run: rm -rf packages/__docs__/__build__
+      - name: Create __build__ folder
+        run: mkdir packages/__docs__/__build__
+      - name: Checkout gh-pages
+        run: git checkout gh-pages
+      - name: Checkout master
+        run: git checkout master
```
matyasf, line 37: this has always the latest commit? dont we need a `git pull` here?

.github/workflows/release.yml
```diff
@@ -0,0 +1,45 @@
...
+    runs-on: ubuntu-latest
+    name: A job to execute release
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Install packages.
+        run: yarn install
+      - name: Set up project.
+        run: yarn build
```
matyasf, line 19: Shouldn't we use `yarn bootstrap` here? That seems to do more checks and other stuff like builing types

.github/workflows/release.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Run tests.
+        run: yarn test
+      - name: Release to NPM
+        env:
+          NPM_TOKEN: ${{secrets.NPM_TOKEN}}
+          NPM_EMAIL: ${{secrets.NPM_EMAIL}}
+          NPM_USERNAME: ${{secrets.NPM_USERNAME}}
+        run: yarn release
```
matyasf, line 27: As I see this is the first 'destructive' operation. What is the plan if this suceeds, but something later (like the `docs:build` fails)?

.github/workflows/release.yml
```diff
@@ -0,0 +1,45 @@
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Install packages.
+        run: yarn install
+      - name: Set up project.
+        run: yarn build
+      - name: Run tests.
+        run: yarn test
```
ke1k0, line 21: I'd consider using `test:patchset` here, becasue it runs all tests (jest too) and does linting too, and this runs on the push builds too.

packages/ui-scripts/lib/publish.js
```diff
@@ -25,43 +25,40 @@ const { getPackageJSON } = require('@instructure/pkg-utils')
...
   process.exit(1)
 }
 
-async function publish(packageName, currentVersion, preidAndTag, config = {}) {
+async function publish(options) {
+  const { packageName, currentVersion, packageConfig } = options
+  //Set up git user.name and user.email
   setupGit()
-  createNPMRCFile(config)
-  checkWorkingDirectory()
```
ke1k0, line 45: is this not needed?

packages/ui-scripts/lib/utils/git.js
```diff
@@ -24,87 +24,31 @@
...
-    runCommandSync('yarn', ['husky:pre-commit'])
-  } catch (err) {
-    error(err.stdout)
-    process.exit(1)
-  }
-
-  return runCommandSync('git-cz')
+exports.setupGit = () => {
+  runGitCommand(['config', 'user.email', 'instructure-ui-ci@instructure.com'])
+  runGitCommand(['config', 'user.name', 'instructure-ui-ci'])
```
ke1k0, line 34: this is a rather simplified version of the former one. can you please tell me more about this on the next standup?

.github/workflows/release.yml
HerrTopi: The first operation clones the repo, so master should be up to date

.github/workflows/release.yml
HerrTopi: We can. It seems to work with build as well and it is quicker, but we can switch to bootstrap, no problem if you think it is meaningful

.github/workflows/release.yml
HerrTopi: We would publish it by hand. We could keep a script for that occasion but we can copy the steps from here and do it by hand.

packages/ui-scripts/lib/publish.js
HerrTopi: Sure, let's rename it

.github/workflows/release.yml
HerrTopi: Good idea, let's do that

packages/ui-scripts/lib/publish.js
HerrTopi: It does (or did) virtually nothing. Checks if the working directory is clean or not, but there is no possible way I can see where it would't.

packages/ui-scripts/lib/utils/git.js
HerrTopi: instui-dev@instructure.com

.github/workflows/release.yml
HerrTopi: It seems meaningful, switched to bootstrap

matyasf (approved):

ke1k0 (approved):

Brailor (approved):

PR #358, opened 2021-01-29
InstUI-2774 drawerlayout Emotion refactor

matyasf:
The removed test util was used to set legacy React context


packages/ui-test-sandbox/src/utils/reactComponentWrapper.js
```diff
@@ -53,11 +53,6 @@ class ReactComponentWrapper {
               wrapper.setChildProps(newProps)
             })
           },
-          setContext(newContext) {
-            return doAsync(() => {
-              wrapper.setChildContext(newContext)
-            })
-          },
```
matyasf, line 60: This was setting legacy context and was used only in one test, which I just refactored

packages/ui-drawer-layout/src/DrawerLayout/DrawerContent/index.js
```diff
@@ -84,16 +90,19 @@ class DrawerContent extends Component {
...
   componentDidUpdate() {
+    /*
     this._timeouts.push(
       setTimeout(() => {
         this.setState({
           shouldTransition: true
         })
       })
-    )
+    )*/
```
Brailor, line 104: I think we should delete unused/commented out code, if we might need it, it is there in the git history.

packages/ui-drawer-layout/src/DrawerLayout/theme.js
```diff
@@ -0,0 +1,35 @@
...
+ * SOFTWARE.
+ */
+
+/**
+ * Generates the theme object for the component from the theme and provided additional information
+ * @param  {Object} theme The actual theme object.
+ * @param  {Object} themeOverride User provided overrides of the default theme mapping.
+ * @return {Object} The final theme object with the overrides and component variables
+ */
+const generateComponentTheme = (theme, themeOverride = {}) => {
```
Brailor, line 31: Do we need this file at all?

packages/ui-drawer-layout/src/DrawerLayout/DrawerTray/__tests__/DrawerTray.test.js
```diff
@@ -96,17 +99,26 @@ xdescribe('<DrawerTray />', async () => {
...
-        }}
-        render={() => {
-          return 'Hello from layout tray'
-        }}
-      />
+        }
+      }
+    }
+    await mount(
+      <EmotionThemeProvider theme={themeOverride}>
```
ke1k0, line 110: Normally wrapping it in EmotionThemeProvider is not needed to add override, would be enough to add `themeOverride={{}}` directly on the component, but this fails here because of another reason. We need to omit "themeOverride" prop in the `omitProps` handler, so it doesn't get passed through: I'll fix it on the main branch.

ke1k0 (approved):

packages/ui-drawer-layout/src/DrawerLayout/DrawerContent/index.js
matyasf: yea, this was left by accident

packages/ui-drawer-layout/src/DrawerLayout/theme.js
matyasf: nope I'll just pass `null` to `withStyle()`

packages/ui-drawer-layout/src/DrawerLayout/DrawerTray/__tests__/DrawerTray.test.js
matyasf: nice find, I'll fix it :)

PR #359, opened 2021-01-29
fix: fix release script (INSTUI-2847)

HerrTopi:
Closes: INSTUI-2847

The docs deploy script ran on every push to master. It only should run on release commit pushes
TEST PLAN:
Push a release and a non release commit to master. On release commit, the "Deploy" job should run,
otherwise it should not

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,7 +27,7 @@ jobs:
         run: yarn release
   deploy:
     needs: release
-    if: "contains(github.event.head_commit.message, 'chore(release)')"
+    if: "startsWith(github.event.head_commit.message, 'chore(release)')"
     runs-on: ubuntu-latest
     name: Deploy docs
     steps:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,6 +26,7 @@ jobs:
           NPM_USERNAME: ${{secrets.NPM_USERNAME}}
         run: yarn release
   deploy:
+    needs: release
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -8,7 +8,7 @@ on:
 jobs:
   release:
     runs-on: ubuntu-latest
-    name: A job to execute release
+    name: Release top npm
```
ke1k0, line 11: I don't really understand this description tbh :D

.github/workflows/release.yml
```diff
@@ -25,6 +25,21 @@ jobs:
           NPM_EMAIL: ${{secrets.NPM_EMAIL}}
           NPM_USERNAME: ${{secrets.NPM_USERNAME}}
         run: yarn release
+  deploy:
+    needs: release
+    if: "contains(github.event.head_commit.message, 'chore(release)')"
```
ke1k0, line 30: Not sure if it makes a difference, but at another place in the code we do this check with "startsWith(...)" instead of "contains()". I'd probably use that here too, just for consistency's sake.

ke1k0 (approved):

.github/workflows/release.yml
HerrTopi: Top wants to be to :D Ty

matyasf (approved):

Brailor (approved):

ke1k0 (approved):

PR #360, opened 2021-01-29
INSTUI-2756 - Progress themeable migration

ke1k0:


packages/ui-progress/src/ProgressBar/theme.js
```diff
@@ -22,77 +22,89 @@
...
+      meterBorderColorInverse: colors?.borderLightest,
+      meterColorBrandInverse: colors?.backgroundLightest,
+      meterColorSuccessInverse: colors?.backgroundLightest,
+      meterColorInfoInverse: colors?.backgroundLightest,
+      meterColorAlertInverse: colors?.backgroundLightest,
+      meterColorWarningInverse: colors?.backgroundLightest,
+      meterColorDangerInverse: colors?.backgroundLightest
+    }
+  }
+  themeSpecificStyle['canvas-a11y'] = themeSpecificStyle['canvas-high-contrast']
```
matyasf, line 49: Is this needed? As I see we don't have a 'canvas-a11y' theme, I guess its an older name for the 'canvas-high-contrast' theme

packages/ui-progress/src/ProgressBar/styles.js
```diff
@@ -0,0 +1,170 @@
...
+ * @param  {Object} state the state of the component, the style is applied to
+ * @return {Object} The final style object, which will be used in the component
+ */
+const generateStyle = (componentTheme, props, state) => {
+  const { valueNow, valueMax, size, color, meterColor } = props
+
+  const meterColorClassName =
+    typeof meterColor === 'function'
+      ? meterColor({ valueNow, valueMax })
+      : meterColor
```
matyasf, line 41: I think the docstring in ProgressBar  should be more clear how this variable works, e.g. describe what kind of method reference can be here. 
Also if its a function type the original code called this function on every `render()` to recalculate the current color, is this updated too?

Brailor (approved):

matyasf (approved):

packages/ui-progress/src/ProgressBar/theme.js
ke1k0: Thanks for pointing this out, I removed it (and from any other usages in next too)!

packages/ui-progress/src/ProgressBar/styles.js
ke1k0: I tried out and it is updating correctly. Since valueNow will change, the style recalculation will happen anyway.

PR #362, opened 2021-01-31
chore: refactor vr validation workflow so its not using docker anymore

Brailor:


package.json
```diff
@@ -24,7 +24,6 @@
     "start": "lerna run start --stream --scope docs-app",
     "start:watch": "lerna run start:watch --stream --scope docs-app",
     "dev": "yarn start:watch",
-    "pretest:patchset": "yarn bootstrap && yarn check",
     "test:patchset": "npm-run-all lint:commit lint:changes test:jest test",
     "pretest:merge": "yarn bootstrap && yarn check",
```
matyasf, line 28: as I see this is not used anywhere so we should remove it (and `yarn check` is deprecated anyway https://classic.yarnpkg.com/en/docs/cli/check/#toc-yarn-check )

matyasf (approved):

matyasf (approved):

package.json
Brailor: I did remove it, for the same reasons you mentioned. :D

HerrTopi (dismissed):

PR #363, opened 2021-02-01
INSTUI-2788, INSTUI-2781, INSTUI-2780 docs app Emotion refactor

matyasf:
Refactor to use Emotion the main app, Guidelines and Header components.
Not pixel-perfect yet, I think for the main app it applies the `instructure` instead of the `canvas` theme in some places for some reason


packages/ui-docs-client/src/index.js
```diff
@@ -47,7 +47,7 @@ function renderDocsClient(data, element) {
 
   ReactDOM.render(
     <EmotionThemeProvider theme={canvas}>
-      <App {...data} />,
```
matyasf, line 50: this was a bug from a previous commit that caused the nav layout to shift up a bit

Brailor (approved):

packages/ui-docs-client/src/Guidelines/index.js
```diff
@@ -22,28 +22,47 @@
...
+  }
+
   render() {
     return (
-      <Flex wrap="wrap" justifyItems="end" alignItems="stretch">
+      <Flex
+        wrap="wrap"
+        justifyItems="end"
+        alignItems="stretch"
+        css={this.props.styles.guidelines}
```
ke1k0, line 64: Hm, as far as I can tell, the root class was never actually added on the element in the previous version. Is it needed? If not, we could probably delete all styling to this element.

packages/ui-docs-client/src/Header/index.js
```diff
@@ -29,13 +29,15 @@ import { Link } from '@instructure/ui-link'
...
 import { Heading } from '../Heading'
 
-import theme from './theme'
+import { withStyle } from '@instructure/emotion'
 
-@themeable(theme)
+@withStyle(
+  (componentTheme, props, state) => ({}),
+  (theme, themeOverride = {}) => ({})
+)
```
ke1k0, line 40: Since we don't use any styling here, the withStyle decorator can be deleted altogether.

packages/ui-docs-client/src/compileMarkdown.js
```diff
@@ -263,9 +265,7 @@ function createRenderer() {
             code: matter.content,
             render: (_el) => {
               el = theme ? (
-                <ApplyTheme theme={ApplyTheme.generateTheme(theme)}>
-                  {_el}
-                </ApplyTheme>
+                <EmotionThemeProvider theme={theme}>{_el}</EmotionThemeProvider>
```
ke1k0, line 268: Sadly this won't work, because `const theme = (matter.data || {}).theme` will be a string (name of the theme, if it is given), and the theme provider needs an object to work. I wrote about this issue on the slack channel, we should discuss it, but I guess it works for now (there are other issues we need to fix too).

packages/ui-docs-client/src/Guidelines/index.js
matyasf: yea, its just dead code, I'll delete it

ke1k0 (approved):

packages/ui-docs-client/src/compileMarkdown.js
matyasf: As per our discussion we will fix this in a future PR

PR #367, opened 2021-02-02
fix(ui-docs-plugin): fix favicon (INSTUI-2837)

HerrTopi:
Closes: INSTUI-2837

### DIFF
--- a/packages/__docs__/index.ejs
+++ b/packages/__docs__/index.ejs
@@ -1,29 +1,48 @@
 <!DOCTYPE html>
 <html lang="en" dir="ltr">
   <head>
-    <meta charset="utf-8">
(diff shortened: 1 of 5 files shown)


packages/__docs__/index.ejs
```diff
@@ -1,29 +1,48 @@
...
+    />
```
matyasf, line 13: While quite hacky (I recommended this solution) it beats the 100+ lines of dependencies added to `yarn.lock` and another plugin/code/etc that can fail for misterious reasons (as it did ~100 commits before)

matyasf (approved):

ke1k0 (approved):

PR #374, opened 2021-02-05
INSTUI-2852 Fix bundle build, build pipeline fixes

matyasf:
Sadly I could not get .mjs build to work, but I found a couple of build pipeline issues meanwhile

### DIFF
--- a/packages/babel-plugin-transform-imports/lib/index.js
+++ b/packages/babel-plugin-transform-imports/lib/index.js
@@ -62,61 +62,52 @@ function transform(transformOption, importName, matches) {
     try {
       transformFn = isFunction ? transformOption : require(transformOption)
     } catch (error) {
-      console.error(
-        `[${transformName}] failed to require transform file ${transformOption}`
-      )
+      throw {
+        err: `[${transformName}] failed to require transform file ${transformOption} ${error}`
+      }
     }
-
     if (typeof transformFn !== 'function') {
-      console.error(
-        `[${transformName}] expected transform function to be exported from ${transformOption}`
-      )
+      throw {
+        err: `[${transformName}] expected transform function to be exported from ${transformOption}`
+      }
     }
-
     let importPath = importName
 
     // Sometimes the import is not located at root level of the src. Examine the source of the specified package and if the designated
     // import is not at the root level, construct a relative path from the source root to it's location.
     if (matches && matches[1]) {
       const packageName = matches[1]
+      const sourceIndex = require.resolve(packageName)
+      const sourceRoot = dirname(sourceIndex)
 
-      try {
-        const sourceIndex = require.resolve(packageName)
-        const sourceRoot = dirname(sourceIndex)
+      const importPaths = globby.sync(
+        [
+          `${sourceRoot}/**/${importName}.js`,
+          `${sourceRoot}/**/${importName}/index.js`
+        ],
+        { cwd: sourceRoot }
+      )
 
-        const importPaths = globby.sync(
-          [
-            `${sourceRoot}/**/${importName}.js`,
-            `${sourceRoot}/**/${importName}/index.js`
-          ],
-          { cwd: sourceRoot }
-        )
-
-        if (!importPaths || importPaths.length === 0) {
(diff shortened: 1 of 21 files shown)


packages/babel-plugin-transform-imports/lib/index.js
```diff
@@ -62,61 +62,52 @@ function transform(transformOption, importName, matches) {
...
+      }
     }
-
     if (typeof transformFn !== 'function') {
-      console.error(
-        `[${transformName}] expected transform function to be exported from ${transformOption}`
-      )
+      throw {
+        err: `[${transformName}] expected transform function to be exported from ${transformOption}`
+      }
```
matyasf, line 72: these are fatal faults in the build, but they were just logging errors

packages/babel-plugin-transform-imports/lib/index.js
```diff
@@ -62,61 +62,52 @@ function transform(transformOption, importName, matches) {
...
-            : importPaths[0]
-        )
-      } catch (error) {
+      if (!importPaths || importPaths.length === 0) {
+        // If there are no import paths found it is the same as if globby or the require.resolve failed, we cannot construct a relative import path
+        // so throw an error and just fall back to using the import name as the path.
         console.error(
           `[${transformName}] no modules match '${importName}' in '${packageName}'. Continuing with '${importName}' as the import path. If that is unexpected or incorrect, make sure '${importName}' exists in '${packageName}' and you have run 'yarn install' to download the package source.`
         )
+        return null
```
matyasf, line 97: This is the main fix.
This plugin looks at the imports and replaces them with `${sourceRoot}/**/${importName}.js` or `${sourceRoot}/**/${importName}/index.js`. If we export things and its filename does not exist, it will just display here an error and continue to make a broken build. After this when actually compiling an app (e.g. making a docs bundle) will the error come out.

packages/babel-plugin-transform-imports/lib/index.js
```diff
@@ -62,61 +62,52 @@ function transform(transformOption, importName, matches) {
...
       }
+      if (importPaths.length > 1) {
+        console.error(
+          `[${transformName}] multiple modules found with the name '${importName}' in '${packageName}'. Continuing using the first match: '${importPaths[0]}'.`
+        )
```
matyasf, line 102: This is also a weird edge case, should we throw an error in this case?

packages/ui-docs-plugin/src/index.js
```diff
@@ -36,32 +36,26 @@ DocsPlugin.prototype.apply = function (compiler) {
...
-      defer: ['ui-docs'] // to ensure that globals loads first
-    })
-  )
+  new SingleEntryPlugin(
+    this.context,
+    `!!${require.resolve('./loaders/docs-loader')}?${JSON.stringify(
+      pluginOptions
+    )}!`,
+    'ui-docs'
+  ).apply(compiler)
```
matyasf, line 45: This is just updating a deprecated syntax, that would throw an error if we upgrade to Webpack 5

packages/ui-webpack-config/config/module/rules.js
```diff
@@ -38,7 +38,7 @@ const babelLoader = {
 
 const rules = [
   {
-    test: /\.js$/,
+    test: /\.m?js$/,
```
matyasf, line 41: this will include .mjs files in webpack processing

packages/ui-buttons/src/ToggleButton/ToggleButtonLocator.js
```diff
@@ -25,8 +25,7 @@
 import { locator } from '@instructure/ui-test-locator'
 import { find, parseQueryArguments } from '@instructure/ui-test-queries'
 
-// eslint-disable-next-line no-restricted-imports
-import { TooltipLocator } from '@instructure/ui-tooltip/es/Tooltip/TooltipLocator'
+import { TooltipLocator } from '@instructure/ui-tooltip'
```
matyasf, line 28: With the Node 12+ one can only access methods that are defined in the `exports` field, it could be that these were not working since a while.

Brailor (approved):

HerrTopi (approved):

PR #377, opened 2021-02-08
INSTUI-2855 - fix randomly applied and changing themes

ke1k0:
Closes: INSTUI-2855

Seems like the ancestor overriding logic was unnecessary and even harmful in our
EmotionThemeProvider, so we switched back using emotion's ThemeProvider directly.

Contains updates for EmotionThemeProvider documentation.


matyasf (approved):

Brailor (approved):
Nice work

packages/ui-alerts/src/Alert/index.js
```diff
@@ -228,13 +228,20 @@ class Alert extends Component {
...
       }
     }
   }
 
+  renderScreenreeaderAlert(div) {
+    const content = this.createScreenreaderContentNode()
+    ReactDOM.render(
+      <EmotionThemeProvider theme={canvas}>{content}</EmotionThemeProvider>,
+      div
+    )
```
matyasf, line 242: why is the `canvas` theme hardcoded here?

matyasf (approved):

packages/ui-alerts/src/Alert/index.js
ke1k0: Well, the thing is, the `ScreenReaderContent` that gets generated inside doesn't really need a theme, but since this block gets rendered outside of our app (it has its own `ReactDOM.render()`), it puts a warning on the console that "no themes found, applying default canvas theme". Other option would be to pass it an "empty" theme, or trying to refactor it completely. Would you prefer the empty theme?

matyasf: no, I think this is better, I guess an empty theme might cause issues (e.g. missing properties). Maybe add a comment here explaining this edge case.

Brailor (approved):

packages/ui-alerts/src/Alert/index.js
ke1k0: added the comment, thanks :)

HerrTopi (approved):

PR #379, opened 2021-02-09
INSTUI-2698 Tooltips don't dismiss when hovered over

matyasf:
https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html

Persistent
The intent of this condition is to ensure users have adequate time to perceive the additional content after it becomes visible. Users with disabilities may require more time for many reasons, such as to change magnification, move the pointer, or simply to bring the new content into their visual field. Once it appears, the content should remain visible until:

The user removes hover or focus from the trigger and the additional content, consistent with the typical user experience

### DIFF
--- a/packages/ui-popover/src/Popover/index.js
+++ b/packages/ui-popover/src/Popover/index.js
@@ -302,10 +302,15 @@ class Popover extends Component {
 
     this._handleMouseOver = handleMouseOverOut.bind(null, (event) => {
       this.show(event)
+      clearTimeout(this.mouseOutTimeout)
     })
     this._handleMouseOut = handleMouseOverOut.bind(null, (event) => {
-      this.hide(event)
+      // this is needed bc the trigger mouseOut fires before tooltip mouseOver
+      this.mouseOutTimeout = setTimeout(() => {
+        this.hide(event)
+      }, 1)
     })
+    this.mouseOutTimeout = undefined
   }
 
   get isTooltip() {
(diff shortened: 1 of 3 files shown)


packages/ui-popover/src/Popover/index.js
```diff
@@ -302,10 +302,15 @@ class Popover extends Component {
...
+      clearTimeout(this.mouseOutTimeout)
     })
     this._handleMouseOut = handleMouseOverOut.bind(null, (event) => {
-      this.hide(event)
+      // this is needed bc the trigger mouseOut fires before tooltip mouseOver
+      this.mouseOutTimeout = setTimeout(() => {
+        this.hide(event)
+      }, 1)
     })
+    this.mouseOutTimeout = undefined
```
matyasf, line 313: not a nice solution, im open to better ideas :)

ke1k0: A nicer way would be to refactor this whole tooltip-popover-dialog-contextview family, but maybe after v8 :D

packages/ui-tooltip/src/Tooltip/README.md
```diff
@@ -74,19 +75,29 @@ class Example extends React.Component {
...
+        >
+          <Link href="#">This link has a tooltip</Link>
+        </Tooltip>
+        </p>
+        <Checkbox label="show tooltip?" variant="toggle"
+                  value="toggled"
+                  onChange={(event) => {
+                    this.setState({isShowingContent: event.target.checked})
+                  }}
+        />
```
ke1k0, line 98: It is probably just my OCD, but can you please fix the indents here? :)
```suggestion
        <Checkbox
          label="show tooltip?"
          variant="toggle"
          value="toggled"
          onChange={(event) => {
            this.setState({isShowingContent: event.target.checked})
          }}
        />
```

ke1k0 (approved):

HerrTopi (approved):

PR #385, opened 2021-02-11
chore: change component example generation so its not using deprecate…

Brailor:
…d props anymore

### DIFF
--- a/packages/ui-list/src/List/index.js
+++ b/packages/ui-list/src/List/index.js
@@ -50,7 +50,9 @@ category: components
 ---
 **/
 @deprecated('8.0.0', {
-  variant: 'List with the isUnstyled boolean or InlineList'
+  variant: 'List with the isUnstyled boolean or InlineList',
+  delimeter:
+    'with delimiter set to [pipe, slash, arrow] will only be available when using [InlineList] as of version 8.0.0.'
 })
 @testable()
 @themeable(theme, styles)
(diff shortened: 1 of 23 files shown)


ke1k0 (dismissed):
Looks good! Shall we make a ticket for making a cleanup version of this for next, after this is merged?

packages/ui-list/src/List/index.js
```diff
@@ -50,7 +50,9 @@ category: components
 ---
 **/
 @deprecated('8.0.0', {
-  variant: 'List with the isUnstyled boolean or InlineList'
+  variant: 'List with the isUnstyled boolean or InlineList',
+  delimeter:
+    'with delimiter set to [pipe, slash, arrow] will only be available when using [InlineList] as of version 8.0.0.'
```
matyasf, line 55: This should not be here. Also there is a type "delimeter"

packages/ui-list/src/List/index.js
```diff
@@ -64,12 +66,14 @@ class List extends Component {
...
-      ({ propValue }) =>
-        `with 'delimiter' set to ${propValue} will only be available when using [InlineList] as of version 8.0.0.`
-    ),
+    delimiter: PropTypes.oneOf([
+      'none',
+      'dashed',
+      'solid',
+      'pipe',
+      'slash',
+      'arrow'
```
matyasf, line 75: these are removed

packages/ui-view/src/View/index.js
```diff
@@ -57,7 +57,12 @@ category: components
...
 @deprecated('8.0.0', {
   focused: 'withFocusOutline',
-  visualDebug: 'withVisualDebug'
+  visualDebug: 'withVisualDebug',
+  background: `In version 8.0.0,
+    the value <default> for background will be changed to <primary>,
+    the value <light> for background will be changed to <secondary>,
+    the value <inverse> for background will be changed to <primary-inverse>.
+    Use these values instead.`
 })
```
matyasf, line 66: is this OK? If we put in here a deprecation notice, it will change in 9.0

ke1k0: After this is merged into master, and master is again merged into next, we need to make cleanup on next to reverse these.

matyasf (approved):

ke1k0 (approved):

PR #389, opened 2021-02-12
refactor(XXLocator): Allow importing from es/ again.

matyasf:
This is a revert of this commit: https://github.com/instructure/instructure-ui/commit/579ab2f4f0bfe9e46590e25a4c5743b28a445761 
This is no longer needed because with module type = commonJS we can import from any place again. Also it broke apps because `ui-test-locator` is declared as a `devDependency `


ke1k0 (approved):

matyasf:
Note: Tests are breaking while https://github.com/instructure/instructure-ui/pull/387 is not merged

Brailor (approved):

PR #390, opened 2021-02-12
INSTUI-2851: template-app: update the code for Emotion theming

matyasf:


matyasf:
I deleted the favicon because the linter kept formatting and breaking it.

packages/template-app/template/src/Banner/index.js.ejs
```diff
@@ -75,4 +84,4 @@ class Banner extends Component {
   }
 }
 
-export default themeable(theme, styles)(Banner)
+export default Banner
```
ke1k0, line 87: all our components have a names export too, probably would be nice here too
```suggestion
export default Banner
export { Banner }
```

ke1k0 (approved):

HerrTopi (approved):

PR #392, opened 2021-02-15
INSTUI-2871 : Fix package imports, fix wrong test call in main package.json

matyasf:
this is the same PR as https://github.com/instructure/instructure-ui/pull/387 just for master.

TEST PLAN:
`bootstrap, build:docs, test:all, dev` works, publishing the package into a private local registry and using it in a test app works an non-exported paths from `@instructure/<component>/es` can be imported.


ke1k0 (dismissed):
I tried with bootstrap, build, build:docs, dev, test:patchset, and all seemed to be fine :)

ke1k0 (approved):

HerrTopi (approved):
Can't really verify, but nevertheless, awesome work :)

PR #397, opened 2021-02-17
INSTUI-2876 - refactor themeOverrides to be handled in withStyle

ke1k0:
Closes: INSTUI-2876

The handling of `themeOverride` props were moved to the `withStyle` decorator. This way the theme.js
of the components are simplified.


HerrTopi (approved):
This was a nice catch. It simplifies theming even further, good job :)

packages/ui-editable/src/InPlaceEdit/styles.js
```diff
@@ -48,7 +48,6 @@ const generateStyle = (componentTheme, props, state) => {
       touchAction: 'manipulation',
       background: 'transparent',
       border: 'none',
-      fontFamily: componentTheme.fontFamily,
```
matyasf, line 51: Why was this deleted?

ke1k0: Because there was no corresponding themeVariable set for it in the theme.js. I looked up the PR from before, remembered we had a discussion about it: https://github.com/instructure/instructure-ui/pull/328. The fontFamily was deleted from the theme.js then, so this can be deleted too.

matyasf: I see, thanks!

matyasf (approved):

PR #403, opened 2021-02-19
InstUI-2877 remove unnecessary themeable babel, postcss and or webpack plugins

matyasf:
Removed the following packages because they are no longer needed when using emotion:
```
babel-plugin-themeable-styles
postcss-themeable-styles
ui-postcss-config
```

This caused some test to break because `displayName` was set in the themeable plugin in compile time. Luckily it turned out that simply using `name` is fine in the broken cases.   
Also other tests were breaking with mysterious errors. Here the root cause was the `ui-test-sandbox` was doing some *really* weird stuff: 
1. Overriding Mocha's global exception handler (this was not working properly for some reason, maybe Mocha changed its internals?) to ignore some errors that were thrown. 
2. and overriding `console.error` to throw errors when a `console.error` message started with `Warning:`.
These were deleted. 

As a result of deleting (1) one can see some harmless messages in the console when the tests are running
As a result of deleting (2) its no longer a valid strategy to surround `jsx`code with `try..catch` because writing something in the console does not cause errors to be thrown. Use `spy(console, 'error')` instead.


packages/ui-test-sandbox/src/utils/sandbox.js
```diff
@@ -68,16 +68,6 @@ class Sandbox {
...
-    // override mocha's onerror handler
-    if (typeof window.onerror === 'function') {
-      window.onerror = overrideWindowOnError(window.onerror)
-    }
-
-    // for prop-type warnings:
-    if (typeof console.error === 'function') {
-      console.error = overrideConsoleError(console.error)
-    }
-
```
matyasf, line 80: this is the really problematic code

ke1k0 (approved):

Brailor (approved):

PR #404, opened 2021-02-22
INSTUI-2862 - write tests for the new emotion package

ke1k0:


matyasf (approved):
Nice work, its always good to see more tests :)

Brailor (approved):

PR #411, opened 2021-03-03
chore(ui-scripts): add tags to bump script (INSTUI-2881)

HerrTopi:
Closes: INSTUI-2881

### DIFF
--- a/packages/ui-scripts/lib/bump.js
+++ b/packages/ui-scripts/lib/bump.js
@@ -70,7 +70,9 @@ async function bump(
     process.exit(1)
   }
 
-  info(`💾  Tagging release commit...`)
+  info(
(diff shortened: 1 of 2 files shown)


Brailor (approved):

packages/ui-scripts/lib/bump.js
```diff
@@ -66,5 +70,13 @@ async function bump(
     process.exit(1)
   }
 
+  info(`💾  Tagging release commit...`)
```
matyasf, line 73: I'd add here some text that the tag needs to be pushed too with e.g. `git push --follow-tags`

matyasf (approved):

Brailor (approved):

PR #415, opened 2021-03-04
fix(ui-scripts): fix bump tagging and rename prerelease builds from n…

HerrTopi:
…ightly to snapshot

### DIFF
--- a/packages/ui-scripts/lib/utils/git.js
+++ b/packages/ui-scripts/lib/utils/git.js
@@ -130,7 +130,7 @@ exports.commitVersionBump = function commitVersionBump(releaseVersion) {
 }
 
 exports.tagCommit = function tagCommit(releaseVersion) {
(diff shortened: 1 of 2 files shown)


packages/ui-scripts/lib/utils/git.js
```diff
@@ -130,7 +130,7 @@ exports.commitVersionBump = function commitVersionBump(releaseVersion) {
 }
 
 exports.tagCommit = function tagCommit(releaseVersion) {
-  runGitCommand(['tag', `v${releaseVersion}`])
+  runGitCommand(['tag', '-a', `v${releaseVersion}`, '-m', `v${releaseVersion}`])
```
matyasf, line 133: I'd add a comment here why we are using the -a and -m flags. Its not trivial that omitting these breaks Lerna :)

Brailor (dismissed):

PR #421, opened 2021-03-08
refactor(ui-docs-client): add Netlify badge, remove newsletter link from the main page

matyasf:


ke1k0:
Not sure it is connected to this change, but noticed when I wanted to check this: at a specific window size the page kinda starts to break
image

packages/ui-docs-client/src/Hero/index.js
```diff
@@ -285,6 +285,17 @@ class Hero extends Component {
             </Link>
           </List.Item>
         </List>
+        <Heading as="h3" level="h2" margin="none none medium">
```
ke1k0, line 288: This block is cool. Should we add other badges? So lonely now, and we do use other products (e.g. Chromatic?)

ke1k0 (approved):

HerrTopi (approved):

PR #422, opened 2021-03-08
chore: update release docs and clear up the release workflow (INSTUI-2893)

HerrTopi:
Closes: INSTUI-2893


docs/getting-started/releases.md
```diff
@@ -9,80 +9,67 @@ order: 8
...
+1. Verify that the [documentation was updated](https://instructure.design/).
+
+### Patching Older Releases
 
-1. Run `git fetch --all --tags`.
-1. Run `git checkout -B your_local_branch --track origin/master`.
-1. Run `git merge origin/stable`
-1. Push the merge commit for review: `git push origin HEAD:refs/for/master`
-1. Review the merge commit, test it and merge it.
+0. You need to branch off from the **last release commit** of the old version and name it v< major-version> -maintenance, if it is non existent. E.g. if you'd like to start v6 maintenance branch:
```
matyasf, line 40:
```suggestion
0. You need to branch off from the **last release commit** of the old version and name it `v< major-version>-maintenance`, if it is non existent. E.g. if you'd like to start v6 maintenance branch:
```

matyasf (dismissed):
nice work!

docs/getting-started/releases.md
```diff
@@ -9,80 +9,67 @@ order: 8
...
+git checkout 35258f25cad1e17e5dbe5dcb4389b5185d709f72
+
+//create new branch from it
+git checkout -b v6-maintenance
+```
+
+1.  Check out the v< major-version >-maintenance branch. E.g. for v7:
+
+```
+git checkout v7-maintenance
```
ke1k0, line 53: It is a bit confusing for me that first you use v6 as an example, then v7 in the next. Which one do I need to checkout now? Both?

HerrTopi: Fixed wording

docs/getting-started/releases.md
HerrTopi: Yeah, it looks better that way :)

matyasf (approved):

PR #428, opened 2021-03-11
INSTUI-2863 -  stop tooltip from "jumping" on show

ke1k0:
Closes: INSTUI-2863

After the "themeable to emotion" migration of the tooltip component it started jumping on show. The
fix was found in `ui-position`: we should hide the component offscreen until it has a non-zero
height and width to measure for the positioning.


matyasf (approved):
nice work!

Brailor (approved):

PR #430, opened 2021-03-11
fix: Add CommonJS build to @instructure/ui-themes

rojlarge:
Looks like `@instructure/ui-themes` is missing a CommonJS build, which Canvas needs since it requires in those themes in several places where we run Jest tests.


matyasf (approved):
Thanks for the PR!

PR #439, opened 2021-03-18
INSTUI-2908 - fix html tag based styling

ke1k0:
During the Emotion Migration we used a workaround to style items with selectors strengthened with html tag selectors, eg.: `input.root[type]`.

This turned out to be a bad decision, because most of our components accept an "as" property (if you want to render the component "as button" instead of the default "as a" for example). If you pass a complex component instead of a string to the "as" prop though, the new solution didn't apply the styles, because the elementType was a function instead of a valid html tag name.

Refactored the code to use the css `&:is(input)` selectors, because they apply as css classes once the DOM is fully rendered and the component is boiled down to the final html element.

### DIFF
--- a/packages/ui-text/src/Text/styles.js
+++ b/packages/ui-text/src/Text/styles.js
@@ -41,8 +41,7 @@ const generateStyle = (componentTheme, props, state) => {
     transform,
     lineHeight,
     letterSpacing,
-    color,
-    as
+    color
   } = props
 
   const colorVariants = {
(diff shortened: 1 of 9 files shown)


packages/ui-text/src/Text/styles.js
```diff
@@ -87,11 +87,11 @@ const generateStyle = (componentTheme, props, state) => {
 
   const letterSpacingVariants = {
     normal: componentTheme.letterSpacingNormal,
-    condensed: componentTheme.letterSpacingNormal,
-    expanded: componentTheme.letterSpacingNormal
+    condensed: componentTheme.letterSpacingCondensed,
+    expanded: componentTheme.letterSpacingExpanded
```
ke1k0, line 91: This is an additional bugfix I found, all letter spacing values were set to normal. Will need to accept this as a new baseline in VRT too.

matyasf (approved):
Looks good, lets try it out!

HerrTopi (approved):

PR #442, opened 2021-03-22
Fix "_interopRequireDefault is not a function" bug

matyasf:
We need at least Babel 7.13.10 because of https://github.com/babel/babel/issues/12854 and at least 7.12 because of https://github.com/babel/babel/pull/10853

TEST PLAN:
run test, run doc app, run SIStemic tests via Verdaccio


package.json
```diff
@@ -67,8 +67,6 @@
   },
   "license": "MIT",
   "resolutions": {
-    "@babel/core": "^7.12.10",
-    "@babel/runtime": "^7.13.10",
```
matyasf, line 71: I think its a bad idea to push these here. Only we are using the main package.json and this can mask issues like this.

packages/ui-babel-preset/package.json
```diff
@@ -28,10 +28,10 @@
     "@babel/plugin-transform-destructuring": "^7.9.5",
     "@babel/plugin-transform-react-constant-elements": "^7.9.0",
     "@babel/plugin-transform-react-display-name": "^7.8.3",
-    "@babel/plugin-transform-runtime": "^7.9.0",
```
matyasf, line 31: according to https://babeljs.io/docs/en/babel-runtime these 2 are used together, It sounds like that we should upgrade them together.

HerrTopi (approved):

Brailor (approved):

PR #450, opened 2021-03-25
feat(instui-cli,ui-upgrade-scripts): add codemod to rename theme prop to themeOverride

matyasf:


packages/instui-cli/lib/commands/codemod-v8.js
```diff
@@ -0,0 +1,99 @@
...
+    describe:
+      'One or multiple glob path patterns for files/directories that will be ignored when the codemods are applied (ex. **/node_modules/**).',
+    default: ['**/node_modules/**']
+  })
+
+  yargs.option('parser', {
+    type: 'string',
+    describe: 'jscodeshift `parser` argument for parsing source files.',
+    choices: ['babel', 'babylon', 'flow', 'ts', 'tsx'],
+    default: 'babylon'
```
Brailor, line 52: The default parser should be `flow` I think.

packages/ui-upgrade-scripts/lib/utils/execute-codemod.js
```diff
@@ -32,20 +32,20 @@ module.exports = ({
...
   parserConfig,
-  isMetaComponentPackageMigration = false
+  isMetaComponentPackageMigration = false,
+  printToStdOut = false,
+  dryRun = false
 } = {}) => {
   info(`Running ${codemodPath}...`)
   info(`Source path: ${sourcePath}`)
-  info(`Config path: ${configPath}`)
+  if (configPath) info(`Config path: ${configPath}`)
```
Brailor, line 41: The content of this if block can be moved under line 44.

packages/instui-cli/lib/commands/codemod-v8.js
HerrTopi: Indeed, the code is tested with `flow`

packages/instui-cli/lib/commands/codemod-v8.js
```diff
@@ -0,0 +1,99 @@
...
+  })
+
+  yargs.option('isDryRun', {
+    alias: 'd',
+    type: 'boolean',
+    describe: 'Whether to make a dry run that does no changes.',
+    default: false
+  })
+
```
HerrTopi, line 61: In hindsight, maybe we should set the flag -  which is responsible for the dry/wet run setting - default dry. We could spare our users of accidental changing runs.

packages/instui-cli/lib/commands/codemod-v8.js
```diff
@@ -0,0 +1,99 @@
...
+  yargs.option('codemodType', {
+    alias: 't',
+    type: 'string',
+    describe:
+      'The codemod to apply. "themeOverride" renames components theme={} prop to themeOverride={}',
+    choices: ['themeOverride'],
+    default: 'themeOverride'
```
HerrTopi, line 68: To be precise, we should't provide defaults for this field. The user should be able to set this one consciously

packages/ui-upgrade-scripts/lib/utils/execute-codemod.js
HerrTopi: Yeah, true, but maybe keeping it here helps the readability of the code ( all info is in one place)

packages/instui-cli/lib/commands/codemod-v8.js
matyasf: hmm I think its fine this way, its in line on how all the other codemods behave. Also I guess people will check what the codemod did before pushing changes..

packages/instui-cli/lib/commands/codemod-v8.js
matyasf: true, fixed

HerrTopi (approved):

Brailor (approved):

PR #462, opened 2021-04-03
refactor(instui-config): refactor docs client, so it does not use Webpack plugins

matyasf:


Brailor (commented):
Looks nice!
I noticed 2 things apart from my review comments:

1. `yarn dev` isn't working in a completely clean project (eg.: the `__docs__/buildScripts/build-docs.js` file has not been executed, thus no `docs-data.json` is present and the documentation won't load at all) because it will call the `start:watch` script and the `prestart` script won't be excecuted inside the `__docs__` package
2. `yarn build:docs` isn't working either because of the same reasons as above, the `prestart` script won't be called and the json will be missing

Brailor (approved):

Brailor (approved):

ke1k0 (approved):
Has a conflict with master, but otherwise approved. :)

PR #469, opened 2021-04-07
INSTUI-2956 - backport fix to remove line extending into root icon

ke1k0:
Closes: INSTUI-2956

Backporting fix from PR #460: Tweaked the vertical line of "large" sized collections not to extend
into the collection icon.
Also tweaked the "small" sized collection to have the line ending closer
to the icon.
Added a size selector in the first example of TreeBrowser to be able to preview all
available sizes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,6 @@
     ],
     "*.css": [
       "stylelint",
-      "prettier --write",
       "git add"
     ],
     "*.{json,jsx,md,mdx,html}": [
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -109,7 +109,6 @@
     ],
     "*.css": [
       "stylelint",
-      "prettier --write",
```
ke1k0, line 112: Had to remove this, because it was linting the css files in the pre-commit, causing them to fail on stylelint in the pre-push check.

matyasf (approved):

Brailor (approved):

PR #470, opened 2021-04-08
refactor: refactor `addResizeListener` usage to ResizeObserver

Brailor:
Closes: INSTUI-2875

Refactor `addResizeListener` to the native DOM ResizeObserver API in all places.
Created v8
deprecation guide.
Add info about deprecation to the `addResizeListener` doc page.
TEST PLAN:
Test the affected components.

### DIFF
--- a/docs/guides/v8-deprecations.md
+++ b/docs/guides/v8-deprecations.md
@@ -1,0 +1,30 @@
+---
+title: Deprecated Utilites, Properties and Components in Version 8.0
+category: Guides
+order: 2
+---
+
+# Deprecated Utilites, Properties and Components in Version 8.0
+
+```javascript
+---
+embed: true
+---
+<ToggleBlockquote
+  summary="Important"
+>
+  <ToggleBlockquote.Paragraph>
+    The following utilities, properties and components are deprecated in v8, and <strong>will be permanently deleted in v9.0.</strong>
+  </ToggleBlockquote.Paragraph>
+</ToggleBlockquote>
+```
+
+The tables below show what will be removed and what are they replaced with. We also marked if there are [codemods](#ui-codemods) available.
+
+- [Deprecated Utilities](#v8-deprecations/#deprecated-utilites)
+
+### Deprecated Utilities
+
+| Utility                                 | Substitute                                                                                                |
+| --------------------------------------- | --------------------------------------------------------------------------------------------------------- |
+| [addResizeListener](#addResizeListener) | Use tha native [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) utility. |
--- a/docs/guides/v8-deprecations.md
+++ b/docs/guides/v8-deprecations.md
@@ -25,6 +25,6 @@ The tables below show what will be removed and what are they replaced with. We a
 
 ### Deprecated Utilities
 
-| Utility                                 | Substitute                                                                                                |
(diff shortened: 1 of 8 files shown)


packages/ui-truncate-text/src/TruncateText/index.js
```diff
@@ -130,21 +130,34 @@ class TruncateText extends Component {
...
+      this._debounced = debounce(this.update, this.props.debounce, {
+        leading: true,
+        trailing: true
+      })
+
+      const { width: origWidth } = getBoundingClientRect(this._ref)
+      this._resizeListener = new ResizeObserver((entries) => {
+        // requestAnimationFrame call is needed becuase some truncatetext test cases
+        // failed due to ResizeObserver was not able to deliver all observations within a single animation frame
+        // see: https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded
```
matyasf, line 142: wont this cause some issue? In the SO discussion they mention something that this causes it to run in the next frame https://stackoverflow.com/a/58701523

Brailor: I haven't noticed any issues with this approach.
There is another way to solve this issue though:

```js
// ignore ResizeObserver loop limit exceeded
// this is ok in several scenarios according to 
// https://github.com/WICG/resize-observer/issues/38
before(() => {
  // called before any tests are run
  const e = window.onerror;
  window.onerror = function(err) {
    if(err === 'ResizeObserver loop limit exceeded') {
      console.warn('Ignored: ResizeObserver loop limit exceeded');
      return false;
    } else {
      return e(...arguments);
    }
  }
});
```

With this we we would simply ignore the error, I'm not sure this one is a better solution, what do you think?

HerrTopi (approved):
It seems legit, nice work

docs/guides/v8-deprecations.md
```diff
@@ -0,0 +1,30 @@
...
+
+The tables below show what will be removed and what are they replaced with. We also marked if there are [codemods](#ui-codemods) available.
+
+- [Deprecated Utilities](#v8-deprecations/#deprecated-utilites)
+
+### Deprecated Utilities
+
+| Utility                                 | Substitute                                                                                                |
+| --------------------------------------- | --------------------------------------------------------------------------------------------------------- |
+| [addResizeListener](#addResizeListener) | Use tha native [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) utility. |
```
ke1k0, line 30: Any chance we can add an example how to convert a code snippet from the old usage to the native one?

packages/ui-responsive/src/addElementQueryMatchListener.js
```diff
@@ -37,9 +33,7 @@ import { debounce } from '@instructure/debounce'
  * Given an object of named queries, listens for changes in the
  * element size and notifies which queries match via a function
  * callback. The callback method is only called when the query
- * matches change, not on all element resizes. (If you are looking
- * to call a method on all element resizes see
- * [addResizeListener](#addResizeListener) instead)
```
ke1k0, line 42: ...use ResizeObserver instead?

packages/ui-truncate-text/src/TruncateText/index.js
matyasf: hmm I like the current one better, lets keep it

matyasf (approved):

ke1k0 (approved):

PR #476, opened 2021-04-12
INSTUI-2964 - list root items too when `showRootCollection` is `false`

ke1k0:
Closes: INSTUI-2964

Fixed the error when the items of the root collection are not rendered when `showRootCollection` is
set to `false`. Added an interactive example on the docs page to show how `showRootCollection`
works.

TEST PLAN:
Root items are rendered too when `showRootCollection` is set to false. The `aria-level` values are
correct on every level.

### DIFF
--- a/packages/ui-tree-browser/src/TreeBrowser/index.js
+++ b/packages/ui-tree-browser/src/TreeBrowser/index.js
@@ -202,14 +202,10 @@ class TreeBrowser extends Component {
   }
 
   get collections() {
-    const { collections, rootId, showRootCollection } = this.props
+    const { collections, rootId } = this.props
 
-    if (typeof rootId !== 'undefined' && showRootCollection) {
+    if (typeof rootId !== 'undefined') {
       return [collections[rootId]]
-    } else if (typeof rootId !== 'undefined') {
-      return collections[rootId].collections
-        .map((id) => collections[id])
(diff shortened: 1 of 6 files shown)


packages/ui-tree-browser/src/TreeBrowser/index.js
```diff
@@ -358,6 +354,11 @@ class TreeBrowser extends Component {
     }
   }
 
+  getIsFlattened = (collection) => {
+    const { rootId, showRootCollection } = this.props
+    return !showRootCollection && rootId && collection.id === rootId
```
matyasf, line 359: I'd put here a warning if showRootCollection is true, but rootId is null/NaN

matyasf (approved):

packages/ui-tree-browser/src/TreeBrowser/index.js
ke1k0: That is a valid case too, actually the default case... Renders all collections at the top level.

prenner:
Are there any plans to backport this into instui 7? or is the plan to keep this with instui 8? 

Thank you for working on this, by the way.

Brailor (approved):

ke1k0:
@prenner yes, we will backport this to v7 :)

PR #480, opened 2021-04-15
INSTUI-2965 - list root items too when `showRootCollection` is `false`

ke1k0:
Closes: INSTUI-2965

Backport of fix on v8 (INSTUI-2964, PR #476). Fixed the error when the items of the root collection are not rendered when `showRootCollection` is
set to `false`. Added an interactive example on the docs page to show how `showRootCollection`
works.

TEST PLAN:
Root items are rendered too when `showRootCollection` is set to false. The `aria-level` values are
correct on every level.


matyasf (approved):
looks good!

Brailor (approved):

PR #491, opened 2021-04-19
TypeScript foundations

Brailor:
This PR lays down the foundations of the migration process of the source code from JS -> TS, it consists:

- building types is now done by `tsc` (typescript compiler) instead of from parsing PropTypes
- type checking is also done by `tsc` and it is now part of our `bootstrap` pipeline
- added `tsconfig.json`, which is mainly for the IDE's to be able to navigate nicely between packages
- added `tsconfig.build.json`, which is for configuring the TS compiler (how loose the type checking should be)
- each package got their own `tsconfig.json` and `tsconfig.build.json` respectively, this enables us to fine tune the TS config for any package
- every package got a new script called `ts:check` which will only do the type checking for that package without any type output (you can run this script from the root as well, and `lerna` will execute it for each package)
- the babel preset is extended with the [official babel typescript preset](https://babeljs.io/docs/en/babel-preset-typescript), so babel will understand and will be able to transform TS code (without type checking)
- this preset is also used by the `babel-loader`, so our `webpack` pipeline is effectively untouched and working as expected


packages/__docs__/package.json
```diff
@@ -25,106 +25,106 @@
   "license": "MIT",
```
matyasf, line 25: this file has no changes, right? In this case I'd revert this

matyasf (approved):

HerrTopi (approved):
NIce work :)

ke1k0 (approved):

PR #495, opened 2021-04-20
Configure ESLint to handle TS

Brailor:


matyasf (approved):
nice work!

packages/ui-avatar/.eslintignore
```diff
@@ -1,3 +1,5 @@
 /node_modules/**
 /lib/**
 /es/**
+/types/**
+/types/**
```
ke1k0, line 5: duplication :)

ke1k0 (approved):
there is one small duplication, otherwise works nice :)

packages/ui-avatar/.eslintignore
Brailor: You are right, removed

PR #509, opened 2021-04-23
TS initial file migration

Brailor:


packages/ui-tray/src/Tray/index.tsx
```diff
@@ -209,23 +246,29 @@ class Tray extends Component {
...
+  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'prevProps' implicitly has an 'any' type... Remove this comment to see the full error message
   componentDidUpdate(prevProps, prevState, snapshot) {
     if (this.props.open !== prevProps.open) {
       this.setState({ transitioning: true })
     }
+    // @ts-expect-error ts-migrate(2722) FIXME: Cannot invoke an object which is possibly 'undefin... Remove this comment to see the full error message
     this.props.makeStyles()
   }
 
+  // @ts-expect-error ts-migrate(1056) FIXME: Accessors are only available when targeting ECMASc... Remove this comment to see the full error message
```
matyasf, line 262: these are unnecessary error messages, we should be targeting browsers that support ES6

Brailor: You are indeed correct.
I set the `target` field in our `tsconfig.build.json` to `ES6` and removed all of the related `@ts-expect-error` directives.

ke1k0 (approved):

HerrTopi (approved):
YOLO

PR #520, opened 2021-04-28
chore(emotion,ui-tree-browser): upgrade to emotion 11.0.0 (INSTUI-2960)

HerrTopi:
Closes: INSTUI-2960

Fix keyframe syntax error in TreeButton and TreeCollection

### DIFF
--- a/packages/emotion/package.json
+++ b/packages/emotion/package.json
@@ -25,7 +25,7 @@
   "license": "MIT",
   "dependencies": {
     "@babel/runtime": "^7.13.10",
-    "@emotion/core": "^10.0.35",
+    "@emotion/react": "11.0.0",
     "@instructure/console": "^8.2.1",
     "@instructure/ui-decorator": "^8.2.1",
     "@instructure/ui-i18n": "^8.2.1",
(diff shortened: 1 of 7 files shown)


matyasf (approved):

packages/emotion/package.json
```diff
@@ -25,7 +25,7 @@
   "license": "MIT",
   "dependencies": {
     "@babel/runtime": "^7.13.10",
-    "@emotion/core": "^10.0.35",
+    "@emotion/react": "11.0.0",
```
Brailor, line 28: I think we should use the `^` syntax so we are getting the features and bugfixes in version 11.

HerrTopi: Since we plan to do constant version bumps, we should know explicitly the version we are on to avoid unnecessary complications

Brailor: I don't see how automatically getting the minor and patch versions for a given package would complicate anything with our constant version bumps, on the contrary, I think it would make our lives easier since then we would not need to have to update the minor and patch versions manually every time we do this reoccuring update task.

Brailor (approved):

PR #526, opened 2021-05-04
chore(ui-scripts): fix publish script (INSTUI-3011)

HerrTopi:


.github/workflows/manual-deploy-to-gh-pages.yml
```diff
@@ -1,4 +1,4 @@
-name: Deploy manually to gh-pages
+name: Docs deploy latest
```
ke1k0, line 1: If we are renaming anyway, how do you like this:
- npm release
- docs deploy
- legacy npm release
- legacy docs deploy
- AUTO npm release and docs deploy
- AUTO Visual Regression Test

This way it is always prefixed and grouped by the category

.github/workflows/release.yml
```diff
@@ -19,7 +19,17 @@ jobs:
         run: yarn bootstrap
       - name: Run tests.
         run: yarn test:patchset
-      - name: Release to NPM
+      - name: Release snapshot to NPM
+        # lerna ususally fails in releasing, but releasing snapshots is too complicated
+        # without it, so we use lerna for snapshot and native npm-cli commands for releases
```
ke1k0, line 24: I think the point of this comment would be to explain the "is too complicated" part, so we remember in few months/years what the exact issue was here, saving a few hours of (re)investigating. Could you please write a more detailed explanation?

packages/ui-scripts/lib/publish-latest.js
```diff
@@ -41,14 +41,14 @@ const { createNPMRCFile } = require('./utils/npm')
 try {
   const pkgJSON = getPackageJSON()
   // Arguments
-  // 1: version to publish. If current version, use 'current' or do not pass any arguments otherwise e.g.: 8.1.3
+  // 1: version to publish. If current version, use 'current' or don't bass anything. Otherwise e.g.: 8.1.3
```
ke1k0, line 44:
```suggestion
  // 1: version to publish. If current version, use 'current' or don't pass anything. Otherwise e.g.: 8.1.3
```

packages/ui-scripts/lib/publish-snapshot.js
```diff
@@ -0,0 +1,81 @@
...
+    currentVersion: pkgJSON.version,
+    packageConfig: getConfig(pkgJSON),
+    releaseType: process.argv[3]
+  })
+} catch (err) {
+  error(err)
+  process.exit(1)
+}
+
+async function publishLegacy(options) {
```
ke1k0, line 47: "publishLegacy"? Shouldn't this be the "publishSnapshot"? Also, if this only publishing the snapshots, shouldn't we simplify the code? Im not sure that checking for release commits and maintenance makes sense here. Or do we need 2 scripts in the first place? Can't we just have the if-else in the publish script?

.github/workflows/manual-deploy-to-gh-pages.yml
HerrTopi: More consistent, I like it

.github/workflows/release.yml
HerrTopi: Added more explanation to publish.js

packages/ui-scripts/lib/publish-latest.js
HerrTopi: Fixed typo

packages/ui-scripts/lib/publish-snapshot.js
HerrTopi: It is a good idea. I refactored the whole stuff

Brailor (approved):

ke1k0 (approved):

PR #530, opened 2021-05-07
docs: fix wrong binaries served for the legacy docs codepens

Brailor:
Closes: INSTUI-3009


ke1k0 (approved):
Works fine :) We just need to remember to change this too when we work on multiple version documentation

matyasf (approved):

PR #531, opened 2021-05-07
INSTUI-3013 : fix: documentation app was not reloading on changes

matyasf:
Also the EsModules parameter was not passed correctly thus not allowing tree shaking


ke1k0 (approved):
Works fine :)

Brailor (approved):

PR #533, opened 2021-05-07
chore: remove ts related comments from built code

Brailor:
Closes: INSTUI-3004

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -23,6 +23,11 @@
  */
 
 module.exports = {
+  shouldPrintComment(value) {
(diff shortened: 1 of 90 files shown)


matyasf:
Not a very elegant solution, IMO we should remove it when we finished the TypeScript migration

matyasf (approved):

ke1k0 (approved):
Not very elegant, indeed :D But when we do the TS typing package-by-package, and there are no more fixme errors in that package, it should be a step of migration to remove this block from the babel config :)

PR #536, opened 2021-05-10
docs(ui-select): fix multiselect example (INSTUI-3019)

HerrTopi:
Closes: INSTUI-3019

### DIFF
--- a/packages/ui-select/src/Select/README.md
+++ b/packages/ui-select/src/Select/README.md
@@ -3,6 +3,7 @@ describes: Select
 ---
 
 `Select` is an accessible, custom styled combobox component for inputting a variety of data types.
+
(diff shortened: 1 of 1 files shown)


packages/ui-select/src/Select/README.md
```diff
@@ -422,9 +426,7 @@ class MultipleSelectExample extends React.Component {
 
   filterOptions = (value) => {
     const { selectedOptionId } = this.state
-    return this.props.options.filter(option => (
-      selectedOptionId.indexOf(option.id) === -1 // ignore selected options removed from list
```
matyasf, line 426: its strange that they even left a comment here for this behaviour

ke1k0 (approved):

PR #538, opened 2021-05-11
INSTUI-3017 - fix page scrolling up when Menu opens

ke1k0:
refs: INSTUI-3017

If a Dialog contains a focusable item, that item gets focused when the dialog opens. The browsers
automatically scroll the the focused element, but if the content and its css is not yet fully
loaded, the position of the focusable item might be not yet calculated, making the page jump to the
top. Now we wait a heartbeat before focusing the element. This fix affects other components using
`Dialog` under the hood: `DrawerLayout`, `Modal`, `Overlay`, `Popover` and `Tray`, and components
using these (e.g. Menu).

### DIFF
--- a/packages/ui-dialog/src/Dialog/index.js
+++ b/packages/ui-dialog/src/Dialog/index.js
@@ -149,9 +149,19 @@ class Dialog extends Component {
     } = this.props
 
     this._raf.push(requestAnimationFrame(() => {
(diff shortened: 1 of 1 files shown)


HerrTopi (approved):
Cant we use componentDidMount for this?

matyasf (approved):

PR #543, opened 2021-05-12
INSTUI-3103 - resolve mocha type errors in tests

ke1k0:
Closes: INSTUI-3103

### DIFF
--- a/packages/ui-a11y-content/package.json
+++ b/packages/ui-a11y-content/package.json
@@ -36,7 +36,8 @@
   },
   "devDependencies": {
     "@instructure/ui-babel-preset": "8.4.0",
-    "@instructure/ui-test-utils": "8.4.0"
+    "@instructure/ui-test-utils": "8.4.0",
+    "@types/mocha": "^8.2.2"
   },
   "peerDependencies": {
     "react": "^16.8"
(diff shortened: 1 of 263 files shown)


Brailor (commented):
I don't think we have to add `@types/mocha` to every package since common dependencies are hoisted to the root `node_modules` by lerna. Thus I would add the `@types/mocha` devDependency to the package that actually has a dependency on mocha (`ui-scripts`), or put in the root package.json.

matyasf (approved):

matyasf:
> Thus I would add the @types/mocha devDependency to the package that actually has a dependency on mocha (ui-scripts), or put in the root package.json.

But if we put them into e.g. the root JSON wont it throw errors for people who use just one package from us?

Brailor:
> > Thus I would add the @types/mocha devDependency to the package that actually has a dependency on mocha (ui-scripts), or put in the root package.json.
> 
> But if we put them into e.g. the root JSON wont it throw errors for people who use just one package from us?

I don't think so, because they can't run our tests since they are not published to npm.

ke1k0:
> > > Thus I would add the @types/mocha devDependency to the package that actually has a dependency on mocha (ui-scripts), or put in the root package.json.
> > 
> > 
> > But if we put them into e.g. the root JSON wont it throw errors for people who use just one package from us?
> 
> I don't think so, because they can't run our tests since they are not published to npm.

Fixed, thanks :)

Brailor (approved):

PR #546, opened 2021-05-13
fix(emotion): add componentId to match with component name in docs (INSTUI-2997)

HerrTopi:
ui-a11y-content,ui-alerts,ui-avatar,ui-badge,ui-billboard,ui-breadcrumb,ui-buttons,ui-byline,ui-calendar,ui-checkbox,ui-code-editor,ui-date-input,ui-dialog,ui-drawer-layout,ui-editable,ui-file-drop,ui-flex,ui-form-field,ui-grid,ui-heading,ui-img,ui-link,ui-list,ui-menu,ui-metric,ui-modal,ui-motion,ui-navigation,ui-number-input,ui-options,ui-overlays,ui-pages,ui-pagination,ui-pill,ui-popover,ui-position,ui-progress,ui-radio-input,ui-range-input,ui-rating,ui-select,ui-simple-select,ui-spinner,ui-svg-images,ui-table,ui-tabs,ui-tag,ui-text-area,ui-text-input,ui-text,ui-time-select,ui-toggle-details,ui-tooltip,ui-tray,ui-tree-browser,ui-truncate-text,ui-view

### DIFF
--- a/packages/emotion/src/EmotionThemeProvider/README.md
+++ b/packages/emotion/src/EmotionThemeProvider/README.md
@@ -130,8 +130,11 @@ The `componentOverrides` can also be nested inside `themeOverrides`.
             infoIconBackground: "darkblue",
             infoBorderColor: "darkblue"
           },
-          'List.Item': {
+          [List.Item.componentId]: {
             color: "red"
+          },
+          'InlineList.Item': {
+            color: "blue"
           }
         },
         themeOverrides: {
@@ -165,6 +168,14 @@ The `componentOverrides` can also be nested inside `themeOverrides`.
         <List.Item>These List.Items have red color.</List.Item>
         <List.Item>These List.Items have red color.</List.Item>
       </List>
+      <div>
+        <InlineList delimiter="pipe" margin="large 0">
+          <InlineList.Item>This text should be blue</InlineList.Item>
+          <InlineList.Item>10pts</InlineList.Item>
+          <InlineList.Item><b>Due:</b> Oct 1, 2019</InlineList.Item>
+          <InlineList.Item><Link href="#">Pipe Separator</Link></InlineList.Item>
+        </InlineList>
+      </div>
     </EmotionThemeProvider>
   </div>
 ```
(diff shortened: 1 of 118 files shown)


Brailor:
Shouldn't we document this usage of the API in the docs somewhere?

packages/emotion/src/withStyle.tsx
```diff
@@ -111,7 +112,7 @@ const withStyle = decorator(
       }
       const themeOverride = getComponentThemeOverride(
         theme,
-        displayName,
+        [displayName, ComposedComponent.componentId],
```
Brailor, line 115: At this point can't we jus set `displayName` to be `ComposedComponent.componentId`?
Whats the point of keeping the `.displayName` and `.name` if we are not even sure they will contain any legit info about the component?

If we only use the `.componentId` then we don't have to do these kind of ugly checks in the `getComponentThemeOverride` function.

ke1k0:
We should add this to the template-app and template-component too.

packages/emotion/src/withStyle.tsx
ke1k0: I think it should be setting the displayName to `ComposedComponent.componentId || ComposedComponent.displayName`, just in case it is missing from somewhere, but primarily using the componentId, yes.

Also, the static is missing at: Calendar.Day, InPlaceEdit, InlineListItem, InlineSVG, SVGIcon (all components with @makeStyle should have it, maybe even the others too, in case we want to use the componentId for something else too.)

ke1k0:
> We should add this to the template-app and template-component too.

We agreed not to add this in the templates, but could you please write an "Parent.Child" example to the EmotionThemeProvider readme and the themeable-emotion migration guide? Actually, adding a test would be nice too in EmotionThemeProvider.

ke1k0 (approved):

packages/emotion/src/EmotionThemeProvider/README.md
```diff
@@ -129,12 +129,15 @@ The `componentOverrides` can also be nested inside `themeOverrides`.
           Alert: {
             infoIconBackground: "darkblue",
             infoBorderColor: "darkblue"
+          },
+          'List.Item': {
+            color: "red"
           }
         },
```
Brailor, line 136: I think we should also add an example showcasing how to use the `Component.componentId` usage as well, so for example:
```js
...,
[List.Item.componentId]: {
    color: "green"
},
....
```

Brailor (approved):

PR #548, opened 2021-05-13
feat(ui-number-input): add inputMode prop to NumberInput

ke1k0:
Closes: INSTUI-3016

This will be put on the underlying input field to hint at intended usage

### DIFF
--- a/packages/ui-number-input/src/NumberInput/__examples__/NumberInput.examples.js
+++ b/packages/ui-number-input/src/NumberInput/__examples__/NumberInput.examples.js
@@ -34,7 +34,14 @@ export default {
       renderLabel: 'A number input'
     }
   },
-  excludeProps: ['required', 'readOnly', 'disabled', 'label', 'inline'],
+  excludeProps: [
+    'required',
+    'readOnly',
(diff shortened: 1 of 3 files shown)


packages/ui-number-input/src/NumberInput/__examples__/NumberInput.examples.js
```diff
@@ -34,7 +34,14 @@ export default {
...
     }
   },
-  excludeProps: ['required', 'readOnly', 'disabled', 'label', 'inline'],
+  excludeProps: [
+    'required',
+    'readOnly',
+    'disabled',
+    'label',
+    'inline',
+    'InputMode'
```
Brailor, line 43: This should be camelcase too

Brailor (approved):

HerrTopi (approved):

packages/ui-number-input/src/NumberInput/__examples__/NumberInput.examples.js
ke1k0: oh indeed! fixed, thanks

PR #549, opened 2021-05-14
Convert ui-test-locator, ui-test-sandbox, ui-test-queries, debounce to TypeScript and type them

matyasf:
The goal was to change code to TS and add proper types. I've left weird code and some bugs in


packages/debounce/src/debounce.ts
```diff
@@ -52,37 +58,41 @@
...
-  let lastArgs, lastThis, result, lastCallTime
+function debounce(
+  func: (...args: unknown[]) => unknown,
+  wait = 0,
+  options: DebounceOptions = {}
+) {
+  let lastArgs: unknown[] | undefined
+  let lastThis: unknown
+  let result: unknown
+  let lastCallTime: number | undefined // TODO this should never be undefined
```
matyasf, line 69: This is a weird pattern, we should fix it in the future.

packages/ui-test-queries/src/utils/bindElementToMethods.ts
```diff
@@ -21,16 +21,21 @@
...
-    }
-  }, {})
+
+export type GenericFunction = (...args: any) => unknown | any
+
+export function bindElementToMethods<T, K>(element: T, methods: K) {
+  return Object.entries(methods).reduce(
+    (bound: Record<string, K>, [key, fn]: [string, K]) => {
+      if (typeof fn === 'function') {
+        bound[key] = fn.bind(null, element) // eslint-disable-line no-param-reassign
```
matyasf, line 31: This is why I could not type the added methods properly. This code adds the first parameter to the function automatically, I dont think we can remove it.

packages/ui-test-queries/src/utils/bindElementToUtilities.ts
```diff
@@ -25,32 +25,45 @@ import { fireEvent } from './events'
...
-
   return {
     ...bindElementToMethods(element, queries),
-    ...bindElementToEvents(element, fireEvent),
+    ...(bindElementToEvents(element, fireEvent) as any), // could not type this :/
     ...bindElementToMethods(element, helpers),
     ...bindElementToMethods(element, customMethods)
-  }
+  } as BoundType[] // would be nicer without this cast
 }
```
matyasf, line 67: When this method is called from the outside it always return an Array. The other values (a single element or null) are returned only from deep inside the recursion, so I applied this hacky solution so it looks good from the outside and does not need a large refactoring

packages/ui-test-queries/src/utils/events.ts
```diff
@@ -347,8 +355,14 @@ function fireEvent(element, event) {
 Object.entries(eventMap).forEach(([key, { EventType, defaultInit }]) => {
   const eventName = key.toLowerCase()
 
-  fireEvent[key] = (node, init = {}) => {
+  // not nice, but this is a very weird code :/
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  fireEvent[key] = (node: Element, init: Record<string, unknown> = {}) => {
     const eventInit = { ...defaultInit, ...init }
```
matyasf, line 362: Again, some really weird code (adding additional function properties to a function) that I could not type properly without refactoring

packages/ui-test-queries/src/utils/queries.ts
```diff
@@ -24,58 +24,78 @@
...
   const { element, selector, options } = parseQueryArguments(...args)
   return find(element, `:withTitle("${selector}")`, options)
 }
 
-async function find(...args) {
-  return firstOrNull(await findAll(...args))
+async function find(...args: QueryArguments) {
+  // TODO remove "as any", but this will cause lots of TSC errors
+  // the main reason is that the function params are bound in bindElementToMethods
+  return firstOrNull(await findAll(...args)) as any
```
matyasf, line 60: Sadly adding this here breaks autocomplete totally. But removing the "any" cast from here would make type checking buggy, since methods in 'helpers' and 'queries' have a parameter (the element its added to) that is added automatically with function.bind().
I dont think we can solve this without large refactorings

packages/ui-test-queries/src/utils/queryResult.ts
```diff
@@ -74,21 +79,34 @@ async function _getQueryResult(
   }
 }
 
-function wrapQueryResult(...args) {
-  return bindElementToUtilities(...args)
+// TODO remove any return type
+function wrapQueryResult<T>(
+  results: T[],
+  customMethods?: Record<string, GenericFunction>
+): any {
```
matyasf, line 86: again, this breaks type autocomplete, because it'd be buggy because of the weird bind

packages/ui-test-queries/src/utils/firstOrNull.ts
```diff
@@ -21,6 +21,7 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-export function firstOrNull(result) {
+
+export function firstOrNull<T>(result: T | T[] | null): T | null {
   return Array.isArray(result) ? result[0] || null : null
```
matyasf, line 26: another code that when properly typed would just result in tons of errors because of the missing null checks. If we refactor this we should do something like 'firstOrError'

packages/ui-test-queries/src/utils/selectors.ts
```diff
@@ -36,51 +36,72 @@ import {
...
 }
 
-function querySelector(...args) {
+function querySelector(...args: QueryArguments) {
   return firstOrNull(querySelectorAll(...args))
 }
 
-function querySelectorParents(element, selector, options) {
-  let results = []
+function querySelectorParents( // TODO type these to QueryFunction
```
matyasf, line 66: I could not figure out how to make a function written this way to implement a type.

HerrTopi (approved):

packages/ui-test-queries/src/utils/bindElementToUtilities.ts
```diff
@@ -0,0 +1,83 @@
...
+type CutFirstArg<F> = F extends (_: any, ...tail: infer TT) => infer R
+  ? (...args: TT) => R
+  : never
+
+type QueryTypes = {
+  [K in Extract<keyof typeof queries, string>]: typeof queries[K]
+}
+type HelperTypes = {
+  [K in Extract<keyof typeof helpers, string>]: CutFirstArg<typeof helpers[K]>
+}
```
matyasf, line 50: this was the tricky part :)

packages/ui-test-locator/src/utils/locator.ts
```diff
@@ -51,23 +62,26 @@ export function locator(containerSelector, customMethods = {}) {
     })
   }
 
-  const find = async (...args) => {
-    return firstOrNull(await findAll(...args))
+  const find = async (...args: QueryArguments) => {
+    // TODO remove "as any". This will result in lots of ts typing errors
+    return firstOrNull(await findAll(...args)) as any
   }
```
matyasf, line 68: Fixing this will solce one of the last typing issues, namely that XYZLocator-s are typed to "any". I did not do in this PR because it would bloat it even more

Brailor (approved):
Looking good, great work!

PR #552, opened 2021-05-18
chore: add netlify docs deploy workflow

Brailor:


.github/workflows/manual-deploy-docs.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Install packages.
+        run: yarn install
+      - name: Set up project.
+        run: yarn bootstrap
+      - name: Build docs-app.
+        run: yarn build:docs
+      - name: Deploy docs-app to netlify
+        env:
+          NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_AUTH_TOKEN }}
+          NETLIFY_SITE_ID: ${{ secrets.NETLIFY_SITE_ID }}
```
matyasf, line 30: Shouldn't these be added to `env.example`?

HerrTopi (approved):

.github/workflows/manual-deploy-docs.yml
```diff
@@ -0,0 +1,31 @@
+name: Docs deploy to Netlify
+on:
+  workflow_dispatch:
+    inputs:
+      docsVersion:
+        description: "Version of the Docs to release. Must be in format: vX eg.: v6 v7 ..."
+        required: true
+        default: "v7"
```
ke1k0, line 8: shouldn't this be "v6" on this branch?

.github/workflows/manual-deploy-docs.yml
Brailor: These env vars are used only in the CI environment.
If we would like to deploy to netlify from a local environment I'd suggest using the [command line authentication](https://docs.netlify.com/cli/get-started/#authentication) which is much more user friendly than obtaining the auth token from netlify, then assigning it to an env var. (everyone would have to do this, since the auth token is exposed exactly 1 time, upon creating it)

Brailor: If you would like to see it in action, I'm more than happy to show it.

.github/workflows/manual-deploy-docs.yml
Brailor: Doesn't really matter, because when you run this workflow on GitHub it will use the default value from the master workflows file.

matyasf (approved):

PR #556, opened 2021-05-19
INSTUI-2982 : fix storybook endlessly reloading in watch mode

matyasf:
Storybook version 6.2+ has a bug that causes it, pin it to the latest 6.1.x version: https://github.com/storybookjs/storybook/issues/14342


HerrTopi (approved):

ke1k0 (approved):
I would put a comment above these 3 to explain why we pinned the versions, so we don't update it accidentally during our "update dependencies" task, but otherwise works well.

matyasf:
> I might put a comment above these 3 to explain why we pinned the versions, so we don't update it accidentally during our "update dependencies" task, but otherwise works well.

Sadly oy cant put comment in .JSON files :/

ke1k0:
> > I might put a comment above these 3 to explain why we pinned the versions, so we don't update it accidentally during our "update dependencies" task, but otherwise works well.
> 
> Sadly you cant put comment in .JSON files :/

Lol true, then we just need to watch out for it :D

PR #557, opened 2021-05-19
INSTUI-3108 : add Babel assumptions that speed up the code

matyasf:
Add some assumptions that likely speed up the code by assuming that we never use some rare JS
functionality. If we use them it will be a nasty bug!

If you think that one of them shoul not be enabled say so in the review

### DIFF
--- a/packages/ui-babel-preset/lib/index.js
+++ b/packages/ui-babel-preset/lib/index.js
@@ -157,7 +157,11 @@ module.exports = function (
     plugins,
     // see https://babeljs.io/docs/en/assumptions
     assumptions: {
-      setPublicClassFields: true
+      constantReexports: true, //assume that re-exported modules dont change
+      ignoreFunctionLength: true, // assume that we never call function.length
+      noDocumentAll: true, // assume we never use document.all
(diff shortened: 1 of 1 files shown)


packages/ui-babel-preset/lib/index.js
```diff
@@ -157,7 +157,11 @@ module.exports = function (
     plugins,
     // see https://babeljs.io/docs/en/assumptions
     assumptions: {
-      setPublicClassFields: true
+      constantReexports: true, //assume that re-exported modules dont change
+      ignoreFunctionLength: true, // assume that we never call function.length
+      noDocumentAll: true, // assume we never use document.all
+      noNewArrows: true, // assume we never instantiate arrow functions with new
+      setPublicClassFields: true // assume public class fields never shadow getters
```
ke1k0, line 164: Just writing down what we discussed last week:
- We should move this to our own `babel.config.js` files, might be breaking if we apply it in the package.
- Or we could make a new package (e.g. ui-babel-preset-instui) that extends the `ui-babel-preset` package, with the overwrites needed for our components, and the other packages could import this new one.

PR #559, opened 2021-05-19
INSTUI-3115 : Only run makestyles if needed

matyasf:
This is a proof-of concept how this change will look like. I'd like to get your opinion before I apply this change to all of our components.

### DIFF
--- a/packages/emotion/src/withStyle.tsx
+++ b/packages/emotion/src/withStyle.tsx
@@ -32,18 +32,19 @@ import { bidirectionalPolyfill } from './styleUtils/bidirectionalPolyfill'
 import { getComponentThemeOverride } from './getComponentThemeOverride'
 import { useTheme } from './useTheme'
 
-export interface MakeStyleConfig {
-  // The previous prop object that will be compared to the current props
-  prevProps: Record<string, unknown>
-  // An array of property names used in styles.js. The style object will
-  // recalculate if any of these properties change In cannot be used to compare
-  // nested properties.
-  styledProps: string[]
+export type MakeStyleConfig = {
+  updateConfig?: {
+    // The previous prop object that will be compared to the current props
+    prevProps: Record<string, unknown>
+    // An array of property names used in styles.js. The style object will
+    // recalculate if any of these properties change. It cannot be used to compare
+    // nested properties.
+    styledProps: string[]
+  }
+  // Extra properties that will be passed to the `generateStyle` method
+  [K: string]: unknown
 }
-export type MakeStyles = (
-  generateStyleArgs?: Record<string, unknown>,
-  updateConfig?: MakeStyleConfig
-) => void
+export type MakeStyles = (config?: MakeStyleConfig) => void
 
 /**
  * ---
@@ -144,21 +145,17 @@ const withStyle = decorator(
       /**
        * This method generates and sets the styles for the component (it calls
        * the `generateStyle` method in the decorator's parameter).
-       * @param generateStyleArgs An object that will be passed to the
-       *     `generateStyle` method.
-       * @param updateConfig An object that you can use to prevent unnecessary
-       *     style recalculations for performance reasons. If not supplied
-       *     everything will be recalculated.
+       * @param config You can use this object to to pass your own data to the
+       * `generateStyle` method and to prevent unnecessary style recalculations
+       * for performance reasons.
        */
-      const makeStyleHandler: MakeStyles = (
-        generateStyleArgs,
-        updateConfig
-      ) => {
-        if (updateConfig) {
+      const makeStyleHandler: MakeStyles = (config) => {
+        if (config && config.updateConfig) {
           let changed = false
-          for (const propToCheck in updateConfig.styledProps) {
+          for (const propToCheck in config.updateConfig.styledProps) {
             if (
-              updateConfig.prevProps[propToCheck] != componentProps[propToCheck]
+              config.updateConfig.prevProps[propToCheck] !=
+              componentProps[propToCheck]
             ) {
               changed = true
               break
@@ -169,7 +166,7 @@ const withStyle = decorator(
           }
         }
         const calculatedStyles = bidirectionalPolyfill(
-          generateStyle(componentTheme, componentProps, generateStyleArgs),
+          generateStyle(componentTheme, componentProps, config),
           dir
         )
         if (!isEqual(calculatedStyles, styles)) {
--- a/packages/emotion/src/withStyle.tsx
+++ b/packages/emotion/src/withStyle.tsx
@@ -31,6 +31,20 @@ import { useTextDirectionContext } from '@instructure/ui-i18n'
 import { bidirectionalPolyfill } from './styleUtils/bidirectionalPolyfill'
 import { getComponentThemeOverride } from './getComponentThemeOverride'
 import { useTheme } from './useTheme'
+
+export interface MakeStyleConfig {
+  // The previous prop object that will be compared to the current props
+  prevProps: Record<string, unknown>
+  // An array of property names used in styles.js. The style object will
+  // recalculate if any of these properties change In cannot be used to compare
+  // nested properties.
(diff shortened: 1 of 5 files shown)


packages/ui-alerts/src/Alert/index.tsx
```diff
@@ -69,6 +70,7 @@ type Props = {
   onDismiss?: (...args: any[]) => any
   transition?: 'none' | 'fade'
   open?: boolean
+  children?: any
```
matyasf, line 73: I dont know why this was not added with the TS migration

packages/ui-avatar/src/Avatar/index.tsx
```diff
@@ -122,13 +127,18 @@ class Avatar extends Component<Props> {
...
 
-  componentDidUpdate() {
-    // @ts-expect-error ts-migrate(2722) FIXME: Cannot invoke an object which is possibly 'undefin... Remove this comment to see the full error message
-    this.props.makeStyles(this.state)
+  componentDidUpdate(prevProps: Props, prevState: State) {
+    let updateConfig: MakeStyleConfig | undefined
+    if (prevState.loaded == this.state.loaded) {
+      updateConfig = {
+        prevProps: prevProps,
+        styledProps: ['size', 'shape', 'src']
```
matyasf, line 138: this is that is used in `styles.ts`

packages/ui-alerts/src/Alert/index.tsx
```diff
@@ -323,10 +324,11 @@ class Alert extends Component<Props> {
...
 
-  // @ts-expect-error ts-migrate(2339) FIXME: Property 'open' does not exist on type 'Readonly<{... Remove this comment to see the full error message
-  componentDidUpdate(prevProps) {
-    // @ts-expect-error ts-migrate(2722) FIXME: Cannot invoke an object which is possibly 'undefin... Remove this comment to see the full error message
-    this.props.makeStyles()
+  componentDidUpdate(prevProps: Props) {
+    this.props.makeStyles(undefined, {
+      prevProps: prevProps,
+      styledProps: ['variant']
+    })
```
matyasf, line 331: This is a simple example where `makeStyles` has no parameter

packages/emotion/src/withStyle.tsx
```diff
@@ -31,6 +31,20 @@ import { useTextDirectionContext } from '@instructure/ui-i18n'
 import { bidirectionalPolyfill } from './styleUtils/bidirectionalPolyfill'
 import { getComponentThemeOverride } from './getComponentThemeOverride'
 import { useTheme } from './useTheme'
+
+export interface MakeStyleConfig {
+  // The previous prop object that will be compared to the current props
+  prevProps: Record<string, unknown>
+  // An array of property names used in styles.js. The style object will
+  // recalculate if any of these properties change In cannot be used to compare
+  // nested properties.
```
ke1k0, line 40:
```suggestion
  // An array of property names used in styles.js. The style object will
  // recalculate if any of these properties change. It cannot be used to compare
  // nested properties.
```

Brailor, line 40: Imho we should be consistent with the `interface` and `type` usage.
By default I think we should stick with the `type` and if we happen to be in a situation where we would need to add a field to a type, then an interface could be used.
https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces

packages/emotion/src/withStyle.tsx
```diff
@@ -31,6 +31,20 @@ import { useTextDirectionContext } from '@instructure/ui-i18n'
...
+  styledProps: string[]
+}
+export type MakeStyles = (
+  generateStyleArgs?: Record<string, unknown>,
+  updateConfig?: MakeStyleConfig
+) => void
```
Brailor, line 46: I'm not sure about this function signature, it could lead to ugly call sites like `makeStyles(undefined, config)`. 
If we keep backwards compatibly in mind and want to preserve the 1 parameter call signature, then we could do something like this:
```js
export type MakeStyleConfig = {
  updateConfig?: {
    prevProps: Record<string, unknown>
    styledProps: string[]
  }
  [K: string]: unknown
}
export type MakeStyles = (config: MakeStyleConfig) => void
```

packages/emotion/src/withStyle.tsx
```diff
@@ -126,9 +140,35 @@ const withStyle = decorator(
...
+       *     style recalculations for performance reasons. If not supplied
+       *     everything will be recalculated.
+       */
+      const makeStyleHandler: MakeStyles = (
+        generateStyleArgs,
+        updateConfig
+      ) => {
+        if (updateConfig) {
+          let changed = false
+          for (const propToCheck in updateConfig.styledProps) {
```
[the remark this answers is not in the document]
Brailor: If we keep the 1 signature `makeStyles` like I mentioned above, then it can be used here like this:
```js
 const makeStyleHandler: MakeStyles = (config) => {
        const { updateConfig, ...generateStyleArgs } = config
...
```

packages/ui-avatar/src/Avatar/index.tsx
```diff
@@ -122,13 +127,18 @@ class Avatar extends Component<Props> {
...
+      }
+    }
+    this.props.makeStyles(this.state, updateConfig)
```
Brailor, line 141: This would be the call site for the 1 parameter `makeStyles`:
```js
componentDidUpdate(prevProps: Props, prevState: State) {
    let updateConfig
    if (prevState.loaded == this.state.loaded) {
      updateConfig = {
        prevProps: prevProps,
        styledProps: ['size', 'shape', 'src']
      }
    }
    this.props.makeStyles({ updateConfig, ...this.state })
  }

```

HerrTopi (approved):

packages/emotion/src/withStyle.tsx
matyasf: OK

packages/ui-avatar/src/Avatar/index.tsx
```diff
@@ -39,9 +39,10 @@ import {
 
 import generateStyle from './styles'
 import generateComponentTheme from './theme'
+import { MakeStyleConfig, MakeStyles } from '@instructure/emotion/src/withStyle'
```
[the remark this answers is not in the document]
matyasf: OK

packages/ui-avatar/src/Avatar/index.tsx
```diff
@@ -132,14 +132,16 @@ class Avatar extends Component<Props, State> {
...
-        styledProps: ['size', 'shape', 'src']
+      makeStyleConfig = {
+        updateConfig: {
+          prevProps: prevProps,
+          styledProps: ['size', 'shape', 'src']
+        }
       }
     }
-    this.props.makeStyles({ ...updateConfig, ...this.state })
+    this.props.makeStyles({ ...makeStyleConfig, ...this.state })
```
Brailor: Just a note:
You don't have to spread the makeStyleConfig (or even give typing to it) if you would rename it back to `updateConfig` and  then simply do:

```js
let updateConfig
if (prevState.loaded == this.state.loaded) {
    updateConfig = { 
      prevProps,
      styledProps: ['size', 'shape', 'src']
    }
  
}
makeStyles({ updateConfig, ...this.state})
```

This works because TS has "shape" typings, so as long the shape of the object you pass to the function matches the shape of the defined type it will compile correctly. What that means is that you don't have to have a proper `MakeStyleConfig` type variable in order to be able to call `makeStyles`, only an object which has the same shape as of `MakeStyleConfig`.

Brailor (approved):

packages/ui-avatar/src/Avatar/index.tsx
matyasf: OK, good idea

matyasf:
I'll close this PR because it does not seem to bring any significant performance improvement.

PR #563, opened 2021-05-20
INSTUI-3112 - add version selector

ke1k0:


packages/__docs__/buildScripts/build-docs.js
```diff
@@ -30,6 +30,9 @@ const path = require('path')
 const getClientProps = require('./utils/getClientProps')
 const processFile = require('./processFile')
 const fs = require('fs')
+//TODO: Rename this env var
+const { DO_THE_VERSION_COPY = '1' } = process.env
+const shouldDoTheVersionCopy = Boolean(parseInt(DO_THE_VERSION_COPY))
```
matyasf, line 35: Why do we need this env var? when should we not copy versions.json?

Brailor: For example when we release v9 and when we deploy the v8 docs to Netlfy, then we do not want this copy to happen.
So for all of the future major version doc updates.

matyasf (approved):

HerrTopi (approved):

PR #564, opened 2021-05-21
INSTUI-3099 ui-utils: rewrite to TypeScript, fix an edge case in hash()

matyasf:


packages/ui-icons-build/lib/util/format-name.js
```diff
@@ -22,6 +22,7 @@
  * SOFTWARE.
  */
 
+// TODO these are implemented in ui-utils, get them from there
```
matyasf, line 25: I've made a ticket for this, INSTUI-3118

packages/ui-utils/src/createChainedFunction.ts
```diff
@@ -33,40 +34,37 @@
...
-        // @ts-expect-error ts-migrate(2683) FIXME: 'this' implicitly has type 'any' because it does n... Remove this comment to see the full error message
-        acc.apply(this, args)
-        // @ts-expect-error ts-migrate(2683) FIXME: 'this' implicitly has type 'any' because it does n... Remove this comment to see the full error message
+      return function chainedFunction(this: any, ...args) {
+        acc!.apply(this, args)
         f.apply(this, args)
       }
-    }, null)
+      // TODO I think it can return null too
+    }, null) as undefined | GenericFunction
```
matyasf, line 67: I think it can return null too, but enabling it would result in tons of code where we'd need null checks. I'd rather modify this to not allow returning null

packages/ui-utils/src/hash.ts
```diff
@@ -52,11 +50,10 @@ function toBase64(input) {
   return output
 }
 
-// @ts-expect-error ts-migrate(7006) FIXME: Parameter 'input' implicitly has an 'any' type.
-function executeHash(input) {
+function executeHash(input: string): string {
   let hash = 0
 
-  if (input.length === 0) return hash
+  if (input.length === 0) return '0'
```
matyasf, line 56: I this case it was returning a number (whereas in every other case a string), I've fixed it

packages/ui-utils/src/ms.ts
```diff
@@ -36,21 +36,20 @@ import { parseUnit } from './parseUnit'
...
  * @returns {Number} Returns numerical representation of milliseconds
  */
-// @ts-expect-error ts-migrate(7006) FIXME: Parameter 'val' implicitly has an 'any' type.
-function ms(val) {
+function ms(val: number | string): number {
+  // TODO: throw an error if val is not number or string
   if (!val || typeof val === 'number') {
-    return val
+    return val as number
   }
```
matyasf, line 46: again some weird code that allows to return edge cases like null or undefined. I've left it this way to not change functionality

packages/ui-utils/src/px.ts
```diff
@@ -39,24 +39,22 @@ import { parseUnit } from './parseUnit'
...
-// @ts-expect-error ts-migrate(7006) FIXME: Parameter 'val' implicitly has an 'any' type.
-function px(val, el) {
+function px(val: string | number, el?: Document | Element | null): number {
   const container = el || document.body
 
+  // TODO !val should not be needed
   if (!val || typeof val === 'number') {
-    return val
+    return val as number
   }
```
matyasf, line 51: Same weirdness as with ms(), a function that should return numbers can return undefined and null

packages/emotion/src/useTheme.ts
```diff
@@ -37,7 +37,7 @@ import { isEmpty } from '@instructure/ui-utils'
 export const useTheme = () => {
   let theme = useEmotionTheme()
 
-  if (isEmpty(theme)) {
+  if (isEmpty(theme as Record<string, unknown>)) {
```
Brailor, line 40: I would move this `as Record<string, unknown>` type casting to be the return type of `useEmotionTheme` and then it would not be required us to cast it here.
What do you think?

matyasf: useEmotionTheme returns a `Theme` type, which comes from Emotion, which is an empty interface. Since this PR is about `ui-utils` I wanted to make minimal changes to the `emotion` package.

packages/ui-utils/src/capitalizeFirstLetter.ts
```diff
@@ -30,8 +30,7 @@
  * @module capitalizeFirstLetter
  * @param {String} str
  */
-// @ts-expect-error ts-migrate(7006) FIXME: Parameter 'str' implicitly has an 'any' type.
-function capitalizeFirstLetter(str) {
+function capitalizeFirstLetter(str: string | undefined) {
```
ke1k0, line 33:
```suggestion
function capitalizeFirstLetter(str?: string) {
```

packages/ui-utils/src/mergeDeep.ts
```diff
@@ -66,25 +66,22 @@ function mergeSourceIntoTarget(target, source) {
...
   return (
     item &&
     (typeof item === 'object' || typeof item === 'function') &&
     !Array.isArray(item)
   )
 }
 
-// @ts-expect-error ts-migrate(7006) FIXME: Parameter 'item' implicitly has an 'any' type.
-function isArray(item) {
-  return item && Array.isArray(item)
```
ke1k0, line 87: isn' the "item && " part needed? it is complaining about it not beeing a boolean, but maybe like this: `!!item && ...`?

matyasf: its not needed, thats why I deleted it :) Array.isArray(item) accepts `any`

ke1k0 (approved):

Brailor (approved):

PR #567, opened 2021-05-25
Add Version selector to V6 docs

Brailor:


packages/ui-docs-client/src/Header/index.js
```diff
@@ -38,11 +43,90 @@ import theme from './theme'
...
+
+  handleSelect = (_e, [ selectedVersion ]) => {
+    const { versionsData } = this.props
+    const { latestVersion } = versionsData
+    const isSelectedLatestVersion = selectedVersion === latestVersion
+    const splittedUrl = window.location.pathname.split('/').filter(Boolean)
+    const [ versionInPath ] = splittedUrl
+    const isOnLatestVersion = splittedUrl.length === 0
+
+    // 1: instructure.design, latest: v8, selected: v8 -> navigate to instructure.design/#index
```
HerrTopi, line 69: Should we hardcode the latest version number in a comment? This could get outdated really soon

packages/ui-docs-client/src/Header/index.js
```diff
@@ -38,11 +43,90 @@ import theme from './theme'
...
+                  <span>
+                    {this.props.name} {this.props.version}
+                  </span>
+                ) || 'Documentation'}
+              </Text>
+              <IconMiniArrowDownLine size="x-small" />
+            </CondensedButton>
+          }
+        >
+          <Menu.Group
```
HerrTopi, line 113: Just curious: what is the benefit of using Menu.Group over Menu?

HerrTopi (approved):

packages/ui-docs-client/src/Header/index.js
ke1k0: You need Menu.Group for a list where 1 or more items can be "selected", with those checkmarks before them.

packages/ui-docs-client/src/Header/index.js
Brailor: Those are just examples of the functionality

ke1k0 (approved):

PR #568, opened 2021-05-25
feat(ui-table): add `stackedSortByLabel` prop to `Table.ColHeader`

ke1k0:
Closes: INSTUI-3116

The new `stackedSortByLabel` prop will allow users to set custom strings to the labels appearing in
the "sort by" select of a sortable table with stacked layout. It is using the `id` prop by default,
which is not always a user-friendly string. It will also allow to display longer, more descriptive
labels.

### DIFF
--- a/packages/ui-table/src/Table/ColHeader/index.tsx
+++ b/packages/ui-table/src/Table/ColHeader/index.tsx
@@ -41,6 +41,7 @@ type Props = {
   makeStyles?: (...args: any[]) => any
   styles?: any
   id: string
+  stackedSortByLabel: string
   width?: string | number
   textAlign?: 'start' | 'center' | 'end'
   sortDirection?: 'none' | 'ascending' | 'descending'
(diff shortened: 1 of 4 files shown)


packages/ui-table/src/Table/ColHeader/index.tsx
```diff
@@ -63,10 +64,16 @@ class ColHeader extends Component<Props> {
...
-     * id is also used as option in combobox.
+     * A unique id for this column. The `id` is also used as option in combobox,
+     * when sortable table is in stacked layout,
+     * and no `stackedSortByLabel` is provided.
      */
     id: PropTypes.string.isRequired,
+    /**
+     * A custom string to display as option in the combobox (instead of `id`),
+     * when sortable table is in stacked layout.
+     */
```
matyasf, line 75: It wasn't clear for me what this property does, I had to rewrite the example code to figure out. I'd rephrase it to:
"A custom string to display as option text in the combobox (instead of using the `id` field), when sortable table is in stacked layout."

matyasf (approved):

packages/ui-table/src/Table/ColHeader/index.tsx
ke1k0: Thank, modified it!

Brailor (approved):

PR #569, opened 2021-05-26
Version selector for V7

Brailor:


packages/ui-docs-client/src/Header/index.js
```diff
@@ -30,6 +30,11 @@ import { InlineSVG } from '@instructure/ui-svg-images'
 import { Text } from '@instructure/ui-text'
 import { View } from '@instructure/ui-view'
 import { themeable } from '@instructure/ui-themeable'
+import { SimpleSelect } from '@instructure/ui-simple-select'
```
ke1k0, line 33: unused import

packages/ui-docs-client/src/Header/index.js
```diff
@@ -39,14 +44,96 @@ import theme from './theme'
...
+        </Menu>
+      </View>
+    )
+  }
   render() {
+    const { versionsData } = this.props
+    const { latestVersion, previousVersions } = versionsData
+    const allVersions = [latestVersion, ...previousVersions]
+
+    const [versionInPath] = window.location.pathname.split('/').filter(Boolean)
```
ke1k0, line 136: these are unused variables

packages/ui-docs-client/src/Header/index.js
```diff
@@ -39,14 +44,96 @@ import theme from './theme'
...
+    return window.location.replace(versionToNavigate)
+  }
+  renderVersionsBlock = () => {
+    const { versionsData } = this.props
+    const { latestVersion, previousVersions } = versionsData
+    const allVersions = [latestVersion, ...previousVersions]
+
+    const [versionInPath] = window.location.pathname.split('/').filter(Boolean)
+
+    const currentVersion = versionInPath || latestVersion
```
ke1k0, line 91: Well, on localhost this comes back with v8, since "/v7" is not in the url. Will probably work in production, but probably not a bad idea to put a comment above it, so the devs know it's not a bug.

packages/ui-docs-client/src/Header/index.js
```diff
@@ -39,14 +44,96 @@ import theme from './theme'
...
-    version: undefined
+    version: undefined,
+    versionsData: undefined
   }
 
+  handleSelect = (e, [selectedVersion]) => {
+    const { versionsData } = this.props
+    const { latestVersion } = versionsData
+    const isSelectedLatestVersion = selectedVersion === latestVersion
+    const splittedUrl = window.location.pathname.split('/').filter(Boolean)
```
ke1k0, line 63: Webstorm shows a typo here, should probably be "splitUrl", or "pathNameParts"

packages/ui-docs-client/src/Header/index.js
Brailor: Removed

packages/ui-docs-client/src/Header/index.js
Brailor: Removed

packages/ui-docs-client/src/Header/index.js
Brailor: Fixed

packages/ui-docs-client/src/Header/index.js
Brailor: Renamed it to `pathNameParts`

packages/__docs__/package.json
```diff
@@ -86,6 +86,21 @@
...
+    "@instructure/ui-grid": "^7.7.0",
+    "@instructure/ui-checkbox": "^7.7.0",
+    "@instructure/ui-list": "^7.7.0",
+    "@instructure/ui-rating": "^7.7.0",
+    "@instructure/ui-text-area": "^7.7.0",
+    "@instructure/ui-flex": "^7.7.0",
+    "@instructure/ui-file-drop": "^7.7.0",
+    "@instructure/ui-link": "^7.7.0",
+    "@instructure/ui-heading": "^7.7.0",
+    "@instructure/ui-progress": "^7.7.0",
```
Brailor, line 103: Had to add these depenecies to the docs-app, because the build failed without them

packages/ui-docs-client/src/Header/index.js
ke1k0: I wouldn't hardcode "v8" here, soon it will be v9, and we won't remember fixing it in v7 :)

packages/ui-docs-client/src/App/index.js
```diff
@@ -102,17 +102,35 @@ class App extends Component {
     this.state = {
       showMenu: showTrayOnPageLoad,
       themeKey: Object.keys(props.themes)[0],
-      layout: 'large'
+      layout: 'large',
+      versionsData: {
+        latestVersion: 'v8',
+        previousVersions: ['v7', 'v6']
```
ke1k0, line 108: I'm not sure about this. I know the fetch will update the data, but probably confusing to hardcode v8 here too. I would set the initial state to null or {}, and handle that in the code (like not rendering the version select until we have the version data, just the current version heading as before)

HerrTopi (approved):

ke1k0 (approved):

PR #577, opened 2021-06-01
INSTUI-2932 : Replace <Flex> with <span> in Button to improve performance

matyasf:
Replace the Flex component with simple span inside BaseButton for better performance. In the testing tool it resulted in ~40% performance improvement testing 100 times with a Button that has an icon and text:

BUTTON_ICON_IUIx100	832.9ms  // old, 8.4.0 code
BUTTON_ICON_IUIx100	504.4ms  // new code in this PR

It should improve performance with other types of button too, albeit to a lesser degree

### DIFF
--- a/packages/ui-buttons/src/BaseButton/index.tsx
+++ b/packages/ui-buttons/src/BaseButton/index.tsx
@@ -39,7 +39,6 @@ import { isActiveElement } from '@instructure/ui-dom-utils'
 
 import { hasVisibleChildren } from '@instructure/ui-a11y-utils'
 import { View } from '@instructure/ui-view'
-import { Flex } from '@instructure/ui-flex'
 
 import {
   withStyle,
@@ -343,7 +342,7 @@ class BaseButton extends Component<Props> {
   }
 
   renderChildren() {
-    const { renderIcon, children, textAlign, isCondensed, styles } = this.props
+    const { renderIcon, children, styles } = this.props
 
     const wrappedChildren = <span css={styles.children}>{children}</span>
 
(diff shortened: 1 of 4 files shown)


Brailor (approved):

packages/ui-buttons/src/BaseButton/index.tsx
```diff
@@ -356,38 +355,21 @@ class BaseButton extends Component<Props> {
     )
 
     const flexChildren = hasOnlyIconVisible ? (
-      <Flex.Item>
+      <span css={styles.flexOnlyIcon}>
```
ke1k0, line 358: Please consider the renaming of the classes I sent you on slack.

packages/ui-buttons/src/BaseButton/styles.ts
```diff
@@ -356,6 +356,42 @@ const generateStyle = (componentTheme, props, state) => {
 
       // @ts-expect-error ts-migrate(7053) FIXME: Element implicitly has an 'any' type because expre... Remove this comment to see the full error message
       ...sizeVariants[size].iconSVG
+    },
+
+    flex: {
```
ke1k0, line 361: Please add the missing css props from view (sent google sheet on slack)

packages/ui-buttons/src/BaseButton/index.tsx
matyasf: done

packages/ui-buttons/src/BaseButton/styles.ts
matyasf: done

ke1k0 (approved):

PR #579, opened 2021-06-02
INSTUI-3045 - ui-dom-utils TS typing

ke1k0:


packages/ui-dom-utils/src/contains.ts
```diff
@@ -32,38 +33,48 @@ import { canUseDOM } from './canUseDOM'
...
 
   if (!container || !node) {
     return false
-  } else if (container.contains) {
+  } else if (!(container instanceof Window) && !(node instanceof Window)) {
     return container.contains(node)
-  } else if (container.compareDocumentPosition) {
-    return (
-      container === node || !!(container.compareDocumentPosition(node) & 16)
-    ) // eslint-disable-line no-bitwise
```
ke1k0, line 51: this was a fallback fro IE

packages/ui-dom-utils/src/elementMatches.ts
```diff
@@ -31,17 +32,17 @@ import { findDOMNode } from './findDOMNode'
...
-function elementMatches(el, selectorString) {
+function elementMatches(
+  el: Node | Window | React.ReactElement | ((...args: any[]) => any),
+  selectorString: string
+) {
   const node = el && findDOMNode(el)
   if (!node) return false
-  return node.matches
-    ? node.matches(selectorString)
-    : node.msMatchesSelector(selectorString)
```
ke1k0, line 44: this was a fallback fro IE: https://developer.mozilla.org/en-US/docs/Web/API/Element/matches#polyfill

packages/ui-dom-utils/src/getFontSize.ts
```diff
@@ -35,32 +36,34 @@ const COMPUTED_CACHE = {}
...
+function getFontSize(
+  el?: Node | Window | React.ReactElement | ((...args: any[]) => any) | null,
+  ignoreCache = false
+) {
   if (!canUseDOM) {
     return 16
   }
 
   const container = el || ownerDocument(el).documentElement
+  const containerString = container.toString()
```
ke1k0, line 52: Should this be JSON.stringify() instead of toString()?

packages/ui-dom-utils/src/handleMouseOverOut.ts
```diff
@@ -38,12 +38,18 @@ import { contains } from './contains'
...
  * @param event {Event} The DOM Event that was fired
  */
-// @ts-expect-error ts-migrate(7006) FIXME: Parameter 'handler' implicitly has an 'any' type.
-function handleMouseOverOut(handler, event) {
+function handleMouseOverOut(
+  handler: (...args: any[]) => any,
+  event: Event & { relatedTarget: EventTarget | null }
+) {
   const target = event.currentTarget
-  const related = event.relatedTarget || event.nativeEvent.toElement
```
ke1k0, line 44: this seems like an IE fallback as well

packages/ui-dom-utils/src/ownerWindow.ts
```diff
@@ -32,14 +33,15 @@ import { ownerDocument } from './ownerDocument'
...
+ * @returns { Window | null } the owner window
  */
-// @ts-expect-error ts-migrate(7006) FIXME: Parameter 'el' implicitly has an 'any' type.
-function ownerWindow(el) {
+function ownerWindow(
+  el?: Node | Window | React.ReactElement | ((...args: any[]) => any) | null
+) {
   const node = el && findDOMNode(el)
   const doc = ownerDocument(node)
-  return doc && (doc.defaultView || doc.parentWindow)
```
ke1k0, line 42: another IE fallback http://help.dottoro.com/ljemhvhd.php

packages/ui-dom-utils/src/addInputModeListener.ts
```diff
@@ -119,20 +113,19 @@ const addDownListeners = () => {
...
 
-// @ts-expect-error ts-migrate(7031) FIXME: Binding element 'onInputModeChange' implicitly has... Remove this comment to see the full error message
-const addInputModeListener = ({ onInputModeChange }) => {
+const addInputModeListener = (handlerObject: {
+  onInputModeChange: (...args: any[]) => any
+}) => {
+  const { onInputModeChange } = handlerObject
   const id = _registeredCount++
 
   if (typeof onInputModeChange === 'function') {
```
matyasf, line 128: this is always a function, no need for this check

packages/ui-dom-utils/src/getActiveElement.ts
```diff
@@ -29,16 +29,16 @@
...
+ * @return { Element | null } the active element
  */
-// @ts-expect-error ts-migrate(7006) FIXME: Parameter 'doc' implicitly has an 'any' type.
-function getActiveElement(doc) {
+function getActiveElement(doc?: Document) {
   try {
     return (doc || document).activeElement
   } catch (e) {
     /* ie throws if no active element */
+    return null
```
matyasf, line 41: This is an old IE workaround https://www.tjvantoll.com/2013/08/30/bugs-with-document-activeelement-in-internet-explorer/
so we can remove it

packages/ui-dom-utils/src/getClassList.ts
```diff
@@ -85,8 +109,7 @@ function shimClassListForIE() {
         HTMLElement.prototype,
         'classList'
       )
-      // @ts-expect-error ts-migrate(2345) FIXME: Argument of type 'PropertyDescriptor | undefined' ... Remove this comment to see the full error message
-      Object.defineProperty(SVGElement.prototype, 'classList', descr)
+      descr && Object.defineProperty(SVGElement.prototype, 'classList', descr)
```
matyasf, line 112: shimClassListforIE is an old IE workaround, we can remove it

packages/ui-dom-utils/src/getComputedStyle.ts
```diff
@@ -35,16 +36,20 @@ import { canUseDOM } from './canUseDOM'
...
+ * @param { Node | Window | React.ReactElement | function } el - component or DOM node
+ * @returns { Object } object containing css properties and values for the element
  */
-// @ts-expect-error ts-migrate(7006) FIXME: Parameter 'el' implicitly has an 'any' type.
-function getComputedStyle(el) {
-  let style = {}
+function getComputedStyle(
+  el: Node | Window | React.ReactElement | ((...args: any[]) => any)
+) {
+  let style: CSSStyleDeclaration | Record<string, never> = {}
```
matyasf, line 45: I think we can safely type this to CSSStyleDeclaration, it wont change its behaviour. So instead of defaulting to `{}` new can write new CSSStyleDeclaration()

packages/ui-dom-utils/src/index.ts
```diff
@@ -48,6 +49,7 @@ export { isVisible } from './isVisible'
 export { ownerDocument } from './ownerDocument'
 export { ownerWindow } from './ownerWindow'
 export { requestAnimationFrame } from './requestAnimationFrame'
+export type { RequestAnimationFrameType } from './requestAnimationFrame'
 export { supportsObjectFit } from './supportsObjectFit'
```
matyasf, line 53: we can remove supportsObjectFit, its an IE hack https://caniuse.com/object-fit
not necessarly here, but if its not removed we should make a ticket to remove it.

packages/ui-dom-utils/src/getFontSize.ts
matyasf: hmm good question. As I see it was using toString(), but this is far form ideal, since it will be just some String like "Element". JSON.stringify would slow this code down. So I'd use a Map https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map , this can use objects as keys.

packages/ui-dom-utils/src/index.ts
ke1k0: Made a ticket for it (INSTUI-3141)

packages/ui-dom-utils/src/getComputedStyle.ts
ke1k0: `new CSSStyleDeclaration` is `Illegal constructor`, so we have to keep the empty object for now

packages/ui-dom-utils/src/addInputModeListener.ts
ke1k0: As we discussed, we should keep this, because not having this check could break the `forEach` above, because it won't be "callable".

packages/ui-dom-utils/src/getClassList.ts
ke1k0: Done, thanks

packages/ui-dom-utils/src/getActiveElement.ts
ke1k0: Removed the try...catch, thanks

packages/ui-dom-utils/src/getFontSize.ts
ke1k0: Converted it to a WeakMap, thanks

matyasf (approved):

packages/ui-focusable/src/Focusable/index.tsx
```diff
@@ -78,9 +78,9 @@ class Focusable extends Component<Props> {
     'datetime-local': true
   }
 
-  _focusListener = null
-  _blurListener = null
-  _inputModeListener = null
+  _focusListener: { remove(): void } | null = null
+  _blurListener: { remove(): void } | null = null
+  _inputModeListener: { isKeyboardMode(): void; remove(): void } | null = null
```
Brailor, line 83: `isKeyboardMode()` should have a return type of `boolean
`

packages/ui-dom-utils/src/getComputedStyle.ts
Brailor, line 45: Not sure about this `never` type here, it seems like a logical error to me.
https://www.typescriptlang.org/docs/handbook/basic-types.html#never

If we don't know the exact type `unknown` would be better I think.

packages/ui-dom-utils/src/getClassList.ts
```diff
@@ -23,10 +23,16 @@
  */
 
 import { findDOMNode } from './findDOMNode'
+import React from 'react'
 
-let classListShimmed = false
+type ClassListApiType = {
```
Brailor, line 28: For me the name `type ClassListApiType` is a bit redundant, I'd drop the Type from it and change it to simply
be `ClassListApi`.

Brailor (approved):

PR #580, opened 2021-06-02
refactor(console): rewrite console to use TypeScript

matyasf:


packages/console/src/console.ts
```diff
@@ -30,14 +30,30 @@ let loggedInitialDeprecationWarning = false
 function getRenderStack() {
   let renderStack = ''
   try {
-    renderStack = React.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.ReactDebugCurrentFrame.getStackAddendum()
+    // @ts-expect-error This does not seem to be typed.
+    renderStack =
+      React.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.ReactDebugCurrentFrame.getStackAddendum()
```
matyasf, line 35: Its so wrong to use this, that its not even included in the type definitions :)

packages/console/src/macro.js
```diff
@@ -54,12 +54,16 @@ function macro({ babel, references, state }) {
 }
 
 module.exports = createMacro(macro)
+// note: error, warn, warnDeprecrated functions have a third ...args:any[]
+// parameter, but this is currently buggy in JSDoc/TS so it cannot be added
+// in a way that IDEs understand it and the generated .d.ts files are correct.
```
matyasf, line 59: This file is a tricky one. Its a Babel macro, which I could not convert to TypeScript. In this case TS tries to be clever and generates type definitions from JSDoc comments. But there is a bug in TypeScript (or in the language server?):
It cannot correctly interpret a `@type` that has 2 or more normal and then a `...rest` arg, it always marks the rest arg as non-optional when typechecking with the IDE (same behaviour in WebStorm and VSCode). Strangely the outputted `.d.ts` file looks OK. As a workaround I had to remove the last ...rest arg (we were only using it in one place, and its just added to the `message` string anyway). Also I've filled a TypeScript bug: https://github.com/microsoft/TypeScript/issues/44386

Brailor (approved):

PR #585, opened 2021-06-07
INSTUI-3129 : Convert ui-react-utils to TS

matyasf:


packages/ui-react-utils/src/getInteraction.ts
```diff
@@ -50,7 +61,7 @@ export function getInteraction({
   // if there's no interaction specified, use disabled or readOnly if they are defined as props (note, disabled will win
   // over readOnly if both are provided)
   if (interactionTypes.includes('disabled') && disabled) return 'disabled'
-  if (interactionTypes.includes('readOnly') && readOnly) return 'readonly'
+  if (interactionTypes.includes('readonly') && readOnly) return 'readonly'
```
matyasf, line 64: I think this was a bug

packages/ui-react-utils/src/getInteraction.ts
```diff
@@ -34,14 +45,14 @@
  * @param {Object} args.props - the component props
  * @param {string} args.props.interaction - specifies the interaction mode, one of 'enabled', 'disabled', or 'readonly'
  * @param {boolean} args.props.disabled - specifies if the component is disabled. Will take precedence over readOnly
- * @param {boolean} args.props.readOnly - specifies if the component is readOnly
+ * @param {boolean} args.props.readonly - specifies if the component is readonly
```
ke1k0, line 48: the name of the PROP is "readOnly", the interactionType option is "readonly"

packages/ui-react-utils/src/ComponentIdentifier.ts
```diff
@@ -64,7 +66,12 @@ Abstract component identifier. Helpful for picking out a specific child.
   ```
   @module ComponentIdentifier
 **/
-class ComponentIdentifier extends Component {
+@deprecated(
+  '8.0.0',
+  undefined,
+  'This component will be removed in InstUI 9.0 since its not used.'
+)
```
ke1k0, line 73: This is causing a lot of warnings on the console. Might not be a problem for our users, because it seems we only use it in our docs app... ¯\_(ツ)_/¯

Brailor (approved):

packages/ui-buttons/src/BaseButton/index.tsx
```diff
@@ -49,6 +49,7 @@ import {
 
 import generateStyles from './styles'
 import generateComponentTheme from './theme'
+import { InteractionType } from '@instructure/ui-react-utils/src/getInteraction'
```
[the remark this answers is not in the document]
matyasf: done

packages/ui-react-utils/src/getInteraction.ts
matyasf: done

ke1k0 (approved):

PR #587, opened 2021-06-07
INSTUI-3131 - ui-testable ts migration

ke1k0:


packages/ui-testable/src/testable.ts
```diff
@@ -34,45 +35,57 @@ const testable =
   // ALWAYS_APPEND_UI_TESTABLE_LOCATORS=1
   // We do this because adding those `data-cid` locators slows things down.
   !process.env.ALWAYS_APPEND_UI_TESTABLE_LOCATORS
-    ? () => (Component) => Component
-    : decorator((ComposedComponent) => {
+    ? () => (Component: any) => Component
+    : decorator((ComposedComponent: any) => {
```
matyasf, line 39: you can type this to ComponentClass, e.g.
```
? () => (Component: ComponentClass) => Component
: decorator((ComposedComponent: ComponentClass<any>) => {
```

packages/ui-testable/src/testable.ts
```diff
@@ -34,45 +35,57 @@ const testable =
...
         const selector = `[${locator.attribute}~="${locator.value}"]`
-        class TestableComponent extends ComposedComponent {
+        class TestableComponent<
+          P = Record<string, unknown>,
+          S = ComponentState
+        > extends ComposedComponent {
           static selector = selector
-          componentDidMount(...args) {
+          private _testableUnmounted: boolean | undefined
+          _locatorTimeout: NodeJS.Timeout | undefined
```
matyasf, line 53: This is a tricky one. Although NodeJS.timeout fits here, its wrong since this package is used in the browser. Ive followed the advice here to solve it: https://stackoverflow.com/questions/51040703/what-return-type-should-be-used-for-settimeout-in-typescript in debounce: https://github.com/instructure/instructure-ui/blob/master/packages/debounce/src/debounce.ts#L71

packages/ui-testable/src/testable.ts
```diff
@@ -34,45 +35,57 @@ const testable =
...
-            this.locatorTimeout = setTimeout(() => {
+            this._locatorTimeout = setTimeout(() => {
               let node
               if (this._testableUnmounted) {
                 return
               }
               try {
                 // Use this.DOMNode for components that render as non-native Portals...
-                node = this.DOMNode || findDOMNode(this)
+                node = this.DOMNode || findDOMNode(this as any)
```
matyasf, line 88: if you type `ComposedComponent` to ComponentClass<any> this cast is not needed

packages/ui-testable/src/testable.ts
```diff
@@ -34,45 +35,57 @@ const testable =
...
-        class TestableComponent extends ComposedComponent {
+        class TestableComponent<
+          P = Record<string, unknown>,
+          S = ComponentState
+        > extends ComposedComponent {
           static selector = selector
-          componentDidMount(...args) {
+          private _testableUnmounted: boolean | undefined
+          _locatorTimeout: NodeJS.Timeout | undefined
+          private DOMNode: any
```
matyasf, line 54: This is also a tricky one. This variable is not defined here, its coming from the super class that `TestableComponent` extends. I'd either leave a TODO here to do this properly, or now make an interface that looks like this:
```
interface ClassWithDOMNode {
  DOMNode?: Element
}
```
and make the classes that have this prop (`ReactPortal` and `SubTreePortal`, maybe more implement it. 
In this class I'd make a cast, then there is no need to define a private property:
```node = (this as ClassWithDOMNode).DOMNode || findDOMNode(this)```

File: packages/ui-testable/src/testable.ts (hunk 1 above)
Comment (ke1k0) [reply packages/ui-testable/src/testable.ts]: Now that `ui-prop-types` is typed, it doesn't give any more warnings.

File: packages/ui-testable/src/testable.ts (hunk 2 above)
Comment (ke1k0) [reply packages/ui-testable/src/testable.ts]: Done, thanks

File: packages/ui-testable/src/testable.ts (hunk 4 above)
Comment (ke1k0) [reply packages/ui-testable/src/testable.ts]: I figured out the possible type of DOMNode, and typed it in the relevant components, but only added a TODO about the common interface. Would like to discuss how we want to do this in the future.

File: packages/ui-tree-browser/src/TreeBrowser/TreeButton/index.tsx [hunk 5]
```diff
@@ -93,8 +93,8 @@ class TreeButton extends Component<Props> {
     selected: PropTypes.bool,
     focused: PropTypes.bool,
     level: PropTypes.number,
-    containerRef: function () {},
-    renderContent: function () {}
```
ke1k0, line 97: This seemed to be buggy, build:types failed on it. Changed it to `PropTypes.func`, because that is what we use everywhere else.

matyasf (approved):

Brailor (approved):

PR #589, opened 2021-06-08
fix(ui-tree-browser): properly hide root collection if rootId === 0

simonista:
TEST PLAN:
Render a TreeBrowser with several collections, one of which has id === 0. Specify rootId={0} and
showRootCollection={false}. The root collection should be properly hidden.


simonista:
FYI, this is blocking canvas-lms upgrading to InstUI 7, so I'm also going to push a PR against v7_maintenance. This is my first time doing this though, so let me know if something else needs to happen.

simonista:
FYI this is blocking canvas-lms upgrading to InstUI 7, so I will open a PR against the maintenance branch as well.

Brailor (approved):

simonista:
Just realized I had the wrong email address on my commit, so the force push was just changing the author email, no other changes.

PR #595, opened 2021-06-09
feat(ui-options): add option to recolor the icon of Option.Item

ke1k0:
Closes: INSTUI-2903

If a function is passed to `renderBeforeLabel` or `renderAfterLabel` prop, the function will have a
`props` parameter. This way the the properties of that `Option.Item` (e.g. the current `variant`) is
accessible in the render function, and can be used to set different colors to that icon for
different option state (variant).


Brailor (approved):

Brailor (approved):

HerrTopi (approved):
NIce docs!

PR #599, opened 2021-06-15
fix(ui-test-sandbox): fix issue with `emotion` can't insert node inside `head` 

Brailor:
stop observing changes being made in the `head` section  of the document, because the teardown of
the sandbox was deleting nodes from it and it messed up the `emotion` styling.


packages/ui-test-sandbox/src/utils/sandbox.ts
```diff
@@ -27,11 +27,6 @@ import { ReactComponentWrapper } from './reactComponentWrapper'
 import initConsole from './initConsole'
 import React from 'react'
 
-// Likely this could be done with /// <reference types
-declare const before: any
-declare const beforeEach: any
-declare const afterEach: any
-
```
Brailor, line 34: These are not needed because TS is able to pick up the type definitions from 'mocha'.

packages/ui-test-sandbox/src/utils/sandbox.ts
```diff
@@ -150,7 +146,6 @@ class Sandbox {
       })
     }
 
-    this._observer.observe(document.head, { childList: true })
```
Brailor, line 153: This is the real change in this PR.
Observing the the `<head>` element changes was not working well with Emotion:
For performance reasons emotion tries to insert the generated css into the `head` element of the document.
If we do this `observe` here, then after every test case the `teardown` method would remove all of the nodes which were injected by emotion.
This caused issues with emotion being able to insert elements to the `head` thus deleting the observation was the only change that allowed us to do the `(node as ChildNode).remove()` below which then only deletes the nodes inside the `body` of the document.

matyasf (approved):
nice work!

HerrTopi (approved):

PR #601, opened 2021-06-15
Convert ui-icons to TypeScript

matyasf:
component.ejs is treated as a TS file and is used to generate the icons and the sourcemap files.


matyasf:
will not do now. Its quite complicated and minimal benefits

PR #605, opened 2021-06-17
fix(ui-popover): recalculate offset and position on prop change

ke1k0:
Closes: INSTUI-3157

The 3 props, `offsetX`, `offsetY` and `placement` were used from the state, and didn't update if any
of these 3 props changed (e.g. when set from the `Responsive` component). They only updated if the
position of the `Popover` changed, now it updates and recalculates correctly on prop change as well.

### DIFF
--- a/packages/ui-popover/src/Popover/index.tsx
+++ b/packages/ui-popover/src/Popover/index.tsx
@@ -466,6 +466,18 @@ class Popover extends Component<Props> {
         this._focusRegion.deactivate()
       }
     }
+
+    // if the props the state is based on change,
+    // we need to recalculate the offset and placement
+    if (
+      this.props.offsetX !== prevProps.offsetX ||
+      this.props.offsetY !== prevProps.offsetY ||
+      this.props.placement !== prevProps.placement
+    ) {
+      this.setState({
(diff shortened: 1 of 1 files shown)


Brailor (approved):

Brailor:
Should we backport this fix to v7 as well?

ke1k0:
> Should we backport this fix to v7 as well?

There are workarounds for it, and it worked for reporter of the issue, so I thought it is enough to solve in v8. Or do you think we should?

packages/ui-popover/src/Popover/index.tsx
```diff
@@ -466,6 +466,18 @@ class Popover extends Component<Props> {
         this._focusRegion.deactivate()
       }
     }
+
+    // if the props the state is based on change,
```
matyasf, line 470: I don't understand what you mean by this, can you please clarify this comment?

ke1k0: Rephrased the comment, is it clearer now?

matyasf: looks good :)

matyasf (approved):

PR #614, opened 2021-06-22
chore(ui-calendar): upgrade packages

HerrTopi:
Closes: INSTUI-3133


matyasf (approved):
So as I see we need to pin TS version because of https://github.com/styleguidist/react-docgen-typescript/issues/356 ... While they released a fix in 2.0.0, we do not use this directly. The dependency tree is:
Storybook -> [@storybook/react](https://github.com/storybookjs/storybook/blob/next/app/react/package.json) -> [@storybook/react-docgen-typescript-plugin](https://github.com/hipstersmoothie/react-docgen-typescript-plugin/blob/master/package.json) -> [react-docgen-typescript](https://github.com/styleguidist/react-docgen-typescript)

We'll need to wait until this update chain finishes and Storybook 6.3 is released because of this issue: https://github.com/storybookjs/storybook/issues/14342

Brailor (approved):

PR #615, opened 2021-06-22
Only build Storybook stories for the changed packages 

Brailor:
This PR changes how the current `vrt.yml` workflow works and also how the current `StoryBook` build behaves:

- in the `vrt` workflow the changed `ui-*` packages will be calculated
- the list of these packages will be passed to the `build:examples` yarn script which will build the props combination for only those packages
- the storybook build process will generate the stories based on this generated props json file
- only the changed package stories will be published to `Chromatic`


packages/ui-view/src/View/index.tsx
```diff
@@ -100,7 +100,7 @@ type Props = {
   withVisualDebug?: boolean
   makeStyles?: (...args: any[]) => any
   styles?: any
-  dir?: any // TODO: PropTypes.oneOf(Object.values(bidirectional.DIRECTION)),
```
Brailor, line 103: This change is there because I wanted to test the feature of the PR, so because this is the only file which has examples in it's package, only the `ui-view` package's stories will be pushed to `Chromatic`.

HerrTopi (approved):
Seems good, hope all the storybook-chromatic magic will work as intended

matyasf (approved):
This will only build changed packages, and not their dependents. e.g. if we change View, it will not re-check all components, that depend on View. 
While this is not ideal, its a good interim solution until we pay for premium Chromatic.

PR #617, opened 2021-06-24
INSTUI-3164 copy configs to app template instead of depending on it

matyasf:
- Copy `ui-babel-preset`, `ui-eslint-config`, `ui-webpack-config`, `ui-stylelint-config`'s files to the newly created app.

- deprecate component and package templates

- `ui-babel-preset`, `ui-eslint-config`: do not load configs, just use the default. We never load configs and they are already very complex for a starter app.

- add a `--version` option to `instui-cli` that returns its version for sanity check

- remove `open-sandbox` from the template app's `package.json`


packages/template-app/template/build-scripts/webpack.js.ejs
```diff
@@ -0,0 +1,88 @@
...
+child.stdout.setEncoding('utf8');
+child.stdout.on('data', function(data) {
+  // eslint-disable-next-line no-console
+  console.log(data)
+})
+
+child.stderr.setEncoding('utf8')
+child.stderr.on('data', function(data) {
+  console.error('ERROR: ' + data)
+})
```
matyasf, line 88: This is mostly a copy of `webpack.js` from `ui-scripts`

HerrTopi (approved):
Seems good, great work :)

Brailor (approved):

PR #618, opened 2021-06-25
Remove proptypes poc

Brailor:
This PR adds an `allowedProps` list to the `View` component as an example.

- TS typing for the `View` component is moved to the `types.ts` file
- `View`'s propTypes is also moved to this file
- added logic to `omitProps` and `pickProps` so they work with the old `propTypes` and with the new `allowedProps` list
- added tests to `omitProps` and `pickProps`
- removed test cases which were essentially testing the inner workings of the `propTypes` library


packages/ui-view/src/ContextView/index.tsx
```diff
@@ -168,12 +168,10 @@ class ContextView extends Component<Props & OtherHTMLAttributes<Props>> {
...
 
   componentDidMount() {
-    // @ts-expect-error ts-migrate(2722) FIXME: Cannot invoke an object which is possibly 'undefin... Remove this comment to see the full error message
-    this.props.makeStyles()
+    this.props.makeStyles?.()
   }
   componentDidUpdate() {
-    // @ts-expect-error ts-migrate(2722) FIXME: Cannot invoke an object which is possibly 'undefin... Remove this comment to see the full error message
-    this.props.makeStyles()
+    this.props.makeStyles?.()
```
matyasf, line 174: Is this the same thing as `this.props?.makeStyles()`

packages/ui-view/src/View/types.ts
```diff
@@ -0,0 +1,295 @@
...
+  ThemeablePropTypes,
+  WithStyleProps,
+  Spacing,
+  BorderWidth,
+  BorderRadii,
+  Shadow,
+  Stacking
+} from '@instructure/emotion'
+
+export type ViewOwnProps = {
```
matyasf, line 40: Is this the correct approach? As we've seen a file that only exports types can cause issues

Brailor: I dont think it will be an issue in this case, because this file also exports a list of possible prop keys which is not a TS only thing (it is a runtime variable).

matyasf (approved):

packages/ui-badge/src/Badge/index.tsx
```diff
@@ -133,11 +133,11 @@ class Badge extends Component<Props> {
     formatOverflowText: (_count: number, countUntil: number) =>
       `${countUntil - 1} +`
   }
-  _defaultId: string
```
ke1k0, line 136: was this removed by accident?

packages/ui-byline/src/Byline/__tests__/Byline.test.tsx
```diff
@@ -82,53 +81,4 @@ describe('<Byline />', async () => {
...
-
-    const ignoreProps = [
-      'theme',
-      'children',
-      'styles',
-      'makeStyles',
-      'elementRef'
-    ]
-
-    Object.keys(View.propTypes)
```
ke1k0, line 108: Shouldn't we keep these tests? We can now iterate through `View.allowedProps` instead of `Object.keys(View.propTypes)`.

Edit: I can see that you made tests for the utils, which is awesome and is probably enough, but I'm afraid to delete tests :) These test it from a little different approach after all.

packages/ui-react-utils/src/passthroughProps.ts
```diff
@@ -48,13 +50,16 @@ const omit = (originalObject: Record<string, unknown>, keys: string[]) => {
   return newObject
 }
 
-function omitProps(
-  props: Record<string, any>,
-  propTypes: Record<string, any>,
+function omitProps<T extends Record<string, any>>(
```
ke1k0, line 53: Now that you know how these utils work exactly, could you please add a few lines of description (js annotations) above them? :) Just so the IDE can help if we hover over the names.

packages/ui-view/src/View/types.ts
```diff
@@ -0,0 +1,295 @@
...
+  focusPosition: PropTypes.oneOf(['offset', 'inset']),
+  focusColor: PropTypes.oneOf(['info', 'inverse', 'success', 'danger']),
+  shouldAnimateFocus: PropTypes.bool,
+  withVisualDebug: PropTypes.bool,
+  makeStyles: PropTypes.func,
+  styles: PropTypes.object,
+  dir: PropTypes.oneOf(Object.values(bidirectional.DIRECTION))
+}
+
+export const allowedProps: AllowedPropKeys = [
```
ke1k0, line 260: I would add some comments above it about what it and why we use it like that

packages/ui-react-utils/src/passthroughProps.ts
ke1k0: Actually, there are separate files (pickProps.ts, omitProps.ts...) next to this... Probably for the doc page to generate pages for them, but then the code should be separated too, what do you think?

packages/ui-react-utils/src/passthroughProps.ts
```diff
@@ -74,13 +79,16 @@ function pick(obj: Record<string, unknown>, keys: string[]) {
   return res
 }
 
-function pickProps(
-  props: Record<string, any>,
-  propTypes: Record<string, any>,
+function pickProps<T extends Record<string, any>>(
```
ke1k0, line 82: can we somehow indicate in these utils, that they return with part of the object that was passed?

packages/ui-view/src/ContextView/index.tsx
Brailor: To be honest, I'm not sure.
According to the [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#optional_chaining_with_function_calls) this the correct way of using optional chaining if we want to call an optional method property.

packages/emotion/src/withStyle.tsx
```diff
@@ -30,6 +31,12 @@ import { useTextDirectionContext } from '@instructure/ui-i18n'
 import { bidirectionalPolyfill } from './styleUtils/bidirectionalPolyfill'
 import { getComponentThemeOverride } from './getComponentThemeOverride'
 import { useTheme } from './useTheme'
+import { CSSObject } from '@emotion/react'
+
+export type WithStyleProps = Partial<{
+  styles: CSSObject
+  makeStyles: (...extraArgs: unknown[]) => void
```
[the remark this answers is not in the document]
Brailor: Right, I'll add it here in my other [PR](https://github.com/instructure/instructure-ui/pull/648)

packages/ui-badge/src/Badge/index.tsx
Brailor: You are indeed correct, it was deleted by mistake

packages/ui-byline/src/Byline/__tests__/Byline.test.tsx
Brailor: I've re-added these kind of tests.

packages/ui-react-utils/src/passthroughProps.ts
Brailor: Done

packages/ui-view/src/View/types.ts
Brailor: I've added some explanation comments, could you check it again please?

packages/ui-react-utils/src/passthroughProps.ts
Brailor: I've tried to do it, but I was not successful because these utils are highly dynamic, maybe we should give it another go some time later?

packages/ui-view/src/View/types.ts
ke1k0: Thanks!

ke1k0 (approved):

packages/ui-react-utils/src/passthroughProps.ts
ke1k0: I bumped into the same issue in other places too :) Let's try to type them once all the components are finalised.

PR #627, opened 2021-06-30
feat(instui-config,ui-dom-utils,ui-img): remove supportObjectFit util (INSTUI-3141)

HerrTopi:
which was necessary only for old, unsupported browsers

Closes: INSTUI-3141


packages/ui-dom-utils/src/index.ts
```diff
@@ -50,6 +50,5 @@ export { ownerDocument } from './ownerDocument'
 export { ownerWindow } from './ownerWindow'
 export { requestAnimationFrame } from './requestAnimationFrame'
 export type { RequestAnimationFrameType } from './requestAnimationFrame'
-export { supportsObjectFit } from './supportsObjectFit'
```
ke1k0, line 53: Hm, removing it now like this is a breaking change, someone might use this util somewhere. Should we mark it deprecated for now and remove it in v9?

ke1k0 (approved):

packages/ui-dom-utils/src/index.ts
ke1k0: Since it was a util needed for IE11, and we dropped support for it a long time ago, this doesn't count as breaking change.

packages/instui-config/codemod-configs/v6/imports.config.js
```diff
@@ -535,7 +535,6 @@ module.exports = {
           'ownerDocument',
           'ownerWindow',
           'requestAnimationFrame',
-          'supportsObjectFit',
```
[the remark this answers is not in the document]
HerrTopi: Good catch, fixed

Brailor (approved):

PR #628, opened 2021-06-30
INSTUI-3102 uid: type everything correctly

matyasf:


packages/uid/src/uid.ts
```diff
@@ -36,30 +36,29 @@ const dictionaryLengthMinus1 = dictionary.length - 1
...
 }
 
-// @ts-expect-error ts-migrate(7006) FIXME: Parameter 'size' implicitly has an 'any' type.
-function _random(size) {
+function _random(size: number) {
   const result = []
   /* eslint-disable-next-line no-param-reassign */
   while (0 < size--) {
-    result.push(Math.floor(Math.random() * 256))
+    // `| 0` is faster than `Math.floor()`.
```
matyasf, line 55: this idea is from https://github.com/ai/nanoid/blob/main/non-secure/index.js

ke1k0: I found [some warnings](https://stackoverflow.com/questions/2526682/why-is-javascripts-math-floor-the-slowest-way-to-calculate-floor-in-javascript) about this method (`Math.floor()` and `n | 0` are not equivalent), but might probably work here, since we know it is a positive number, etc.

ke1k0 (approved):

Brailor (approved):

PR #629, opened 2021-07-01
INSTUI-3055 ui-i18n: add correct TypeScript typing

matyasf:


packages/ui-i18n/src/bidirectional.tsx
```diff
@@ -21,12 +21,28 @@
...
-// @ts-expect-error ts-migrate(7016) FIXME: Could not find a declaration file for module 'hois... Remove this comment to see the full error message
 import hoistNonReactStatics from 'hoist-non-react-statics'
 
+// This is a workaround because TS cannot take type information from
+// decorators into account. This type needs to be added to every component,
+// that uses the bidirectional decorator.
+// see https://github.com/microsoft/TypeScript/issues/4881
+export type BidirectionalProps = {
+  dir?: 'ltr' | 'rtl'
+}
```
matyasf, line 35: How do you like this pattern? Its a workaround, but I could not think of anything better

Brailor: I did something similar with the the `withStyle` decorator in my PR [here](https://github.com/instructure/instructure-ui/blob/8c83dcc9c1174ad469e4f5e90042572754196b26/packages/emotion/src/withStyle.tsx#L37-L40,  so I think this is a reasonable solution, but in the long run I think it would be the best to rewrite all of our current decorators to [Higher Order Components](https://reactjs.org/docs/higher-order-components.html) and hooks for the functional components.

packages/ui-drawer-layout/src/DrawerLayout/DrawerTray/index.tsx
```diff
@@ -37,6 +37,7 @@ import { mirrorHorizontalPlacement } from '@instructure/ui-position'
 import generateStyle from './styles'
 import generateComponentTheme from './theme'
 import { DrawerLayoutContext } from '../index'
+import { BidirectionalProps } from '@instructure/ui-i18n/src/bidirectional'
```
Brailor, line 40: I think we should export the `BidirectionalProps` from the index of the `ui-i18n` package

Brailor (approved):

HerrTopi (approved):

packages/ui-drawer-layout/src/DrawerLayout/DrawerTray/index.tsx
ke1k0: I agree, we shouldn't import anything from /src

matyasf: yep, fixed

PR #633, opened 2021-07-06
chore(ui-responsive): add TS types to `ui-responsive` package

ke1k0:
Closes: INSTUI-3077

### DIFF
--- a/packages/ui-responsive/src/addElementQueryMatchListener.ts
+++ b/packages/ui-responsive/src/addElementQueryMatchListener.ts
@@ -21,10 +21,14 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-import { parseQuery } from './parseQuery'
+import React from 'react'
+
+import { parseQuery, Size } from './parseQuery'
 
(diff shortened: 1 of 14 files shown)


packages/ui-responsive/src/addElementQueryMatchListener.ts
```diff
@@ -76,27 +80,35 @@ import { debounce } from '@instructure/debounce'
...
+  query: BreakpointQueries,
+  el:
+    | Node
+    | Window
+    | React.ReactElement
+    | React.Component
+    | ((
+        ...args: any[]
+      ) => Node | Window | React.ReactElement | React.Component),
+  cb: (queriesMatching: QueriesMatching) => any
```
matyasf, line 98: `cb` is `UpdateMatches` type (from index.tsx)

The signature of this method should be the same as `addMediaQueryMatchListener`, I'd extract it to a type.

matyasf: otherwise it looks good!

ke1k0: `addMediaQueryMatchListener` sadly has an extra parameter `matchMedia` :S I couldn't find a way to extract it in a way so one can have an extra parameter and the other doesn't.

Brailor (approved):

matyasf (approved):
looks good!

PR #634, opened 2021-07-06
Tests can be run directly from the IDE

matyasf:
Tested in Webstorm, you can now use the small green arrows in test files to run or debug tests.

Now the little play buttons beside tests can be used to run and debug tests. The following changes were needed:
-Removed the Webpack test loader plugin. This had an extra functionality that it was adding a test to each package that just checked that it loads, this is gone.
-commented out the non-working generateA11yTests() test, because it was silently throwing errors making debugging a
pain
-Replaced the DEBUG env var with NO_DEBUG that does the opposite, so tests are generated as debug builds by default
-small changes to karma.config to get debugging to work
-slight refactor to ReactComponentWrapper, so it doesnt spam warnings
TEST PLAN:
run tests in normal and watch mode, run a single test in these, try to debug a test in WebStorm and
in the browser


packages/ui-karma-config/lib/index.js
```diff
@@ -21,20 +21,20 @@
...
 
 const noLaunchers = process.argv.some((arg) => arg === '--no-launch')
 const noHeadless = process.argv.some((arg) => arg === '--no-headless')
 const randomizeTestOrder = process.argv.some((arg) => arg === '--randomize')
 
 const baseWebpackConfig = require('@instructure/ui-webpack-config')
 
-const DEBUG = process.env.DEBUG
+const NO_DEBUG = !!process.env.NO_DEBUG
 const withCoverage = process.env.COVERAGE
```
matyasf, line 36: Now tests are made with withe the DEBUG flag by default, so breakpoints can be placed.

packages/ui-test-sandbox/src/utils/reactComponentWrapper.tsx
```diff
@@ -44,7 +44,9 @@ class WrapperComponent extends React.Component<WrapperProp, WrapperProp> {
 
   constructor(props: WrapperProp) {
     super(props)
-    this.state = this.props
+    // wrapping props and destructuring is a workaround to silence the
+    // "It is not recommended to assign props directly to state" warning
+    this.state = { ...this.props }
```
matyasf, line 49: this is unrelated, just a fix for React spamming the console with warnings.

packages/ui-test-utils/src/utils/generateA11yTests.tsx
```diff
@@ -45,21 +41,24 @@ type ExampleSection = {
...
-}: ComponentExample) => <Component key={key} {...componentProps} />
+//const renderExample = ({
+//  Component,
+//  componentProps,
+//  key
+//}: ComponentExample) => <Component key={key} {...componentProps} />
 
 // this is coming from generateComponentExamples.js in __examples__
 // TODO This is broken. It was using the examples-loader to preprocess the data.
 export function generateA11yTests(
```
Brailor, line 52: This is technically a breaking change, but since nobody should be using this utility anyways it is okay in this case to merge it.

Brailor (approved):

Brailor:
Unfortunately neither the run test cases from the IDE and neither the breakpoint/debugger feature doesn't seem to work for me in VS Code. 

https://user-images.githubusercontent.com/20640237/124713001-42e17d00-df00-11eb-9122-54802350d4c6.mov

matyasf:
@Brailor I would not expect the breakpoints to work if you just run the tests from the console, they must be somehow launched from the IDE. I was thinking of some plugin that adds some nice buttons: https://github.com/Raagh/angular-karma_test-explorer

ke1k0 (approved):
Seems to work fine now :)

PR #635, opened 2021-07-06
chore(ui-portal): add TS types to `ui-portal` package

ke1k0:
Closes: INSTUI-3071


packages/ui-text-input/src/TextInput/index.tsx
```diff
@@ -65,7 +65,7 @@ type Props = {
   onFocus?: (...args: any[]) => any
   makeStyles?: (...args: any[]) => any
   styles?: any
-}
+} & InputHTMLAttributes<Props>
```
ke1k0, line 68: I think we will need this on every Component, that we add all available props that can be added on that element (eg. aria tags)

packages/ui-portal/src/Portal/ReactPortal.tsx
```diff
@@ -21,21 +21,28 @@
...
 
+import { PortalNode } from './PortalType'
+
 type Props = {
   open?: boolean
-  onOpen?: (DOMNode: HTMLSpanElement | null) => any
-  onClose?: (...args: any[]) => any
-  mountNode?: any // TODO: PropTypes.oneOfType([element, PropTypes.func]),
+  onOpen?: (DOMNode: PortalNode) => any
+  onClose?: () => any
```
matyasf, line 37: I think it would be nicer if we'd have `void` as a return type for these callbacks since we're not awaiting any return value from these methods.

packages/ui-portal/src/Portal/index.tsx
```diff
@@ -22,25 +22,29 @@
...
 
 import { element } from '@instructure/ui-prop-types'
 
 import { ReactPortal } from './ReactPortal'
 import { SubtreePortal } from './SubtreePortal'
+import { BidirectionalProps } from '@instructure/ui-i18n'
+
+import { PortalNode } from './PortalType'
 
 const IS_CREATE_PORTAL_SUPPORTED = typeof ReactDOM.createPortal === 'function'
```
matyasf, line 37: As I see this is supported since React 16, so we can delete SubTreePortal 
https://newbedev.com/how-to-use-reactdom-createportal-in-react-16
https://instructure.atlassian.net/browse/INSTUI-2287

packages/ui-portal/src/Portal/ReactPortal.tsx
```diff
@@ -21,21 +21,28 @@
...
+  mountNode?: Element | (() => Element) | null
   insertAt?: 'bottom' | 'top'
-  elementRef?: (...args: any[]) => any
+  elementRef?: (el: PortalNode) => any
```
matyasf, line 40: as I see elementRef is never null (see index.tsx) so this is not optional. This will remove the ugly non-null assertions too.

packages/ui-portal/src/Portal/ReactPortal.tsx
ke1k0: done, thanks!

packages/ui-portal/src/Portal/index.tsx
ke1k0: Nice catch! Removed it and simplified the Portal to use the native React portal.

packages/ui-portal/src/Portal/ReactPortal.tsx
ke1k0: True, but now that I moved this component to index.tsx, it remains an optional prop :)

packages/ui-react-utils/src/passthroughProps.ts
```diff
@@ -84,8 +84,10 @@ function pickProps(
   return pick(props, combined)
 }
 
-function passthroughProps(props: Record<string, any>) {
-  const validProps: Record<string, unknown> = {}
+function passthroughProps<Props extends Record<string, any>>(
+  props: Props
+): Partial<Props> {
```
Brailor, line 89: I'm not sure about this `Partial<Props>` return type here, becuase it could lead to incorrect typing on the consuming side.
For example if you have a required property in `Props` and you pass the whole props object to `passthroughProps` then that required property now will display as an optional property.
Another problem is that the `style, className, children, styles, makeStyles` will be shown as actual properties on the returned object even tough `passthroughProps` clearly filters those properties out.

I know it is really hard to type these kind of really dynamic solutions without rewriting them alltogheter, but I think we could do a type casting for the returned value, such as:

```ts
...
return validProps as Omit<Props, 'style' | 'className' | 'styles' | 'makeStyles'>
```

This will keep the required props being required and at least will not show those style and className related props.

matyasf (approved):

packages/ui-portal/src/Portal/index.tsx
```diff
@@ -22,24 +22,32 @@
...
-  onOpen?: (DOMNode: HTMLSpanElement | null) => any
-  onClose?: (...args: any[]) => any
-  mountNode?: any // TODO: PropTypes.oneOfType([element, PropTypes.func]),
+  onOpen?: (DOMNode: PortalNode) => void
+  onClose?: () => void
+  mountNode?: Element | (() => Element) | null
   insertAt?: 'bottom' | 'top'
-  elementRef?: (...args: any[]) => any
+  elementRef?: (el: PortalNode) => void
+} & HTMLAttributes<Props> &
```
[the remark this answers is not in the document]
ke1k0: Thanks!

packages/ui-react-utils/src/passthroughProps.ts
ke1k0: I reverted this change, because it was not needed for the typing of Portal, and it seems more complicated. Let's wait until we have the correct typing of Components and decorators/HOCs, and come back to this.

Brailor (approved):

PR #640, opened 2021-07-11
chore(ui-i18n): make docs search more fluid

HerrTopi:
Closes: INSTUI-2818


packages/ui-i18n/package.json
```diff
@@ -35,6 +35,7 @@
     "@instructure/ui-prop-types": "8.6.0",
     "@instructure/ui-react-utils": "8.6.0",
     "@instructure/ui-utils": "8.6.0",
+    "dayjs": "^1.10.6",
```
Brailor, line 38: I don't think this is needed here

HerrTopi: Nice catch, removed

Brailor (approved):

PR #642, opened 2021-07-12
docs: render ts based types instead of propTypes

Brailor:
Closes: INSTUI-3183

### DIFF
--- a/packages/__docs__/src/Figure/index.js
+++ b/packages/__docs__/src/Figure/index.js
@@ -23,10 +23,10 @@
  */
 
 /** @jsx jsx */
-import React, { Component } from 'react'
+import React, { Component, ReactElement } from 'react'
 import PropTypes from 'prop-types'
 
-import { omitProps, ComponentIdentifier } from '@instructure/ui-react-utils'
+import { omitProps, ensureSingleChild } from '@instructure/ui-react-utils'
 import { Children } from '@instructure/ui-prop-types'
 import {
   IconCheckMarkSolid,
@@ -44,8 +44,22 @@ import generateComponentTheme from './theme'
 
 import { Heading } from '../Heading'
 
-class FigureItem extends ComponentIdentifier {
+class FigureItem extends Component {
   static displayName = 'FigureItem'
+
+  static propTypes = {
+    children: PropTypes.node
+  }
+
+  static defaultProps = {
+    children: null
+  }
+
+  render() {
+    const { children, ...props } = this.props
+
+    return ensureSingleChild(children, props)
+  }
 }
 
 @withStyle(generateStyle, generateComponentTheme)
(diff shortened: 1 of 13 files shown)


ke1k0:
Update: marked it as draft, because we talked about it with Viktor, and there is still room for improvement here. I volunteered to take over this ticket and try to play with it a bit more.

packages/__docs__/src/Figure/index.js
```diff
@@ -44,8 +44,22 @@ import generateComponentTheme from './theme'
 
 import { Heading } from '../Heading'
 
-class FigureItem extends ComponentIdentifier {
+class FigureItem extends Component {
```
ke1k0, line 47: I refactored this, because ComponentIdentifier was marked as deprecated, and was throwing warnings.

packages/__docs__/src/Properties/index.js
```diff
@@ -92,6 +107,44 @@ class Properties extends Component {
...
+        elements.forEach((element) => {
+          if (element.name !== 'literal') {
+            isEnum = false
+          }
+        })
+        return isEnum ? 'enum' : 'union'
+      case 'signature':
+        switch (type) {
+          case 'function':
+            return 'func'
```
ke1k0, line 133: I tried to call them the same way as before, also shorter names help a lot with the width of the table column

packages/ui-portal/src/Portal/index.tsx
```diff
@@ -56,42 +80,20 @@ type State = {
 category: components/utilities
 ---
 @module Portal
+@tsProps
```
ke1k0, line 83: This is the first notable change: if you want to turn on the Properties generation from the ts props, add this label in the description

packages/ui-portal/src/Portal/index.tsx
```diff
@@ -34,15 +34,39 @@ import { bidirectional, BidirectionalProps } from '@instructure/ui-i18n'
 /**
  * The DOM Node of the Portal. It is created as a `span` element.
  */
-type PortalNode = HTMLSpanElement | null
+type PortalNode = HTMLSpanElement | null | undefined
 
 type OwnProps = {
+  /**
```
ke1k0, line 40: Moved comments here from PropTypes, we should keep them here if we want to get rid of PropTypes eventually (this could be a step in the ts typing.)

packages/__docs__/src/Figure/index.js
matyasf: yep, that was annoying

packages/ui-portal/src/Portal/index.tsx
matyasf: yep, I agree

matyasf (approved):
looks good!

HerrTopi (approved):

PR #647, opened 2021-07-13
INSTUI-3205 - avatar updates

ke1k0:


matyasf:
Chromatic tests are looking different, are they OK?

ke1k0:
> Chromatic tests are looking different, are they OK?

@matyasf well yes... the changes in Avatar were intended (and approved), and the Menu and Popover changes are just the usual positioning issues that Chromatic has with Popover... still waiting for a working fix, that second one....

matyasf:
Likely offtopic, but the default box-shadow is so weak, that its practically visible, it has 12% alpha
image
 (both in v7 and v8), we should investigate if its intentional. On the right image I've set the alpha to 100%

matyasf (approved):

ke1k0:
> Likely offtopic, but the default box-shadow is so weak, that its practically invisible, it has 12% alpha
> image
> (both in v7 and v8), we should investigate if its intentional. On the rightmost image I've set the alpha to 100%, its still a quite soft shadow

I made the changes based on the figma link in the ticket, and even asked to double check, so 12% is intentional. I find it a bit weak too, but it has s different effect based on the color of the image (looks stronger on a lighter image), so I made it "overridable" in the theme variables, if needed.

Brailor (approved):

PR #648, opened 2021-07-14
Create types for EmotionOverride's `theme` property

Brailor:


packages/types/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "license": "MIT",
+  "main": "src/index.ts",
+  "types": "dist/index.d.ts",
+  "private": true,
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/instructure/instructure-ui.git"
+  },
+  "scripts": {
+    "build:types": "tsc -p tsconfig.build.json"
```
matyasf, line 16: dont we need a lint script here?

packages/emotion/src/EmotionThemeProvider/index.tsx
```diff
@@ -79,16 +104,15 @@ function EmotionThemeProvider({ children, theme = {} }: any) {
...
  * the overrides merged together.
  *
- * @param {object} theme - A full theme or an override object
+ * @param {object} themeOrOverride - A full theme or an override object
  * @returns {function} A function that returns with the theme object for the [ThemeProvider](https://emotion.sh/docs/theming#themeprovider-reactcomponenttype)
+ * @module getTheme
  */
-const getTheme = (theme: any) => (ancestorTheme = {}) => {
-  // themeable themes have a 'key' property (= name of the theme),
-  // so without it it's just an overrides objects
```
matyasf, line 87: I'd leave this comment above the `isBaseTheme` method so its clear why it works.

packages/types/tsconfig.build.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../../tsconfig.build.json",
+  "compilerOptions": {
+    "outDir": "./dist"
```
matyasf, line 4: This should output into a ./types folder, like the other packages IMO. This current solution confuses my IDE a bit (for some reason it thinks that code in ./dist is not autogenerated

packages/ui-avatar/src/Avatar/index.tsx
```diff
@@ -80,7 +80,7 @@ category: components
 @withStyle(generateStyle, generateComponentTheme)
 @testable()
 class Avatar extends Component<Props> {
-  static componentId = 'Avatar'
+  static componentId = 'Avatar' as const
```
matyasf, line 83: Why is this needed? Why not use `static readonly `?

packages/types/src/ComponentThemeMap.ts
```diff
@@ -0,0 +1,118 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+import { ThemeVariables } from './ComponentThemeVariables'
+
+type ComponentThemeVariables = {
+  Avatar: ThemeVariables['Avatar']
```
matyasf, line 28: as I see ThemeVariables is only used here. Instead of maintaining these 2 large arrays wouldnt it be better to just use stuff here directly from `ComponentThemeVariables`? for example: 

```
import * from './ComponentThemeVariables'

type ComponentThemeVariables = {
  Avatar: Avatar
```

packages/types/src/ComponentThemeVariables.ts
```diff
@@ -0,0 +1,1333 @@
...
+import { Colors } from './Colors'
+
+export type Avatar = {
+  color: Colors['textBrand']
+  background: Colors['backgroundLightest']
+  borderWidthSmall: Border['widthSmall']
+  borderWidthMedium: Border['widthMedium']
+  borderColor: Colors['borderMedium']
+  fontFamily: Typography['fontFamily']
+  fontWeight: Typography['fontWeightBold']
```
ke1k0, line 48: Avatar got an update since, has a few extra theme variables, please add those too. Also, this is a good example why we should type the component theme variable object too, so the IDE can warn us:

```js
// Avatar theme.js
import { Avatar } from '@instructure/types'

const generateComponentTheme = (theme) => {
    const componentVariables: Avatar = { ... }
    ...
}
```

I know it will be a pain to add it to all components, but I can help with it if you want to. I think we should do it in this PR.

packages/types/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@instructure/types",
+  "version": "8.7.0",
+  "description": "Shared TypeScript typings for Instructure UI packages.",
+  "author": "Instructure, Inc. Engineering and Product Design",
+  "homepage": "https://instructure.design/#types",
+  "license": "MIT",
+  "main": "src/index.ts",
+  "types": "dist/index.d.ts",
+  "private": true,
```
matyasf, line 10: As we've discussed this could be private, but we should state in the Readme that they should not use types that are for internal APIs because they might change.

packages/ui-select/src/Select/index.tsx
```diff
@@ -656,25 +656,28 @@ class Select extends Component<Props> {
...
+      visibleOptionsCount,
+      children
+    } = this.props
 
     let lastWasGroup = false
     const viewProps = isShowingOptions
       ? {
-          display: 'block',
-          overflowY: 'auto',
+          display: 'block' as const,
```
ke1k0, line 669: I'm not sure why this is needed here... I mean I can see it is needed, just don't completely understand how this works, can you please explain later? :)

packages/types/src/ComponentThemeVariables.ts
ke1k0: Also, because of this usage, I would rename these types to "AvatarTheme" etc, just so it indicates theme variables, not props or sth else.

packages/emotion/src/EmotionThemeProvider/index.tsx
```diff
@@ -79,16 +104,15 @@ function EmotionThemeProvider({ children, theme = {} }: any) {
...
-  const overrides = !theme.key ? theme : null
-
-  if (!overrides) {
-    return theme
+const getTheme = (themeOrOverride: ThemeOrOverride) => (
+  ancestorTheme: BaseTheme = canvas
```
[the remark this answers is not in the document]
Brailor: I did it this way because `{}` is not assignable to the `BaseTheme` type and since we are already default to `canvas` in the `useTheme` hook, I thought this will be fine to do.

packages/types/src/ComponentThemeMap.ts
```diff
@@ -0,0 +1,118 @@
...
+  CondensedButton: ThemeVariables['BaseButton']
+  IconButton: ThemeVariables['BaseButton']
+  Byline: ThemeVariables['Byline']
+  'Calendar.Day': ThemeVariables['CalendarDay']
+  Calendar: ThemeVariables['Calendar']
+  CheckboxFacade: ThemeVariables['CheckboxFacade']
+  ToggleFacade: ThemeVariables['ToggleFacade']
+  CodeEditor: ThemeVariables['CodeEditor']
+  'DrawerLayout.Content': ThemeVariables['DrawerLayoutContent']
+  'DrawerLayout.Tray': ThemeVariables['DrawerLayoutTray']
```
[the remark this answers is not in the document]
Brailor: You are right, they could be using these components directly, I only used the `componentId` property for these kind of components, but it would be better to list all of the possible variants.

packages/types/src/ComponentThemeVariables.ts
Brailor: Yeah, that is the reason I've made this "central" interface called `ThemeVariables` which contains all of the components theme variable typings, so one would use it like this: `ThemeVariables.Avatar` which I think is a bit less redundant than naming every component with <ComponentName>ThemeVariables.

Also I did not do the theme generator function typings on purpose in this PR, I it should be done in a different one, this PR is already pretty big.

packages/types/tsconfig.build.json
Brailor: Fixed

packages/ui-avatar/src/Avatar/index.tsx
Brailor: This was a mistake by me, I've made this change when I was exploring how to map the components to their generateTheme function, I've deleted it.

packages/types/src/ComponentThemeMap.ts
Brailor: You are correct, changed

packages/types/package.json
Brailor: I've changed this to be public and will explain in the package's readme that it should be used with caution because of the internal usage.

packages/emotion/src/EmotionThemeProvider/index.tsx
```diff
@@ -97,12 +121,19 @@ const getTheme = (theme: any) => (ancestorTheme = {}) => {
   const themeName = currentTheme.key
 
   // we pick the overrides for the current theme from the override object
-  const currentThemeOverrides = overrides?.themeOverrides?.[themeName] || {}
+  const currentThemeOverrides =
+    //@ts-expect-error fix this
```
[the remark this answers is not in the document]
Brailor: Done

packages/ui-select/src/Select/index.tsx
Brailor: I added this because of some strange TS error, but it has gone away, so this is not needed anymore, deleted.

packages/types/package.json
Brailor: Added the lint script

packages/emotion/src/EmotionThemeProvider/index.tsx
Brailor: Done

matyasf (approved):

packages/emotion/src/EmotionThemeProvider/index.tsx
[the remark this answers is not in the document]
ke1k0: Well, not sure it is a widely used use-case, but if someone wanted to use the EmotionThemeProvider on the top level, and wrap their app in their own InstUI compatible theme, it will always merge on top of canvas then, and might cause some unwanted results.

packages/types/src/ComponentThemeVariables.ts
ke1k0: Cool, thanks! Tell me if you need help with batch-typing the generators :)

packages/ui-select/src/Select/index.tsx
ke1k0: Weird :D Thanks!

packages/emotion/src/EmotionThemeProvider/index.tsx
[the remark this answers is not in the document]
Brailor: You are right, changed back to be `{}` and did a type casting instead

packages/types/src/ComponentThemeVariables.ts
Brailor: I forgot to include the new Avatar theme variables, added it now

Brailor: Also, yeah it would be nice to do the batch-typing parallel :)

ke1k0 (approved):
Nice job!

PR #654, opened 2021-07-20
Storybook build transitive dependents

Brailor:


packages/__examples__/buildScripts/__tests__/generateTransitiveDependents.test.js
```diff
@@ -0,0 +1,76 @@
+/*
```
matyasf, line 1: I think these can be .ts files since they are processed as the other tests

packages/__examples__/buildScripts/parseTransitiveDependents.js
```diff
@@ -0,0 +1,61 @@
...
+          '--json',
+          '--loglevel',
+          'silent'
+        ])
+      })
+    )
+
+    return rawPackages.map((output) => JSON.parse(output.stdout)).flat()
+  } catch (error) {
+    console.error('Possible undifned package given as input!')
```
matyasf, line 54: this has a type; I'd also output the error

matyasf (approved):
nice!

HerrTopi (approved):

packages/__examples__/buildScripts/parseTransitiveDependents.js
Brailor: Done

PR #655, opened 2021-07-22
INSTUI-3079: ui-selectable TS typing

matyasf:


packages/emotion/package.json
```diff
@@ -25,7 +25,7 @@
   "license": "MIT",
   "dependencies": {
     "@babel/runtime": "^7.13.10",
-    "@emotion/react": "11.0.0",
+    "@emotion/react": "11.4.0",
```
matyasf, line 28: This made it possible to make props that have `defaultProps` non optional

packages/ui-select/src/Select/index.tsx
```diff
@@ -520,7 +522,7 @@ class Select extends Component<Props> {
                 // @ts-expect-error ts-migrate(2345) FIXME: Argument of type 'null' is not assignable to param... Remove this comment to see the full error message
                 const index = this._optionIds.indexOf(highlightedOptionId)
                 highlightId =
-                  index > -1 ? this._optionIds[index + direction] : null
+                  index > -1 ? this._optionIds[index + direction!] : null
```
ke1k0, line 525: Hm, as far as I can tell, direction is not always provided, more like... it needs EITHER and id OR a direction. Probably not worth typing it correctly now, but could you please leave a comment for later?

packages/ui-selectable/src/Selectable/index.tsx
```diff
@@ -54,61 +152,17 @@ tags: autocomplete, typeahead, combobox, dropdown, search
 **/
 class Selectable extends Component<Props> {
```
ke1k0, line 153: It won't do anything for now, but could you please add the `@tsProps` flag to the description?
```
/**
---
category: components
tags: autocomplete, typeahead, combobox, dropdown, search
---
@tsProps
**/
```

This will tell the docs app to use the types for the docs page properties, once this PR is merged: https://github.com/instructure/instructure-ui/pull/642

ke1k0 (approved):

packages/ui-selectable/src/Selectable/index.tsx
matyasf: done

packages/ui-select/src/Select/index.tsx
matyasf: done

Brailor (approved):

PR #657, opened 2021-07-22
INSTUI-3215: Make storybook story generation better

matyasf:
story generation is much faster, now it does not hang/take minutes on some components. These were generating 0 stories, so there will be Chromatic diffs.

Better error messages when a story cannot be generated, that includes a stack trace

### DIFF
--- a/packages/__examples__/.storybook/stories/generateComponentExamples.js
+++ b/packages/__examples__/.storybook/stories/generateComponentExamples.js
@@ -24,30 +24,12 @@
 const { uid } = require('@instructure/uid')
 const generatePropCombinations = require('./generatePropCombinations')
 
-module.exports = function generateComponentExamples(
-  Component,
-  config = {
-    sectionProp: null,
-    maxExamplesPerPage: null,
-    propValues: {},
-    excludeProps: [],
-    getComponentProps: (props, index) => {
-      return {}
-    },
-    getExampleProps: (props, index) => {
-      return {}
-    },
-    getParameters: (examples, pageIndex) => {
-      return {}
-    },
-    filter: (props) => false
-  }
-) {
+module.exports = function generateComponentExamples(Component, config) {
   const { sectionProp, excludeProps, filter } = config
 
   const PROPS_CACHE = []
   const sections = []
-  const maxExamples = config.maxExamples | 500
+  const maxExamples = config.maxExamples
   let exampleCount = 0
   let propValues = {}
 
@@ -61,16 +43,18 @@ module.exports = function generateComponentExamples(
     return parameters
   }
 
+  /**
+   * Merges the auto-generated props with ones in the examples files,
+   * a prop returned by getComponentProps() takes priority
+   */
   const getComponentProps = (props) => {
     let componentProps = props
-
     if (typeof config.getComponentProps === 'function') {
       componentProps = {
         ...componentProps,
         ...config.getComponentProps(props)
       }
     }
-
     return componentProps
   }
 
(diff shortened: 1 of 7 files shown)


packages/__examples__/.storybook/stories/generateComponentExamples.js
```diff
@@ -24,30 +24,12 @@
...
-    },
-    getExampleProps: (props, index) => {
-      return {}
-    },
-    getParameters: (examples, pageIndex) => {
-      return {}
-    },
-    filter: (props) => false
-  }
-) {
```
matyasf, line 45: This was not needed since this method always gets a value

packages/__examples__/.storybook/stories/generateComponentExamples.js
```diff
@@ -132,12 +116,14 @@ module.exports = function generateComponentExamples(
...
-
-    if (!ignore && !PROPS_CACHE.includes(propsString)) {
+    if (ignore) {
+      return
+    }
+    const propsString = Object.entries(componentProps).sort().toString() //JSON.stringify(componentProps)
+    if (!PROPS_CACHE.includes(propsString)) {
+      const key = uid()
+      const exampleProps = getExampleProps(props)
       exampleCount++
```
matyasf, line 127: This method was the culprit for the bad performance. Its called >5000 times for some components and JSON.stringify is slow for such numbers. 
Now it exits early if something is filtered, and the Object.entries() method is way faster.

packages/__examples__/.storybook/stories/generateComponentExamples.js
```diff
@@ -132,12 +116,14 @@ module.exports = function generateComponentExamples(
 
   const maybeAddExample = (props) => {
```
matyasf, line 117: This method was the culprit for the bad performance. Its called >5000 times for some components and JSON.stringify is slow for such numbers.
Now it exits early if something is filtered, and the Object.entries() method is way faster.

packages/ui-alerts/src/Alert/__examples__/Alert.examples.ts
```diff
@@ -28,22 +28,17 @@ export default {
...
       margin: 'medium',
       screenReaderOnly: false,
       liveRegionPoliteness: 'polite',
       transition: 'none',
       open: true
     }
-  },
-  // @ts-expect-error ts-migrate(6133) FIXME: 'props' is declared but its value is never read.
-  filter: (props) => {
-    return props.screenReaderOnly || props.open === false
```
matyasf, line 47: If something is defined in getComponentProps, it can have that single value only

packages/__examples__/.storybook/stories/generateComponentExamples.js
ke1k0: To be honest, I was using this to see what options it accepts :D Well, it is available in the readme too. Do you think we should make a TS type for it so we can use that in the `xxx.examples.tsx` files?

ke1k0 (approved):
Waaay faster, thanks :) Are the changes in Chromatic all valid too?

packages/__examples__/.storybook/stories/generateComponentExamples.js
matyasf: I've added a JSDoc about the parameters, the Readme has detailed docs

matyasf:
> Waaay faster, thanks :) Are the changes in Chromatic all valid too?

Some of them changed. The original code was not generating some examples for some components (Alert, Flex, TextInput), because it took too much time. Now they generate examples and it turns out that they were generating way too many, so I needed to change their filters resulting in Chromatic changes.

Brailor (approved):

PR #659, opened 2021-07-23
INSTUI-3114 Test locator custom methods are type checked and autocompleted, better typing for test utils.

matyasf:


packages/ui-billboard/src/Billboard/__tests__/Billboard.test.tsx
```diff
@@ -115,7 +115,7 @@ describe('<Billboard />', async () => {
 
       const subject = await mount(<Billboard onClick={onClick} disabled />)
       const billboard = within(subject.getDOMNode())
-      await billboard.click(null, { clickable: false })
+      await billboard.click(undefined, { clickable: false })
```
matyasf, line 118: These were not seen by TS until now. The parameters are optional, so one has to use `undefined` instead of `null`, I've indicated this in the commit message

packages/ui-breadcrumb/src/Breadcrumb/BreadcrumbLocator.ts
```diff
@@ -30,10 +30,10 @@ export { BreadcrumbLinkLocator }
 
 // @ts-expect-error ts-migrate(2339) FIXME: Property 'selector' does not exist on type 'typeof... Remove this comment to see the full error message
 export const BreadcrumbLocator = locator(Breadcrumb.selector, {
-  findAllLinks: (...args) => {
+  findAllLinks: (...args: any[]) => {
```
matyasf, line 33: again these were not seen by TS before

packages/ui-test-locator/src/utils/locator.ts
```diff
@@ -45,26 +52,32 @@ export function locator(
...
         ...customMethods,
         ...options.customMethods
       }
-    })
+      // Adding here the generic T argument makes TS able to autocomplete custom
+      // methods (if they are a compile time constant)
+      // Typing this properly would require to make QueryArguments generic,
+      // which would introduce a tons of complexity, so a refactor of this
+      // library would be preferred.
+    }) as Promise<QueriesHelpersEventsType[] & ObjWithCutFirstArg<T>>
```
matyasf, line 75: This makes it possible to use autocomplete with custom methods in locators.

packages/ui-test-queries/src/utils/bindElementToEvents.ts
```diff
@@ -26,42 +26,65 @@ import keycode from 'keycode'
...
+    ) => Promise<Event | void>
+  } &
+  {
+    // keyboard event triggers handled by fireKeyboardEvent are a bit different
+    [K in 'keyDown' | 'keyPress' | 'keyUp']: (
+      whichKey?: string | number,
+      init?: FireEventInit,
+      options?: Record<string, unknown>
+    ) => Promise<Event>
+  } // TODO refactor this to make it more uniform
```
matyasf, line 55: This is a horror looking type, but I did not want to introduce breaking changes. We should refactor this code to be nicer and more TS friendly, it has tons of unnecessary abstractions.

matyasf, line 55: This is a horror looking type, but I did not want to introduce breaking changes. We should refactor this code to be nicer and more TS friendly, it has tons of unnecessary abstractions.

packages/ui-tree-browser/src/TreeBrowser/TreeNode/__tests__/TreeNode.test.tsx
```diff
@@ -90,7 +90,7 @@ describe('<TreeNode />', async () => {
       const item = await TreeNodeLocator.find()
       expect(item).to.exist()
       await wait(() => {
-        expect(item.getComputedStyle(item, ':after').opacity).to.equal('1')
+        expect(item.getComputedStyle().opacity).to.equal('1')
```
matyasf, line 93: This was a bug, this method does not need parameters

packages/ui-test-locator/src/utils/locator.ts
```diff
@@ -30,11 +30,18 @@ import {
...
-import { GenericFunction } from '@instructure/ui-test-queries/src/utils/bindElementToMethods'
+import { QueriesHelpersEventsType } from '@instructure/ui-test-queries/src/utils/bindElementToUtilities'
 
-export function locator(
+// Cut the first argument of function that's an object's value
+type ObjWithCutFirstArg<T> = {
+  [K in keyof T]: T[K] extends (_: any, ...tail: infer TT) => infer R // if a function
+    ? (...args: TT) => R // cut first argument
+    : never // or use the original value
+}
```
Brailor, line 40: This type is kinda is re-declared in the file `bindElementsToMethods.ts`, I think we should just export it from one place and import the type where its needed.

packages/ui-test-queries/src/utils/bindElementToMethods.ts
```diff
@@ -24,9 +24,26 @@
 
 export type GenericFunction = (...args: any) => unknown | any
 
-export function bindElementToMethods<K>(element: Element, methods: K) {
+// In an object where keys are functions remove their first argument
+type ObjWithCutFirstArg<T> = {
+  [K in keyof T]: T[K] extends (_: unknown, ...tail: infer TT) => infer R // if a function
+    ? (...args: TT) => R // cut first argument
+    : never // do not allow it, the method will throw an error
+}
```
Brailor, line 32: This is the duplicate type I mentioned in my earlier comment.
Also as I see we have a utility type which can be used in this type as well:
```ts
//bindElementToUtilites.ts
type CutFirstArg<F> = F extends (_: any, ...tail: infer TT) => infer R
  ? (...args: TT) => R
  : never
```

So we could reuse this type like this:

```ts
type ObjWithCutFirstArg<T> = {
  [K in keyof T]: CutFirstArg<T[K]>
}
```
What do you think?

packages/ui-test-locator/src/utils/locator.ts
Brailor: To be honest I'm not completely sure how this type works, as I see the Promise will contain an array and an object as well?
How that would look like in a real world example?

HerrTopi (approved):
Reviewed with the best of my ts knowledge

packages/ui-test-locator/src/utils/locator.ts
matyasf: hmm this was a bug, I fixed it

packages/ui-test-locator/src/utils/locator.ts
matyasf: yep, fixed

packages/ui-test-queries/src/utils/bindElementToMethods.ts
matyasf: yes, good idea

Brailor (approved):

PR #661, opened 2021-07-28
INSTUI-3191 - improve emotion performance caching

ke1k0:


matyasf:
We are closing it because it does not seem to bring real performance improvements

PR #662, opened 2021-07-28
INSTUI-3060: Type Metric and MetricGroup properly

matyasf:


packages/ui-react-utils/src/callRenderProp.ts
```diff
@@ -22,37 +22,53 @@
...
-      )
-    ) {
-      return value(props)
+    // fat arrow functions don't have a prototype
+    if (!(value.prototype && value.prototype.isReactComponent)) {
+      return (value as any)(props)
     }
-    return React.createElement(value, props)
+    // TODO type 'value' properly
+    return React.createElement<P>(value as any, props)
```
matyasf, line 71: The `value` prop here is a nightmare to type. `React.createElement` does not accept `ReactNode`, yet we seem to be passing `ReactNode`s to this method regularly.

packages/ui-metric/src/MetricGroup/__examples__/MetricGroup.examples.tsx
```diff
@@ -25,8 +25,7 @@
 import React from 'react'
 import { Metric } from '../../Metric'
 
-// @ts-expect-error ts-migrate(7006) FIXME: Parameter 'textAlign' implicitly has an 'any' type... Remove this comment to see the full error message
-const getMetric = (textAlign) => {
+const getMetric = (textAlign: any) => {
```
ke1k0, line 28: I'm wondering if it worth typing these... We know that e.g. here the value are `'start' | 'end' | 'center'`, we could even export this type from Metric to use here, but maybe we don't win anything by it.

packages/ui-metric/src/MetricGroup/index.tsx
```diff
@@ -42,6 +43,7 @@ type Props = {
 ---
 category: components
 ---
+@tsProps
```
ke1k0, line 46: The same as with Metric, adding the div attributes to the type would be useful.

packages/ui-metric/src/Metric/index.tsx
```diff
@@ -35,16 +35,21 @@ import generateComponentTheme from './theme'
...
+   * aria labels get set
+   */
+  isGroupChild: boolean
 }
 
 /**
 ---
 category: components
 ---
+@tsProps
```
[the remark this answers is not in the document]
matyasf: As we discussed the `class Metric extends Component<Props & OtherHTMLAttributes<Props>>` signature is the way to go here

packages/ui-metric/src/MetricGroup/__examples__/MetricGroup.examples.tsx
matyasf: we could export `Metric`'s props here, but I'm inclined not to do it, because it would not be used in other places

ke1k0 (approved):

HerrTopi (approved):

PR #663, opened 2021-07-29
Inst UI 3169/fix a11y test generator

Brailor:


packages/ui-test-utils/src/utils/generateComponentExamples.js
```diff
@@ -22,8 +22,9 @@
  * SOFTWARE.
  */
 
-const { uid } = require('@instructure/uid')
-const generatePropCombinations = require('./generatePropCombinations')
+import { nanoid } from 'nanoid'
```
Brailor, line 25: Had to use a 3rd party lib for generating the keys, because the `generateComponentExamples` utility was moved from the `__examples__` package, to the `ui-test-utils` which is a dependency for the `@instructure/uid` package, and that would cause a dependency cycle.

packages/ui-test-utils/src/utils/generateA11yTests.tsx
```diff
@@ -31,35 +35,28 @@ type ComponentExample = {
...
-// this is coming from generateComponentExamples.js in __examples__
-// TODO This is broken. It was using the examples-loader to preprocess the data.
+/**
+ *
+ * @param Component
+ * @param componentExample
+ * @param only
+ *
+ * @module generateA11yTests
+ * @private
```
matyasf, line 51: it would be nice to add a sentence here on what the parameters do

matyasf: Also better typing: `componentExample: Record<string, any>`

matyasf (approved):

HerrTopi (approved):

PR #667, opened 2021-08-03
docs: fix imported types not showing up in docs

Brailor:
This PR fix the issue with our current solution for importing propTypes and TS types not showing up in the documentation.
It uses an upcoming feature of `react-docgen`, thus I bumped its version from 5 to 6.alpha
For more info: https://github.com/reactjs/react-docgen/pull/352

### DIFF
--- a/packages/__docs__/buildScripts/utils/getReactDoc.js
+++ b/packages/__docs__/buildScripts/utils/getReactDoc.js
@@ -35,7 +35,8 @@ module.exports = function getReactDoc(source, fileName, error) {
       reactDocgen.resolver.findAllExportedComponentDefinitions,
       null,
       {
-        filename: fileName
+        filename: fileName,
+        importer: reactDocgen.importers.makeFsImporter()
       }
     )
     if (Array.isArray(doc)) {
--- a/packages/__docs__/package.json
+++ b/packages/__docs__/package.json
@@ -124,7 +124,7 @@
     "html-webpack-plugin": "^4",
     "jsdoc-api": "^6.0.0",
     "mkdirp": "^1",
-    "react-docgen": "^5",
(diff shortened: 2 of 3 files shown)


packages/__docs__/buildScripts/utils/getReactDoc.js
```diff
@@ -35,7 +35,8 @@ module.exports = function getReactDoc(source, fileName, error) {
       reactDocgen.resolver.findAllExportedComponentDefinitions,
       null,
       {
-        filename: fileName
+        filename: fileName,
+        importer: reactDocgen.importers.makeFsImporter()
```
Brailor, line 39: This new upcoming feature of `react-docgen` allows us to operate on imported types as well.

packages/__docs__/package.json
```diff
@@ -124,7 +124,7 @@
     "html-webpack-plugin": "^4",
     "jsdoc-api": "^6.0.0",
     "mkdirp": "^1",
-    "react-docgen": "^5",
+    "react-docgen": "^6.0.0-alpha.0",
```
Brailor, line 127: This new importer feature is not fully released in production yet, but according to the maintainers of the repo v6 won't contain any breaking changes, so it seems like a safe thing for us to use the pre-release version of the lib.

ke1k0: The yarn.lock is missing, please add it :) I couldn't find any problems with it, the View props display correctly now, and it could recognise the "query" and "props" prop of Responsive as "object" (typeof this and that) too. Well, it cannot go deeper than one level, but that's why we need that type-linking ticket too :)

ke1k0 (approved):

HerrTopi (approved):

PR #668, opened 2021-08-04
INSTUI-3200 - easy way to remove shadow from inline alerts

ke1k0:


packages/ui-checkbox/src/Checkbox/index.tsx
```diff
@@ -76,6 +76,7 @@ type Props = {
 /**
 ---
 category: components
+tags: toggle, switch
```
ke1k0, line 79: This is not relevant to Alert, but added these tags to Checkbox, so it shows up as a result if someone is typing "toggle" or "switch" to the search input (saw it multiple times on slack that new ppl cannot find this component)

Brailor (approved):

HerrTopi (approved):

PR #670, opened 2021-08-04
chore(ui-expandable,ui-toggle-details): add better TS types to `Expandable`

ke1k0:
Closes: INSTUI-3048

### DIFF
--- a/packages/ui-expandable/src/Expandable/index.tsx
+++ b/packages/ui-expandable/src/Expandable/index.tsx
@@ -26,69 +26,112 @@ import PropTypes from 'prop-types'
 import { controllable } from '@instructure/ui-prop-types'
 import { uid } from '@instructure/uid'
 import { createChainedFunction } from '@instructure/ui-utils'
-const toggleExpanded = ({ expanded }: any) => ({ expanded: !expanded })
-type ExpandableProps = {
-  expanded?: any // TODO: controllable(PropTypes.bool, 'onToggle', 'defaultExpanded')
-  defaultExpanded?: boolean
-  onToggle?: (...args: any[]) => any
-  render?: (...args: any[]) => any
+
+const toggleExpanded = ({ expanded }: { expanded: boolean }) => ({
+  expanded: !expanded
+})
+
+// TODO: try to type it so that the return object contains the prop object
+type GetToggleProps = (
+  props?: Record<string, any> & { onClick?: (...args: any[]) => any }
+) => {
+  'aria-controls': string
+  'aria-expanded': boolean
+  onClick?: (...args: any[]) => any
+  [p: string]: any
 }
-type ExpandableState = any
+
+type RenderProps = {
+  expanded: boolean
+  /**
+   * Props to be spread onto the trigger element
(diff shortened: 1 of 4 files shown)


packages/ui-expandable/src/Expandable/index.tsx
```diff
@@ -26,69 +26,112 @@ import PropTypes from 'prop-types'
 import { controllable } from '@instructure/ui-prop-types'
 import { uid } from '@instructure/uid'
 import { createChainedFunction } from '@instructure/ui-utils'
-const toggleExpanded = ({ expanded }: any) => ({ expanded: !expanded })
+
+const toggleExpanded = ({ expanded }: { expanded: boolean }) => ({
+  expanded: !expanded
+})
+
+// TODO: try to type it so that the return object contains the prop object
```
ke1k0, line 34: It was a nightmare to type, so in the end I just added a TODO. If you have a tip how to do it without making everything Generic, let me know please :)

matyasf: You could do something like:
```
type GetToggleProps = <P>(props: P & { onClick?: (...args: any[]) => any }) => {
  'aria-controls': string
  'aria-expanded': boolean
  onClick?: (...args: any[]) => any
} & P
```
[playground](https://www.typescriptlang.org/play?noUnusedParameters=true&target=3&inlineSourceMap=false&inlineSources=false&isolatedModules=false&noFallthroughCasesInSwitch=true&ts=4.3.5#code/C4TwDgpgBA4hwBUD2BzFAbCAFATksAzlALxQA8WAfABRh6EBcUWUAZFAN5RIB2AwugCWAYwDWAfibUAdLICGOFASZyeIANoBdAJQlKUVSCgBfXcX0cAUFCgByBYLkBaYb2B50BW0wLvBPFGs7B2cIAA8wVQATCCjvKAAjJCRMVSDeAREJKVlpBSUVNS0zfUNLYzZmS0tXHl8oYAhfJjhEVAxseiJSWi6STgq5IkMSziCceABXHB4xmxtcunwCABogm3scRxc3Dy8mWwBGQ4AmWzX54K3QiOjY+IAzOU8IC5Ny6sbfag4hqIYACymPIEKKWL7AH5-QHAv4PBJRIA)

matyasf: note that you will need a `as any` cast here: 
`getToggleProps: (props = {} as any) => {` because the default parameter would determine the generic type otherwise.

matyasf: Also onClick should have this signature: ` onClick?: (event: Event) => void`

matyasf: https://stackoverflow.com/questions/54225839/required-vs-inferred-generic-types-in-typescript

ke1k0: Thanks!

matyasf (approved):

Brailor (approved):

PR #676, opened 2021-08-11
try vol3. on emotion performance improvement

ke1k0:
Closes: INSTUI-3191


matyasf:
While this would bring ~10% performance improvement, its very hard to maintain

PR #678, opened 2021-08-12
Fix storybook story generation

matyasf:
SB stories were not generated when the .example.ts files had JSX children
Also made TS typing better and added a couple of tests.


packages/ui-test-utils/src/utils/generateComponentExamples.ts
```diff
@@ -169,22 +172,43 @@ export function generateComponentExamples<Props>(
...
+  const maybeAddExample = (props: Props): void => {
     const componentProps = mergeComponentPropsFromConfig(props)
     const ignore = typeof filter === 'function' ? filter(componentProps) : false
     if (ignore) {
       return
     }
-    // TODO this is wrong, it thinks that objects with object children are the same
-    const propsString = Object.entries(componentProps).sort().toString()
-    //const propsString = JSON.stringify(componentProps)
+    const propsString = fastSerialize(componentProps)
```
matyasf, line 204: This is the bugfix. `Object.entries(componentProps).sort().toString()` was just outputting something like `[Symbol]` for JSX nodes in example props. The current solution is faster that simply doing JSON serialization.

Brailor (approved):

ke1k0:
I tried running it, and it seems fine on first glance (Calendar seems good), but the Breadcrumb is still breaking. Can you check it please? Or wasn't @Brailor doing a fix for it?

matyasf:
> I tried running it, and it seems fine on first glance (Calendar seems good), but the Breadcrumb is still breaking. Can you check it please? Or wasn't @Brailor doing a fix for it?

yea, that will be fixed in a different PR

HerrTopi (approved):

PR #682, opened 2021-08-13
INSTUI-3072 - TS typing ui-position

ke1k0:


packages/ui-dom-utils/src/ownerDocument.ts
```diff
@@ -36,7 +36,13 @@ import React from 'react'
  * @returns { Document } the owner document
  */
 function ownerDocument(
-  el?: Node | Window | React.ReactElement | ((...args: any[]) => any) | null
+  el?:
+    | Node
+    | Window
+    | React.ReactElement
+    | React.Component
+    | ((...args: any[]) => any)
```
matyasf, line 44: I think this function should return `Node | Window | null | undefined` too. Also as I see in `findDOMNode` it has actually 0 arguments, so the signature should be `(() => Node | Window | null | undefined)`

packages/ui-position/src/Position/index.tsx
```diff
@@ -243,44 +172,43 @@ class Position extends Component<Props> {
   }
 
   componentWillUnmount() {
-    // @ts-expect-error ts-migrate(2339) FIXME: Property 'cancel' does not exist on type '() => vo... Remove this comment to see the full error message
+    // @ts-expect-error TODO: figure out if this is valid
     this.position.cancel()
```
matyasf, line 176: This is valid code (but pretty weirdly written), Ive checked in the debugger.
`position` is a medhod first (the one that calls `setState`), but in the constructor this variable is reassigned to a function returned by `debounce`, and this one has a `cancel` method. Its more clear in the compiled ES6 code what is happening:

```
  constructor(props) {
    ...
    this.position = () => {
      this.setState({
        positioned: true,
        ...this.calculatePosition(this.props)
      });
    };
    this.position = debounce(this.position, 0, {
      leading: false,
      trailing: true
    });
  }
```

IMO we should refactor this to get rid of this confusing code (its OK to do it in another ticket)

packages/ui-position/src/mirrorPlacement.ts
```diff
@@ -32,13 +42,17 @@ const mirror = {
   offscreen: 'offscreen'
 }
 
-// @ts-expect-error ts-migrate(7006) FIXME: Parameter 'placement' implicitly has an 'any' type... Remove this comment to see the full error message
-function mirrorPlacement(placement, delimiter) {
+function mirrorPlacement<D extends string | undefined = undefined>(
```
matyasf, line 45: Is this typed correctly?
When I try the examples in the bottom of this file they give me TS errors:
```
mirrorPlacement("start end") // error
mirrorPlacement('top start', ' ') // error
mirrorPlacement(["start", "end"]) // OK
```

packages/ui-dom-utils/src/ownerDocument.ts
ke1k0: I ended up making a new `UIElement` type in the shared-types package and used that everywhere where relevant.

packages/ui-position/src/Position/index.tsx
ke1k0: For now I made a `Debounced` type that I could use here.

packages/ui-position/src/mirrorPlacement.ts
ke1k0: True, I accidentally deleted something during a refactor. I even made tests for the 2 mirror utils, they didn't have one before.

matyasf (approved):

Brailor (approved):

PR #683, opened 2021-08-13
NSTUI-3159 better types for View and refactoring for OmitProps/passtroughProps

matyasf:
Refactor omitViewProps and omitProps so its more clear what they do. This resulted in a few small issues coming up (e.g. null used instead of undefined)
remove "?" from View props that have default value

### DIFF
--- a/packages/ui-options/src/Options/index.tsx
+++ b/packages/ui-options/src/Options/index.tsx
@@ -181,8 +181,7 @@ class Options extends Component<Props> {
           margin="none"
           padding="none"
           background="primary"
-          // TODO renderLabel is returned is renderLabel is a function, this is likely a bug
-          aria-labelledby={(renderLabel as string) && this._labelId}
+          aria-labelledby={renderLabel ? this._labelId : undefined}
         >
           {this.renderChildren()}
         </View>
--- a/packages/ui-options/src/Options/index.tsx
+++ b/packages/ui-options/src/Options/index.tsx
@@ -162,14 +162,12 @@ class Options extends Component<Props> {
   }
 
   render() {
-    // @ts-expect-error ts-migrate(2339) FIXME: Property 'omitViewProps' does not exist on type 't... Remove this comment to see the full error message
     const passthroughProps = View.omitViewProps(
       omitProps(this.props, Options.propTypes),
       Options
     )
 
     const { as, role, elementRef, renderLabel, styles } = this.props
-
     return (
       <div css={styles.options} role="presentation">
         {renderLabel && this.renderLabel()}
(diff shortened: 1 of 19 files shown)


packages/ui-options/src/Options/index.tsx
```diff
@@ -183,7 +181,8 @@ class Options extends Component<Props> {
           margin="none"
           padding="none"
           background="primary"
-          aria-labelledby={renderLabel && this._labelId}
+          // TODO renderLabel is returned is renderLabel is a function, this is likely a bug
+          aria-labelledby={(renderLabel as string) && this._labelId}
```
matyasf, line 185: Likely a bug.. ir renderLabel is a function its returned. I think it should return _labelId if renderLabel is not undefined.

packages/ui-react-utils/src/omitProps.ts
```diff
@@ -26,34 +26,35 @@
...
+ * @param propsToOmit list disallowed prop keys or an object whose
+ *        keys will be omitted.
+ * @param exclude an optional array of disallowed prop names to omit
+ * @returns props object without the excluded props
  * @module omitProps
  */
 function omitProps<T extends Record<string, any>>(
   props: T,
-  propTypesOrAllowedPropList: Record<string, any> | string[],
+  propsToOmit: Record<string, any> | string[],
```
matyasf, line 43: this made it so hard to understand, this param specifies what to omit, not what is allowed

packages/ui-react-utils/src/omitProps.ts
```diff
@@ -26,34 +26,35 @@
...
-    ? propTypesOrAllowedPropList
-    : Object.keys(propTypesOrAllowedPropList || {})
-  const combined = exclude ? propKeys.concat(exclude) : propKeys
-
-  return omit(props, combined)
+  const propKeys = Array.isArray(propsToOmit)
+    ? propsToOmit
+    : Object.keys(propsToOmit || {})
+  const keysToOmit = exclude ? propKeys.concat(exclude) : propKeys
+  return omit(props, keysToOmit) as Record<string, any>
```
matyasf, line 50: While it returns `Partial<P>` this causes lots of issues since it will try to put conflicting props where its used.

packages/ui-table/src/Table/index.tsx
```diff
@@ -168,14 +168,13 @@ class Table extends Component<Props> {
...
       <View
-        // @ts-expect-error ts-migrate(2339) FIXME: Property 'omitViewProps' does not exist on type 't... Remove this comment to see the full error message
         {...View.omitViewProps(omitProps(this.props, Table.propTypes), Table)}
         as={isStacked ? 'div' : 'table'}
         margin={margin}
         elementRef={elementRef}
         css={styles.table}
         role={isStacked ? 'table' : null}
-        aria-label={isStacked ? caption : null}
+        aria-label={isStacked ? (caption as string) : null}
```
matyasf, line 177: same as above, likely a bug

packages/ui-react-utils/src/omitProps.ts
ke1k0: I'd leave a TODO here to try to type it later once all props are in place

ke1k0 (approved):

Brailor (approved):

PR #685, opened 2021-08-18
INSTUI-3216 Clean up defaultprops in all components

matyasf:
Remove any `undefined`, `null` and empty function (e.g. `()=>{}`)
Add null checks where needed.
There should be no functional difference.

### DIFF
--- a/packages/ui-calendar/src/Calendar/index.tsx
+++ b/packages/ui-calendar/src/Calendar/index.tsx
@@ -145,12 +145,8 @@ class Calendar extends Component<Props> {
   }
 
   static defaultProps = {
-    children: null,
-    renderNextMonthButton: undefined,
-    renderPrevMonthButton: undefined,
-    renderNavigationLabel: undefined,
-    onRequestRenderNextMonth: () => {},
-    onRequestRenderPrevMonth: () => {},
+    onRequestRenderNextMonth: () => {}, // TODO check if its OK if this is removed
+    onRequestRenderPrevMonth: () => {}, // TODO check if its OK if this is removed
     as: 'span',
     role: 'table'
   }
--- a/packages/ui-drawer-layout/src/DrawerLayout/DrawerContent/index.tsx
+++ b/packages/ui-drawer-layout/src/DrawerLayout/DrawerContent/index.tsx
@@ -73,7 +73,11 @@ class DrawerContent extends Component<Props> {
   }
 
   static defaultProps = {
-    role: 'region'
+    role: 'region',
+    // @ts-expect-error ts-migrate(6133) FIXME: 'el' is declared but its value is never read.
+    contentRef: (el: any) => {}, // TODO if this is set to undefined _content is never set
+    // @ts-expect-error ts-migrate(6133) FIXME: 'node' is declared but its value is never read.
+    onSizeChange: (node: any) => {} // TODO this should not need a default value
   }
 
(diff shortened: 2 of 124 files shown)


packages/ui-calendar/src/Calendar/index.tsx
```diff
@@ -145,12 +145,8 @@ class Calendar extends Component<Props> {
...
 
   static defaultProps = {
-    children: null,
-    renderNextMonthButton: undefined,
-    renderPrevMonthButton: undefined,
-    renderNavigationLabel: undefined,
-    onRequestRenderNextMonth: () => {},
-    onRequestRenderPrevMonth: () => {},
+    onRequestRenderNextMonth: () => {}, // TODO check if its OK if this is removed
+    onRequestRenderPrevMonth: () => {}, // TODO check if its OK if this is removed
```
matyasf, line 149: I tried to be as safe as possible and did not remove anything that could have side effects. (but this is still something we should fix when converting these components to TS, because the user can set it intentionally to `undefined`

packages/ui-drawer-layout/src/DrawerLayout/DrawerContent/index.tsx
```diff
@@ -73,12 +73,11 @@ class DrawerContent extends Component<Props> {
...
   static defaultProps = {
-    children: null,
+    role: 'region',
     // @ts-expect-error ts-migrate(6133) FIXME: 'el' is declared but its value is never read.
-    contentRef: (el: any) => {},
+    contentRef: (el: any) => {}, // TODO if this is set to undefined _content is never set
     // @ts-expect-error ts-migrate(6133) FIXME: 'node' is declared but its value is never read.
-    onSizeChange: (node: any) => {},
-    role: 'region'
+    onSizeChange: (node: any) => {} // TODO this should not need a default value
```
matyasf, line 80: this was a tricky one, the component breaks when these are set to null/undefined. We should fix this in the future.

packages/ui-motion/src/Transition/BaseTransition.ts
```diff
@@ -38,14 +38,37 @@ const STATES = {
...
+  onTransition?: (...args: any[]) => any
+  onEnter?: (...args: any[]) => any
+  onEntering?: (...args: any[]) => any
+  onEntered?: (...args: any[]) => any
+  onExit?: (...args: any[]) => any
+  onExiting?: (...args: any[]) => any
+  onExited?: (...args: any[]) => any
+  children?: any
+  className?: string
+}
```
matyasf, line 63: For some reason this was not typed during the TS conversion, I added here some simple typings so its more clear how this works.

packages/ui-calendar/src/Calendar/index.tsx
ke1k0: Hm, that's true, we should look out for these.

ke1k0 (commented):
I started reviewing this, but after adding the first few comments I started having doubts about this PR:
- I think we would loose functionality (function signatures on docs pages),
- would risk minor changes/bugs (functions could be manually set to undefined),
- and also would have a huge conflict with another PR,
but we wouldn't gain much with it. So I think we should wait with this change until the components are properly typed.

matyasf:
Closing this for now, undefined-s are removed now, we will remove empty functions in a new PR

PR #688, opened 2021-08-19
fix: link correct source code on docs page for legacy versions

ke1k0:
Closes: INSTUI-3232


Brailor (approved):
Nice! This has to be backported to the v6 and v7 docs as well though.

HerrTopi (approved):

PR #699, opened 2021-08-30
chore(ui-text): add typing to ui-text (INSTUI-3090)

HerrTopi:
Closes: INSTUI-3090

### DIFF
--- a/packages/ui-text/babel.config.js
+++ b/packages/ui-text/babel.config.js
@@ -23,11 +23,6 @@
  */
 
 module.exports = {
-  shouldPrintComment(value) {
-    const isTsRelatedComment = /@ts-expect-error|FIXME/.test(value)
-
-    return !isTsRelatedComment
-  },
   presets: [
     [
       require('@instructure/ui-babel-preset'),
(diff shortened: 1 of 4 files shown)


packages/ui-text/src/Text/types.ts
```diff
@@ -36,14 +43,17 @@ export type TextProps = {
...
   fontStyle?: 'italic' | 'normal'
-  letterSpacing?: 'normal' | 'condensed' | 'expanded'
+  letterSpacing: 'normal' | 'condensed' | 'expanded'
   lineHeight?: 'default' | 'fit' | 'condensed' | 'double'
-  size?: 'x-small' | 'small' | 'medium' | 'large' | 'x-large' | 'xx-large'
+  size: 'x-small' | 'small' | 'medium' | 'large' | 'x-large' | 'xx-large'
   transform?: 'none' | 'capitalize' | 'uppercase' | 'lowercase'
   weight?: 'normal' | 'light' | 'bold'
-  wrap?: 'normal' | 'break-word'
+  wrap: 'normal' | 'break-word'
```
Brailor, line 56: I don't think these are required props. (`letterSpacing, size, wrap`)
If you check out the README it shows that the `Text` component doesn't require any props in order to be working.
Also they have defaultProps default values set.

Brailor:
Also, please delete the `shouldPrintComment` config from the babel.config.js.

matyasf (approved):

packages/ui-text/src/Text/types.ts
HerrTopi: Nice catch, thanks

matyasf (approved):

ke1k0 (approved):

PR #700, opened 2021-08-30
chore(ui-pill): add typing to ui-pill (INSTUI-3069) 

HerrTopi:


packages/ui-pill/src/Pill/index.tsx
```diff
@@ -45,7 +45,7 @@ category: components
 **/
 @withStyle(generateStyle, generateComponentTheme)
```
Brailor, line 46: The `@tsProps` annotation is missing.

Brailor:
Again, the `shouldPrintComment` config should be deleted from babel.config.js.

packages/ui-pill/src/Pill/index.tsx
HerrTopi: Done, added.

packages/ui-tooltip/src/Tooltip/types.ts
```diff
@@ -29,16 +29,21 @@ import type {
   PositionMountNode
 } from '@instructure/ui-position'
 
-export type ChildrenArgs = {
+export type TooltipRenderChildrenArgs = {
```
ke1k0, line 32: Renamed this, because I think it's best if we prefix the globally exported component types.

packages/ui-tooltip/src/Tooltip/index.tsx
```diff
@@ -191,15 +191,13 @@ class Tooltip extends Component<TooltipProps> {
         </Trigger>
       )
     } else if (typeof children === 'function') {
-      return children({
+      // TODO: try to remove cast when typing Tooltip
+      return (children as RenderChildren)({
```
ke1k0, line 195: I'm really not sure why, but TS couldn't correctly calculate the children type, so I used a cast here... maybe it can be sorted out when we fully type Tooltip.

ke1k0 (approved):

Brailor (approved):

PR #702, opened 2021-09-02
INSTUI-3043 : Better TypeScript typing for ui-decorator

matyasf:
While not perfect, I doubt that we can make it significantly better without some large refactoring

### DIFF
--- a/packages/emotion/src/withStyle.tsx
+++ b/packages/emotion/src/withStyle.tsx
@@ -101,9 +101,9 @@ export type WithStyleProps = Partial<{
  *
  * @module withStyle
  *
- * @param generateStyle - The function that returns the component's style object
- * @param generateComponentTheme - The function that returns the component's theme variables object
- * @returns The decorated WithStyle Component
+ * @param {function} generateStyle - The function that returns the component's style object
+ * @param {function} generateComponentTheme - The function that returns the component's theme variables object
+ * @returns {ReactElement} The decorated WithStyle Component
  */
 const withStyle = decorator(
   (ComposedComponent, generateStyle: any, generateComponentTheme: any) => {
--- a/packages/emotion/src/withStyle.tsx
+++ b/packages/emotion/src/withStyle.tsx
@@ -22,7 +22,7 @@
  * SOFTWARE.
  */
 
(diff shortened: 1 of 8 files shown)


packages/ui-decorator/src/decorator.ts
```diff
@@ -22,19 +22,24 @@
...
  */
-function decorator(decorate: any) {
-  return (...args: any[]) => {
-    return (ComposedComponent: any) => {
+function decorator(
+  decorate: (toDecorate: ComponentClass<any>, ...args: any[]) => any
+) {
+  return (...args: unknown[]) => {
+    // parameters when the decorator is used
+    return (ComposedComponent: ComponentClass<any>) => {
```
matyasf, line 42: I've tried to introduce generics in many ways, so modules who use this function are giving errors when called with wrong parameters (e.g. `withStyle` is called zero parameters), but could not do it. It seems to me that TS gives up on interpreting what is going on when decorators have generics.

packages/emotion/src/withStyle.tsx
```diff
@@ -101,15 +101,15 @@ export type WithStyleProps = Partial<{
  *
  * @module withStyle
  *
- * @param {function} generateStyle - The function that returns the component's style object
- * @param {function} generateComponentTheme - The function that returns the component's theme variables object
- * @returns {ReactElement} The decorated WithStyle Component
```
ke1k0, line 106: Removing the types here removes them from the doc page too. https://613078228ea68c6766e640f1--preview-instui.netlify.app/#withStyle

packages/ui-decorator/src/decorator.ts
```diff
@@ -22,19 +22,24 @@
...
+
 /**
  * ---
  * category: utilities
  * ---
  * Wraps (decorates) a React component to add additional functionality
  *
  * @module decorator
- * @param {Function} decorator a function that returns a decorated component
- * @returns {Class} a decorated component class
```
ke1k0, line 33: These types are gone from the doc page too. (And at the other methods where you deleted them too :D)

Brailor (approved):

ke1k0 (approved):

PR #705, opened 2021-09-02
Add correct typing to `getElementType`, remove passed down html attributes being `nullable`

Brailor:


packages/shared-types/src/UtilityTypes.ts
```diff
@@ -56,4 +56,4 @@ export type CSSShorthandValue<Value extends string> =
 export type OtherHTMLAttributes<
   Props,
   Attributes extends React.HTMLAttributes<Props> = React.AllHTMLAttributes<Props>
-> = Nullable<Omit<Attributes, keyof Props>>
```
Brailor, line 59: This had to be changed to be not nullable because when we have a component which uses the `getElementType` utility, the returned element from the `getElementType` could be a native DOM element, which essentially has the same typing as `React.HTMLAttributes` but without the attributes on it being nullable.

So what that means is when we have a scenario for example with the `View` component:

```tsx

render() {
  ...
     <View as="div" className={null}>some content</View>
  ...
```
The rendered element in the DOM will be a `div` element and its attributes can only take value types or `undefined` but not the value `null`.

packages/ui-a11y-content/src/AccessibleContent/index.tsx
```diff
@@ -44,15 +44,15 @@ class AccessibleContent extends Component<
...
+    children: PropTypes.node,
+    as: PropTypes.elementType
+  } as const
 
   static defaultProps = {
     alt: undefined,
     as: 'span',
     children: null
-  }
+  } as const
```
Brailor, line 55: Casting `propTypes` and `defaultProps` to `const` was the only way keep the current `defaultProps` functionality (required props being not really required when they have `defaultProps` entries) while getting `getElementType` utility work properly.

packages/shared-types/src/UtilityTypes.ts
ke1k0: Seems nice!

packages/ui-a11y-content/src/AccessibleContent/index.tsx
ke1k0: Hm, this way it will loose the link between the propTypes and the Props type, but hopefully the combination of the refact we are doing in the allowedProps ticket will solve this. Anyway, we'll have to test it well when we merge the two.

packages/ui-buttons/src/BaseButton/types.ts
```diff
@@ -25,8 +25,9 @@
 import { AsElementType } from '@instructure/shared-types'
 import type { InteractionType } from '@instructure/ui-react-utils'
 import type { Spacing } from '@instructure/emotion'
+import React from 'react'
 
-export type BaseButtonProps = {
+export type BaseButtonProps = React.PropsWithChildren<{
```
ke1k0, line 30: Just a note: I found it better to list children among the props instead of using this type, because
- it is easier to read
- and sometimes we have to manually type it anyway (e.g. it is required or not just simply "ReactNode")

packages/ui-dialog/src/Dialog/index.tsx
```diff
@@ -240,15 +239,16 @@ class Dialog extends Component<DialogProps & OtherHTMLAttributes<DialogProps>> {
   }
 
   render() {
-    const ElementType = getElementType(Dialog, this.props)
+    const ElementType = getElementType<DialogProps>(Dialog, this.props)
```
ke1k0, line 242: Why is `<DialogProps>` needed here?

packages/ui-react-utils/src/getElementType.ts
```diff
@@ -80,7 +75,7 @@ function getElementType<T extends ComponentWithAsProp>(
     return 'button'
   }
 
-  return Component.defaultProps.as || 'span'
+  return Component.defaultProps?.as ?? 'span'
```
matyasf, line 78: just FYI this refactor makes it behave slightly differently: 

```
class Component {
  static defaultProps = {as: ''}
}
const a = Component.defaultProps.as || 'span'
console.log("AA", a) // logs "AA span"
const b = Component.defaultProps?.as ?? 'span'
console.log("BB:", b) // logs "BB"
```
But I dont think this is an issue here

matyasf (approved):

PR #706, opened 2021-09-06
Inst UI 3230 make an allowed props array for every component

Brailor:


ke1k0 (approved):

packages/__examples__/package.json
```diff
@@ -49,6 +49,6 @@
     "@instructure/ui-test-utils": "8.8.0",
     "globby": "^11",
     "prop-types": "^15",
-    "react-docgen": "^5"
+    "react-docgen": "^6.0.0-alpha.0"
```
ke1k0, line 52: this was needed for the examples to generate well: react-docgen 5 doesn't parse the props/types in other files, and we just moved then to a different file...

matyasf (approved):

PR #713, opened 2021-09-17
refactor(ui-dom-utils): remove `ReactDOM.findDOMNode` usage, replace with refs

Brailor:


packages/ui-dom-utils/src/findDOMNode.ts
```diff
@@ -49,7 +63,23 @@ function findDOMNode(el?: UIElement) {
...
+      }
+      return refElement
+    } else {
+      console.warn(
+        'Warning: ReactDOM.findDOMNode is deprecated in Strict mode, consider using refs instead.\n See more here: https://instructure.design/#focus-management'
+      )
+
+      return ReactDOM.findDOMNode(node as ReactInstance)!
+    }
   }
```
Brailor, line 83: This is the main change:
Our own `findDOMNode` utility was refactored in a way that it is now tries to utilize react refs instead of directly calling `ReactDOM.findDOMNode`.
If a react instance has the `.ref` property on it then this function will try to get the underlying HTML element. 
If the `ref` points to another react instance, then it will recursively try to get the HTML element.

matyasf: I dont understand this part: "See more here: https://instructure.design/#focus-management'"
It does not mention how to use refs

matyasf (approved):

ke1k0 (approved):

PR #715, opened 2021-09-17
INSTUI-3174 Refactor code, so propTypes is not used in code logic

matyasf:
Now users can use tools like [babel-plugin-transform-react-remove-prop-types](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types)  to make their bundle size smaller.
It also activates an ESLint rule that forbids the usage of `propTypes`

Note that while this PR allows prop type removers run, these will not actually shrink the bundle size by a significant amount because it only removes these lines: `static propTypes = propTypes`.  A further PR is needed, that moves back propTypes to the component itself.

### DIFF
--- a/packages/__docs__/src/Figure/index.js
+++ b/packages/__docs__/src/Figure/index.js
@@ -77,6 +77,17 @@ class Figure extends Component {
     children: Children.oneOf(['FigureItem'])
   }
 
+  static allowedProps = [
+    'makeStyles',
+    'styles',
+    'title',
+    'caption',
+    'recommendation',
+    'iconTitle',
+    'float',
+    'children'
+  ]
+
   static defaultProps = {
     recommendation: 'none',
     float: 'none',
@@ -116,7 +127,7 @@ class Figure extends Component {
 
     return (
       <View
-        {...omitProps(mergedProps, Figure.propTypes, ['padding'])}
+        {...omitProps(mergedProps, Figure.allowedProps, ['padding'])}
         as="figure"
         css={styles.figure}
       >
--- a/packages/__docs__/src/Figure/index.js
+++ b/packages/__docs__/src/Figure/index.js
@@ -78,8 +78,6 @@ class Figure extends Component {
   }
 
   static allowedProps = [
-    'makeStyles',
-    'styles',
     'title',
     'caption',
     'recommendation',
--- a/packages/emotion/src/withStyle.tsx
+++ b/packages/emotion/src/withStyle.tsx
@@ -153,9 +153,8 @@ const withStyle = decorator(
       )
     })
     hoistNonReactStatics(WithStyle, ComposedComponent)
-    // we have to pass these on, because sometimes we need to
+    // we have to pass these on, because sometimes users
     // access propTypes of the component in other components
-    // (mainly in the `omitProps` method)
     WithStyle.propTypes = ComposedComponent.propTypes
     WithStyle.defaultProps = ComposedComponent.defaultProps
     // @ts-expect-error These static fields exist on InstUI components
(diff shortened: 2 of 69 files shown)


packages/ui-i18n/src/bidirectional.tsx
```diff
@@ -95,7 +95,10 @@ const bidirectional: BidirectionalType = decorator((ComposedComponent) => {
   }
   hoistNonReactStatics(BidirectionalForwardingRef, ComposedComponent)
   BidirectionalForwardingRef.defaultProps = ComposedComponent.defaultProps
+  // eslint-disable-next-line react/forbid-foreign-prop-types
   BidirectionalForwardingRef.propTypes = ComposedComponent.propTypes
+  // @ts-expect-error These static fields exist on InstUI components
+  BidirectionalForwardingRef.allowedProps = ComposedComponent.allowedProps
```
matyasf, line 101: this is also a fix for a likely bug

packages/emotion/src/withStyle.tsx
```diff
@@ -153,11 +153,13 @@ const withStyle = decorator(
...
     hoistNonReactStatics(WithStyle, ComposedComponent)
-    // we have to pass these on, because sometimes we need to
+    // we have to pass these on, because sometimes users
     // access propTypes of the component in other components
-    // (mainly in the `omitProps` method)
+    // eslint-disable-next-line react/forbid-foreign-prop-types
     WithStyle.propTypes = ComposedComponent.propTypes
     WithStyle.defaultProps = ComposedComponent.defaultProps
+    // @ts-expect-error These static fields exist on InstUI components
+    WithStyle.allowedProps = ComposedComponent.allowedProps
```
matyasf, line 162: again, before this fix I think `@withStyle` was erasing allowedProps

packages/__docs__/src/Figure/index.js
```diff
@@ -77,6 +77,17 @@ class Figure extends Component {
     children: Children.oneOf(['FigureItem'])
   }
 
+  static allowedProps = [
```
ke1k0, line 80: Oh, yes, I forgot to add allowedProps to the __docs__ components... thanks :)

packages/__docs__/src/Figure/index.js
```diff
@@ -77,6 +77,17 @@ class Figure extends Component {
...
+    'makeStyles',
+    'styles',
```
ke1k0, line 82: These shouldn't be allowedProps, it is always a bad idea to pass the styles to another component.

packages/emotion/src/withStyle.tsx
ke1k0: Nice catch!

packages/ui-billboard/src/Billboard/props.ts
```diff
@@ -113,19 +113,19 @@ const propTypes: PropValidators<PropKeys> = {
 }
 
 const allowedProps: AllowedPropKeys = [
+  'hero',
```
ke1k0, line 116: I know you just reordered these, but was there a reason to it? Just curious, because when I started the allowedProps refact, in the beginning I started ordering them alphabetically, but gave up on it later... I hope I didn't break something with it.

packages/ui-motion/src/Transition/props.ts
```diff
@@ -59,15 +59,17 @@ type AllowedPropKeys = Readonly<Array<PropKeys>>
 
 type TransitionProps = TransitionOwnProps & WithStyleProps
 
+const transitionTypes = PropTypes.oneOf([
```
ke1k0, line 62: I would either call this "TransitionTypePropType", or export only the array of values and use it like `PropTypes.oneOf(transitionTypes)` in the other place(s).

packages/ui-billboard/src/Billboard/props.ts
matyasf: I just ordered them the same way Props are ordered, so its easy to check that they are the same

Brailor (approved):

packages/ui-motion/src/Transition/props.ts
matyasf: fixed

ke1k0 (approved):

Issue #707, opened 2021-09-06
Additional properties for components


milkeshaa:
### Context

Components: **Tabs**, **Tabs.Panel**, **Modal** (**Mask** which is a child of **Modal** one).

### Summary

Requested functionality:
1. Please, add _height_ property to **Tabs** and **Tabs.Panel** components. 
  **Tabs** component does not have its own _height_ property, but when it's provided it's passed through **View** component which is the origin of the **Tabs** component. Could you add this property to **Tabs** component as well.
  **Tabs.Panel** also does not have its own _height_ property, I want to ask you to make it available and when provided to be applied to the div element which is the origin of the **Tabs.Panel** component and to the **View** component which is inside of this div element.  
  
2. Please, add _overflowX_ and _overflowY_ properties to **Tabs.Panel** component.
  **Tabs.Panel** also does not have its own _overflowX_ and _overflowY_ properties, I want to ask you to make them available and when provided to be applied to the **View** component which is inside of **Tabs.Panel** component. 
  
3. Please, make it possible to change _flexDirection_ property of the origin component of the **Modal** one (**Mask** component).
  Currently it's set to 'row' by default, but it seems like there is no possibility to change it.
  
### Justification

  Because sometimes it is necessary to match the _height_ of the parent element, so that the child elements of the **Tabs.Panel** component also have the opportunity to expand in _height_. Especially when you are using elements with absolute position as children of the **Tabs.Panel** component (just for example). So while using heights we should be able to manipulate _overflow_ (hide it or show it and so on). 
  Adding posibility to change _flexDirection_ is also important, since we should be able to change the direction of expanding the **Modal** component HTML element (with row direction it can only expand horizontally).

### Additional Details

  It's very important to implement this functionality, since that currently when you need to set these properties for these components you have no other way except traversing and manipulating DOM elements using JavaScript.

Brailor:
Hi there!

Thank you for your feedback/suggestions!
We will look into them and get back to you with a more detailed response, however this could take us a while to do, so please bear with us.

Brailor:
Hi, just a quick update, we are going to discuss these suggestions today with the design team to see if they are feasible.
In the meantime could you tell me which version of inst-ui are you using?

milkeshaa:
Hi, thanks for informing me! 
We are using the latest version (8.8.0) of inst-ui.

Yes, here are some examples with comments in them:

-  **With Tabs & Tabs.Panel components:**
[...]

PR #716, opened 2021-09-17
chore(ui-tabs): add 'fixHeight' prop to Tabs

ke1k0:
Closes: INSTUI-3258

Refactored the DOM structure and CSS of Tabs component to have an extra wrapper for panels and use
flex layout under the hood. This makes it easier to handle the fix, min and max heights of the
component. Added a `fixHeight` prop to set a fix height, which can be set to `100%` now to fill the
parent element. Also added a few new examples.


Brailor:
#707

matyasf (approved):
looks nice!
A test for this new prop would be nice, but not a must-have IMO

HerrTopi (approved):

PR #722, opened 2021-09-28
fix(ui-popover): popover does not update on certain prop changes (INSTUI-3261)

HerrTopi:
Closes: INSTUI-3261

Popover did not update when either plavement, shouldAlignArrow, or withArrow was changed. It
calculated placementOffset incorrectly when withArrow was false

### DIFF
--- a/packages/ui-popover/src/Popover/index.tsx
+++ b/packages/ui-popover/src/Popover/index.tsx
@@ -237,7 +237,9 @@ class Popover extends Component<PopoverProps> {
     if (
       this.props.offsetX !== prevProps.offsetX ||
       this.props.offsetY !== prevProps.offsetY ||
-      this.props.placement !== prevProps.placement
+      this.props.placement !== prevProps.placement ||
+      this.props.shouldAlignArrow !== prevProps.shouldAlignArrow ||
+      this.props.withArrow !== prevProps.withArrow
     ) {
       this.setState({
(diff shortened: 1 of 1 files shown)


packages/ui-popover/src/Popover/index.tsx
```diff
@@ -237,7 +237,9 @@ class Popover extends Component<PopoverProps> {
     if (
       this.props.offsetX !== prevProps.offsetX ||
       this.props.offsetY !== prevProps.offsetY ||
-      this.props.placement !== prevProps.placement
+      this.props.placement !== prevProps.placement ||
+      this.props.shouldAlignArrow !== prevProps.shouldAlignArrow ||
+      this.props.withArrow !== prevProps.withArrow
```
matyasf, line 242: You could write a test for this, for example like here: https://github.com/instructure/instructure-ui/blob/datetimeinput/packages/ui-date-time-input/src/DateTimeInput/__tests__/DateTimeInput.test.tsx#L442

NamNguyenInstructure (approved):
That's great! Thank you for fixing this.

ke1k0 (approved):

packages/ui-popover/src/Popover/index.tsx
HerrTopi: Unfortunately, our current testing lib does not allow for state checking, so the testing not really viable

PR #728, opened 2021-10-04
Popover adding examples

NamNguyenInstructure:


packages/ui-popover/src/Popover/README.md
```diff
@@ -105,21 +152,136 @@ render: false
...
+                'none',
+                'resting',
+                'above',
+                'topmost'
+            ],
+        }
+    }
+    changePlacement = (e, { id, placement }) => this.setState({ placement: this.state.placementsValues[id] })
+    changeShadow = (e, { id, shadow }) => this.setState({ shadow: this.state.shadowValues[id] })
+    toggleWithArrow = (event) => {
```
HerrTopi, line 194: For consistency sake, you should format this line as you did with the others:`
```suggestion
   toggleWithArrow = (event) => this.setState({ withArrow: !this.state.withArrow })
```

packages/ui-popover/src/Popover/README.md
```diff
@@ -105,21 +152,136 @@ render: false
...
-        >
-          <Heading>Small<br/>Target</Heading>
-        </Popover>
-      </div>
-  )
-}
+    constructor(props) {
+        super(props)
+
+        this.state = {
```
HerrTopi, line 158: You only should store those variables in state, which you'd like to change inside the component. For example placementsValues is a constant list, it will never change. So is ShadowValues. They should be a static variable on the class.

packages/ui-popover/src/Popover/README.md
```diff
@@ -105,21 +152,136 @@ render: false
...
+            ],
+            shadowValues: [
+                'none',
+                'resting',
+                'above',
+                'topmost'
+            ],
+        }
+    }
+    changePlacement = (e, { id, placement }) => this.setState({ placement: this.state.placementsValues[id] })
```
HerrTopi, line 192: Simple select passes to its onchange function 2 parameters. The first is indeed an event, the second is always {value:"",id:""} shape. You access placement and shadow, (which are not in the second parameter, and you don't use them in the function body. Since you have access to the value, you can do something like this:
```suggestion
    changePlacement = (e, { value }) => this.setState({ placement: value})
```

packages/ui-popover/src/Popover/README.md
NamNguyenInstructure: I will fix it and create pull request again

packages/ui-popover/src/Popover/README.md
```diff
@@ -9,25 +9,72 @@ Popovers hide or show content as a result of user interaction, such as clicking,
...
-  <View padding="x-small" display="block" as="div" id="tip">
-    Hello World
-  </View>
-</Popover>
+class Example extends React.Component {
+    state = {
+        withArrow: true,
+        shouldAlignArrow: true,
+        color : "primary"
+    }
```
ke1k0, line 19: Please use the 2 space indent instead of 4 space indent (and try to set up Prettier in VSCode, that will help you fix most of these issues automatically)

packages/ui-popover/src/Popover/README.md
```diff
@@ -9,25 +9,72 @@ Popovers hide or show content as a result of user interaction, such as clicking,
...
+    toggleWithArrow = (event) => this.setState({ withArrow: !this.state.withArrow })
+    toggleAlignArrow = (event) => this.setState({ shouldAlignArrow: !this.state.shouldAlignArrow })
+    changeColor = (event,color) =>  {this.setState({color})}
+    render() {
+        return (
+            <>
+                <FormFieldGroup description="Uncontrolled Popover Example">
```
ke1k0, line 26: I would suggest making and extra example (called "Popover playground" or something like this) where the users can play with all the properties. The examples under the texts should focus on the the prop mentioned in the text only. 

(Also, as we talked about it, maybe you could move the inputs on a white panel to separate them from the example, using the View and Flex components)

packages/ui-popover/src/Popover/props.ts
```diff
@@ -38,6 +38,32 @@ import type {
 import type { BidirectionalProps } from '@instructure/ui-i18n'
 import type { PropValidators } from '@instructure/shared-types'
 
+const placementPropValues = [
```
ke1k0, line 41: Please put a comment above this:

```js
// Copied list from "PositionPropTypes.placement" so that it appears as options in the properties table. 
// TODO: try to use "PositionPropTypes.placement" again once Popover is fully typed and uses @tsProps
```

ke1k0 (commented):
Good job, looks and works nice! I added the comments we talked about :)

packages/ui-popover/src/Popover/README.md
NamNguyenInstructure: I just have done it, you should take a look.

packages/ui-popover/src/Popover/props.ts
NamNguyenInstructure: Yes.

ke1k0: Cool, thanks!

packages/ui-popover/src/Popover/README.md
```diff
@@ -170,6 +226,148 @@ class Example extends React.Component {
 render(<Example />)
 ```
 
+#### Popover playground
```
ke1k0, line 229: This looks pretty cool now too! I suggest putting the color picker input here too (the `primary` and `primary-inverse` radio you put in the Uncontrolled Popover example too)! :)

packages/ui-popover/src/Popover/README.md
```diff
@@ -170,6 +226,148 @@ class Example extends React.Component {
...
+      'top center',
+      'center end',
+      'center start'
+    ],
+    shadowValues: [
+      'none',
+      'resting',
+      'above',
+      'topmost'
+    ],
```
ke1k0, line 269: As Topi said, these shouldn't be in the state, because we never change them. You could move them to the class, and then you can use them as `this.shadowValues`. 

```js
state = { ... }
shadowValues = [ ...  ]

// and then later in the render:
{
  this.shadowValues.map((shadow, index) => ( ... ))
}
```

NamNguyenInstructure: I have change them to static variables

packages/ui-popover/src/Popover/README.md
NamNguyenInstructure: Great idea! I will do it soon

HerrTopi (approved):
Good job!

PR #730, opened 2021-10-04
Menu tooltip position fix

matyasf:
Fix Menu tooltip positioning and add tooltip VRT tests for most components


packages/ui-menu/src/Menu/index.tsx
```diff
@@ -531,6 +529,7 @@ class Menu extends Component<MenuProps> {
           // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'el' implicitly has an 'any' type.
           ref: (el) => {
             this._trigger = el
+            this.ref = el
```
matyasf, line 532: This is the actual bugfix

packages/ui-tooltip/src/Tooltip/__examples__/Tooltip.examples.tsx
```diff
@@ -21,35 +21,34 @@
...
       textAlign: 'center'
-    }
-  },
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'props' implicitly has an 'any' type.
-  filter: (props) => {
-    if (!props.defaultIsShowingContent) return true
-
-    return false
+    } as const
   }
```
matyasf, line 53: no actual changes here, just some refactoring

packages/__examples__/.storybook/stories/TooltipPositioning.tsx
```diff
@@ -0,0 +1,213 @@
+import { ApplyLocale } from '@instructure/ui-i18n'
+import React from 'react'
+import { Alert } from '@instructure/ui-alerts/src/Alert'
```
ke1k0, line 3: I think (almost) all these components could be imported from the `ui` package. Also, we avoid importing from `/src` everywhere else too.

packages/__examples__/.storybook/stories/TooltipPositioning.tsx
```diff
@@ -0,0 +1,213 @@
...
+    <Text key={33}>Text component</Text>,
+    <TextArea key={34} label="This is a TextArea" />,
+    <TextInput key={35} renderLabel="This is a TextInput" />,
+    <View key={36}>This is a View</View>
+  ]
+  return (
+    <ApplyLocale locale="en-US">
+      <br />
+      {components.map(function (component, key) {
+        return wrapComponent(component, key.toString())
```
ke1k0, line 192: I would suggest wrapping the examples in a `<div key={key}>`, rather than manually setting the keys on the components in the array, because it will be a pain to modify the array like this :D

Edit: as I see you've already done that in the wrapComponent function, are those keys still needed then?

packages/__examples__/.storybook/stories/stories.ts
```diff
@@ -107,5 +108,11 @@ console.log(
     }
   }
 )
+
+// Stories not tied to one component
+storiesOf('Tooltip positioning', module).add('Tooltip positions', () =>
+  TooltipPositioning()
+)
+numStories++
```
ke1k0, line 116: This is probably good for now, but let's try to write code that adds these extra stories automatically later (once we have more). 👍🏻

packages/ui-tooltip/src/Tooltip/__examples__/Tooltip.examples.tsx
ke1k0: it did change the examples in the VRT, are those valid changes?

packages/__examples__/.storybook/stories/TooltipPositioning.tsx
```diff
@@ -0,0 +1,213 @@
...
+      })}
+    </ApplyLocale>
+  )
+}
+
+function wrapComponent(component: string | JSX.Element, key: string) {
+  return (
+    <View key={key} display="block" padding="large">
```
ke1k0, line 200: Can you please double-check the VRT examples, that they are as they should be? The examples with "display: block" seem kinda funky with the tooltip floating in the middle of the page (like checkbox), but thats probably fine :)

packages/__examples__/.storybook/stories/TooltipPositioning.tsx
matyasf: Yep, I missed this. Also I think the ESLint doesn't process these files for some reason

PR #734, opened 2021-10-05
INSTUI-3056 - TS typing `ui-img`

ke1k0:


packages/ui-img/babel.config.js
```diff
@@ -23,11 +23,6 @@
  */
 
 module.exports = {
-  shouldPrintComment(value) {
```
HerrTopi, line 26: There is still a @ts-expect-error in the ui-img folder ( in ImgLocator.ts). Shouldn't we wait with the removal of shouldprintcomment?

ke1k0: I put it back as we discussed!

PR #736, opened 2021-10-07
refactor(emotion): remove `bidirectionalPolyfill`

Brailor:
This util polyfilled Bi-directional [CSS proposal from W3C](https://drafts.csswg.org/css-logical-props/) to support direction-sensitive rules, a.k.a Left-To-Right (LTR) and Right-To-Left (RTL) in all browsers.

With the release of Safari 15 (Sep 20, 2021) all of our supported browsers (last 2 major versions) can handle [CSS Logical Properties](https://caniuse.com/css-logical-props) without any polyfills and prefixes, so this util was no longer needed.

### DIFF
--- a/packages/ui-form-field/src/FormFieldLayout/__tests__/FormFieldLayout.test.tsx
+++ b/packages/ui-form-field/src/FormFieldLayout/__tests__/FormFieldLayout.test.tsx
@@ -50,7 +50,7 @@ describe('<FormFieldLayout />', async () => {
 
     const formFieldLayout = within(subject.getDOMNode())
     const label = await formFieldLayout.findWithText('Username')
-    expect(label.getComputedStyle().textAlign).to.equal('right')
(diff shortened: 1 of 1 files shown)


packages/ui-form-field/src/FormFieldLayout/__tests__/FormFieldLayout.test.tsx
```diff
@@ -50,7 +50,7 @@ describe('<FormFieldLayout />', async () => {
 
     const formFieldLayout = within(subject.getDOMNode())
     const label = await formFieldLayout.findWithText('Username')
-    expect(label.getComputedStyle().textAlign).to.equal('right')
+    expect(label.getComputedStyle().textAlign).to.equal('end')
```
matyasf, line 53: Do you know why this test (and the other) needed to be changed?

Issue #729, opened 2021-10-04
`width="fit-content"` for <View /> not working in Firefox 


gromaco:
### Background Information

**Package Version(s)**:
ui-view v8.8.0 
**Browser**:

**Component**:
View

### Describe the Bug
`<View width="fit-content" /> `
`fit-content` doesn't work in Firefox.
It's not prefixed with `-moz-`
 
### Steps To Reproduce

   1. Go to [https://codepen.io/gromaco/pen/BaZeqKZ?editors=001](https://codepen.io/gromaco/pen/BaZeqKZ?editors=001)
   2. Check it in Chrome
   3. Check it in Firefox

### Expected Behavior
`-moz-fix-content` is set for `View`

### Additional Information

Currently, width(and a lot of other props) is set as inline style

https://github.com/instructure/instructure-ui/blob/d3ef347fa5e8b3126659528defa4a6e4408bf01e/packages/ui-view/src/View/index.tsx#L222

https://github.com/instructure/instructure-ui/blob/d3ef347fa5e8b3126659528defa4a6e4408bf01e/packages/ui-view/src/View/styles.ts#L602
```
inlineStyles: {
      ...spacingStyle,
      ...borderStyle,
      ...offsetStyle,
      width,
      height,
      minWidth,
      minHeight,
      maxWidth,
      maxHeight,
      ...getStyleProps(props)
    }

```
So it's not processed by `Stylis` (https://github.com/thysultan/stylis.js) unlike styles applied by Emotion `css` prop.

**Are you willing to submit a PR to fix?**
- [ ] Yes, if needed, but not sure at the moment what solution should be applied

**Requested Priority:**
High

Brailor:
Hi!
Thank you for the detailed bug report.
We will look into it and get back to you.

PR #741, opened 2021-10-11
INSTUI-3271 - feat(ui-view): vendor prefix all view style props

HerrTopi:
Closes: INSTUI-3271

Adding prop-based styling had no vendor prefixing.
Added inline styles to emotion's css prop and
added !important to every rule in it.

closes https://github.com/instructure/instructure-ui/issues/729

### DIFF
--- a/packages/ui-view/src/View/styles.ts
+++ b/packages/ui-view/src/View/styles.ts
@@ -571,6 +571,15 @@ const generateStyle = (
 
   const focusStyles = getFocusStyles(props, componentTheme)
 
+  const important = (styles: any) =>
+    Object.entries(styles).reduce(
+      (acc, entry) =>
+        entry[1] === undefined
+          ? acc
+          : { ...acc, [entry[0]]: `${entry[1]} !important` },
+      {}
+    )
+
   return {
     view: {
       label: 'view',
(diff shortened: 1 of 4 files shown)


packages/ui-view/src/View/styles.ts
```diff
@@ -600,16 +609,16 @@ const generateStyle = (
...
+      ...important(spacingStyle),
+      ...important(borderStyle),
+      ...important(offsetStyle),
+      width: `${width} !important`,
+      height: `${height} !important`,
+      minWidth: `${minWidth} !important`,
+      minHeight: `${minHeight} !important`,
+      maxWidth: `${maxWidth} !important`,
+      maxHeight: `${maxHeight} !important`,
+      ...important(getStyleProps(props))
```
ke1k0, line 621: I would wrap the whole inlineStyles into the important method instead of one-by-one. It is needed too, because width and height etc. can be undefined too, and it would result in invalid `undefined !important`. I tried it out, and almost halves the render time too.

packages/ui-view/src/View/styles.ts
```diff
@@ -571,6 +571,15 @@ const generateStyle = (
 
   const focusStyles = getFocusStyles(props, componentTheme)
 
+  const important = (styles: any) =>
```
ke1k0, line 574: This shouldn't be `any`, this will always be a Record<string, string | number>

matyasf:
One test is failing. Also I'd add a test that would check if the `!important` is added

packages/ui-view/src/View/styles.ts
```diff
@@ -600,16 +600,18 @@ const generateStyle = (
...
-      ...borderStyle,
-      ...offsetStyle,
-      width,
-      height,
-      minWidth,
-      minHeight,
-      maxWidth,
-      maxHeight,
-      ...getStyleProps(props)
+      '&&&&&&&&&&': {
```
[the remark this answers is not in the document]
ke1k0: Yeah I guess a comment would be nice :D

ke1k0 (approved):

PR #748, opened 2021-10-14
Reduce size of `Flex`, `SVGImage` examples to speed up Storybook example generation

Brailor:


packages/__examples__/.storybook/stories/stories.ts
```diff
@@ -50,6 +50,7 @@ const componentsContext = require.context(
 )
 
 let numStories = 0
+console.time('storybook-build-examples')
```
matyasf, line 53: I did not knew that `console` could do such thing, cool

packages/ui-svg-images/src/SVGIcon/__examples__/SVGIcon.examples.tsx
```diff
@@ -25,6 +25,7 @@ import React from 'react'
 
 export default {
   sectionProp: 'size',
+  maxExamples: 200,
```
matyasf, line 28: I dont think this was a performance issue, rather here we need better filtering

packages/ui-flex/src/Flex/__examples__/Flex.examples.tsx
```diff
@@ -61,7 +61,8 @@ const grow = [
 ]
 
 export default {
-  maxExamplesPerPage: 50,
+  maxExamplesPerPage: 20,
+  maxExamples: 100,
```
matyasf, line 65: While not an elegant solution, it will do for now. We need a better example filtering here in the future; a few hundred examples of `Flex` must be enough for a thorough visual test.

matyasf (approved):

packages/__examples__/.storybook/main.js
```diff
@@ -23,5 +23,9 @@
  */
 
 module.exports = {
-  stories: ['./stories/stories.ts']
+  stories: ['./stories/stories.ts'],
+  typescript: {
+    check: false,
+    reactDocgen: false
```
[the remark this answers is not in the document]
Brailor: If we have this setting enabled the whole example generation takes around ~13 sec, if it is turned off: **745 ms**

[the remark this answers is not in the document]
matyasf: nice one!

packages/ui-flex/src/Flex/__examples__/Flex.examples.tsx
Brailor: You are right, fixed

PR #750, opened 2021-10-14
fix(emotion): fix EmotionThemeProvider falling back to using Canvas t…

ke1k0:
…heme

Closes: INSTUI-3288

The `theme` prop is now correctly optional. If no theme is provided, or an invalid theme is provided
on the root level, we fall back to using Canvas an throw a warning.

### DIFF
--- a/packages/shared-types/src/BaseTheme.ts
+++ b/packages/shared-types/src/BaseTheme.ts
@@ -133,6 +133,19 @@ type BaseThemeVariables = {
   typography: Typography
 }
 
+type BaseThemeVariableKeys = [
+  'borders',
+  'breakpoints',
+  'colors',
+  'forms',
+  'media',
+  'shadows',
+  'spacing',
+  'stacking',
+  'transitions',
+  'typography'
+]
+
 type BaseTheme = {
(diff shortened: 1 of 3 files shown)


packages/shared-types/src/BaseTheme.ts
```diff
@@ -133,13 +133,27 @@ type BaseThemeVariables = {
...
+  'breakpoints',
+  'colors',
+  'forms',
+  'media',
+  'shadows',
+  'spacing',
+  'stacking',
+  'transitions',
+  'typography'
+]
```
matyasf, line 147: you don't need to type out here, I'd use just 
```
const baseThemeProps = [
  'borders',
...
] as const
```
(unless `BaseThemeVariableKeys` is something that will be used by our users)

ke1k0: I didn't want to put a `const` here, since this package contains types only, and couldn't find a better place to put it to... so that's how I solved it :) What do you think?

matyasf: OK :)

PR #756, opened 2021-10-15
InstUI 3255 convert components to typescript

NamNg1121:


packages/__docs__/src/ColorBand/props.ts
```diff
@@ -0,0 +1,52 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+import PropTypes from 'prop-types'
+import { PropValidators } from '@instructure/shared-types'
```
matyasf, line 26: The build is failing because `shared-types` is not included as a depencency (you need to add it to `__docs__`'s `package.json`.  Also if something is a type you should use `import type ...`

PR #757, opened 2021-10-15
  INSTUI-3038: refactor(ui-calendar): add detailed TS typing

matyasf:


packages/ui-calendar/src/Calendar/props.ts
```diff
@@ -93,27 +71,51 @@ const propTypes: PropValidators<PropKeys> = {
...
-  renderWeekdayLabels: PropTypes.arrayOf(
-    PropTypes.oneOfType([PropTypes.node, PropTypes.func])
-  ).isRequired,
+  renderWeekdayLabels: (React.ReactNode | ((...args: any[]) => any))[]
   /**
    * Callback fired when the next month button is clicked in the navigation
    * header, requesting to render the next month.
    */
-  onRequestRenderNextMonth: PropTypes.func,
+  onRequestRenderNextMonth?: (e?: Event) => void
```
ke1k0, line 79: I think it will always pass an event, shouldn't be optional.

ke1k0 (approved):

PR #758, opened 2021-10-18
InstUI-3255 convert components to typescript

NamNguyenInstructure:


packages/__docs__/src/ComponentTheme/props.ts
```diff
@@ -0,0 +1,60 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+import PropTypes from 'prop-types'
+import type { ComponentStyle } from '@instructure/emotion'
+import type { PropValidators } from '@instructure/shared-types'
+
+type ComponentThemeOwnProps = {
+  componentTheme: Record<string, any>
```
ke1k0, line 30: We have a special type for this in the `shared-types` package called `ComponentTheme`.

packages/__docs__/src/ComponentTheme/index.tsx
```diff
@@ -24,29 +24,20 @@
...
-    makeStyles: PropTypes.func,
-    // eslint-disable-next-line react/require-default-props
-    styles: PropTypes.object
-  }
-
-  mapColors(colorKey) {
-    const map = {}
+class ComponentTheme extends Component<ComponentThemeProps> {
+  static propTypes = propTypes
+  mapColors(colorKey: any) {
```
ke1k0, line 39: We actually know that this will have a type of `BaseColors`. But because TS is not very good with typing `Object.keys().forEach()`, you will need to have a special type cast there too:
```js
  mapColors(colorKey: BaseColors) {
    const map: Record<string, keyof BaseColors> = {}
    ;(Object.keys(colorKey) as Array<keyof BaseColors>).forEach((color) => {
      const hex = colorKey[color]
      if (typeof map[hex] === 'undefined') {
        map[hex] = color
      }
    })
    return map
  }
```

packages/__docs__/src/ContentWrap/props.ts
```diff
@@ -0,0 +1,50 @@
...
+
+import PropTypes from 'prop-types'
+import type { PropValidators } from '@instructure/shared-types'
+import { ThemeablePropTypes } from '@instructure/emotion'
+import type { Spacing } from '@instructure/emotion'
+
+type ContentWrapOwnProps = {
+  children?: typeof PropTypes.node
+  maxWidth?: string | number | undefined
+  padding?: Spacing | undefined
```
ke1k0, line 33: The correct TS type for children is `children?: React.ReactNode`. Also, you don't have to put ` | undefined` at the end of the optional properties, because the `?` already means "it can be undefined" :)

packages/__docs__/src/ColorBand/index.tsx
```diff
@@ -24,31 +24,22 @@
...
-
+class ColorBand extends Component<ColorBandProps> {
+  static propTypes = propTypes
+  static allowedProps = allowedProps
   componentDidMount() {
-    this.props.makeStyles()
-  }
-
-  componentDidUpdate(prevProps, prevState, snapshot) {
-    this.props.makeStyles()
```
ke1k0, line 51: Oh soooory, I wasn't clear about this. Please don't delete the whole `componentDidUpdate` method, it is needed for the component, just please delete the unused parameters: 

```js
componentDidUpdate() {
    this.props.makeStyles?.()
}
```

packages/__docs__/src/ColorCard/props.ts
```diff
@@ -0,0 +1,46 @@
...
+
+type PropKeys = keyof ColorCardOwnProps
+
+type AllowedPropKeys = Readonly<Array<PropKeys>>
+
+type ColorCardProps = ColorCardOwnProps
+
+const propTypes: PropValidators<PropKeys> = {
+  hex: PropTypes.string.isRequired,
+  name: PropTypes.string,
```
ke1k0, line 40: You forgot to add the `.isRequired` to this propType, it has to match the TS type.

ke1k0 (approved):

PR #772, opened 2021-10-26
chore: add bundle size to PRs

Brailor:
Note: this solution will likely not work for PRs which are coming from forked repos.
Since we don't get that much PRs from outside collaborators I think this is not a huge issue right now.

### DIFF
--- a/scripts/calculateBundleSize.js
+++ b/scripts/calculateBundleSize.js
@@ -1,0 +1,59 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015 - present Instructure, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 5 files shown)


scripts/calculateBundleSize.js
```diff
@@ -0,0 +1,59 @@
...
+  })
+
+  const analysis = await esbuild.analyzeMetafile(result.metafile)
+
+  return { bundle: result, outFile, analysis }
+}
+const gzip = async (fileName) => {
+  const file = path.resolve(fileName)
+  await execAsync(`gzip -k ${file}`)
+
```
ke1k0, line 50: Oh, this looks nasty! (Didn't even know GitHub can do this...)

Brailor: Well, when we run a github action essentially we get a fully fletched linux at our hands, so it is not GitHub that does this rather the underlying linux vm.
And it might look like a security issue (which it is) but this kind of thing is in the same as the other [(ReDos)](https://owasp.org/www-community/attacks/Regular_expression_Denial_of_Service_-_ReDoS#) security vulnerabilities. Basically if an attacker somehow gets access to our source code then, he/she may be able to shut down our CI vm's.

HerrTopi (approved):

Issue #752, opened 2021-10-14
Add sort property to TreeBrowser collections


cjlangston713:
### Context
Currently the `collections` property in the `TreeBrowser` accepts a hash of collection objects where the hash key is responsible for the display order.  Instead of the hash key determining the display order, a `sort` property is requested to handle the display order of `collections`.

### Summary
Since the hash key & `collection.id` needs to be unique, most use cases will use an identifier that will guarantee it is unique.  For example, if the collection objects are return from an api, then perhaps you would use the `id` of said object which is typically an integer.  This is just fine until you want to sort on another object's property such as a title. As stated in this [Stack Overflow](https://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order/23202095#23202095) (thank you Matyas!), if a hash key is parsed as an integer, it will order the hash objects in ascending not matter if you have applied a sort or not. 😢 The only way to maintain the sort order is to have a hash key as a string that does not parse as an integer.

### Justification
Our team uses the TreeBrowser to display a taxonomy that contains multiple groups with nested groups that have nested groups and so on and so forth.  The best identifier that we have available is the group id.  Not only because it is unique, it is used for all CRUD operations that the user can perform. The issue that we are trying to solve is displaying the groups alphabetically.  Since the group id is a parsable integer, we cannot supply a sorted hash of collections. Being able to send a sort property that will allow the collections to be sorted by a property would solve our sorting woes. 

### Additional Details
For more information around the specific issue, please refer to this Slack conversation that contains codepen examples and more detailed explanation.

matyasf:
Thanks for the feature request! We will get back to you if there is anything to report

NamNguyenInstructure:
Hello, I added new property to sort collections and items, please check it [here](https://github.com/instructure/instructure-ui/pull/776) and feedback if you want to modify anything.
Now I just apply same sorting method for both collections and items.

matyasf:
We have merged this feature, it will be in the next InstUI v8 release

ke1k0:
Hi @cjlangston713 ! This feature has been backported and released for V7 as well :)
https://instructure.design/v7/#CHANGELOG

cjlangston713:
Yay! This is fantastic news.  Thank you so much for all your hard work.  The outcomes team is stoked!

PR #776, opened 2021-11-02
INSTUI-3304 ui-tree-browser: add 'sortOrder' property to TreeBrowser component

NamNguyenInstructure:
According to [this issue](https://github.com/instructure/instructure-ui/issues/752), I have implemented a 'sortOrder' property to rearranged the appearance order of collections and items base on users' need. Closes https://github.com/instructure/instructure-ui/issues/752

### DIFF
--- a/packages/ui-tree-browser/src/TreeBrowser/README.md
+++ b/packages/ui-tree-browser/src/TreeBrowser/README.md
@@ -383,7 +383,7 @@ render(<Example/>)
 
 ### Change the order of appearance of items and collection
 
-By default, the order of collections and items depends on the order of `collections` and `items` array. We can override it by add new `sortOrder` comparision function.
+By default, the order of collections and items depend on the order of `collections` and `items` array. We can override it by providing a `sortOrder` comparison function.
 
 ---
 
@@ -402,29 +402,22 @@ class Example extends React.Component {
   constructor (props) {
     super(props)
     this.state = {
-      size: 'medium'
+      size: 'medium',
+      sorted: false
     }
-
-    this.sizes = ['small', 'medium', 'large']
   }
-
-  handleSizeSelect = (e, size) => {
-    this.setState({ size })
-  };
-
+  toogleSort = (event)=>{ this.setState({sorted:!this.state.sorted}) }
   render () {
     return (
       <>
         <View display="block" margin="none none medium">
-          <RadioInputGroup
-            name="treeBrowserSize"
-            defaultValue="medium"
-            description={<ScreenReaderContent>TreeBrowser size selector</ScreenReaderContent>}
-            variant="toggle"
-            onChange={this.handleSizeSelect}
-          >
-            {this.sizes.map((size) => <RadioInput key={size} label={size} value={size} />)}
-          </RadioInputGroup>
+         <FormFieldGroup description="Turn of/off sorting">
+              <Checkbox
+                checked={this.state.sorted}
+                label="Sort"
+                onChange={this.toogleSort}
+              />
+               </FormFieldGroup>
         </View>
 
         <TreeBrowser
@@ -451,9 +444,7 @@ class Example extends React.Component {
           }}
           defaultExpanded={[1, 3]}
           rootId={1}
-         sortOrder={(a, b) => {
-            return a.name.localeCompare(b.name) // sort by alphabetical order
-          }}
+         sortOrder={this.state.sorted? (a,b)=>{return(a.name).localeCompare(b.name)}: (a,b)=>{return 0} }
         />
       </>
     )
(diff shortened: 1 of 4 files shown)


cjlangston713:
I played around with the new sortOrder property in CodePen and it works great!  Thank you so much for adding it!

matyasf (commented):
Looks good! Some text in the Readme would be nice on the sort property.

packages/ui-tree-browser/src/TreeBrowser/index.tsx
```diff
@@ -73,7 +73,11 @@ class TreeBrowser extends Component<TreeBrowserProps> {
     // @ts-expect-error ts-migrate(6133) FIXME: 'id' is declared but its value is never read.
     onCollectionClick: function (id, collection) {},
     // @ts-expect-error ts-migrate(6133) FIXME: 'collection' is declared but its value is never re... Remove this comment to see the full error message
-    onCollectionToggle: function (collection) {}
+    onCollectionToggle: function (collection) {},
+    // @ts-expect-error ts-migrate(6133) FIXME: 'foo' and 'bar' are declared but their values are never read.
+    sortOrder: function (obj1, obj2) {
+      return 0
+    }
```
matyasf, line 80: Do not add a default value if it does nothing, rather add null checks (e.g. with `?.`) where this is used.

matyasf (approved):

packages/ui-tree-browser/src/TreeBrowser/index.tsx
```diff
@@ -172,6 +176,9 @@ class TreeBrowser extends Component<TreeBrowserProps> {
     return this.getExpanded(this.state, this.props)
   }
 
+  get sortOrder() {
+    return this.props.sortOrder
+  }
```
ke1k0, line 181: I'm not sure a getter is needed here, you can just use `this.props.sortOrder` where you need it. A getter is useful when you want to add some special logic in it.

packages/ui-tree-browser/src/TreeBrowser/props.ts
```diff
@@ -52,6 +52,7 @@ type TreeBrowserOwnProps = {
   onItemClick?: (...args: any[]) => any
   treeLabel?: string
   renderContent?: (...args: any[]) => any
+  sortOrder?: (arg1: any, arg2: any) => number
```
ke1k0, line 55: I would name these arguments `obj1` and `obj2` here as well, like you did in the defaultProps. That is more consistent, and indicates that these are objects.

packages/ui-tree-browser/src/TreeBrowser/__tests__/TreeBrowser.test.tsx
```diff
@@ -909,4 +909,63 @@ describe('<TreeBrowser />', async () => {
       expect((await tree.findAllItems()).length).to.equal(4)
     })
   })
+
+  describe('sorting', async () => {
```
ke1k0, line 913: It is nice that you wrote a test for it! Please add an example to the docs page as well, so that ppl can see how it works. Thanks!

ke1k0 (changes requested):

packages/ui-tree-browser/src/TreeBrowser/index.tsx
NamNguyenInstructure: Yes, you're right. I will delete it.

packages/ui-tree-browser/src/TreeBrowser/README.md
```diff
@@ -381,6 +381,89 @@ class Example extends React.Component {
 render(<Example/>)
 ```
 
+### Change the order of appearance of items and collection
+
+By default, the order of collections and items depends on the order of `collections` and `items` array. We can override it by add new `sortOrder` comparision function.
```
ke1k0, line 386:
```suggestion
By default, the order of collections and items depend on the order of `collections` and `items` array. We can override it by providing a `sortOrder` comparison function.
```

packages/ui-tree-browser/src/TreeBrowser/README.md
```diff
@@ -381,6 +381,89 @@ class Example extends React.Component {
...
+    this.state = {
+      size: 'medium'
+    }
+
+    this.sizes = ['small', 'medium', 'large']
+  }
+
+  handleSizeSelect = (e, size) => {
+    this.setState({ size })
+  };
```
ke1k0, line 413: For this example the size setters don't seem relevant, this is already displayed in another example. If you would like to make it interactive, I suggest having a "Sort collections alphabetically" checkbox where you can turn the sorting on and off, and that would display what this property does very nicely :)

ke1k0 (approved):
Nice, thanks! The new sorting example on the docs page looks nice too!

PR #781, opened 2021-11-06
INSTUI-3294 - more customization options for avatars

ke1k0:


Brailor (approved):

packages/ui-avatar/src/Avatar/index.tsx
```diff
@@ -135,7 +144,7 @@ class Avatar extends Component<AvatarProps, AvatarState> {
           onLoad={this.handleImageLoaded}
           aria-hidden="true"
         />
-        {!this.state.loaded && this.renderInitials()}
+        {!this.state.loaded && this.renderContent()}
```
matyasf, line 147: A potential issue here: If I set the src prop, it will cause loaded set to `true` when the image loads. Then if I set the `src` to `undefined` to display the initials/icon it will not show anything because loaded is never set to `false`

ke1k0: It has been like this from the start, so not a new issue (not very likely problem either), but added the logic for it, just in case. Thanks!

matyasf (approved):

PR #784, opened 2021-11-06
docs(ui-modal): update Modal Header spacing example

ke1k0:
Closes: INSTUI-3320

The title in the Modal.Header has to remain a h2 heading for a11y reasons.

### DIFF
--- a/packages/ui-modal/src/Modal/README.md
+++ b/packages/ui-modal/src/Modal/README.md
@@ -483,8 +483,8 @@ class Example extends React.Component {
           <Modal.Header spacing={this.state.smallViewport ? 'compact' : 'default'}>
             {this.renderCloseButton()}
             {this.state.smallViewport
-              ? <Text size="large">This Modal is optimized for small viewport</Text>
-              : <Heading>This is a deafult size Modal</Heading>
+              ? <Heading as="h2" level="h3" themeOverride={{ h3FontWeight: 400 }}>This Modal is optimized for small viewport</Heading>
+              : <Heading as="h2">This is a deafult size Modal</Heading>
             }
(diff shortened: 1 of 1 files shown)


packages/ui-modal/src/Modal/README.md
```diff
@@ -483,8 +483,8 @@ class Example extends React.Component {
           <Modal.Header spacing={this.state.smallViewport ? 'compact' : 'default'}>
             {this.renderCloseButton()}
             {this.state.smallViewport
-              ? <Text size="large">This Modal is optimized for small viewport</Text>
-              : <Heading>This is a deafult size Modal</Heading>
+              ? <Heading as="h2" level="h3" themeOverride={{ h3FontWeight: 400 }}>This Modal is optimized for small viewport</Heading>
+              : <Heading as="h2">This is a deafult size Modal</Heading>
```
Brailor, line 487: In the #783 the other heading is using the default `as` value, while here it is set to `h2`.
Is this intentional?
https://github.com/instructure/instructure-ui/pull/783/files#diff-73d2fd61dfc7313935095d082609d6ec977fbf3720bb60952f60a683c60da853R482

ke1k0: Oh, nice catch, thanks :) Well, the "h2" is the default, but I wanted to highlight that it shouldn't change and explicitly passed the prop, should do so in both versions, thanks!

ke1k0: I updated it in the V7 version.

HerrTopi (approved):

Brailor (approved):

PR #786, opened 2021-11-08
INSTUI-3307: convert typescript letters H,I,L

NamNguyenInstructure:


packages/__docs__/src/Heading/index.tsx
```diff
@@ -26,8 +26,9 @@ import React from 'react'
 import { Heading as UIHeading } from '@instructure/ui-heading'
 import { instructure } from '@instructure/ui-themes'
 import { EmotionThemeProvider } from '@instructure/emotion'
+import type { HeadingProps } from '@instructure/ui-heading/src/Heading/props'
```
matyasf, line 29: We should not use the full import path, always import from the top level of the package, e.g. in this case:
`import type { HeadingProps } from '@instructure/ui-heading'`. This is to prevent errors or big changes when we move packages around.

ke1k0:
Please rebase the branch and resolve the conflicts with master :)

packages/__docs__/src/Icons/index.tsx
```diff
@@ -23,7 +23,7 @@
  */
 
 /** @jsx jsx */
-import { Component } from 'react'
+import { Component, SyntheticEvent } from 'react'
 import PropTypes from 'prop-types'
```
matyasf, line 27: i get an ESLint error that this is not used, you should delete it.

packages/__docs__/src/Icons/index.tsx
```diff
@@ -47,20 +47,14 @@ import { Heading } from '../Heading'
...
-    makeStyles: PropTypes.func,
-    // eslint-disable-next-line react/require-default-props
-    styles: PropTypes.object
-  }
+class Icons extends Component<IconsProps, IconsState> {
+  static propTypes = propTypes
+  searchTimeout: any
 
-  constructor(props) {
+  constructor(props: any) {
```
matyasf, line 57: this is `IconsProps`

packages/__docs__/src/Icons/index.tsx
```diff
@@ -112,24 +106,28 @@ class Icons extends Component {
...
-  handleFormatChange = (e, o) => {
+  handleFormatChange = (o: Record<string, any>) => {
     window.location.hash = `#${o.value}`
   }
 
-  handleVariantClick = (e, name, variant, glyph) => {
+  handleVariantClick = (
+    name: string,
+    variant: string,
+    glyph: Record<string, any>
```
matyasf, line 125: `glyph` can be typed properly when you rebase this PR to the latest master commit, because it will contain your typings for it.

matyasf (commented):
looks good so far :) there are just some minor issues, and `Icons` needs to be typed after a rebase to the latest master

packages/__docs__/src/Header/props.ts
```diff
@@ -0,0 +1,54 @@
...
+const propTypes: PropValidators<PropKeys> = {
+  name: PropTypes.string.isRequired,
+  version: PropTypes.string.isRequired,
+  versionsData: PropTypes.shape({
+    latestVersion: PropTypes.string.isRequired,
+    previousVersions: PropTypes.arrayOf(PropTypes.string).isRequired
+  })
+}
+
+const allowedProps: AllowedPropKeys = ['name', 'version', 'versionsData']
```
ke1k0, line 51: You defined the `allowedProps` here, but forgot to add it to the component in the `index.tsx` :) Also the same problem with ComponentTheme, ContentWrap, Document, Glyph, Guidelines, Header, Hero and Icon components.

packages/__docs__/src/Icons/index.tsx
```diff
@@ -112,24 +106,24 @@ class Icons extends Component {
...
+    const query = (e.target as HTMLInputElement).value
 
     clearTimeout(this.searchTimeout)
     this.searchTimeout = setTimeout(() => {
       this.setState({ query })
     }, 500)
   }
 
-  handleFormatChange = (e, o) => {
+  handleFormatChange = (o: Record<string, any>) => {
```
ke1k0, line 118: As I mentioned before, the order of the parameters count, so you cannot remove `e` as a first parameter, it will break the code. If it is not used, mark it as unused:
`handleFormatChange = (_e: SyntheticEvent, o: Record<string, any>) => { ...  }`

packages/__docs__/src/Icons/index.tsx
```diff
@@ -112,24 +106,24 @@ class Icons extends Component {
...
     window.location.hash = `#${o.value}`
   }
 
-  handleVariantClick = (e, name, variant, glyph) => {
+  handleVariantClick = (name: string, variant: string, glyph: GlyphData) => {
```
ke1k0, line 122: As I mentioned before, the order of the parameters count, so you cannot remove `e` as a first parameter, it will break the code. If it is not used, mark it as unused:
`handleVariantClick = (_e: SyntheticEvent, name: string, variant: string, glyph: GlyphData) => {`

packages/__docs__/src/Icons/theme.ts
```diff
@@ -22,12 +22,14 @@
...
 
+import type { Theme } from '@instructure/ui-themes'
+
 /**
  * Generates the theme object for the component from the theme and provided additional information
  * @param  {Object} theme The actual theme object.
  * @return {Object} The final theme object with the overrides and component variables
  */
-const generateComponentTheme = (theme) => {
+const generateComponentTheme = (theme: Theme) => {
```
ke1k0, line 32: You can tell TypeScript that you expect `IconsTheme` as a return type.

```suggestion
const generateComponentTheme = (theme: Theme): IconsTheme => {
```

ke1k0 (changes requested):

ke1k0 (approved):
Nice, thank you!

matyasf (approved):
looks good!

PR #790, opened 2021-11-10
Improve `bootstrap` performance

Brailor:
This PR aims to reduce the time `yarn bootstrap` takes by leveraging features as:
- [typescript project references](https://www.typescriptlang.org/docs/handbook/project-references.html), which allows us to spin up only 1 instance of `tsc` to build the whole project, instead of spinning up as many `tsc` instances as many packages we have
- run the `tsc` compiler in parallel with `babel`, since (apart from the `ui-icons` package) the `babel` build and `tsc` does not depend on each others result we can run them in parallel instead, this speeds up the whole process
- remove not necessary abstractions:
    - the `clean` script previously created a new process for each package because it was called from a JS lib abstraction
    - now cleaning the unneeded directories is done with a unix native `rm` command, which is much faster 

With this PR local bootstrap is faster by around 2-3 minutes, in CI settings it can be faster by around 4-5 minutes than the previous solution.


packages/ui-a11y-content/tsconfig.build.json
```diff
@@ -1,7 +1,21 @@
...
+    { "path": "../console/tsconfig.build.json" },
+    { "path": "../emotion/tsconfig.build.json" },
+    { "path": "../shared-types/tsconfig.build.json" },
+    { "path": "../ui-dom-utils/tsconfig.build.json" },
+    { "path": "../ui-prop-types/tsconfig.build.json" },
+    { "path": "../ui-react-utils/tsconfig.build.json" },
+    { "path": "../ui-utils/tsconfig.build.json" },
+    { "path": "../uid/tsconfig.build.json" },
+    { "path": "../ui-babel-preset/tsconfig.build.json" },
+    { "path": "../ui-test-utils/tsconfig.build.json" }
```
matyasf, line 19: we will need to keep this in sync package.json dependencies & devDependencies?

package.json
```diff
@@ -44,15 +44,15 @@
     "commit": "ui-scripts --commit",
     "install:react": "ui-scripts --install-react",
     "install:packages": "lerna bootstrap --reject-cycles",
-    "bootstrap": "npm-run-all clean:modules install:packages clean build build:tokens build:types",
+    "bootstrap": "./scripts/bootstrap.sh",
```
matyasf, line 47: These 2 are not doing the same thing, I think the new script should delete the same folders too. Also instead of a shell script a .js "cleanAll" script would be better (unless the shell script has much better performance.)

packages/ui-a11y-content/tsconfig.build.json
Brailor: Basically yes, when I did this work I wrote a helper script which generated these reference paths from the package's `package.json` dependencies and devDependencies entries.

package.json
Brailor: I rewrote the scripts to nodejs, and although they seem to be a lit slower it is still much faster than the previous solution.

matyasf (approved):
looks good!

HerrTopi (approved):
And it REALLY makes it faster! Very nice work

packages/ui-a11y-content/tsconfig.build.json
ke1k0: Can you rerun that script with the fresh master? I'm pretty sure I've added some new deps in my PR-s since then, maybe others have too, so would be nice to update these too. Or if you don't have the script anymore, I'll try to find the ones that were added since :D

Also, just a question: what happens, if a dep is missing from here?

packages/ui-scripts/package.json
```diff
@@ -19,8 +19,8 @@
   "scripts": {
     "lint": "ui-test --lint",
     "lint:fix": "ui-test --lint --fix",
-    "build": "tsc -p tsconfig.build.json",
```
ke1k0, line 22: shouldn't we keep the build scripts? (there was another place where it was deleted)

package.json
```diff
@@ -44,15 +44,15 @@
...
+    "bootstrap": "node scripts/bootstrap.js",
     "bootstrap:docs": "npm-run-all clean:modules install:packages clean build",
     "build": "lerna run build --stream",
     "build:watch": "lerna run build:watch --stream",
     "build:docs": "lerna run bundle --stream --scope docs-app",
     "build:types": "lerna run build:types --stream",
     "build:tokens": "ui-token-scripts generate-all-tokens",
-    "build:ts": "yarn build && yarn build:types",
-    "build:node-scripts": "lerna run build --scope @instructure/ui-scripts --include-dependencies",
+    "build:ts": "lerna run prepare-build --stream --scope @instructure/ui-icons && tsc -b tsconfig.references.json --verbose",
```
ke1k0, line 54: This is just a question: is it possible to somehow filter or color code the output of tsc? We have so many packages, and really hard to pinpoint from the log what packages were updated and what packages weren't. It's a nice-to-have, but otherwise seems to work fine :)

packages/ui-a11y-content/tsconfig.build.json
Brailor: Sadly I deleted the script, but I agree that keeping these in sync is a pain in the ass to do manually, so maybe it should be checked automatically at some point of our CI process. (maybe before commit/push?)

To answer the second question:
If no other package is referencing that dep, then it will not get built by `tsc` I guess, I haven't tried it out to be honest.

packages/ui-scripts/package.json
Brailor: I haven't deleted the build scripts, rather just renamed them so they won't be rebuilded when we run `yarn run build`.

package.json
Brailor: There should be a way to somehow color the output of tsc, but TBH I'm not sure how at the moment, so if it is okay for you then I will create a ticket for this and investigate a little bit on how to do it.

packages/ui-a11y-content/tsconfig.build.json
ke1k0: I sent you the dep changes I found on slack. Some of those might be added already, didn'T know how far back shall I check. 

Can you please make tickets for trying to do this automatically and for the output coloring options? Thanks!

ke1k0 (approved):

PR #792, opened 2021-11-12
fix(ui-time-select): fix timeselect highlight behaviour (INSTUI-3308)

HerrTopi:
Closes: INSTUI-3308

Sometimes when there is a selected time and the select reopens, the
highlight not starts from the
selection, but rather from the first element
TEST PLAN:

- Pick a time
- set the focus to somewhere above the component
- with keyboard, navigate to the select
- open it
- navigate with arrows and see if the highlight starts from the selected element or not


ke1k0 (approved):

matyasf (approved):
nice work, this must've been quite hard to hunt down

Brailor (approved):

PR #796, opened 2021-11-15
chore(ui-options): type ui-options (INSTUI-3065)

HerrTopi:
Closes: INSTUI-3065

### DIFF
--- a/packages/ui-options/src/Options/Item/index.tsx
+++ b/packages/ui-options/src/Options/Item/index.tsx
@@ -44,6 +44,7 @@ import { allowedProps, propTypes } from './props'
 parent: Options
 id: Options.Item
 ---
+@tsProps
 **/
 @withStyle(generateStyles, generateComponentTheme)
 @testable()
@@ -97,8 +98,14 @@ class Item extends Component<OptionsItemProps> {
   }
 
   render() {
-    const { as, role, styles, renderBeforeLabel, renderAfterLabel, children } =
-      this.props
+    const {
+      as,
+      role,
+      styles,
+      renderBeforeLabel,
+      renderAfterLabel,
+      children
+    } = this.props
 
     const ElementType = getElementType(Item, this.props, () => as!)
     const passthroughProps = omitProps(this.props, Item.allowedProps)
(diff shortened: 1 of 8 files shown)


packages/ui-options/src/Options/Item/props.ts
```diff
@@ -34,10 +34,27 @@ import type {
...
   role?: string
+  /**
+   * Content to render before the label
+   * (if you pass a function, it has the `props` as its parameter)
+   */
   renderBeforeLabel?: React.ReactNode | ((...args: any[]) => any)
+  /**
+   * Content to render after the label
+   * (if you pass a function, it has the `props` as its parameter)
+   */
```
ke1k0, line 57: The label render types are not complete here. We know from the index what params are passed here, and we also expect a ReactNode in the return, so we should indicate in the props. Needs a bit of refactor to avoid circular dependency, but something like this:

(Note that the children type is changed too, it has no parameters.)

```js
type ItemProps = {
  children?: React.ReactNode | (() => React.ReactNode)
  /**
   * Element type to render as
   */
  as?: AsElementType
  /**
   * The style variant of the item
   */
  variant?: 'default' | 'highlighted' | 'selected' | 'disabled'
  /**
   * The aria role of the element
   */
  role?: string
}

type OptionsItemOwnProps = ItemProps & {
  /**
   * Content to render before the label
   * (if you pass a function, it has the `props` as its parameter)
   */
  renderBeforeLabel?: React.ReactNode | ((props: ItemProps) => React.ReactNode)
  /**
   * Content to render after the label
   * (if you pass a function, it has the `props` as its parameter)
   */
  renderAfterLabel?: React.ReactNode | ((props: ItemProps) => React.ReactNode)
}
```

packages/ui-options/src/Options/Item/index.tsx
```diff
@@ -72,8 +73,12 @@ class Item extends Component<OptionsItemProps> {
     this.props.makeStyles?.()
   }
 
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'renderLabel' implicitly has an 'any' ty... Remove this comment to see the full error message
-  renderContent(renderLabel, contentVariant) {
+  renderContent(
+    renderLabel: OptionsItemProps['renderBeforeLabel'],
```
ke1k0, line 77: The types are the same, but maybe we could indicate that it is either this or the After version
```suggestion
    renderLabel: OptionsItemProps['renderBeforeLabel'] | OptionsItemProps['renderAfterLabel'],
```

ke1k0 (approved):

Brailor (approved):

PR #797, opened 2021-11-16
INSTUI-3051 Focusable typing

matyasf:
TS typing for Focusable. I have rewritten its internal logic a little, because it was really hard to understand.

### DIFF
--- a/packages/ui-focusable/src/Focusable/README.md
+++ b/packages/ui-focusable/src/Focusable/README.md
@@ -11,26 +11,27 @@ a `render` or `children` method that returns content which must have exactly one
 example: true
 ---
 <Focusable>
-{({ focused }) => (
-  <span>
-    <div>
-      <Button>Focus me!</Button>
-    </div>
-    {focused && (
-      <ContextView
-        margin="small 0"
-        placement="bottom"
-        as="div"
-        background="primary"
-        padding="small"
-        borderWidth="small"
-        display="block"
-      >
-        I&#39;m focused!
-      </ContextView>
-    )}
-  </span>
-)}
+  {(options) => {
+    console.log(options)
+    return <span>
+      <div>
+        <Button>Focus me!</Button>
+      </div>
+      {options.focused && (
+        <ContextView
+          margin="small 0"
+          placement="bottom"
+          as="div"
+          background="inverse"
+          padding="small"
+          borderWidth="small"
+          display="block"
+        >
+          I&#39;m focused!
+        </ContextView>
+      )}
+    </span>
+  }}
 </Focusable>
 ```
 
--- a/packages/ui-focusable/src/Focusable/__tests__/Focusable.test.tsx
+++ b/packages/ui-focusable/src/Focusable/__tests__/Focusable.test.tsx
@@ -34,22 +34,20 @@ import {
 } from '@instructure/ui-test-utils'
 
 import { Focusable } from '../index'
+import type { FocusableRenderOptions } from '../props'
 
 describe('<Focusable />', async () => {
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'spy' implicitly has an 'any' type.
-  const lastCall = (spy) => spy.lastCall.args[0]
+  const lastCall = (spy: any) => spy.lastCall.args[0]
 
   it('should render', async () => {
     const subject = await mount(
       <Focusable>{() => <button>hello world</button>}</Focusable>
     )
-
     expect(subject.getDOMNode()).to.exist()
   })
 
   it('should call children function with focused when element receives focus', async () => {
     const renderSpy = spy()
-
     await mount(
       <Focusable>
         {(args) => {
@@ -115,8 +113,7 @@ describe('<Focusable />', async () => {
     const props = {
       /* eslint-disable react/display-name */
       /* eslint-disable react/prop-types */
-      // @ts-expect-error ts-migrate(7031) FIXME: Binding element 'focused' implicitly has an 'any' ... Remove this comment to see the full error message
-      render: ({ focused }) => {
+      render: ({ focused }: { focused: boolean }) => {
         return (
           <div>
             {focused ? (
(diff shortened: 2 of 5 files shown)


packages/ui-focusable/src/Focusable/__tests__/Focusable.test.tsx
```diff
@@ -331,32 +327,31 @@ describe('<Focusable />', async () => {
...
 
   it('should provide a focus method', async () => {
-    let focusable
+    let focusable: Focusable
     await mount(
       <Focusable
-        // @ts-expect-error ts-migrate(2769) FIXME: No overload matches this call.
-        componentRef={(el) => {
+        //@ts-expect-error TODO this is coming from ReactComponentWrapper
+        componentRef={(el: Focusable) => {
```
matyasf, line 334: this prop is added only during testing by ReactComponentWrapper, I had no good idea who to type it.

packages/ui-focusable/src/Focusable/index.tsx
```diff
@@ -73,12 +83,12 @@ class Focusable extends Component<FocusableProps> {
 
   state = {
     focused: false,
-    focusable: false
-  }
+    focusable: undefined
```
matyasf, line 86: This was really messy. It could either have a boolean value (I think only false?) or the `HTMLElement` that is focusable. I've rewritten the code so its value can be `HtmlElement` or `undefined`

packages/ui-focusable/src/Focusable/index.tsx
```diff
@@ -194,34 +194,31 @@ class Focusable extends Component<FocusableProps> {
...
+    const focusableArr = findFocusable(this, () => true, true) || []
+    const focusableCount = (focusableArr && focusableArr.length) || 0
 
     warn(
       focusableCount === 1,
       `[Focusable] Exactly one focusable child is required (${focusableCount} found).`
     )
 
-    // @ts-expect-error ts-migrate(2322) FIXME: Type 'unknown' is not assignable to type 'unknown[... Remove this comment to see the full error message
-    focusable = focusable ? focusable[0] : false
```
matyasf, line 206: Again some mess with types. I've introduced a new variable for the array of `focusable`s

packages/ui-focusable/src/Focusable/index.tsx
ke1k0: Just a question: what is the practical difference between this (`as FocusableState`) and this?
```
state: FocusableState = {
  focused: false,
  focusable: undefined
}
```
As far as I could tell so far the second is telling the variable what type to allow, the `as` cast is a more forceful action, and can even hide possible tiny bugs, that's why I'm trying to avoid using it whenever I can. 

Can we talk about this on the next standup? It probably isn't a breaking issue here, just wanna discuss it in general (Interesting article: https://dev.to/ylerjen/typescript-cast-is-a-type-breaker-1jbh)

packages/ui-focusable/src/Focusable/index.tsx
```diff
@@ -266,13 +258,16 @@ class Focusable extends Component<FocusableProps> {
       const rendered = render({
         focused,
         focusable,
-        focusVisible: this.isFocusVisible(focusable, focused),
+        focusVisible: this.isFocusVisible(focused, focusable),
         attachRef: this.attachRef
       })
-
+      type ElementWithRef = { ref?: () => unknown }
```
ke1k0, line 264: I would probably expand the type in the `props.ts`, but not sure that's the better approach.

```
type FocusableRenderFunction = (
  opts: FocusableRenderOptions
) => (
  | ReactElement
  | DOMElement<any, any>
  | CElement<any, any>
  | FunctionComponentElement<any>
) & { ref?: () => unknown }
```

Actually, only `ReactElement` doesn't have this prop, and all the others (`CElement` and `FunctionComponentElement` and `DOMElement`) are extending it, so is it needed?

packages/ui-focusable/src/Focusable/props.ts
```diff
@@ -37,16 +73,7 @@ type AllowedPropKeys = Readonly<Array<PropKeys>>
 type FocusableProps = FocusableOwnProps
 
 const propTypes: PropValidators<PropKeys> = {
-  /**
-   * @param {Object} renderProps
-   * @param {Boolean} renderProps.focused - Is the element focused (via keyboard only)?
-   * @param {HTMLElement} renderProps.focusable - The focusable element
-   * @param {Boolean} renderProps.focusVisible - Whether the focus state should be visible or not
```
ke1k0, line 44: By removing this block this options don't show up on the docs page anymore, it only says `opts: FocusableRenderOptions` :/ It is not nice, but maybe we should add it back until find a way to render the docs better.

packages/ui-focusable/src/Focusable/README.md
```diff
@@ -11,26 +11,27 @@ a `render` or `children` method that returns content which must have exactly one
...
-        borderWidth="small"
-        display="block"
-      >
-        I&#39;m focused!
-      </ContextView>
-    )}
-  </span>
-)}
+  {(options) => {
+    console.log(options)
```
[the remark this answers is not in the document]
matyasf: yes, I wanted users to see what the options object contains

HerrTopi (approved):

packages/ui-focusable/src/Focusable/index.tsx
matyasf: this is here by accident, I've added it as normal type not as a cast

packages/ui-focusable/src/Focusable/index.tsx
```diff
@@ -36,12 +36,22 @@ import { propTypes, allowedProps } from './props'
 import type { FocusableProps } from './props'
 import { createChainedFunction } from '@instructure/ui-utils'
 
+type FocusableState = {
+  focused: boolean
+  focusable?: HTMLElement
+}
```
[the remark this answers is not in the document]
matyasf: done

packages/ui-focusable/src/Focusable/index.tsx
matyasf: sadly that's still not enough becuase `ref` needs to be a function, and sometimes its not one. I've added the ref to props.

packages/ui-focusable/src/Focusable/props.ts
matyasf: I've added it back, but it looks wrong, I cant seem to specify an elements in an object properly (same issue in e.g. Editable/children). The original was even worse, it suggested that these functions have these as separate parameters when actually there is one object with these fields.

ke1k0 (approved):

PR #806, opened 2021-11-19
INSTUI-3029 INSTUI-3044 ui-a11y-utils and ui-dialog typing

matyasf:
These needed to be typed together because they share props.


packages/ui-a11y-utils/src/FocusRegion.ts
```diff
@@ -36,97 +35,79 @@ import { logError as error } from '@instructure/console'
...
 
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'event' implicitly has an 'any' type.
-  handleKeyUp = (event) => {
+  handleKeyUp = (event: KeyboardEvent) => {
     if (
-      // @ts-expect-error ts-migrate(2339) FIXME: Property '_options' does not exist on type 'FocusR... Remove this comment to see the full error message
       this._options.shouldCloseOnEscape &&
-      // @ts-expect-error ts-migrate(2339) FIXME: Property 'escape' does not exist on type 'CodesMap... Remove this comment to see the full error message
-      event.keyCode === keycode.codes.escape &&
+      event.keyCode === keycode.codes.esc &&
```
matyasf, line 103: actually both are OK, just `escape` is not typed.

packages/ui-a11y-utils/src/FocusRegion.ts
```diff
@@ -144,63 +125,59 @@ class FocusRegion {
...
 
-      // @ts-expect-error ts-migrate(2339) FIXME: Property '_options' does not exist on type 'FocusR... Remove this comment to see the full error message
       if (this._options.shouldCloseOnDocumentClick) {
         this._listeners.push(
-          // @ts-expect-error ts-migrate(2345) FIXME: Argument of type '{ remove(): void; }' is not assi... Remove this comment to see the full error message
-          addEventListener(doc, 'click', this.captureDocumentClick, true)
+          addEventListener(
+            doc,
+            'click',
+            this.captureDocumentClick as EventListener,
```
matyasf, line 136: I don't know why TS needs this cast, its issue is that `(e:MouseEvent)=>void` is not compatible with `(e:Event)=>void`

packages/ui-a11y-utils/src/FocusRegionManager.ts
```diff
@@ -45,60 +55,57 @@ class FocusRegionManager {
...
       if (region) {
         // If the active region is triggering a new focus region that does not have
         // keyboard focusable content, don't deactivate the active region's keyboard
         // focus region
-        region.deactivate(
-          region.focused && !keyboardFocusable && { keyboard: false }
-        )
+        const keyboard =
+          region.focused && !keyboardFocusable ? { keyboard: false } : undefined
+        region.deactivate(keyboard)
```
matyasf, line 103: just some refactoring because TS did not get the previous code

packages/ui-dom-utils/src/index.ts
```diff
@@ -41,7 +41,6 @@ export type { RectType } from './getBoundingClientRect'
 export { getClassList } from './getClassList'
 export { getComputedStyle } from './getComputedStyle'
 export { getFontSize } from './getFontSize'
-export { getFrameDocumentSafe } from './getFrameDocumentSafe'
```
matyasf, line 44: I deleted this because it was a trivial function, was used in 1 place, was not shown in the documentation and no one uses it internally.

packages/ui-a11y-utils/src/KeyboardFocusRegion.ts
```diff
@@ -41,55 +41,49 @@ import { logError as error } from '@instructure/console'
...
-      onBlur: (event) => {},
-      // @ts-expect-error ts-migrate(6133) FIXME: 'event' is declared but its value is never read.
-      onDismiss: (event) => {},
       defaultFocusElement: null
     }
-
-    // @ts-expect-error ts-migrate(2339) FIXME: Property '_options' does not exist on type 'Keyboa... Remove this comment to see the full error message
     if (this._options.shouldReturnFocus) {
-      this._focusLaterElement = this.activeElement
+      this._focusLaterElement = getActiveElement(this.doc)
```
ke1k0, line 65: Just curious: why did you remove the getter for this?

packages/ui-a11y-utils/src/KeyboardFocusRegion.ts
```diff
@@ -41,55 +41,49 @@ import { logError as error } from '@instructure/console'
...
+  private _focusLaterElement: Element | null = null
+  private _needToFocus = false
+  private _listeners: ReturnType<typeof addEventListener>[] = []
+  private _raf: RequestAnimationFrameType[] = []
+  private _active = false
+  private _wasDocumentClick?: boolean
+  public _contextElement?: Node
+
+  constructor(element: UIElement, options: FocusRegionOptions) {
+    this._contextElement = findDOMNode(element) as Node
```
ke1k0, line 58: Can this somehow be `Window` in some cases?

packages/ui-a11y-utils/src/KeyboardFocusRegion.ts
```diff
@@ -268,62 +230,55 @@ class KeyboardFocusRegion {
...
               'keydown',
-              this.handleFirstTabbableKeyDown
+              this.handleFirstTabbableKeyDown as EventListener
             )
           )
           this._listeners.push(
-            // @ts-expect-error ts-migrate(2345) FIXME: Argument of type '{ remove(): void; }' is not assi... Remove this comment to see the full error message
             addEventListener(
-              this.lastTabbable || defaultFocusElement,
+              this.lastTabbable || defaultFocusElement!,
```
ke1k0, line 267: This made me think, that the `findFocusable` util might not be typed correctly, since it might give back and empty array, might it not? Seems logical, that there are no focusable/tabbable elements on a page. If so, then `this.tabbable` might be empty and both `firstTabbable` and `lastTabbable` can be undefined. (might not be the scope of this ticket)

packages/ui-dialog/src/Dialog/props.ts
```diff
@@ -28,30 +28,34 @@ import PropTypes from 'prop-types'
...
-  shouldCloseOnDocumentClick?: boolean
-  shouldCloseOnEscape?: boolean
-  shouldFocusOnOpen?: boolean
-  elementRef?: (element: Element | null) => void
-}
+  /**
+   * An element or a function returning an element that wraps the content of the `<Dialog />`
+   */
+  contentElement?: UIElement
+} & FocusRegionOptions
```
ke1k0, line 58: I like that this is used from FocusRegion, but the problem is, that the current doc generation cannot read the comments from imported files, so it doesn't appear on Dialogs Properties block. For a workaround (until this is fixed somehow) I suggest listing the props like this: 
```js
...
  /**
   * An element, function returning an element, or array of elements that will not be hidden from
   * the screen reader when the `<Dialog />` is open
   */
  liveRegion?: FocusRegionOptions['liveRegion']

  /**
   * Function called when tab focus leaves the `<Dialog />` focusable content. This only
   * occurs when `shouldContainFocus` is set to false.
   */
  onBlur?: FocusRegionOptions['onBlur']
...
```
etc...

What do you think?

ke1k0 (commented):
Nice job! Just had a few questions and one suggestion. :)

packages/ui-a11y-utils/src/KeyboardFocusRegion.ts
matyasf: It was not used anywhere

packages/ui-a11y-utils/src/KeyboardFocusRegion.ts
matyasf: I think it can, but I dont see a use for using this on the whole `Window`. Later we could go over these casts and throw errors instead (now if you try to use it in a `Window` it will crash). But this will likely introduce breaking changes.

packages/ui-a11y-utils/src/KeyboardFocusRegion.ts
matyasf: yes, you're right, I've added that they can be `undefined`

packages/ui-a11y-utils/src/KeyboardFocusRegion.ts
ke1k0: It was used in `focusDefaultElement() {}` :) I prefer writing getters if something is used in 2+ places, but can live with it like this :)

packages/ui-a11y-utils/src/KeyboardFocusRegion.ts
ke1k0: Ok, thanks!

packages/ui-a11y-utils/src/KeyboardFocusRegion.ts
ke1k0: Shall we make a ticket for fixing "findFocusable"?

packages/ui-dialog/src/Dialog/props.ts
matyasf: hmm I'm not a fan of it, we're duplicating code because of a buggy tool :/ Also its not making things much better, because `LiveRegion` is in `shared-types`. 
And the code documentation is already readable in the IDE autocomplete popup.

ke1k0: Then another idea: if you put back the original comments for those props in the propTypes block, the doc generation can pick those up, and we don't have to touch the types.

HerrTopi (approved):

packages/ui-dialog/src/Dialog/props.ts
matyasf: OK, did this, the comments are now showing up.

PR #807, opened 2021-11-19
chore(ui-heading): type heading

HerrTopi:
Closes: INSTUI-3054


packages/ui-heading/src/Heading/__examples__/Heading.examples.ts
```diff
@@ -32,26 +34,19 @@ export default {
...
         ? 'primary-inverse'
         : 'primary'
     }
-  },
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'props' implicitly has an 'any' type.
-  filter: (props) => {
-    return (
-      (props.type === 'notification' && props.countUntil) ||
-      (props.standalone && props.placement !== 'start top')
-    )
```
ke1k0, line 55: Lol these are the props of Badge, seems like this has been like this forever, since the component was added :D Good thing that TS catches these :)

ke1k0 (approved):

Brailor (approved):

PR #809, opened 2021-11-19
chore(ui-grid): add TS types to `ui-grid` package

ke1k0:
Closes: INSTUI-3053

### DIFF
--- a/packages/ui-grid/src/GridCol/styles.ts
+++ b/packages/ui-grid/src/GridCol/styles.ts
@@ -26,6 +26,8 @@ import type { GridTheme } from '@instructure/shared-types'
 import type { GridBreakpoints } from '../GridTypes'
 import type { GridColProps, GridColStyle } from './props'
 
+type BreakPoints = NonNullable<GridBreakpoints>
+
 /**
  * ---
  * private: true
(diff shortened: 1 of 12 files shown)


packages/ui-grid/src/GridCol/styles.ts
```diff
@@ -111,31 +113,29 @@ const generateStyle = (
...
-  const getColSize = (breakpoint: GridBreakpoints) => {
+  const getColSize = (breakpoint: BreakPoints) => {
     let { width } = props
 
     if (!width) return
 
     if (width && typeof width === 'object') {
-      // @ts-expect-error TODO: type when typing component
-      width = width[breakpoint]
+      width = width[breakpoint === 'x-large' ? 'xLarge' : breakpoint]
```
matyasf, line 126: was this a bug that you found while doing the types?

packages/ui-grid/src/GridRow/props.ts
```diff
@@ -38,6 +38,9 @@ import type { WithStyleProps, ComponentStyle } from '@instructure/emotion'
 import type { GridBreakpoints } from '../GridTypes'
 
 type GridRowOwnProps = {
+  /**
+   * On of: `Grid.Col`, `ScreenReaderContent`
```
matyasf, line 42: typo: On of

matyasf (approved):

packages/ui-grid/src/GridCol/styles.ts
ke1k0: Yeah, it seems so, but there were no complaints about it before.

ke1k0: I didn't think it was important to mention as a "fix" commit and show it in the Changelog, do you think I should?

packages/ui-grid/src/GridRow/props.ts
ke1k0: Ooops, thanks, corrected!

HerrTopi (approved):

PR #811, opened 2021-11-22
chore(ui-pages): add TS types to `ui-pages` package

ke1k0:
Closes: INSTUI-3067

### DIFF
--- a/packages/ui-pages/src/Pages/index.tsx
+++ b/packages/ui-pages/src/Pages/index.tsx
@@ -170,7 +170,9 @@ class Pages extends Component<PagesProps> {
   }
 
   render() {
-    return this.activePage ? (
+    const { activePage } = this
+
+    return activePage ? (
       <PagesContext.Provider
         value={{
           history: this._history,
@@ -187,7 +189,7 @@ class Pages extends Component<PagesProps> {
           role="region"
           elementRef={this.handleRef}
         >
-          {this.activePage}
+          {activePage}
         </View>
       </PagesContext.Provider>
     ) : null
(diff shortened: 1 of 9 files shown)


packages/ui-pages/babel.config.js
```diff
@@ -23,11 +23,6 @@
  */
 
 module.exports = {
-  shouldPrintComment(value) {
-    const isTsRelatedComment = /@ts-expect-error|FIXME/.test(value)
-
-    return !isTsRelatedComment
-  },
```
HerrTopi, line 30: Should we remove this, since the locator file still has the ts-expect-errors?

HerrTopi (approved):

packages/ui-pages/babel.config.js
ke1k0: The Pages component doesn't have locators, so there are no more ts-expect-errors in the package.

packages/ui-pages/src/Pages/index.tsx
```diff
@@ -59,38 +59,39 @@ class Pages extends Component<PagesProps> {
   static defaultProps = {
     children: null,
     defaultPageIndex: null,
```
matyasf, line 61: I think that these 2 are not needed

ke1k0: Removed them, thanks!

matyasf (commented):
looks good, just one thing: Its Readme.md has no text at all what this component is, just some examples. Could you add 1-2 sentences?

matyasf (approved):

PR #816, opened 2021-11-24
chore(ui-a11y-utils,ui-popover,ui-position,ui-view): type Popover

ke1k0:
Closes: INSTUI-3070

### DIFF
--- a/packages/__docs__/src/Properties/index.js
+++ b/packages/__docs__/src/Properties/index.js
@@ -85,11 +85,11 @@ class Properties extends Component {
           !propsToIgnore.includes(name)
         )
       })
-      .map((name) => {
+      .map((name, idx) => {
         const prop = props[name]
 
         return (
-          <Table.Row key={name}>
+          <Table.Row key={idx}>
             <Table.Cell>
               <code>{name}</code>
             </Table.Cell>
@@ -206,8 +206,8 @@ class Properties extends Component {
       return <span>{type.value}</span>
     }
 
-    const values = type.value.map(({ value }) => (
-      <li css={styles.listItem} key={value}>
+    const values = type.value.map(({ value }, idx) => (
+      <li css={styles.listItem} key={idx}>
         <code>{this.unquote(value)}</code>
       </li>
     ))
@@ -230,8 +230,8 @@ class Properties extends Component {
     if (!Array.isArray(type.value)) {
       return <span>{type.value}</span>
     }
-    const values = type.value.map((value) => (
-      <li css={styles.listItem} key={value.name}>
+    const values = type.value.map((value, idx) => (
+      <li css={styles.listItem} key={idx}>
         <code>{this.renderType(value)}</code>
       </li>
     ))
@@ -259,13 +259,13 @@ class Properties extends Component {
     // so we have to parse the raw data
     const elements = tsType.raw.split('|').filter((item) => item !== '')
 
-    const values = elements.map((rawValue) => {
+    const values = elements.map((rawValue, idx) => {
       let value = rawValue.trim()
       if (this.isTsGeneralFunction(value)) {
         value = 'function'
       }
       return (
-        <li css={styles.listItem} key={value}>
+        <li css={styles.listItem} key={idx}>
           <code>{value ? this.unquote(value) : value}</code>
         </li>
       )
--- a/packages/ui-a11y-utils/src/ScreenReaderFocusRegion.ts
+++ b/packages/ui-a11y-utils/src/ScreenReaderFocusRegion.ts
@@ -29,13 +29,13 @@ class ScreenReaderFocusRegion {
   private _parents: HTMLElement[] = []
   private _nodes: Element[] = []
   private _liveRegion: (Element | undefined)[]
-  private _contextElement?: Element | Node
+  private _contextElement: Element | Node | null
   private _options: FocusRegionOptions
   private _observer: MutationObserver | null = null
   private _attributes: [Element, string, string][] = []
 
   constructor(
-    element: Element | Node,
+    element: Element | Node | null,
     options: FocusRegionOptions = {
       shouldContainFocus: true,
       liveRegion: []
--- a/packages/ui-popover/src/Popover/index.tsx
+++ b/packages/ui-popover/src/Popover/index.tsx
@@ -39,16 +39,20 @@ import {
   requestAnimationFrame,
   handleMouseOverOut
 } from '@instructure/ui-dom-utils'
-import type { RequestAnimationFrameType } from '@instructure/ui-dom-utils'
 
 import { safeCloneElement, callRenderProp } from '@instructure/ui-react-utils'
 import { createChainedFunction, shallowEqual, px } from '@instructure/ui-utils'
 import { logError as error } from '@instructure/console'
 import { uid } from '@instructure/uid'
 import { testable } from '@instructure/ui-testable'
-
 import { FocusRegion } from '@instructure/ui-a11y-utils'
-import type { PopoverProps } from './props'
+
+import type { RequestAnimationFrameType } from '@instructure/ui-dom-utils'
+import type { ViewProps, ContextViewProps } from '@instructure/ui-view'
+import type { PositionProps } from '@instructure/ui-position'
+import type { DialogProps } from '@instructure/ui-dialog'
+
+import type { PopoverProps, PopoverState } from './props'
 import { allowedProps, propTypes } from './props'
 
 /**
@@ -56,10 +60,11 @@ import { allowedProps, propTypes } from './props'
 category: components
 tags: overlay, portal, dialog
 ---
+@tsProps
 **/
 @textDirectionContextConsumer()
 @testable()
-class Popover extends Component<PopoverProps> {
+class Popover extends Component<PopoverProps, PopoverState> {
   static readonly componentId = 'Popover'
 
   static allowedProps = allowedProps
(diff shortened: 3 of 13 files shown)


packages/__docs__/src/Properties/index.js
```diff
@@ -206,8 +206,8 @@ class Properties extends Component {
       return <span>{type.value}</span>
     }
 
-    const values = type.value.map(({ value }) => (
-      <li css={styles.listItem} key={value}>
+    const values = type.value.map(({ value }, idx) => (
+      <li css={styles.listItem} key={idx}>
```
ke1k0, line 210: These needed to be fixed, because the values could clash as identical keys in the .map()

packages/ui-popover/src/Popover/index.tsx
```diff
@@ -86,72 +89,50 @@ class Popover extends Component<PopoverProps> {
...
       placement: props.placement,
       offsetX: props.offsetX,
-      offsetY: props.offsetY
-    }
-
-    if (typeof props.isShowingContent === 'undefined') {
-      // @ts-expect-error ts-migrate(2339) FIXME: Property 'isShowingContent' does not exist on type... Remove this comment to see the full error message
-      this.state.isShowingContent = props.defaultIsShowingContent
+      offsetY: props.offsetY,
+      isShowingContent:
```
ke1k0, line 102: Just a tiny refact because TS was complaining that it is not the way to set the state.

packages/ui-popover/src/Popover/index.tsx
```diff
@@ -506,12 +452,14 @@ class Popover extends Component<PopoverProps> {
...
+        expanded = undefined
       }
 
       trigger = safeCloneElement(trigger, {
-        // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'el' implicitly has an 'any' type.
-        ref: (el) => (this._trigger = el),
+        // TODO: try to type `ref` better, LegacyRef<T> was not compatible
+        ref: (el: any) => {
+          this._trigger = el
+        },
```
ke1k0, line 462: Typing the trigger was a nightmare, because it can be anything really. I also couldn't type it correctly, because it was always complaining about the LegacyRef not compatible with this callback. I' gladly accept any ideas how to do this better :)

packages/ui-popover/src/Popover/index.tsx
```diff
@@ -565,62 +512,56 @@ class Popover extends Component<PopoverProps> {
...
-      let ViewElement
       const color = this.props.color
 
-      let viewProps = {
-        // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'c' implicitly has an 'any' type.
-        ref: (c) => (this._view = c),
-        // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'el' implicitly has an 'any' type.
-        elementRef: (el) => {
-          // @ts-expect-error ts-migrate(2339) FIXME: Property '_contentElement' does not exist on type ... Remove this comment to see the full error message
+      let viewProps: (Partial<ViewProps> | Partial<ContextViewProps>) & {
```
ke1k0, line 517: I made a little refactor here too. It isn't easy to type, because this can be either the props of View of ContextView, and can return View os ContextView.

packages/ui-popover/src/Popover/props.ts
```diff
@@ -33,270 +33,300 @@ import type { Shadow, Stacking } from '@instructure/emotion'
...
    * A function that returns a reference to the content element
    */
-  contentRef: PropTypes.func,
+  contentRef?: (contentElement: Element | null) => void
+
   /**
    * An element or a function returning an element to focus by default
    */
-  defaultFocusElement: PropTypes.oneOfType([PropTypes.element, PropTypes.func]),
+  defaultFocusElement?: UIElement
```
ke1k0, line 90: I noticed sth strange here: TS can read the types imported from shared-types well and they appear in the Props object correctly, but any type imported from other (component) packages does not. It might be worth exploring.

packages/ui-a11y-utils/src/ScreenReaderFocusRegion.ts
```diff
@@ -29,13 +29,13 @@ class ScreenReaderFocusRegion {
   private _parents: HTMLElement[] = []
   private _nodes: Element[] = []
   private _liveRegion: (Element | undefined)[]
-  private _contextElement?: Element | Node
+  private _contextElement: Element | Node | null
```
matyasf, line 32: Just a sidenote: I dont like that our code is using `undefined` and `null` to indicate non-existence of a value seemingly randomly. We should either stick to one or have some kind of rule when we use one.

packages/ui-popover/src/Popover/index.tsx
```diff
@@ -86,72 +89,50 @@ class Popover extends Component<PopoverProps> {
...
     })
     this._handleMouseOut = handleMouseOverOut.bind(null, (event) => {
       // this is needed bc the trigger mouseOut fires before tooltip mouseOver
-      // @ts-expect-error ts-migrate(2339) FIXME: Property 'mouseOutTimeout' does not exist on type ... Remove this comment to see the full error message
       this.mouseOutTimeout = setTimeout(() => {
         this.hide(event)
       }, 1)
     })
-    // @ts-expect-error ts-migrate(2339) FIXME: Property 'mouseOutTimeout' does not exist on type ... Remove this comment to see the full error message
     this.mouseOutTimeout = undefined
```
matyasf, line 121: this line is not needed

packages/ui-popover/src/Popover/index.tsx
```diff
@@ -86,72 +89,50 @@ class Popover extends Component<PopoverProps> {
...
+
+  _id: string
   _raf: RequestAnimationFrameType[] = []
+  _trigger?: React.ReactElement
+  _view: View | ContextView | null = null
+  _dialog: Dialog | null = null
+  _contentElement: Element | null = null
+  _focusRegion?: FocusRegion
+
+  mouseOutTimeout?: ReturnType<typeof setTimeout>
```
matyasf, line 135: I'd make all these `private`, so autocomplete does not offer them if someone gets a `ref` to this component

packages/ui-popover/src/Popover/index.tsx
```diff
@@ -315,124 +279,114 @@ class Popover extends Component<PopoverProps> {
...
       this.props.shouldFocusContentOnTriggerBlur
     ) {
       return
     }
     this.hide(event)
   }
 
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'event' implicitly has an 'any' type.
-  handleTriggerKeyDown = (event) => {
+  handleTriggerKeyDown = (event: Event) => {
```
matyasf, line 353: This can be typed as KeyboardEvent then you dont need the cast below

packages/ui-popover/src/Popover/index.tsx
```diff
@@ -315,124 +279,114 @@ class Popover extends Component<PopoverProps> {
...
           this._dialog && this._dialog.focus()
         })
       )
     }
   }
 
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'event' implicitly has an 'any' type.
-  handleTriggerKeyUp = (event) => {
-    if (event.keyCode === keycode.codes.esc && this.shown && this.isTooltip) {
+  handleTriggerKeyUp = (event: Event) => {
```
matyasf, line 371: This can be typed as KeyboardEvent then you dont need the cast below

packages/ui-popover/src/Popover/index.tsx
matyasf: For me its accepting `ref: (el: React.ReactElement) => {`, isnt this OK?

packages/ui-popover/src/Popover/index.tsx
```diff
@@ -86,72 +89,50 @@ class Popover extends Component<PopoverProps> {
...
       this.mouseOutTimeout = setTimeout(() => {
         this.hide(event)
       }, 1)
     })
-    // @ts-expect-error ts-migrate(2339) FIXME: Property 'mouseOutTimeout' does not exist on type ... Remove this comment to see the full error message
     this.mouseOutTimeout = undefined
   }
 
+  _handleMouseOver: (event: MouseEvent) => void
+  _handleMouseOut: (event: MouseEvent) => void
```
matyasf, line 125: I've been playting with typing these properly.  Since there are on `View` they are actually `React.MouseEvent<any>`. If you type them this way you dont need the cast on line 536, also I'd type `handleMouseOverOut` as `event: { currentTarget: EventTarget | null, relatedTarget?: EventTarget | null}`

matyasf (approved):

packages/ui-popover/src/Popover/index.tsx
ke1k0: Yeah, I tried it before and was throwing errors, but works fine now, thanks :)

packages/ui-a11y-utils/src/ScreenReaderFocusRegion.ts
ke1k0: True, it is probably worth making a ticket for it to decide and solve it globally.

packages/ui-popover/src/Popover/index.tsx
ke1k0: removed, thanks

packages/ui-popover/src/Popover/index.tsx
ke1k0: done

packages/ui-popover/src/Popover/index.tsx
ke1k0: React.MouseEventHandler did the trick, thanks!

Brailor (approved):

PR #821, opened 2021-11-29
Ts type ui-progress (INSTUI-3073)

HerrTopi:
Closes: INSTUI-3073

### DIFF
--- a/packages/ui-progress/src/ProgressBar/__examples__/ProgressBar.examples.tsx
+++ b/packages/ui-progress/src/ProgressBar/__examples__/ProgressBar.examples.tsx
@@ -22,7 +22,7 @@
  * SOFTWARE.
  */
 import React from 'react'
-import { ProgressBarProps } from '../props'
+import type { ProgressBarProps, Values } from '../props'
 
 const valueMax = 100
 
@@ -32,13 +32,9 @@ export default {
     valueNow: [0, 40, 80, 100],
     renderValue: [
       null,
-      ({
-        valueNow,
-        valueMax
-      }: {
-        valueNow: ProgressBarProps['valueNow']
-        valueMax: ProgressBarProps['valueMax']
-      }) => <span>{Math.round((valueNow! / valueMax!) * 100)}</span>
+      ({ valueNow, valueMax }: Values) => (
+        <span>{Math.round((valueNow / valueMax) * 100)}</span>
+      )
     ]
   },
   getComponentProps: () => {
@@ -53,6 +49,5 @@ export default {
         ? 'primary-inverse'
         : 'primary'
     }
-  },
-  filter: () => {}
+  }
 }
--- a/packages/ui-progress/src/ProgressBar/__examples__/ProgressBar.examples.tsx
+++ b/packages/ui-progress/src/ProgressBar/__examples__/ProgressBar.examples.tsx
@@ -22,6 +22,7 @@
  * SOFTWARE.
  */
 import React from 'react'
+import { ProgressBarProps } from '../props'
 
 const valueMax = 100
 
@@ -31,28 +32,27 @@ export default {
     valueNow: [0, 40, 80, 100],
     renderValue: [
       null,
-      // @ts-expect-error ts-migrate(7031) FIXME: Binding element 'valueNow' implicitly has an 'any'... Remove this comment to see the full error message
-      // eslint-disable-next-line react/display-name
-      ({ valueNow, valueMax }) => (
-        <span>{Math.round((valueNow / valueMax) * 100)}</span>
-      )
+      ({
+        valueNow,
+        valueMax
+      }: {
+        valueNow: ProgressBarProps['valueNow']
+        valueMax: ProgressBarProps['valueMax']
+      }) => <span>{Math.round((valueNow! / valueMax!) * 100)}</span>
     ]
   },
-  // @ts-expect-error ts-migrate(6133) FIXME: 'props' is declared but its value is never read.
-  getComponentProps: (props) => {
+  getComponentProps: () => {
     return {
       screenReaderLabel: 'Passing students',
       valueMax
     }
   },
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'props' implicitly has an 'any' type.
-  getExampleProps: (props) => {
+  getExampleProps: (props: ProgressBarProps) => {
     return {
-      background: props.color.includes('inverse')
+      background: props.color!.includes('inverse')
         ? 'primary-inverse'
         : 'primary'
     }
   },
-  // @ts-expect-error ts-migrate(6133) FIXME: 'props' is declared but its value is never read.
-  filter: (props) => {}
+  filter: () => {}
 }
--- a/packages/ui-progress/src/ProgressBar/props.ts
+++ b/packages/ui-progress/src/ProgressBar/props.ts
@@ -67,7 +67,7 @@ type ProgressBarOwnProps = {
   /**
    * A function for formatting the text provided to screen readers via `aria-valuenow`
    */
-  formatScreenReaderValue?: (...args: any[]) => any
+  formatScreenReaderValue?: (values: Values) => string
   /**
    * A function to format the displayed value. If null the value will not display.
    * Takes `valueNow` and `valueMax` as parameters.
--- a/packages/ui-progress/src/ProgressBar/props.ts
+++ b/packages/ui-progress/src/ProgressBar/props.ts
@@ -46,18 +46,55 @@ export type ProgressBarMeterColor =
   | 'brand'
 
 type ProgressBarOwnProps = {
+  /**
+   * A label is required for accessibility
+   */
   screenReaderLabel: string
+  /**
+   * Control the height of the progress bar
+   */
   size?: 'x-small' | 'small' | 'medium' | 'large'
+  /**
+   * Maximum value (defaults to 100)
+   */
   valueMax?: number
+  /**
+   * Receives the progress of the event
+   */
   valueNow?: number
+  /**
+   * A function for formatting the text provided to screen readers via `aria-valuenow`
+   */
   formatScreenReaderValue?: (...args: any[]) => any
+  /**
+   * A function to format the displayed value. If null the value will not display.
+   * Takes `valueNow` and `valueMax` as parameters.
+   */
   renderValue?: ((...args: any[]) => any) | React.ReactNode
+  /**
+   * Controls the overall color scheme of the component
+   */
   color?: 'primary' | 'primary-inverse'
+  /**
+   * Control the color of the progress meter. Defaults to showing theme success
+   * color on completion, based on `valueNow` and `valueMax`.
+   */
   meterColor?:
     | ((...args: any[]) => ProgressBarMeterColor)
     | ProgressBarMeterColor
+  /**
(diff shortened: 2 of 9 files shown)


packages/ui-progress/src/ProgressBar/props.ts
```diff
@@ -45,85 +45,87 @@ export type ProgressBarMeterColor =
...
   /**
    * Receives the progress of the event
    */
-  valueNow: PropTypes.number,
+  valueNow?: Values['valueNow']
   /**
    * A function for formatting the text provided to screen readers via `aria-valuenow`
    */
-  formatScreenReaderValue: PropTypes.func,
+  formatScreenReaderValue?: (...args: any[]) => any
```
ke1k0, line 70: This can be typed too: `(values: Values) => string`

ke1k0: Also, after typing this, there is a lil change required in render:

```js
const valueText =
      typeof formatScreenReaderValue === 'function'
        ? formatScreenReaderValue({ valueNow: valueNow!, valueMax: valueMax! })
        : undefined
```
Mind that refactoring it to ternary is needed too, because "false" is not accpeted as value of `aria-valuetext`.

packages/ui-progress/src/ProgressBar/styles.ts
```diff
@@ -39,7 +39,7 @@ const generateStyle = (
   componentTheme: ProgressBarTheme,
   props: ProgressBarProps
 ): ProgressBarStyle => {
-  const { valueNow, valueMax, size, color, meterColor } = props
+  const { valueNow = 0, valueMax = 100, size, color, meterColor } = props
```
ke1k0, line 42: It is a bit risky to set default values for these here too, because they can get out of sync with the defaults set in defaultProps. But logically, these values will always stay like this, so probably not so risky in this case :)

packages/ui-progress/src/ProgressBar/__examples__/ProgressBar.examples.tsx
```diff
@@ -31,28 +32,27 @@ export default {
...
-      ({ valueNow, valueMax }) => (
-        <span>{Math.round((valueNow / valueMax) * 100)}</span>
-      )
+      ({
+        valueNow,
+        valueMax
+      }: {
+        valueNow: ProgressBarProps['valueNow']
+        valueMax: ProgressBarProps['valueMax']
+      }) => <span>{Math.round((valueNow! / valueMax!) * 100)}</span>
```
ke1k0, line 41: You can use the `Values` type here so that you don't need the `!`-s

packages/ui-progress/src/ProgressCircle/__examples__/ProgressCircle.examples.tsx
```diff
@@ -31,28 +32,23 @@ export default {
     valueNow: [0, 40, 80, 100],
     renderValue: [
       null,
-      // @ts-expect-error ts-migrate(7031) FIXME: Binding element 'valueNow' implicitly has an 'any'... Remove this comment to see the full error message
-      // eslint-disable-next-line react/display-name
-      ({ valueNow, valueMax }) => (
+      ({ valueNow, valueMax }: { valueNow: number; valueMax: number }) => (
```
ke1k0, line 35: You can use the `Values` type here

packages/ui-progress/src/ProgressCircle/__examples__/ProgressCircle.examples.tsx
```diff
@@ -31,28 +32,23 @@ export default {
...
       screenReaderLabel: 'Passing students',
       valueMax
     }
   },
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'props' implicitly has an 'any' type.
-  getExampleProps: (props) => {
+  getExampleProps: (props: ProgressCircleProps) => {
     return {
-      background: props.color.includes('inverse')
+      background: props?.color?.includes('inverse')
```
ke1k0, line 48: Just for consistency's stake: In ProgressBar you typed it as `props.color!.includes()`, because color is always set by defaultProps, so can be like this here too :)

packages/ui-progress/src/ProgressBar/__examples__/ProgressBar.examples.tsx
```diff
@@ -31,28 +32,27 @@ export default {
...
     return {
-      background: props.color.includes('inverse')
+      background: props.color!.includes('inverse')
         ? 'primary-inverse'
         : 'primary'
     }
   },
-  // @ts-expect-error ts-migrate(6133) FIXME: 'props' is declared but its value is never read.
-  filter: (props) => {}
+  filter: () => {}
```
ke1k0, line 57: I think the filter can be deleted if not used.

packages/ui-progress/src/ProgressCircle/__examples__/ProgressCircle.examples.tsx
```diff
@@ -31,28 +32,23 @@ export default {
...
         ? 'primary-inverse'
         : 'primary'
     }
   },
-  // @ts-expect-error ts-migrate(6133) FIXME: 'props' is declared but its value is never read.
-  filter: (props) => {}
+  filter: () => {}
```
ke1k0, line 53: I think the filter can be deleted if not used.

packages/ui-progress/src/ProgressCircle/index.tsx
```diff
@@ -135,20 +132,17 @@ class ProgressCircle extends Component<
 
     const valueText =
       typeof formatScreenReaderValue === 'function' &&
-      formatScreenReaderValue({ valueNow, valueMax })
+      formatScreenReaderValue({ valueNow: valueNow!, valueMax: valueMax! })
```
ke1k0, line 135: I think this needs a bugfix here. Unlike ProgressBar, the original type of this prop allows a func OR node, but the node is never processed. So I think we should make it the same type as ProgressBar and only accept a function that returns a string.

packages/ui-progress/src/ProgressCircle/index.tsx
```diff
@@ -135,20 +132,17 @@ class ProgressCircle extends Component<
...
     // NVDA does not read aria-valuetext: https://github.com/nvaccess/nvda/issues/913
     // But leaving aria-valuetext because JAWS ignores aria-label
     const labelAndValueText = `${screenReaderLabel} ${valueText}`
 
     const value = callRenderProp(renderValue, { valueNow, valueMax })
 
-    // @ts-expect-error ts-migrate(2339) FIXME: Property 'animateOnMount' does not exist on type '... Remove this comment to see the full error message
-    const style = this.state.animateOnMount
-      ? null
-      : {
```
ke1k0, line 149: This is possibly another bug :D But I checked in v7, same issue here, and there was no complaint about it, so hopefully not an issue.

packages/ui-progress/src/ProgressCircle/props.ts
```diff
@@ -84,62 +125,25 @@ type ProgressCircleStyle = ComponentStyle<
   | 'meter'
   | 'radii'
   | 'dashOffset'
```
ke1k0, line 127: `radii` can be deleted from this union because you define it later

packages/ui-progress/src/ProgressBar/props.ts
HerrTopi: Nice catch, thanks :)

packages/ui-progress/src/ProgressBar/styles.ts
HerrTopi: Couldn't find another sensible solution, and to moving the defaults to a separate variable and exporting it seemed a bit of an overkill. Other than that, since there are always defaults to these values, even if they get out of sync, these values in styles.js will be overwritten every time

packages/ui-progress/src/ProgressBar/__examples__/ProgressBar.examples.tsx
HerrTopi: Darn, missed this. The Values idea was an afterthought, thanks for catching it

packages/ui-progress/src/ProgressBar/__examples__/ProgressBar.examples.tsx
HerrTopi: Honestly, I didn't know if the functions are needed or not, so thanks I know now

packages/ui-progress/src/ProgressCircle/index.tsx
HerrTopi: It is indeed an unused state variable. Probably no one needs it

packages/ui-progress/src/ProgressCircle/index.tsx
```diff
@@ -65,20 +65,18 @@ class ProgressCircle extends Component<
...
 
     // default to showing `success` color on completion
-    // @ts-expect-error ts-migrate(7031) FIXME: Binding element 'valueNow' implicitly has an 'any'... Remove this comment to see the full error message
-    meterColor: ({ valueNow, valueMax }) =>
+    meterColor: ({ valueNow, valueMax }: Values) =>
       valueNow / valueMax >= 1 ? 'success' : 'brand'
   }
 
-  _timeouts = []
+  _timeouts: NodeJS.Timeout[] = []
```
Brailor, line 72: Just a small note:
I know that for some reason TypeScript thinks this is a Node package (probably because the packages.json's `type` field is set to `commonjs`) and thus giving back Node.js related types for apis such as: `setTimeout`, but this code will not run in a Node environment, so this type is not correct here.

A platform independent way of doing the typing for such apis is this:

```typescript

  _timeouts: Array<ReturnType<typeof setTimeout>> = []

```

Brailor: Another note:
this will still set the underlying type to `NodeJS.Timeout`, since TS thinks this is a Node environment, so we should be careful with these kind of situations, and have to remember to not use any kind of Node functionality. (even though the type would allow such things, but the runtime for it is simply not there)

packages/ui-progress/src/ProgressCircle/index.tsx
ke1k0: I meant that maybe ages ago this was needed for the "animate on mount" style, but later the state variable got renamed and this wasn't refactored here by accident. But even when using the current state variable I couldn't see any difference, so probably it is not a problem to remove it (instead of renaming the variable).

packages/ui-progress/src/ProgressCircle/props.ts
```diff
@@ -82,64 +121,26 @@ type ProgressCircleStyle = ComponentStyle<
...
-   * Receives the progress of the event
-   */
   valueNow: PropTypes.number,
-  /**
-   * A function for formatting the text provided to screen readers via `aria-valuenow`
-   */
   formatScreenReaderValue: PropTypes.oneOfType([
     PropTypes.func,
     PropTypes.node
   ]),
```
ke1k0, line 135: Thanks for changing the type for this, please change the propType too!

packages/ui-progress/src/ProgressCircle/index.tsx
HerrTopi: Interesting notes, thanks for explaining it to me (:

ke1k0 (approved):

Issue #810, opened 2021-11-22
TextArea doesn't remember scroll position


buckett:
### Background Information

**Package Version(s)**:
ui-text-area  8.12.0

**Browser**:
Chrome

**OS:**
Mac OS

**Component**:
`<TextArea>`

### Describe the Bug

If you have a TextArea component that has a `maxHeight` set as outlined in https://instructure.design/#TextArea then if you have a `maxHeight` property then when editing content near the start of the textarea it always scrolls to the end.

### Steps To Reproduce
    Steps to reproduce the behavior, e.g.:
    1. Create a `<TextArea>` with a `maxHeight`, this can be seen on https://instructure.design/#TextArea
    2. Paste in some content so that there is a scroll bar in the component, any text will do.
    3. Scroll up to the first line of text and attempt to edit the text.
    4. The component will scroll to show the last line and so you can no longer see the text you are editing.

### Expected Behavior

The component should show the cursor instead of scrolling to the end.

### Screenshots

This shows appending to the last line is fine, but when I try to type on the first line is scrolls to the last line, hiding the cursor.

https://user-images.githubusercontent.com/5921/142864259-695027aa-6a68-452b-a79a-5854add43df1.mov

### Additional Information

**Current Workaround(s):**

Don't use `maxHeight` property.

**Are you willing to submit a PR to fix?**
- [ ] Yes, I'm willing to submit a PR

**Requested Priority:**

HerrTopi:
Thanks for the report! We are looking into this and will let you know whet it's fixed

matyasf:
@buckett We have released a bix for this in InstUI 8.13.0, did this solve you issue? https://github.com/instructure/instructure-ui/releases/tag/v8.13.0

PR #822, opened 2021-11-29
fix(ui-text-area): fix jumping scroll issue (INSTUI-3332)

HerrTopi:
Closes: INSTUI-3332, https://github.com/instructure/instructure-ui/issues/810

when textarea has maxhight set and there are enough lines to have a scrollbar,
if the component is
scrolled up and typed into, it will scroll down on every input.

removing the seemingly useless
scrollheight setter from the code fixes the issue

### DIFF
--- a/packages/ui-text-area/src/TextArea/index.tsx
+++ b/packages/ui-text-area/src/TextArea/index.tsx
@@ -183,7 +183,7 @@ class TextArea extends Component<TextAreaProps> {
     let height = ''
 
     // Notes:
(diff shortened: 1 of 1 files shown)


ke1k0 (approved):
I'm not sure why this force scrollTop was there, might have been some another a11y issue they tried to solve. It might be nice to ask someone (Ákos?) about it, but seems to work fine.

matyasf (approved):
This line of code was added by https://github.com/instructure/instructure-ui/commit/682d336 . I tested everything I could and it worked fine + the tests are passing, so I think its OK to delete.

PR #823, opened 2021-11-29
chore(ui-text-area): type ui-text-area (INSTUI-3088)

HerrTopi:
Closes: INSTUI-3088

### DIFF
--- a/packages/ui-text-area/src/TextArea/index.tsx
+++ b/packages/ui-text-area/src/TextArea/index.tsx
@@ -71,16 +71,17 @@ class TextArea extends Component<TextAreaProps> {
     required: false
   }
 
-  _listener: { remove(): void } | null = null
-  _request: RequestAnimationFrameType | undefined
-  _defaultId: string
-  _textareaResizeListener?: ResizeObserver
-  _debounced?: Debounced
-  _textarea?: HTMLTextAreaElement | null
-  _container?: HTMLDivElement
-  _height?: string
-  _node?: FormField | null
-  _manuallyResized?: boolean
+  private _listener?: { remove(): void }
+  private _request?: RequestAnimationFrameType
+  private _defaultId: string
+  private _textareaResizeListener?: ResizeObserver
+  private _debounced?: Debounced
+  private _textarea: HTMLTextAreaElement | null = null
+  private _container: HTMLDivElement | null = null
+  private _height?: string
+  //TODO remove in v9 since it seems to be unused
+  private _node?: FormField | null = null
+  private _manuallyResized?: boolean
 
   constructor(props: TextAreaProps) {
     super(props)
@@ -173,9 +174,15 @@ class TextArea extends Component<TextAreaProps> {
     this._textarea.style.overflowY = 'hidden' // hide scrollbars for autoGrow textareas
     height = this._textarea.scrollHeight + offset + 'px'
 
-    const maxHeight = px(this.props.maxHeight!, this._container)
+    const maxHeight = this.props.maxHeight
+      ? px(this.props.maxHeight, this._container)
+      : undefined
 
-    if (this.props.maxHeight && this._textarea.scrollHeight > maxHeight) {
+    if (
+      this.props.maxHeight &&
+      maxHeight !== undefined &&
+      this._textarea.scrollHeight > maxHeight
+    ) {
       this._textarea.style.overflowY = 'auto' // add scroll if scrollHeight exceeds maxHeight in pixels
     } else if (this.props.height) {
       if (this._textarea.value === '') {
@@ -190,7 +197,7 @@ class TextArea extends Component<TextAreaProps> {
 
     // preserve container height to prevent scroll jumping on long textareas,
     // but make sure container doesn't exceed maxHeight prop
-    const heightExceedsMax = px(height) > maxHeight
+    const heightExceedsMax = maxHeight !== undefined && px(height) > maxHeight
     if (!heightExceedsMax) {
       this._container!.style.minHeight = height
     }
@@ -221,7 +228,7 @@ class TextArea extends Component<TextAreaProps> {
     }
   }
 
-  handleContainerRef = (node: HTMLDivElement) => {
+  handleContainerRef = (node: HTMLDivElement | null) => {
     this._container = node
   }
 
@@ -291,7 +298,7 @@ class TextArea extends Component<TextAreaProps> {
         id={this.id}
         required={required}
         aria-required={required}
-        aria-invalid={this.invalid ? 'true' : 'false'}
+        aria-invalid={this.invalid ? 'true' : undefined}
         disabled={disabled || readOnly}
         css={this.props.styles?.textArea}
         onChange={this.handleChange}
@@ -300,8 +307,8 @@ class TextArea extends Component<TextAreaProps> {
 
     return (
       <FormField
-        label={null}
         {...pickProps(this.props, FormField.allowedProps)}
+        label={this.props.label}
         vAlign="top"
         id={this.id}
         ref={(el) => {
--- a/packages/ui-text-area/src/TextArea/index.tsx
+++ b/packages/ui-text-area/src/TextArea/index.tsx
@@ -48,6 +48,7 @@ import { allowedProps, propTypes } from './props'
 ---
 category: components
 ---
+@tsProps
 **/
 @withStyle(generateStyle, generateComponentTheme)
 @testable()
@@ -65,20 +66,25 @@ class TextArea extends Component<TextAreaProps> {
     messages: [],
     disabled: false,
     readOnly: false,
-    // @ts-expect-error ts-migrate(6133) FIXME: 'textarea' is declared but its value is never read... Remove this comment to see the full error message
-    textareaRef: function (textarea) {},
+    textareaRef: () => {},
     layout: 'stacked',
     required: false
   }
 
   _listener: { remove(): void } | null = null
   _request: RequestAnimationFrameType | undefined
+  _defaultId: string
+  _textareaResizeListener: any
+  _debounced: any
+  _textarea?: HTMLTextAreaElement | null
+  _container?: HTMLDivElement
+  _height?: string
+  _node?: FormField | null
+  _manuallyResized?: boolean
 
-  constructor() {
-    // @ts-expect-error ts-migrate(2554) FIXME: Expected 1-2 arguments, but got 0.
-    super()
+  constructor(props: TextAreaProps) {
+    super(props)
 
-    // @ts-expect-error ts-migrate(2339) FIXME: Property '_defaultId' does not exist on type 'Text... Remove this comment to see the full error message
     this._defaultId = uid('TextArea')
   }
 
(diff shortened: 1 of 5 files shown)


packages/ui-text-area/src/TextArea/index.tsx
```diff
@@ -98,116 +103,86 @@ class TextArea extends Component<TextAreaProps> {
...
-    // @ts-expect-error ts-migrate(2339) FIXME: Property '_textarea' does not exist on type 'TextA... Remove this comment to see the full error message
     this._textarea.style.height = 'auto'
-    // @ts-expect-error ts-migrate(2339) FIXME: Property '_textarea' does not exist on type 'TextA... Remove this comment to see the full error message
     this._textarea.style.overflowY = 'hidden' // hide scrollbars for autoGrow textareas
-    // @ts-expect-error ts-migrate(2339) FIXME: Property '_textarea' does not exist on type 'TextA... Remove this comment to see the full error message
     height = this._textarea.scrollHeight + offset + 'px'
 
-    // @ts-expect-error ts-migrate(2339) FIXME: Property '_container' does not exist on type 'Text... Remove this comment to see the full error message
-    const maxHeight = px(this.props.maxHeight, this._container)
+    const maxHeight = px(this.props.maxHeight!, this._container)
```
ke1k0, line 176: `maxHeight` has no default value, so the `!` is not exactly true here. So I would refactor the code here a bit to make sense when it is undefined.

packages/ui-text-area/src/TextArea/index.tsx
```diff
@@ -251,15 +221,12 @@ class TextArea extends Component<TextAreaProps> {
     }
   }
 
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'node' implicitly has an 'any' type.
-  handleContainerRef = (node) => {
-    // @ts-expect-error ts-migrate(2339) FIXME: Property '_container' does not exist on type 'Text... Remove this comment to see the full error message
+  handleContainerRef = (node: HTMLDivElement) => {
```
ke1k0, line 224: Since this is coming from e `ref=""`, the node can be `HTMLDivElement | null`.

packages/ui-text-area/src/TextArea/index.tsx
```diff
@@ -318,31 +282,29 @@ class TextArea extends Component<TextAreaProps> {
...
+            textareaRef.apply(this, [textarea, ...args])
+          }
         }}
-        // @ts-expect-error ts-migrate(2322) FIXME: Type '{ width: string | undefined; resize: "none" ... Remove this comment to see the full error message
         style={style}
         id={this.id}
         required={required}
         aria-required={required}
-        aria-invalid={this.invalid ? 'true' : null}
+        aria-invalid={this.invalid ? 'true' : 'false'}
```
ke1k0, line 294: Hm, I would rather set this to `undefined`, because it might be an a11y change if we explicitly set it to `false`.

packages/ui-text-area/src/TextArea/index.tsx
```diff
@@ -318,31 +282,29 @@ class TextArea extends Component<TextAreaProps> {
         defaultValue={defaultValue}
         placeholder={placeholder}
         ref={(textarea, ...args) => {
-          // @ts-expect-error ts-migrate(2339) FIXME: Property '_textarea' does not exist on type 'TextA... Remove this comment to see the full error message
           this._textarea = textarea
-          // @ts-expect-error ts-migrate(2532) FIXME: Object is possibly 'undefined'.
-          textareaRef.apply(this, [textarea].concat(args))
+          if (typeof textareaRef === 'function') {
+            textareaRef.apply(this, [textarea, ...args])
```
ke1k0, line 287: I have no clue why they wrote it like this XD I really doubt that the `ref=""` would pass any more args, so there is nothing else to pass to the textareaRef other than the node. I would refactor and simplify it, but ask someone for second opinion on this :)

packages/ui-text-area/src/TextArea/index.tsx
```diff
@@ -318,31 +282,29 @@ class TextArea extends Component<TextAreaProps> {
...
         css={this.props.styles?.textArea}
         onChange={this.handleChange}
       />
     )
 
     return (
-      // @ts-expect-error ts-migrate(2769) FIXME: No overload matches this call.
       <FormField
+        label={null}
         {...pickProps(this.props, FormField.allowedProps)}
```
ke1k0, line 304: I bumped into this before too. TS doesn't know that pickProps will give back the `label` props, so I just put it there explicitly:

```js
<FormField
  {...pickProps(this.props, FormField.allowedProps)}
  label={this.props.label}
```

packages/ui-text-area/src/TextArea/props.ts
```diff
@@ -39,110 +39,110 @@ import { TextareaHTMLAttributes } from 'react'
...
    * when autoGrow is true, the textarea will never grow beyond this value
    */
-  maxHeight: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+  maxHeight?: number | string
   /**
    * object with shape: `{
    * text: PropTypes.node,
    * type: PropTypes.oneOf(['error', 'hint', 'success', 'screenreader-only'])
    *   }`
    */
```
ke1k0, line 74: Matyi said in one of my PR-s that this would be more readable as:

```suggestion
  /**
   * Array of objects with shape: `{
   *   text: React.ReactNode,
   *   type: One of: ['error', 'hint', 'success', 'screenreader-only']
   * }`
   */
```

packages/ui-text-area/src/TextArea/props.ts
```diff
@@ -39,110 +39,110 @@ import { TextareaHTMLAttributes } from 'react'
...
   /**
    * Sets the required property on the underlying native textArea
    */
-  required: PropTypes.bool,
+  required?: boolean
   /**
    * a function that provides a reference to the actual textarea element
    */
-  textareaRef: PropTypes.func,
+  textareaRef?: (textarea: HTMLTextAreaElement | null, ...args: any[]) => any
```
ke1k0, line 97: The return here is `void`, and also the extra args might not be needed if you refactor the code in index.

packages/ui-text-area/src/TextArea/props.ts
```diff
@@ -39,110 +39,110 @@ import { TextareaHTMLAttributes } from 'react'
...
+  defaultValue?: string
   /**
    * the selected value (must be accompanied by an `onChange` prop)
    */
-  value: controllable(PropTypes.string),
+  value?: string // TODO: controllable(PropTypes.string)
   /**
    * when used with the `value` prop, the component will not control its own state
    */
+  onChange?: (event: React.ChangeEvent<HTMLTextAreaElement>) => any
```
ke1k0, line 109:
```suggestion
  onChange?: (event: React.ChangeEvent<HTMLTextAreaElement>) => void
```

packages/ui-text-area/src/TextArea/index.tsx
HerrTopi: I did not refactor it by choice, so it can be as minimal chance to change the behaviour as possible. Weird for sure :D

packages/ui-text-area/src/TextArea/index.tsx
HerrTopi: Nice idea, did not think of that

packages/ui-text-area/src/TextArea/index.tsx
HerrTopi: It should not be a problem, but better safe than sorry

packages/ui-text-area/src/TextArea/index.tsx
```diff
@@ -98,142 +104,113 @@ class TextArea extends Component<TextAreaProps> {
...
-    // @ts-expect-error ts-migrate(2339) FIXME: Property '_textarea' does not exist on type 'TextA... Remove this comment to see the full error message
     this._textarea.style.height = 'auto'
-    // @ts-expect-error ts-migrate(2339) FIXME: Property '_textarea' does not exist on type 'TextA... Remove this comment to see the full error message
     this._textarea.style.overflowY = 'hidden' // hide scrollbars for autoGrow textareas
-    // @ts-expect-error ts-migrate(2339) FIXME: Property '_textarea' does not exist on type 'TextA... Remove this comment to see the full error message
     height = this._textarea.scrollHeight + offset + 'px'
 
-    // @ts-expect-error ts-migrate(2339) FIXME: Property '_container' does not exist on type 'Text... Remove this comment to see the full error message
-    const maxHeight = px(this.props.maxHeight, this._container)
+    const maxHeight = this.props.maxHeight
```
ke1k0, line 177: Looks good, but how about just wrapping this block in an `if`? Seems a bit more readable, and it shows the block is only relevant if the user provided a `maxHeight` prop :)

```js
  if (this.props.maxHeight) {
      const maxHeight = px(this.props.maxHeight, this._container)

      if (this._textarea.scrollHeight > maxHeight) {
        this._textarea.style.overflowY = 'auto' // add scroll if scrollHeight exceeds maxHeight in pixels
      } else if (this.props.height) {
        if (this._textarea.value === '') {
          height = this.props.height
        } else if (
          px(this.props.height, this._container) > this._textarea.scrollHeight
        ) {
          this._textarea.style.overflowY = 'auto' // add scroll if scrollHeight exceeds height in pixels
          height = this.props.height
        }
      }

      // preserve container height to prevent scroll jumping on long textareas,
      // but make sure container doesn't exceed maxHeight prop
      const heightExceedsMax = px(height) > maxHeight
      if (!heightExceedsMax) {
        this._container!.style.minHeight = height
      }
    }
```

packages/ui-text-area/src/TextArea/index.tsx
```diff
@@ -65,20 +67,25 @@ class TextArea extends Component<TextAreaProps> {
...
+  private _listener?: { remove(): void }
+  private _request?: RequestAnimationFrameType
+  private _defaultId: string
+  private _textareaResizeListener?: ResizeObserver
+  private _debounced?: Debounced
+  private _textarea: HTMLTextAreaElement | null = null
+  private _container: HTMLDivElement | null = null
+  private _height?: string
+  //TODO remove in v9 since it seems to be unused
+  private _node?: FormField | null = null
```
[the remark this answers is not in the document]
HerrTopi: We talked about it with @matyasf  and we decided to remove it only in v9, maybe someone uses it somewhere

packages/ui-text-area/src/TextArea/index.tsx
HerrTopi: When this.props.maxHeight `falsy`, maxHeight is also `falsy`, do heightExceedsMax is `falsy`, so the
` if (!heightExceedsMax) {
        this._container!.style.minHeight = height
      }`
should run, which will not if we put the `if` there

packages/ui-text-area/src/TextArea/index.tsx
HerrTopi: Well alright then (:

packages/ui-text-area/src/TextArea/index.tsx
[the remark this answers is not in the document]
ke1k0: Ok, but if we want to deprecate it in v9, then let's remove the `private` and make a getter for it (like Viktor did for the old refs) and give a deprecation warning if used.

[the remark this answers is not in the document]
ke1k0: It still throws an error, and can you please deprecate or delete it?

ke1k0 (approved):

PR #824, opened 2021-11-29
chore: remove bundle analysis from comment, put it inside job console

Brailor:
This PR fixes a couple of things:
- comments by github bot being overridden by the `maintain-one` github action, now instead the `pr-validation` workflow uses an another github action to post bot comments which enables multiple bot comments
- bundle size analysis will not be included in the comment, instead it will be logged out inside the `Calculate bundle size` step
- added threshold variable


matyasf (approved):
yay no spam

ke1k0 (approved):
I wanted to ask removing the long log myself, thanks for fixing it (and the overriding bot comments)!

HerrTopi (approved):

PR #827, opened 2021-11-30
chore(ui-form-field): add TS types to `ui-form-field` package

ke1k0:
Closes: INSTUI-3052

### DIFF
--- a/packages/__docs__/src/Document/index.tsx
+++ b/packages/__docs__/src/Document/index.tsx
@@ -122,10 +122,12 @@ class Document extends Component<
   renderDescription(doc: SingleChildrenType, description: string) {
     const { id, title } = doc
 
+    const filteredDescription = description.replace('@tsProps', '')
+
     return this.props.description ? (
       <Description
         id={`${id}Description`}
-        content={description}
+        content={filteredDescription}
         title={title}
       />
     ) : null
--- a/packages/ui-form-field/src/FormField/index.tsx
+++ b/packages/ui-form-field/src/FormField/index.tsx
@@ -35,6 +35,7 @@ import type { FormFieldProps } from './props'
 ---
 category: components
 ---
+@tsProps
 **/
 class FormField extends Component<FormFieldProps> {
   static readonly componentId = 'FormField'
@@ -45,8 +46,7 @@ class FormField extends Component<FormFieldProps> {
     inline: false,
     layout: 'stacked',
     labelAlign: 'end',
-    vAlign: 'middle',
-    children: null
+    vAlign: 'middle'
   }
 
   ref: Element | null = null
@@ -66,7 +66,7 @@ class FormField extends Component<FormFieldProps> {
       <FormFieldLayout
         {...omitProps(this.props, FormField.allowedProps)}
         {...pickProps(this.props, FormFieldLayout.allowedProps)}
-        // @ts-expect-error ts-migrate(2322) FIXME: Remove this comment to see the full error message
+        label={this.props.label}
         vAlign={this.props.vAlign}
         as="label"
         htmlFor={this.props.id}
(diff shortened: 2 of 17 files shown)


packages/__docs__/src/Document/index.tsx
```diff
@@ -122,10 +122,12 @@ class Document extends Component<
   renderDescription(doc: SingleChildrenType, description: string) {
     const { id, title } = doc
 
+    const filteredDescription = description.replace('@tsProps', '')
```
ke1k0, line 125: This was a "fix" for docs, because the "child" components had @tsProps written out in their descriptions.

packages/ui-form-field/src/FormField/index.tsx
```diff
@@ -66,7 +66,7 @@ class FormField extends Component<FormFieldProps> {
       <FormFieldLayout
         {...omitProps(this.props, FormField.allowedProps)}
         {...pickProps(this.props, FormFieldLayout.allowedProps)}
-        // @ts-expect-error ts-migrate(2322) FIXME: Remove this comment to see the full error message
+        label={this.props.label}
```
ke1k0, line 69: It was passed from the pickProps, but had to add it manually because TS was complaining (it cannot see what comes back from pickProps)

packages/ui-form-field/src/FormFieldLayout/index.tsx
```diff
@@ -121,9 +118,8 @@ class FormFieldLayout extends Component<FormFieldLayoutProps> {
     return this.props.inline && this.props.layout === 'inline'
   }
 
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'node' implicitly has an 'any' type.
-  handleInputContainerRef = (node) => {
-    if (this.props.inputContainerRef) {
+  handleInputContainerRef = (node: Element | null) => {
+    if (typeof this.props.inputContainerRef === 'function') {
```
ke1k0, line 122: Here I just strengthened the `if`.

packages/ui-form-field/src/FormFieldLayout/props.ts
```diff
@@ -36,16 +36,35 @@ import type { FormMessage } from '../FormPropTypes'
...
   messages?: FormMessage[]
+  /**
+   * id for the form field messages
+   */
   messagesId?: string
   children?: React.ReactNode
   inline?: boolean
   layout?: 'stacked' | 'inline'
   labelAlign?: 'start' | 'end'
+  vAlign?: 'top' | 'middle' | 'bottom'
```
ke1k0, line 62: As stated below in the allowedProps comment, I decided to add this prop to the component because it can be passed (and forwarded to Grid), our components pass it too. Also, there is a small chance someone would use this component by itself.

HerrTopi (approved):
Seems good to me, nice work (:

packages/ui-form-field/src/FormFieldLayout/index.tsx
```diff
@@ -194,8 +189,7 @@ class FormFieldLayout extends Component<FormFieldLayoutProps> {
         ])}
         css={styles?.formFieldLayout}
         style={{ width }}
-        // @ts-expect-error ts-migrate(2339) FIXME: Property '_messagesId' does not exist on type 'For... Remove this comment to see the full error message
-        aria-describedby={this.hasMessages ? this._messagesId : null}
+        aria-describedby={this.hasMessages ? this._messagesId : undefined}
         ref={this.handleRef}
```
matyasf, line 193: strangely here the `ref` causes the  `TS2590: Expression produces a union type that is too complex to represent.` error, but just in the IDE. Likely we cannot fix it, its some TS language server issue.

matyasf (approved):
looks good!

packages/ui-form-field/src/FormFieldLayout/index.tsx
ke1k0: Yes, we encountered this issue before with these <ElementType> components. Hopefully some TS update will solve it later or maybe we can type these better later.

PR #834, opened 2021-12-02
chore(ui-menu): type ui-menu

HerrTopi:
Closes: INSTUI-3059

### DIFF
--- a/packages/ui-menu/src/Menu/MenuItem/index.tsx
+++ b/packages/ui-menu/src/Menu/MenuItem/index.tsx
@@ -39,42 +39,38 @@ import generateStyle from './styles'
 import generateComponentTheme from './theme'
 
 import { propTypes, allowedProps } from './props'
-import type { MenuItemProps } from './props'
+import type { MenuItemProps, MenuItemState } from './props'
 
 /**
 ---
 parent: Menu
 id: Menu.Item
 ---
+@tsProps
 **/
 @withStyle(generateStyle, generateComponentTheme)
 @testable()
-class MenuItem extends Component<MenuItemProps> {
+class MenuItem extends Component<MenuItemProps, MenuItemState> {
   static readonly componentId = 'Menu.Item'
 
   static propTypes = propTypes
   static allowedProps = allowedProps
   static defaultProps = {
     type: 'button',
-    disabled: false,
-    // @ts-expect-error ts-migrate(6133) FIXME: 'e' is declared but its value is never read.
-    onSelect: function (e, value, selected, item) {}
+    disabled: false
   } as const
 
   static contextType = MenuContext
 
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'props' implicitly has an 'any' type.
-  constructor(props) {
-    // @ts-expect-error ts-migrate(2554) FIXME: Expected 1-2 arguments, but got 0.
-    super()
+  constructor(props: MenuItemProps) {
+    super(props)
 
     if (typeof props.selected === 'undefined') {
       this.state = {
-        selected: props.defaultSelected
+        selected: !!props.defaultSelected
       }
     }
 
-    // @ts-expect-error ts-migrate(2339) FIXME: Property 'labelId' does not exist on type 'MenuIte... Remove this comment to see the full error message
     this.labelId = uid('MenuItem__label')
   }
 
@@ -85,14 +81,16 @@ class MenuItem extends Component<MenuItemProps> {
 
     return this.ref
   }
+
+  labelId: string
   ref: Element | null = null
+
   handleRef = (el: Element | null) => {
     this.ref = el
   }
 
   componentDidMount() {
-    // @ts-expect-error ts-migrate(2722) FIXME: Cannot invoke an object which is possibly 'undefin... Remove this comment to see the full error message
-    this.props.makeStyles()
+    this.props.makeStyles?.()
     const context = this.context
 
     if (context && context.registerMenuItem) {
@@ -101,8 +99,7 @@ class MenuItem extends Component<MenuItemProps> {
   }
 
   componentDidUpdate() {
-    // @ts-expect-error ts-migrate(2722) FIXME: Cannot invoke an object which is possibly 'undefin... Remove this comment to see the full error message
-    this.props.makeStyles()
+    this.props.makeStyles?.()
   }
 
   componentWillUnmount() {
@@ -113,8 +110,7 @@ class MenuItem extends Component<MenuItemProps> {
     }
   }
 
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'e' implicitly has an 'any' type.
-  handleClick = (e) => {
+  handleClick = (e: React.MouseEvent) => {
     const { onSelect, onClick, disabled, value } = this.props
     const selected = !this.selected
 
@@ -137,8 +133,7 @@ class MenuItem extends Component<MenuItemProps> {
     }
   }
 
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'e' implicitly has an 'any' type.
-  handleKeyDown = (e) => {
+  handleKeyDown = (e: React.KeyboardEvent) => {
     const spaceKey = e.keyCode === keycode.codes.space
     const enterKey = e.keyCode === keycode.codes.enter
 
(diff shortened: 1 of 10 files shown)


Brailor (approved):

packages/ui-menu/src/Menu/props.ts
```diff
@@ -29,175 +29,169 @@ import {
...
    * Is the `<Menu />` open (should be accompanied by `onToggle`)
    */
-  show: controllable(PropTypes.bool, 'onToggle', 'defaultShow'),
+  show?: boolean // TODO: controllable(PropTypes.bool, 'onToggle', 'defaultShow')
   /**
    * Callback fired when the `<Menu />` is toggled open/closed. When used with `show`,
    * the component will not control its own state.
    */
-  onToggle: PropTypes.func,
+  onToggle?: (show: boolean, subMenu: Menu) => void
```
ke1k0, line 82: Is it the `subMenu`? Not the main Menu itself?

packages/ui-menu/src/Menu/index.tsx
```diff
@@ -102,51 +82,45 @@ class Menu extends Component<MenuProps> {
 
   state = { hasFocus: false }
   _rootNode = null
-  _menuItems = []
-  _popover = null
+  _menuItems: MenuItem[] = []
+  _popover: Popover | null = null
   _trigger = null
```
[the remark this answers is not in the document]
ke1k0: https://github.com/instructure/instructure-ui/commit/ee57f812ebc89866b0c314666d93a6cb26e8e3a0#

Yep it was deleted here, probably a mistake

packages/ui-menu/src/Menu/MenuItem/index.tsx
```diff
@@ -85,14 +84,17 @@ class MenuItem extends Component<MenuItemProps> {
 
     return this.ref
   }
+
+  labelId: string
   ref: Element | null = null
-  handleRef = (el: Element | null) => {
+
+  //TODO cannot-type
+  handleRef = (el: any) => {
```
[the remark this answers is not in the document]
HerrTopi: It was underlining it for me. Probably my built in ts interpreter failed to do its job.

packages/ui-menu/src/Menu/index.tsx
```diff
@@ -231,87 +199,79 @@ class Menu extends Component<MenuProps> {
...
   }
 
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'event' implicitly has an 'any' type.
-  show = (event) => {
+  show = (event?: React.MouseEvent | React.KeyboardEvent) => {
     if (this._popover) {
-      // @ts-expect-error ts-migrate(2531) FIXME: Object is possibly 'null'.
-      this._popover.show(event)
+      //TODO make events consistents
+      this._popover.show(event as unknown as Event)
```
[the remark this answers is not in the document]
HerrTopi: INSTUI-3346

packages/ui-menu/src/Menu/MenuItemGroup/props.ts
```diff
@@ -42,15 +42,36 @@ import type { WithStyleProps, ComponentStyle } from '@instructure/emotion'
...
+  /**
+   * an array of the values (or indices by default) for the selected items on initial render
+   */
+  defaultSelected?: (string | number)[]
+  /**
+   * call this function when a menu item is selected
+   */
   onSelect?: (...args: any[]) => any
   onMouseOver?: (...args: any[]) => any
   onKeyDown?: (...args: any[]) => any
```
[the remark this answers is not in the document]
HerrTopi: Well, I knew I forgot something. :D

packages/ui-menu/src/Menu/MenuItemGroup/props.ts
```diff
@@ -38,19 +38,46 @@ import type {
...
+  /**
+   * call this function when a menu item is selected
+   */
+  onSelect?: (
+    e: React.MouseEvent,
+    updated: MenuItemProps['value'][],
+    selected: MenuItemProps['selected'],
+    item: MenuItem
+  ) => void
+  onMouseOver?: MenuItemProps['onMouseOver']
```
ke1k0, line 68: I can see that `onKeyDown` was removed because it was unused, but it is an API change if we simply remove it, so I'd recommend deprecating it.

ke1k0: Also, for now I think we should copy the types in cases like this, because the doc parser cannot resolve imported types and in this case it renders: `onMouseOver: intersection['onMouseOver']` in the Properties block. That sounds a bit lame, so I'd just copy  `(e: React.MouseEvent, args: MenuItem) => void` for now.

packages/ui-menu/src/Menu/index.tsx
```diff
@@ -231,87 +199,83 @@ class Menu extends Component<MenuProps> {
     this.setState({ hasFocus: this.focusedIndex >= 0 })
   }
 
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'event' implicitly has an 'any' type.
-  handleMenuItemMouseOver = (event, menuItem) => {
-    // @ts-expect-error ts-migrate(2531) FIXME: Object is possibly 'null'.
-    if (this._activeSubMenu && menuItem !== this._activeSubMenu._trigger) {
+  handleMenuItemMouseOver: MenuItemProps['onMouseOver'] = (event, menuItem) => {
+    if (this._activeSubMenu) {
```
ke1k0, line 203: The `_trigger` part is missing from the if.

packages/ui-menu/src/Menu/index.tsx
```diff
@@ -231,87 +199,83 @@ class Menu extends Component<MenuProps> {
...
+        this._menu !== null && typeof this._menu.focus === 'function'
+
+      error(hasFocusOnMenu, '[Menu] Could not focus the menu.')
+
+      if (hasFocusOnMenu) {
+        this._menu!.focus()
+      }
+      if (this._trigger !== null && typeof this._trigger.focus === 'function') {
+        this._trigger.focus()
+      }
```
ke1k0, line 258: Imho this if-else block got a bit overcomplicated in the back-and-forth refactoring, and it is hard to tell wether if it is still equivalent to the original. This method basically tries to focus the menu, and if it cannot, throws an error and tries to focus the trigger instead, and a warning if that is not possible either. I suggest writing it like this, seems clean and simple and is equal to the original:

```js
  focus() {
    if (this.shown) {
      error(!!this._menu?.focus, '[Menu] Could not focus the menu.')
      this._menu!.focus()
    } else {
      error(!!this._trigger?.focus, '[Menu] Could not focus the trigger.')
      this._trigger!.focus!()
    }
  }
```

packages/ui-menu/src/Menu/index.tsx
```diff
@@ -231,87 +199,83 @@ class Menu extends Component<MenuProps> {
...
+      error(false, '[Menu] Could not focus the trigger.')
     }
   }
 
   focused() {
     if (this.shown) {
       return containsActiveElement(this._menu) || this.state.hasFocus
     } else {
-      return containsActiveElement(this._trigger)
+      return false
```
ke1k0, line 268: The original check is missing.

```suggestion
      return containsActiveElement(this._trigger)
```

ke1k0 (changes requested):

packages/ui-menu/src/Menu/index.tsx
HerrTopi: The other checks are more secure, but probably won't be an issue either way. 

I'll replace the block with this, just to be more readable

packages/ui-menu/src/Menu/index.tsx
HerrTopi: This condition will always return 'true' since the types 'MenuItem' and '(ReactElement<any, string | JSXElementConstructor<any>> & { focus?: (() => void) | undefined; }) | null' have no overlap.

If the check is correct (and I think it is), than we don't need that part

packages/ui-menu/src/Menu/MenuItemGroup/props.ts
HerrTopi: I already did this with onBlur somewhere else. I don't think we can consider this an API change really

ke1k0 (approved):

PR #837, opened 2021-12-03
chore(ui-billboard): add TS types to `ui-billboard` package

NamNguyenInstructure:
Closes: INSTUI-3034

### DIFF
--- a/packages/ui-billboard/babel.config.js
+++ b/packages/ui-billboard/babel.config.js
@@ -23,11 +23,6 @@
  */
 
 module.exports = {
-  shouldPrintComment(value) {
-    const isTsRelatedComment = /@ts-expect-error|FIXME/.test(value)
-
-    return !isTsRelatedComment
-  },
   presets: [
     [
       require('@instructure/ui-babel-preset'),
--- a/packages/ui-billboard/src/Billboard/__tests__/Billboard.test.tsx
+++ b/packages/ui-billboard/src/Billboard/__tests__/Billboard.test.tsx
@@ -30,10 +30,8 @@ import {
   within,
   generateA11yTests
 } from '@instructure/ui-test-utils'
-import type { StoryConfig } from '@instructure/ui-test-utils'
 import { Billboard } from '../index'
 import BillboardExamples from '../__examples__/Billboard.examples'
-import { BillboardProps } from '../props'
 
 describe('<Billboard />', async () => {
   it('should render', async () => {
--- a/packages/ui-billboard/src/Billboard/__tests__/Billboard.test.tsx
+++ b/packages/ui-billboard/src/Billboard/__tests__/Billboard.test.tsx
@@ -43,10 +43,7 @@ describe('<Billboard />', async () => {
   })
 
   describe('with generated examples', async () => {
-    generateA11yTests(
-      Billboard,
-      BillboardExamples as StoryConfig<BillboardProps>
-    )
+    generateA11yTests(Billboard, BillboardExamples)
   })
 
   it('should render a heading with the correct tag', async () => {
--- a/packages/ui-billboard/src/Billboard/__tests__/Billboard.test.tsx
+++ b/packages/ui-billboard/src/Billboard/__tests__/Billboard.test.tsx
@@ -30,9 +30,10 @@ import {
   within,
   generateA11yTests
 } from '@instructure/ui-test-utils'
-
+import type { StoryConfig } from '@instructure/ui-test-utils'
 import { Billboard } from '../index'
 import BillboardExamples from '../__examples__/Billboard.examples'
+import { BillboardProps } from '../props'
 
 describe('<Billboard />', async () => {
   it('should render', async () => {
@@ -42,7 +43,10 @@ describe('<Billboard />', async () => {
   })
 
   describe('with generated examples', async () => {
-    generateA11yTests(Billboard, BillboardExamples)
+    generateA11yTests(
+      Billboard,
+      BillboardExamples as StoryConfig<BillboardProps>
+    )
   })
 
   it('should render a heading with the correct tag', async () => {
--- a/packages/ui-billboard/src/Billboard/index.tsx
+++ b/packages/ui-billboard/src/Billboard/index.tsx
@@ -23,7 +23,7 @@
  */
 
 /** @jsx jsx */
-import { Component } from 'react'
+import { Component, MouseEvent } from 'react'
 
 import { Heading } from '@instructure/ui-heading'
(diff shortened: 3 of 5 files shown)


packages/ui-billboard/src/Billboard/index.tsx
```diff
@@ -115,8 +113,7 @@ class Billboard extends Component<BillboardProps> {
   }
 
   renderHero() {
-    if (this.heroIsFunction) {
-      // @ts-expect-error ts-migrate(2722) FIXME: Cannot invoke an object which is possibly 'undefin... Remove this comment to see the full error message
+    if (typeof this.props.hero === 'function') {
```
ke1k0, line 116: The original `heroIsFunction` getter can be deleted since it is not used anymore.

packages/ui-billboard/src/Billboard/props.ts
```diff
@@ -37,93 +37,94 @@ import type {
...
-type BillboardStyle = ComponentStyle<
-  'billboard' | 'content' | 'hero' | 'heading' | 'message'
->
-
-const propTypes: PropValidators<PropKeys> = {
   /**
    * Provide an <Img> component or Instructure Icon for the hero image
    */
-  hero: PropTypes.oneOfType([PropTypes.element, PropTypes.func]),
+  hero?: React.ReactElement | ((size: 'medium' | 'x-large' | 'large') => string)
```
ke1k0, line 46: The function can return `ReactElement` too. Also, I would make a type here: `type HeroIconSize = 'medium' | 'x-large' | 'large'` and use it here and and in the index too: 

```js
hero?: React.ReactElement | ((iconSize: HeroIconSize) => React.ReactElement)
```

```js
get SVGIconSize(): HeroIconSize {
    const size = this.props.size

    // serve up appropriate SVGIcon size for each Billboard size
    if (size === 'small') {
      return 'medium'
    } else if (size === 'large') {
      return 'x-large'
    } else {
      return 'large'
    }
  }
```

packages/ui-billboard/src/Billboard/props.ts
```diff
@@ -37,93 +37,94 @@ import type {
...
-  headingLevel: PropTypes.oneOf(['h1', 'h2', 'h3', 'h4']),
+  headingLevel?: 'h1' | 'h2' | 'h3' | 'h4'
   /**
    * Instructions or information for the Billboard. Note: you should not pass
    * interactive content to this prop if you are also providing an `href` or
    * `onClick`. That would cause the Billboard to render as a button or link
    * and would result in nested interactive content.
    */
-  message: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
+  message?: React.ReactNode
```
ke1k0, line 78: It allows a function as well, with no parameters. The `callRenderProp` will handle it in the `index.tsx` :)

```suggestion
  message?: React.ReactNode | (() => React.ReactNode)
```

ke1k0 (changes requested):

packages/ui-billboard/babel.config.js
```diff
@@ -23,11 +23,6 @@
  */
 
 module.exports = {
-  shouldPrintComment(value) {
-    const isTsRelatedComment = /@ts-expect-error|FIXME/.test(value)
-
-    return !isTsRelatedComment
-  },
```
ke1k0, line 30: There are a few ts-expect-errors left in the `Billboard.examples.tsx` :) Please resolve those as well! When typing, we have to check the test and examples files as well.

packages/ui-billboard/src/Billboard/index.tsx
```diff
@@ -23,7 +23,7 @@
  */
 
 /** @jsx jsx */
-import { Component } from 'react'
+import { Component, MouseEvent } from 'react'
```
ke1k0, line 26: On line 49, add `@tsProps` to the annotation. We have to add it to each component we fully typed so the doc generation knows that it can render the props from the TS types safely instead of the old PropTypes :)

```js
/**
---
category: components
---
@tsProps
**/
```

packages/ui-billboard/babel.config.js
ke1k0: Please resolve the `ts-expect-error` comments in the `Billboard.examples.tsx` too!

packages/ui-billboard/src/Billboard/__tests__/Billboard.test.tsx
```diff
@@ -42,7 +43,10 @@ describe('<Billboard />', async () => {
   })
 
   describe('with generated examples', async () => {
-    generateA11yTests(Billboard, BillboardExamples)
+    generateA11yTests(
+      Billboard,
+      BillboardExamples as StoryConfig<BillboardProps>
```
ke1k0, line 48: You can put the `as StoryConfig<BillboardProps>` cast at the end of the export at `Billboard.examples.tsx` too. That way it is already exported as a StoryConfig and you don't have to use the cast here :)

ke1k0 (approved):

packages/ui-billboard/src/Billboard/__tests__/Billboard.test.tsx
NamNguyenInstructure: Great idea! I will change it.

packages/ui-billboard/src/Billboard/__tests__/Billboard.test.tsx
```diff
@@ -43,10 +43,7 @@ describe('<Billboard />', async () => {
   })
 
   describe('with generated examples', async () => {
-    generateA11yTests(
-      Billboard,
-      BillboardExamples as StoryConfig<BillboardProps>
-    )
+    generateA11yTests(Billboard, BillboardExamples)
```
ke1k0: Just remove the unused imports here (BillboardProps and StoryConfig) and it should be fine :) Good job!

ke1k0 (approved):

packages/ui-billboard/src/Billboard/__tests__/Billboard.test.tsx
```diff
@@ -30,9 +30,10 @@ import {
   within,
   generateA11yTests
 } from '@instructure/ui-test-utils'
-
+import type { StoryConfig } from '@instructure/ui-test-utils'
```
matyasf, line 33: StoryConfig and BillboradProps are unused imports, please delete them

matyasf (changes requested):
Besides the small change to remove some imports it looks good :)

matyasf (approved):

PR #838, opened 2021-12-03
chore(ui-select): add TS types to `ui-select` package

ke1k0:
Closes: INSTUI-3078

### DIFF
--- a/packages/ui-select/src/Select/index.tsx
+++ b/packages/ui-select/src/Select/index.tsx
@@ -23,10 +23,11 @@
  */
 
 /** @jsx jsx */
-import { Children, Component } from 'react'
+import React, { Children, Component } from 'react'
 
 import { createChainedFunction } from '@instructure/ui-utils'
 import { testable } from '@instructure/ui-testable'
+import { uid } from '@instructure/uid'
 import {
   matchComponentTypes,
   omitProps,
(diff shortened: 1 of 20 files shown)


packages/ui-select/src/Select/index.tsx
```diff
@@ -126,19 +119,20 @@ class Select extends Component<SelectProps> {
     hasInputRef: false
   }
 
-  _defaultId = uid('Select')
-  _assistId = uid('Select-assistiveText')
```
ke1k0, line 130: this wasn't used anywhere

packages/ui-select/src/Select/index.tsx
```diff
@@ -458,53 +442,50 @@ class Select extends Component<SelectProps> {
...
           {...getListProps({ as: 'ul', elementRef: this.handleListRef })}
         >
           {isShowingOptions
-            ? // @ts-expect-error ts-migrate(7030) FIXME: Not all code paths return a value.
-              Children.map(children, (child, index) => {
+            ? Children.map(children, (child, index) => {
                 if (!child || !matchComponentTypes(child, [Group, Option])) {
                   return // ignore invalid children
                 }
                 if (matchComponentTypes(child, [Option])) {
```
matyasf, line 478: We should modernize this code in the future, there is lots of boilerplate going on just for these comparisons (if I understand correctly, this code is the sole reason for Select.Group and Select.Option's existence)

matyasf (approved):
looks good!

matyasf:
There are lots of change is Chromatic, do you know why?

Brailor (approved):

ke1k0:
> There are lots of change is Chromatic, do you know why?

@matyasf I checked and the number of examples are the same on master and here, but since I shuffled the props around a lot, I assume only the order of the examples changed.

packages/ui-select/src/Select/index.tsx
ke1k0: It is indeed pretty weird, but we already started the talks about writing a new Select component and retire this one...

PR #840, opened 2021-12-06
INSTUI-3035 - TS typing `ui-breadcrumb`

NamNguyenInstructure:


packages/ui-breadcrumb/src/Breadcrumb/index.tsx
```diff
@@ -63,22 +63,17 @@ class Breadcrumb extends Component<BreadcrumbProps> {
...
+    this.props.makeStyles?.()
   }
 
   static Link = BreadcrumbLink
 
   renderChildren() {
     const { styles, children } = this.props
-    // @ts-expect-error ts-migrate(2339) FIXME: Property 'length' does not exist on type 'string |... Remove this comment to see the full error message
-    const numChildren = children ? children.length : 0
+    const numChildren = children ? React.Children.count(children) : 0
```
ke1k0, line 76: Nice solution! The ternary is not needed here anymore, because `React.Children.count(children)` will give back `0` if there are no children anyway :)

`const numChildren = React.Children.count(children)`

packages/ui-breadcrumb/babel.config.js
```diff
@@ -23,11 +23,6 @@
  */
 
 module.exports = {
-  shouldPrintComment(value) {
-    const isTsRelatedComment = /@ts-expect-error|FIXME/.test(value)
-
-    return !isTsRelatedComment
-  },
```
ke1k0, line 30: Please resolve the ts-expect-errors in the test and example files as well :)

ke1k0 (changes requested):

packages/ui-breadcrumb/src/Breadcrumb/BreadcrumbLink/index.tsx
```diff
@@ -60,8 +60,6 @@ class BreadcrumbLink extends Component<BreadcrumbLinkProps> {
 
     return (
       <Link
-        // @ts-expect-error ts-migrate(2339) FIXME: Property 'element' does not exist on type 'Breadcr... Remove this comment to see the full error message
```
ke1k0, line 63: On line 40, add `@tsProps` to the annotation. We have to add it to each component we fully typed so the doc generation knows that it can render the props from the TS types safely instead of the old PropTypes :)

/**
---
parent: Breadcrumb
id: Breadcrumb.Link
---
@tsProps
**/

packages/ui-breadcrumb/src/Breadcrumb/index.tsx
```diff
@@ -63,22 +63,17 @@ class Breadcrumb extends Component<BreadcrumbProps> {
   }
 
   componentDidMount() {
-    // @ts-expect-error ts-migrate(2722) FIXME: Cannot invoke an object which is possibly 'undefin... Remove this comment to see the full error message
-    this.props.makeStyles()
+    this.props.makeStyles?.()
```
ke1k0, line 66: On line 45, add the `@tsProps` flag!

ke1k0 (changes requested):
I found another typing issue: in the `BreadcrumbLink.test.tsx`, on line 53 there is still a `ts-expect-error`. It is because we silently allow a "to" parameter for BreadcrumbLink, just like for the Link component. These are usually used for [React Router](https://v5.reactrouter.com/web/guides/quick-start). 

To fix this, add modify the type like this (just like in `LinkProps`):
```js
type BreadcrumbLinkProps = BreadcrumbLinkOwnProps &
  OtherHTMLAttributes<BreadcrumbLinkOwnProps> & {
    // React Router might add a `to` prop
    to?: string
  }
``` 

After this you can remove that error from the test.

packages/ui-breadcrumb/src/Breadcrumb/BreadcrumbLink/props.ts
```diff
@@ -29,13 +29,32 @@ import type {
...
-  renderIcon?: React.ReactNode | ((...args: any[]) => any)
+  /**
+   * Add an icon to the Breadcrumb.Link
+   */
+  renderIcon?: React.ReactNode | (() => React.ReactNode)
+  /**
+   * Place the icon before or after the text in the Breadcrumb.Link
+   */
   iconPlacement?: 'start' | 'end'
 }
```
matyasf, line 59: I see this in the test:
```
    //@ts-expect-error fix this, but how? `to` is not a prop BreadcrumbLink
    await mount(<BreadcrumbLink to="/example">Test</BreadcrumbLink>)
```

When I check `Link` (where the props are forwarded) there is this explanation:

```
type LinkProps = LinkOwnProps &
  WithStyleProps<LinkTheme, LinkStyle> &
  OtherHTMLAttributes<LinkOwnProps> & {
    // React Router might add a `to` prop
    to?: string
  }
```

This means to me, that we should allow this `to` prop in `BreadcrumbLink` too.

ke1k0: Yeah I just commented the same thing :)

packages/ui-breadcrumb/src/Breadcrumb/index.tsx
```diff
@@ -63,22 +64,17 @@ class Breadcrumb extends Component<BreadcrumbProps> {
...
   }
 
   static Link = BreadcrumbLink
 
   renderChildren() {
     const { styles, children } = this.props
-    // @ts-expect-error ts-migrate(2339) FIXME: Property 'length' does not exist on type 'string |... Remove this comment to see the full error message
-    const numChildren = children ? children.length : 0
+    const numChildren = React.Children.count(children)
```
matyasf, line 77: oh nice, a bugfix :)

matyasf (approved):

PR #841, opened 2021-12-07
chore(ui-tag): add TS types to `ui-tag` package

ke1k0:
Closes: INSTUI-3085

### DIFF
--- a/packages/ui-tag/src/Tag/index.tsx
+++ b/packages/ui-tag/src/Tag/index.tsx
@@ -22,10 +22,11 @@
  * SOFTWARE.
  */
 /** @jsx jsx */
-import { Component } from 'react'
+import React, { Component } from 'react'
 
 import { IconXSolid } from '@instructure/ui-icons'
 import { View } from '@instructure/ui-view'
+import type { ViewProps } from '@instructure/ui-view'
 import { omitProps } from '@instructure/ui-react-utils'
 import { isActiveElement } from '@instructure/ui-dom-utils'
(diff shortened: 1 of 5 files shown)


packages/ui-tag/src/Tag/index.tsx
```diff
@@ -57,27 +59,25 @@ class Tag extends Component<TagProps> {
...
 
   get focused() {
-    // @ts-expect-error ts-migrate(2339) FIXME: Property '_container' does not exist on type 'Tag'... Remove this comment to see the full error message
-    return isActiveElement(this._container)
+    return isActiveElement(this.ref)
   }
 
   focus = () => {
-    // @ts-expect-error ts-migrate(2339) FIXME: Property '_container' does not exist on type 'Tag'... Remove this comment to see the full error message
-    this._container && this._container.focus()
```
ke1k0, line 76: This seemed to be a bug since the container would be View, and it doesn't have a focus method. So I ended up using the ref element and try to focus that. Also it doesn't seem that it is used anywhere and a normal (span) Tag cannot be focused anyway...

HerrTopi (approved):

packages/ui-tag/src/Tag/index.tsx
ke1k0: Well it can be used from the outside, and since the button version of the Tag CAN be focused, there might be someone using this.

Brailor (approved):

PR #842, opened 2021-12-09
INSTUI-3346 - refactor event handlers and refs

ke1k0:


packages/ui-utils/src/createChainedFunction.ts
```diff
@@ -38,8 +38,8 @@ type GenericFunction = (...args: any[]) => any
  * @returns {function|null}
  */
 
-function createChainedFunction(
-  ...funcs: (null | undefined | GenericFunction)[]
+function createChainedFunction<F extends GenericFunction = GenericFunction>(
```
ke1k0, line 41: I added this to ensure that all functions passed have the same/compatible types.

packages/ui-menu/src/Menu/index.tsx
```diff
@@ -470,9 +470,9 @@ class Menu extends Component<MenuProps> {
           }
         }}
         renderTrigger={safeCloneElement(trigger as ReactElement, {
-          ref: (el: ReactElement) => {
+          ref: (el: React.ReactInstance & { ref?: Element }) => {
             this._trigger = el
-            this.ref = el
+            this.ref = el.ref || (el as Element)
```
ke1k0, line 475: I refactored the ref-s as well, because (I think) these are ReactInstance type (Component<any> | Element). Also added a bit of extra logic here, because if the ref is a Component, we need its `.ref`, not the component itself.

matyasf:
Chromatic is breaking for `Menu` with "cannot read property 'ref' of null"

Brailor (approved):

packages/ui-dom-utils/src/addEventListener.ts
```diff
@@ -32,22 +34,26 @@ import { findDOMNode } from './findDOMNode'
...
 function addEventListener(
   el: Node | Window,
   event: keyof WindowEventMap,
-  handler: EventListenerOrEventListenerObject,
+  handler: React.EventHandler<React.SyntheticEvent>,
   capture?: boolean
 ) {
   const node = el === window || el === document ? el : findDOMNode(el)
-  node?.addEventListener(event, handler, capture)
+  node?.addEventListener(event, handler as unknown as EventListener, capture)
```
matyasf, line 48: this is a bit ugly... maybe something like this is nicer:
```
  type ReactEvent = <T extends Event>(event: T)=>void
  node?.addEventListener(event, handler as ReactEvent, capture)
```

ke1k0: Oh I like it! Changed it, thanks!

packages/ui-utils/src/createChainedFunction.ts
matyasf: oh this is cool :)

matyasf (approved):

PR #843, opened 2021-12-10
chore(ui-buttons): add TS types to `ui-buttons` package

ke1k0:
Closes: INSTUI-3036

### DIFF
--- a/packages/ui-buttons/src/BaseButton/index.tsx
+++ b/packages/ui-buttons/src/BaseButton/index.tsx
@@ -23,9 +23,10 @@
  */
 
 /** @jsx jsx */
-import { Component } from 'react'
+import React, { Component } from 'react'
 import keycode from 'keycode'
 
+import { warn } from '@instructure/console'
 import { testable } from '@instructure/ui-testable'
 import {
   getElementType,
@@ -36,6 +37,7 @@ import {
 import { isActiveElement } from '@instructure/ui-dom-utils'
 import { hasVisibleChildren } from '@instructure/ui-a11y-utils'
 import { View } from '@instructure/ui-view'
+import type { ViewProps } from '@instructure/ui-view'
 
 import { withStyle, jsx } from '@instructure/emotion'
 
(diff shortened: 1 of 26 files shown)


packages/ui-buttons/src/BaseButton/index.tsx
```diff
@@ -131,6 +128,20 @@ class BaseButton extends Component<BaseButtonProps> {
     return this.interaction === 'enabled'
   }
 
+  // TODO: delete once the type of tabIndex is changed to number only
+  get tabIndex() {
```
ke1k0, line 132: The HTML tabIndex accepts numbers only, so I decided to deprecate the string version.

packages/ui-buttons/src/Button/props.ts
```diff
@@ -29,110 +29,135 @@ import { ThemeablePropTypes } from '@instructure/emotion'
...
-  renderIcon: PropTypes.oneOfType([PropTypes.node, PropTypes.func])
+  renderIcon?: React.ReactNode | (() => React.ReactNode)
+
+  /**
+   * Callback fired when the `Button` is clicked.
+   */
+  onClick?: (
+    event: React.KeyboardEvent<ViewProps> | React.MouseEvent<ViewProps>
+  ) => void
+}
```
ke1k0, line 127: I explicitly added the `onClick` prop to all Buttons since they will get passed to the underlying BaseButton anyway, and this way we can pass the type of the handler.

packages/ui-buttons/src/ToggleButton/props.ts
```diff
@@ -37,90 +37,116 @@ import type {
...
-  onClick: PropTypes.func,
+  onClick?: (
+    event: React.KeyboardEvent<ViewProps> | React.MouseEvent<ViewProps>
+  ) => void
+
   /**
    * The color in which to display the icon
    */
-  color: PropTypes.string,
+  color?: 'primary' | 'primary-inverse' | 'secondary' | 'success' | 'danger'
```
ke1k0, line 93: I specified the type for "color", because it used to be `string`, but it gets passed to IconButton, which has this type above. It shouldn't be breaking because if they used a wrong string before, the IconButton prop check caught it.

Brailor (approved):

matyasf (approved):
looks good!

PR #844, opened 2021-12-13
docs(ui-modal): readme

jacekk:
A small and first, but not last :wink: contribution.

### DIFF
--- a/packages/ui-modal/README.md
+++ b/packages/ui-modal/README.md
@@ -32,7 +32,7 @@ const MyModal = () => {
   return (
     <Modal>
       <Modal.Header>My Modal</Modal.Header>
-      <Modal.Content>Hello world</Modal.Content>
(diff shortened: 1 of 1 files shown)


Brailor:
Hi @jacekk!
Thanks for the contribution, although one of our CI checks have failed, it is not because of your change. (we recently added a check to monitor bundle sizes and it looks like it does not work properly for outside collaborators.)

Brailor (approved):

jacekk:
> Hi `jacekk`! Thanks for the contribution, although one of our CI checks have failed, it is not because of your change. (we recently added a check to monitor bundle sizes and it looks like it does not work properly for outside collaborators.)

@Brailor Yup! I'd checked the job, but I did not have time to find a proper fix. FYI, I am an Impact developer, and most likely, I'll contribute more in future.

PR #845, opened 2021-12-13
fix(ui-drawer-layout): fix tray direction calculation (INSTUI-3348)

HerrTopi:
Closes: INSTUI-3348

### DIFF
--- a/packages/ui-drawer-layout/src/DrawerLayout/index.tsx
+++ b/packages/ui-drawer-layout/src/DrawerLayout/index.tsx
@@ -108,7 +108,7 @@ class DrawerLayout extends Component<DrawerLayoutProps> {
       matchComponentTypes(child, [DrawerTray])
     )[0]
     // @ts-expect-error ts-migrate(2339) FIXME: Property 'props' does not exist on type 'ReactChil... Remove this comment to see the full error message
-    return tray.props
+    return { ...this.props, ...tray.props }
   }
 
   get trayPlacement() {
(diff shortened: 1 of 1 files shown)


packages/ui-drawer-layout/src/DrawerLayout/index.tsx
```diff
@@ -108,7 +108,7 @@ class DrawerLayout extends Component<DrawerLayoutProps> {
       matchComponentTypes(child, [DrawerTray])
     )[0]
     // @ts-expect-error ts-migrate(2339) FIXME: Property 'props' does not exist on type 'ReactChil... Remove this comment to see the full error message
-    return tray.props
+    return { ...this.props, ...tray.props }
```
ke1k0, line 111: I checked the problem too, and it probably has something to do with how we refactored the bidirectional context since V7. The problem is that the DrawerTray knows it has a dir prop that is coming from a decorator, but asking for it's props from the DrawerLayout doesn't show it. But mixing the Layout's props with the Tray's might be an overkill here and potentially dangerous if we want to use this getter for something else later, so I'd recommend handling this where the problem is:

In the `get trayPlacement` I would just give a default to the dir prop: 
```js
const { placement, dir = this.props.dir } = this.trayProps
```

But even like this is not the best solution, so we might need to dig into this deeper and see how our text direction contexts are working, and find a way to be able to access the child's `dir` prop too.

HerrTopi: On second thought: do we really want to use the child's dir prop for the parent? It seems a bit interesting

ke1k0: It is the other way around, we use the parent's `dir` for the Tray if it doesn't explicitly has one passed

ke1k0: Well I guess I know what you mean, but it's not the first time a component looks into the child's prop to decide how to render sth :D But it would be a big refactor or a whole redo to change these...

matyasf (approved):

ke1k0 (approved):

PR #850, opened 2021-12-14
chore(emotion,ui-i18n): throw warning when 'auto' dir (INSTUI-3306)

HerrTopi:
throw warning when 'auto' dir is passed to the TextDirectionContext

Closes: INSTUI-3306

### DIFF
--- a/packages/ui-i18n/src/ApplyTextDirection/__tests__/ApplyTextDirection.test.tsx
+++ b/packages/ui-i18n/src/ApplyTextDirection/__tests__/ApplyTextDirection.test.tsx
@@ -92,19 +92,6 @@ describe('<ApplyTextDirection />', async () => {
     expect(text.getAttribute('data-dir')).to.equal('ltr')
   })
 
-  it('should use the "auto" dir value correctly', async () => {
-    const subject = await mount(
-      <ApplyTextDirection dir="auto">
-        <span>
-          هذه الفقرة باللغة العربية ، لذا يجب الانتقال من اليمين إلى اليسار.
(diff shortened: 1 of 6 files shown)


matyasf (changes requested):
BidirectionalProps still has a comment about 'auto' direction; ApplyTextDirectionProps still accepts 'auto'

matyasf (approved):

ke1k0 (changes requested):
In the `ApplyTextDirection.test.tsx` it throws an error for the "auto" dir, please add a `ts-expect-error` that we intentionally pass it. It's also weird that this test passes, so we DO handle the auto value :)

packages/ui-i18n/src/ApplyTextDirection/__tests__/ApplyTextDirection.test.tsx
```diff
@@ -92,19 +92,6 @@ describe('<ApplyTextDirection />', async () => {
...
-        <span>
-          هذه الفقرة باللغة العربية ، لذا يجب الانتقال من اليمين إلى اليسار.
-        </span>
-      </ApplyTextDirection>
-    )
-    expect(
-      getComputedStyle(subject.getDOMNode().childNodes[0] as Element).direction
-    ).to.equal('rtl')
-  })
-
```
matyasf, line 107: if getTextDirection can return 'auto' we could move this test there

ke1k0 (approved):

matyasf (approved):
looks good, let's release

PR #852, opened 2021-12-15
INSTUI-3081-ts-typing-ui-spinner

NamNguyenInstructure:


packages/ui-spinner/src/Spinner/index.tsx
```diff
@@ -55,6 +55,7 @@ class Spinner extends Component<SpinnerProps> {
   }
 
   ref: Element | null = null
+  titleId: string | undefined
```
ke1k0, line 58: We decided to mark these as "private" props (not the `ref`, because it has to be accessible from the outside, but the other props that we use internally):

```js
ref: Element | null = null

private titleId: string | undefined
```

Also, this prop in particular is only set in the constructor, so it can be "readonly" too:

```js
private readonly titleId: string | undefined
```

And just one more thing: the way you wrote it is fine, just wanted to show that there is another way to mark it as "possibly undefined" or "optional" with the `?` mark.

```js
private readonly titleId?: string
```

packages/ui-spinner/src/Spinner/index.tsx
```diff
@@ -66,11 +67,8 @@ class Spinner extends Component<SpinnerProps> {
     }
   }
 
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'props' implicitly has an 'any' type.
-  constructor(props) {
+  constructor(props: SpinnerProps) {
```
ke1k0, line 70: Don't forget to add the `@tsProps` flag in the annotations! :)

Also, now that the typing is done, you can remove this block from the `babel.config.js` that removes the unnecessary `@ts-expect-errors`:

```js
  shouldPrintComment(value) {
    const isTsRelatedComment = /@ts-expect-error|FIXME/.test(value)

    return !isTsRelatedComment
  },
```

ke1k0 (changes requested):
Seems nice, just needs the @tsProps flag :)

Brailor (approved):
Good Work!

packages/ui-rating/src/RatingIcon/index.tsx
```diff
@@ -41,6 +41,7 @@ import { allowedProps, propTypes } from './props'
 parent: Rating
 id: Rating.Icon
 ---
+@tsProps
```
ke1k0, line 44: I remember from last year that you had this 2 branches mixed up a bit, but now accidentally you have a change for RatingIcon in the spinner PR :) Can you please revert this file on this branch?

packages/ui-spinner/src/Spinner/index.tsx
```diff
@@ -55,6 +56,7 @@ class Spinner extends Component<SpinnerProps> {
   }
 
   ref: Element | null = null
+  private readonly titleId?: string | undefined
```
ke1k0, line 59: The second part is not needed because the `?` already makes it optional :)

```suggestion
  private readonly titleId?: string
```

ke1k0 (changes requested):

ke1k0 (approved):

PR #853, opened 2021-12-15
chore(ui-range-input): add TS types to `ui-range-input` package

ke1k0:
Closes: INSTUI-3075

### DIFF
--- a/packages/ui-range-input/src/RangeInput/index.tsx
+++ b/packages/ui-range-input/src/RangeInput/index.tsx
@@ -23,7 +23,7 @@
  */
 
 /** @jsx jsx */
-import { Component } from 'react'
+import React, { Component } from 'react'
 
 import { ContextView } from '@instructure/ui-view'
 import { FormField } from '@instructure/ui-form-field'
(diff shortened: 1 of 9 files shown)


Brailor (approved):

NamNguyenInstructure (approved):

packages/ui-range-input/src/RangeInput/index.tsx
```diff
@@ -169,12 +166,18 @@ class RangeInput extends Component<RangeInputProps> {
...
             css={this.props.styles?.rangeInputInputValue}
           >
-            {/* @ts-expect-error ts-migrate(2722) FIXME: Cannot invoke an object which is possibly 'undefin... Remove this comment to see the full error message */}
-            {this.props.formatValue(this.value)}
+            {/**
+             * Value can be undefined when no value or defaultValue is given,
+             * and then it renders an empty ContextView
+             * TODO: Ask a designer what the intended behaviour is
+             */}
+            {this.props.formatValue!(this.value, this.props.max)}
```
ke1k0, line 174: Opened a ticket about this: INSTUI-3356

ke1k0: Little more explanation: the VRT made me notice this bug after typing it correctly, but I wasn't sure what the best solution is to fix it, so I reverted this part, added a comment and made a ticket.

PR #854, opened 2021-12-15
Create git pre-commit hook that checks if the dependencies listed in `package.json` does not match TS references

Brailor:


scripts/checkTSReferences.js
```diff
@@ -0,0 +1,135 @@
...
+      packageDepenecies: instuiDeps,
+      packageReferences
+    }
+  })
+  .filter(
+    ({ packageDepenecies, packageReferences }) =>
+      !eqaulSets(packageReferences, packageDepenecies)
+  )
+
+if (packageDepsAndRefs.length > 0) {
```
ke1k0, line 85: Well technically `packageDepsAndRefs` returns the the refs and deps for only those that have diffs (because of the filter), so it is hard to understand for first time reading. I would separate the 2 and make a new const for readablity, sth like: 

```js
const depsAndRefsWithDiff = packageDepsAndRefs.filter(
  ({ packageDepenecies, packageReferences }) =>
    !eqaulSets(packageReferences, packageDepenecies)
)

if (depsAndRefsWithDiff.length > 0) { ... }
```

Also I'm not even sure the `if` is needed, since the `for () {}` will do nothing if it has to iterate through an empty array, or?

scripts/checkTSReferences.js
```diff
@@ -0,0 +1,135 @@
...
+      `Please make sure the package.json 'dependecies' and 'devDependecies' entries and the tsconfig.build.json 'references' are synchronized!`
+    )
+    console.warn(
+      'These are the dependencies for the package:\n',
+      chalk.green([...packageDepenecies])
+    )
+    console.warn(
+      'And these are the project references listed in tsconfig.build.json:\n',
+      chalk.blue([...packageReferences])
+    )
```
ke1k0, line 100: I'm not sure it is needed to print these out to the console, makes it kinda hard to read when there are tons of deps, and we can compare the 2 in the IDE too :) But let's hear the opinion of the others too. Can we at least put a line-break between 2 `Dependency mismatch` blocks?

scripts/checkTSReferences.js
```diff
@@ -0,0 +1,135 @@
...
+  .filter(
+    ({ packageDepenecies, packageReferences }) =>
+      !eqaulSets(packageReferences, packageDepenecies)
+  )
+
+if (packageDepsAndRefs.length > 0) {
+  for (let packageInfo of packageDepsAndRefs) {
+    const { packageDepenecies, packageReferences, package } = packageInfo
+
+    console.warn(`Dependecy mismatch in package: ${chalk.red(package)}!`)
```
ke1k0, line 89:
```suggestion
    console.warn(`Dependency mismatch in package: ${chalk.red(package)}!`)
```

scripts/checkTSReferences.js
```diff
@@ -0,0 +1,135 @@
...
+    console.warn(
+      `Please make sure the package.json 'dependecies' and 'devDependecies' entries and the tsconfig.build.json 'references' are synchronized!`
+    )
```
ke1k0, line 92:
```suggestion
    console.warn(
      `Please make sure the package.json 'dependencies' and 'devDependencies' entries and the tsconfig.build.json 'references' are synchronized!`
    )
```

scripts/checkTSReferences.js
```diff
@@ -0,0 +1,135 @@
...
+        ...symmetricDifference(packageDepenecies, packageReferences)
+      ])
+    )
+  }
+  process.exit(1)
+}
+
+process.exit(0)
+
+function eqaulSets(setA, setB) {
```
matyasf, line 113: typo, should be "equal"

scripts/checkTSReferences.js
```diff
@@ -0,0 +1,135 @@
...
+          // check if the dependent package is a ts package at all
+          return fs.existsSync(
+            path.resolve(`./packages/${depName}/tsconfig.build.json`)
+          )
+        })
+    )
+
+    return {
+      package,
+      packageDepenecies: instuiDeps,
```
matyasf, line 76: type, should be "Dependencies"

scripts/checkTSReferences.js
```diff
@@ -0,0 +1,135 @@
...
+ * SOFTWARE.
+ */
+/* eslint-disable @typescript-eslint/no-var-requires */
+/* eslint-disable no-console */
+
+/**
+ * This simple script will check if there is any discrepancies between the listed instui related
+ * dependencies of a given package and it's typescript references listed in the package's tsconfig.build.json config.
+ *
+ * If there is any it will print out the differences to the console.
```
matyasf, line 32: I'd add that it also exits with an error

scripts/checkTSReferences.js
matyasf: I think its good that we print what exactly the error is, a line break would be nice

matyasf (approved):
besides the small typos it looks good

scripts/checkTSReferences.js
Brailor: You are right, it was confusing, I refactored it.
And the `if` condition is needed for determining the status code for the process.exit(code), so if there are any package with diffs, then we should exit the process with a code other than 0 (thus breaking the pre-commit hook).

scripts/checkTSReferences.js
```diff
@@ -0,0 +1,135 @@
...
+const packageDepsAndRefs = tsReferences
+  .map((reference) => {
+    const [, dir, package] = reference.path.split('/')
+    const packageJson = require(path.resolve(
+      `./${dir}/${package}/package.json`
+    ))
+    const tsConfigJson = require(path.resolve(`${reference.path}`))
+    const { devDependencies, dependencies } = packageJson
+
+    const packageReferences = new Set(
```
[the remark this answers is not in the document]
Brailor: Fixed

ke1k0 (approved):

PR #858, opened 2021-12-16
chore(ui-tooltip): add TS types to `ui-tooltip` package

ke1k0:
Closes: INSTUI-3095

### DIFF
--- a/packages/ui-tooltip/src/Tooltip/props.ts
+++ b/packages/ui-tooltip/src/Tooltip/props.ts
@@ -31,7 +31,8 @@ import { element } from '@instructure/ui-prop-types'
 import type {
   PropValidators,
   AsElementType,
-  TooltipTheme
+  TooltipTheme,
+  OtherHTMLAttributes
 } from '@instructure/shared-types'
 import type {
   PlacementPropValues,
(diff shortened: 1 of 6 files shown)


HerrTopi (approved):

packages/ui-tooltip/src/Tooltip/props.ts
```diff
@@ -31,127 +31,177 @@ import { element } from '@instructure/ui-prop-types'
...
-   * @param {Boolean} renderProps.focused - Is the Tooltip trigger focused?
-   * @param {Function} renderProps.getTriggerProps - Props to be spread onto the trigger element
+   * A ReactNode or a function that returns a ReactNode with the following params:
+   *
+   * @param {Boolean} focused - Is the Tooltip trigger focused?
+   * @param {Function} getTriggerProps - Props to be spread onto the trigger element
    */
-  children: PropTypes.oneOfType([PropTypes.node, PropTypes.func]).isRequired,
+  children: React.ReactNode | TooltipRenderChildren
+
```
Brailor, line 64: Not something that is concerning this PR, but JSDoc comments for TS types (PropTypes too for the matter) does not render nicely in our `Properties` table:
Screenshot 2022-01-03 at 10 07 02

Maybe it would be a good idea to investigate this later?

Brailor (approved):

packages/ui-tooltip/src/Tooltip/props.ts
ke1k0: Yeah, well, this seems to have been like this forever, it is the same in V7 :/ Also, I think that once we have better control over our TS types in the docs, and can resolve stuff like "TooltipRenderChildren" syntax, then the JSDocs won't be needed at all.

PR #859, opened 2021-12-17
INSTUI-3350  ui-tag: position text and close "x" better

matyasf:


HerrTopi (approved):

packages/ui-tag/src/Tag/styles.ts
```diff
@@ -59,7 +59,7 @@ const generateStyle = (componentTheme: TagTheme, props: TagProps): TagStyle => {
       text: {
         lineHeight: `calc(${componentTheme.heightMedium} - (${componentTheme.defaultBorderWidth} * 2))`
       },
-      icon: { fontSize: '0.75rem' }
```
ke1k0, line 62: The "small" version is still a few px lower :) Also, I guess this PR needs a +1 from Ákos as well :D

matyasf: Akos has given his +1, he told me that we only use the "normal" size

ke1k0 (approved):

PR #860, opened 2021-12-17
chore(ui-rating): add TS types to `ui-rating` package

NamNguyenInstructure:
Closes: INSTUI-3076

### DIFF
--- a/packages/ui-rating/src/Rating/index.tsx
+++ b/packages/ui-rating/src/Rating/index.tsx
@@ -39,6 +39,7 @@ import { allowedProps, propTypes } from './props'
 ---
 category: components
 ---
+@tsProps
 **/
 @withStyle(generateStyle, null)
 @testable()
@@ -50,8 +51,8 @@ class Rating extends Component<RatingProps> {
 
   static defaultProps = {
     animateFill: false,
-    // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'filled' implicitly has an 'any' type.
-    formatValueText: (filled, iconCount) => `${filled} / ${iconCount}`,
+    formatValueText: (filled: string, iconCount: string) =>
+      `${filled} / ${iconCount}`,
     iconCount: 3,
     size: 'medium',
     valueNow: 0
@@ -77,15 +78,12 @@ class Rating extends Component<RatingProps> {
     const { valueNow, iconCount, valueMax } = this.props
 
     // prevent divide by zero errors
-    // @ts-expect-error ts-migrate(2532) FIXME: Object is possibly 'undefined'.
-    const max = valueMax > 0 ? valueMax : iconCount
+    const max = !!valueMax && valueMax > 0 ? valueMax : iconCount
 
-    // @ts-expect-error ts-migrate(2532) FIXME: Object is possibly 'undefined'.
-    const filledIcons = Math.round((valueNow * iconCount) / max)
+    const filledIcons = Math.round((valueNow! * iconCount!) / max!)
 
     // Handle edge case where valueNow is greater than valueMax
-    // @ts-expect-error ts-migrate(2532) FIXME: Object is possibly 'undefined'.
-    if (filledIcons > iconCount) {
+    if (filledIcons > iconCount!) {
       return iconCount
     } else {
       return filledIcons
@@ -93,22 +91,14 @@ class Rating extends Component<RatingProps> {
   }
 
   get empty() {
-    // @ts-expect-error ts-migrate(2532) FIXME: Object is possibly 'undefined'.
-    return this.props.iconCount - this.filled
+    return this.props.iconCount! - this.filled!
   }
 
   render() {
-    const {
-      iconCount,
-      animateFill,
-      size,
-      margin,
-      label,
-      formatValueText
-    } = this.props
+    const { iconCount, animateFill, size, margin, label, formatValueText } =
+      this.props
 
-    // @ts-expect-error ts-migrate(2722) FIXME: Cannot invoke an object which is possibly 'undefin... Remove this comment to see the full error message
-    const valueText = label + ' ' + formatValueText(this.filled, iconCount)
+    const valueText = label + ' ' + formatValueText?.(this.filled, iconCount)
 
     const passthroughProps = View.omitViewProps(
       omitProps(this.props, Rating.allowedProps),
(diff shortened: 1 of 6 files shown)


packages/ui-rating/src/Rating/index.tsx
```diff
@@ -77,38 +78,27 @@ class Rating extends Component<RatingProps> {
     const { valueNow, iconCount, valueMax } = this.props
 
     // prevent divide by zero errors
-    // @ts-expect-error ts-migrate(2532) FIXME: Object is possibly 'undefined'.
-    const max = valueMax > 0 ? valueMax : iconCount
+    const max = !!valueMax && valueMax > 0 ? valueMax : iconCount
```
ke1k0, line 81: Adding the `!` at the end of iconCount will fix the tpye of `max` to be always be a number (since iconCount has a defaultValue of 3).

```suggestion
    const max = !!valueMax && valueMax > 0 ? valueMax : iconCount!
```

packages/ui-rating/src/Rating/index.tsx
```diff
@@ -77,38 +78,27 @@ class Rating extends Component<RatingProps> {
...
 
-    // @ts-expect-error ts-migrate(2532) FIXME: Object is possibly 'undefined'.
-    const filledIcons = Math.round((valueNow * iconCount) / max)
+    const filledIcons = Math.round((valueNow! * iconCount!) / max!)
```
ke1k0, line 83: If you fix the type of `max`, the `!` is not needed here

```suggestion
    const filledIcons = Math.round((valueNow! * iconCount!) / max)
```

packages/ui-rating/src/Rating/index.tsx
```diff
@@ -77,38 +78,27 @@ class Rating extends Component<RatingProps> {
...
 
     // Handle edge case where valueNow is greater than valueMax
-    // @ts-expect-error ts-migrate(2532) FIXME: Object is possibly 'undefined'.
-    if (filledIcons > iconCount) {
+    if (filledIcons > iconCount!) {
       return iconCount
```
ke1k0, line 87: Add a `!` here too because it will never return undefined (iconCount has a default). This will make the whole function type fixed and will say it always returns a number. Because of this you can remove the `!` from `this.filled!` in other places where it is used.

```suggestion
      return iconCount!
```

packages/ui-rating/src/Rating/index.tsx
```diff
@@ -50,8 +51,8 @@ class Rating extends Component<RatingProps> {
 
   static defaultProps = {
     animateFill: false,
-    // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'filled' implicitly has an 'any' type.
-    formatValueText: (filled, iconCount) => `${filled} / ${iconCount}`,
+    formatValueText: (filled: string, iconCount: string) =>
```
ke1k0, line 54: Once you fix the types below, this will be: 

```suggestion
    formatValueText: (filled: number, iconCount: number) =>
```

packages/ui-rating/src/Rating/props.ts
```diff
@@ -36,13 +36,39 @@ import type {
...
 
 type RatingOwnProps = {
+  /**
+   * A label is required for accessibility
+   */
   label: string
+  /**
+   * A function that returns the current value formatted for screen readers
+   */
   formatValueText?: (...args: any[]) => any
```
ke1k0, line 46: We know the type of this from the index file:

```suggestion
  /**
   * A function that returns the current value formatted for screen readers
   */
  formatValueText?: (filled: number, iconCount: number) => string
```

packages/ui-rating/src/RatingIcon/__tests__/RatingIcon.test.tsx
```diff
@@ -26,9 +26,8 @@ import React from 'react'
 import { expect, mount, within } from '@instructure/ui-test-utils'
 
 import { InstUISettingsProvider } from '@instructure/emotion'
-// @ts-expect-error ts-migrate(7016) FIXME: Could not find a declaration file for module '@ins... Remove this comment to see the full error message
-// eslint-disable-next-line no-restricted-imports
-import { TransitionLocator } from '@instructure/ui-motion/es/Transition/TransitionLocator'
```
ke1k0, line 31: Hm, I1m not sure about this, might be a reason why they use the built version here. @matyasf or @Brailor you guys have any idea?

packages/ui-rating/src/RatingIcon/index.tsx
```diff
@@ -58,23 +58,20 @@ class RatingIcon extends Component<RatingIconProps, RatingIconState> {
 
   ref: Element | null = null
```
ke1k0, line 59: Actually we can tell that it is HTMLSpanElement:
```suggestion
  ref: HTMLSpanElement | null = null
```

packages/ui-rating/src/Rating/index.tsx
```diff
@@ -39,6 +39,7 @@ import { allowedProps, propTypes } from './props'
 ---
 category: components
 ---
+@tsProps
```
ke1k0, line 42: Add `@tsProps` to the RatingIcon too!

ke1k0 (changes requested):

packages/ui-rating/src/RatingIcon/__tests__/RatingIcon.test.tsx
ke1k0: ping @matyasf és @Brailor :)

HerrTopi (approved):

packages/ui-rating/src/RatingIcon/__tests__/RatingIcon.test.tsx
matyasf: hmm I don't think it has any special meaning, I think it will get this file from `/es/` by default. And if something would be wrong it would fail with some kind of "package not found" error.

ke1k0: I would put it back as it was, this really seems like sth that was deliberately made this way. Later we can have a look at these and fix everywhere, but I wouldn't risk breaking sth as part of a TS typing task :)

ke1k0: Please add this back as it was :)

ke1k0: I think this is the only thing left from this PR

ke1k0 (changes requested):

ke1k0 (approved):

PR #861, opened 2021-12-17
INSTUI-3032 refactor(ui-axe-check): better TypeScript types

matyasf:


packages/ui-axe-check/src/runAxeCheck.ts
```diff
@@ -21,30 +21,39 @@
...
+  element: string,
+  options: AxeCheckOptions = {}
+) {
+  let result: Error | true = true
+  const context: axeCore.ContextObject = {
     include: [element],
-    // @ts-expect-error ts-migrate(2339) FIXME: Property 'exclude' does not exist on type '{}'.
     exclude: options.exclude
   }
   const config = {
```
ke1k0, line 45: I'd rather do this than a cast:
```suggestion
  const config: axeCore.RunOptions = {
```

packages/ui-axe-check/src/runAxeCheck.ts
```diff
@@ -21,30 +21,39 @@
...
+  exclude?: string | string[]
+  /**
+   * Violations to ignore/filter out. 'color-contrast' is automatically excluded
+   * because we test color contrast in theme tests.
+   */
+  ignores?: string[]
+}
+
+export default async function runAxe(
+  element: string,
```
ke1k0, line 37: From the build error it seems that we pass HTMLElements to the `runAxeCheck(element, options)` in the ui-test-queries `accessible` util.

packages/ui-axe-check/src/runAxeCheck.ts
```diff
@@ -79,20 +87,19 @@ export default async function runAxe(element, options = {}) {
...
+        violation.nodes
+          .map(function (node) {
+            return node.target.toString()
+          })
+          .join('\n'),
+        violation.description,
+        `${violation.helpUrl}\n`
+      ].join('\n')
+    })
+    .join('\n')
```
ke1k0, line 104: Is this a bug fix?

matyasf: yep, it had the wrong type

packages/ui-axe-check/src/runAxeCheck.ts
matyasf: fixed

packages/ui-axe-check/src/runAxeCheck.ts
matyasf: its a bug in axe-check's typing I've opened a bug and did a cast as a workaround

packages/ui-axe-check/src/runAxeCheck.ts
```diff
@@ -21,30 +21,39 @@
...
  */
+
 import axeCore from 'axe-core'
 
-// @ts-expect-error ts-migrate(7006) FIXME: Parameter 'element' implicitly has an 'any' type.
-export default async function runAxe(element, options = {}) {
-  let result = true
-  const context = {
+type AxeCheckOptions = {
+  exclude?: string | string[]
```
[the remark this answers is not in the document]
matyasf: fixed it

packages/ui-axe-check/src/runAxeCheck.ts
ke1k0: oh cool, thanks!

ke1k0 (approved):

PR #864, opened 2022-01-03
chore: fix bootstrap not failing when sub process exits with non 0

Brailor:


scripts/bootstrap.js
```diff
@@ -23,25 +23,43 @@
...
-const util = require('util')
-const { execSync, fork, spawn } = require('child_process')
+const { execSync, fork, spawn, exec } = require('child_process')
 const path = require('path')
-const spawnAsync = util.promisify(spawn)
 
 const opts = { stdio: 'inherit' }
-async function buildProject() {
-  spawnAsync('yarn', ['build:ts'], opts)
+function buildProject() {
```
ke1k0, line 30: I tried it out, and have a question: I broke a type in View which is pretty high on the dep list, and it indeed killed the babel process but only once the whole tsBuild finished on all the components. Shouldn't we (can we?) stop and kill it as soon as any type breaks?

Brailor: Well, I don't think this is possible sadly, because in the bootstrap process the tsc buildinfo gets deleted, so it has to recreate the info the whole project.

But if you bootstrap the project once and after you run the `run build:ts` then it will fail much faster.

matyasf (approved):
looks good

ke1k0 (approved):

PR #865, opened 2022-01-04
chore(ui-pagination): add TS types to `ui-pagination` package

ke1k0:
Closes: INSTUI-3068

### DIFF
--- a/packages/ui-pagination/src/Pagination/PaginationButton/index.tsx
+++ b/packages/ui-pagination/src/Pagination/PaginationButton/index.tsx
@@ -56,12 +56,14 @@ class PaginationButton extends Component<PaginationPageProps> {
 
   render() {
     const exclude = this.props.current ? ['onClick', 'href'] : []
+
     const props = omitProps(this.props, PaginationButton.allowedProps, exclude)
+
     return (
       <BaseButton
         color="primary"
-        withBackground={this.props.current ? true : false}
-        withBorder={this.props.current ? true : false}
+        withBackground={this.props.current}
+        withBorder={this.props.current}
         {...props}
         aria-current={this.props.current ? 'page' : undefined}
         elementRef={this.handleRef}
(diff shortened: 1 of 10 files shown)


packages/ui-pagination/src/Pagination/index.tsx
```diff
@@ -402,8 +398,7 @@ class Pagination extends Component<PaginationProps> {
     if (!this.props.children) return null
 
     const currentPageIndex = fastFindIndex(
-      this.props.children,
-      // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'p' implicitly has an 'any' type.
+      this.childPages,
```
matyasf, line 401: is this a bugfix?

matyasf (approved):
looks good!

packages/ui-pagination/src/Pagination/index.tsx
ke1k0: Technically not, it was working before too, it is just typed better this way.

packages/ui-pagination/src/Pagination/PaginationButton/index.tsx
```diff
@@ -56,12 +56,14 @@ class PaginationButton extends Component<PaginationPageProps> {
...
+
     const props = omitProps(this.props, PaginationButton.allowedProps, exclude)
+
     return (
       <BaseButton
         color="primary"
-        withBackground={this.props.current ? true : false}
-        withBorder={this.props.current ? true : false}
+        withBackground={!!this.props.current}
+        withBorder={!!this.props.current}
```
Brailor, line 66: I don't really understand the need for the previous version of this code, but why is this `!!` needed?
Since `props.current` is a boolean why are we double negating it?

ke1k0: Well there is no "need", but since TS cannot really handle default prop values, it was `boolean | undefined` type... Which is still a totally valid value for the withBackground/withBorder prop, but wanted to indicate that it has a default value, and I didn't like `withBackground={this.props.current!}` :D I guess I'll just remove it, since it is really not needed...

ke1k0: removed it :)

Brailor (approved):

PR #869, opened 2022-01-06
InstUI-3357 Treebrowser typing

matyasf:


packages/ui-tree-browser/src/TreeBrowser/TreeCollection/props.ts
```diff
@@ -28,40 +28,25 @@ import PropTypes from 'prop-types'
...
-  collectionIconExpanded?: React.ReactNode | ((...args: any[]) => any)
-  itemIcon?: React.ReactNode | ((...args: any[]) => any)
-  getItemProps?: (...args: any[]) => any
-  getCollectionProps?: (...args: any[]) => any
-  onItemClick?: (...args: any[]) => any
-  onCollectionClick?: (...args: any[]) => any
-  onKeyDown?: (...args: any[]) => any
+  onKeyDown?: (e: React.KeyboardEvent, data: CollectionData) => void
+  // TODO it would be nice to use the same handler as in TreeBrowser
+  onItemClick?: (e: React.MouseEvent, data: CollectionData) => void
```
matyasf, line 43: the one in TreeBrowser does not pass the event

packages/ui-tree-browser/src/TreeBrowser/TreeCollection/index.tsx
```diff
@@ -100,45 +87,38 @@ class TreeCollection extends Component<TreeBrowserCollectionProps> {
 
   get itemsLevel() {
     const { level, isCollectionFlattened } = this.props
-    // @ts-expect-error ts-migrate(2532) FIXME: Object is possibly 'undefined'.
-    return isCollectionFlattened ? level : level + 1
+    return isCollectionFlattened ? level : level! + 1
```
ke1k0, line 90: I had to search for the reason why this is always not null. Shall we add a comment here, that the root collection always be level=1 or shall we make it required on the TreeCollection?

packages/ui-tree-browser/src/TreeBrowser/props.ts
```diff
@@ -31,28 +31,87 @@ import type {
...
+
+// props shared between TreeBrowser, TreeCollection
+type TreeBrowserBaseProps = {
+  /**
+   * A function called with each item's props as an argument. The return value of this function is a
+   * props object which will be passed to the item when it is rendered. This is useful for situations where
+   * you need to render the item differently depending on it's props. For example, if you would like to
+   * display a different icon for items with a certain name.
+   */
+  getItemProps?: (props: Record<string, any>) => Record<string, any> // cant use generics here :/
```
ke1k0, line 94: We could technically tell what the args are, but would be pretty complicated :D

packages/ui-tree-browser/src/TreeBrowser/props.ts
```diff
@@ -31,28 +31,87 @@ import type {
...
+  /**
+   * A function called with each collection's props as an argument. The return value of this function is a
+   * props object which will be passed to the collection when it is rendered. This is useful for situations where
+   * you need to render the collection differently depending on it's props. For example, if you would like to
+   * display a different icon for collections with a certain name.
+   */
+  getCollectionProps?: (props: Record<string, any>) => Record<string, any>
```
ke1k0, line 101: Same here, but we know for sure that it has to return `TreeBrowserButtonProps` because we end up passing it to the a TreeButton: `<TreeButton {...collectionProps} />`

packages/ui-tree-browser/src/TreeBrowser/index.tsx
```diff
@@ -342,46 +316,46 @@ class TreeBrowser extends Component<TreeBrowserProps> {
...
-        position={1}
-        renderContent={this.props.renderContent}
-      />
-    ))
+    return this.collections
+      .sort(this.props.sortOrder)
+      .map((collection, i) => (
+        <TreeCollection
+          key={i}
+          {...pickProps(this.props, TreeCollection.allowedProps)}
```
ke1k0, line 347: I think the `omitProps` here was supposed to cut all the automatically excluded props (e.g. withStyle props, className, etc). We could make the second parameter optional :) It might be good to keep it, because if in the future we add "className" as an allowedProp on TreeCollection, then TreeBrowser will pass it's own themeOverrides to TreeCollection, etc

ke1k0 (commented):
Oh wow this is one super (over)complicated component to type :D

packages/ui-tree-browser/src/TreeBrowser/props.ts
```diff
@@ -31,28 +31,87 @@ import type {
...
+// props shared between TreeBrowser, TreeCollection, TreeButton
+type TreeBrowserCommonProps = {
+  size?: 'small' | 'medium' | 'large'
+  variant?: 'folderTree' | 'indent'
+  collectionIcon?: React.ReactNode | ((props: unknown) => React.ReactNode)
+  collectionIconExpanded?:
+    | React.ReactNode
+    | ((props: unknown) => React.ReactNode)
+  itemIcon?: React.ReactNode | ((props: unknown) => React.ReactNode)
+  renderContent?: (props: any) => JSX.Element
```
ke1k0, line 114: Isn't the props here `TreeBrowserButtonProps`? It seems to be used in TreeButton's render method.

packages/ui-tree-browser/src/TreeBrowser/TreeCollection/props.ts
```diff
@@ -28,40 +28,25 @@ import PropTypes from 'prop-types'
...
   numChildren?: number
-  level?: number
   position?: number
-  renderBeforeItems?: any // TODO: Children.oneOf([TreeNode])
-  renderAfterItems?: any // TODO: Children.oneOf([TreeNode])
-  containerRef?: (...args: any[]) => any
```
ke1k0, line 61: Just a note here: there are some types that show up correctly on the docs page (eg TreeCollection's containerRef) because of the imports back and forth :/

packages/ui-tree-browser/src/TreeBrowser/TreeButton/props.ts
```diff
@@ -30,26 +30,24 @@ import type {
...
   id?: string | number
   name?: string
   descriptor?: string
-  type?: string
-  size?: 'small' | 'medium' | 'large'
-  variant?: 'folderTree' | 'indent'
-  collectionIcon?: React.ReactNode | ((...args: any[]) => any)
-  collectionIconExpanded?: React.ReactNode | ((...args: any[]) => any)
-  itemIcon?: React.ReactNode | ((...args: any[]) => any)
+  type?: 'collection' | 'item'
```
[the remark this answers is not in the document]
matyasf: hmm I'll add `| string` for custom nodes :)

packages/ui-tree-browser/src/TreeBrowser/TreeCollection/index.tsx
matyasf: good catch, as I see it always has a value for us (otherwise it'd be a bug). I'll make it mandatory

packages/ui-tree-browser/src/TreeBrowser/props.ts
matyasf: nice, done

packages/ui-tree-browser/src/TreeBrowser/index.tsx
matyasf: yep, you are right. I've made the second param of `omitProps` optional

packages/ui-tree-browser/src/TreeBrowser/props.ts
matyasf: yep, fixed

packages/ui-tree-browser/src/TreeBrowser/TreeCollection/props.ts
matyasf: I could not get the text show up for this one.. I've still added some text for the IDE

packages/ui-tree-browser/src/TreeBrowser/TreeCollection/props.ts
```diff
@@ -38,12 +38,12 @@ import type { WithStyleProps, ComponentStyle } from '@instructure/emotion'
 import { CollectionProps, TreeBrowserBaseProps } from '../props'
 
 type TreeBrowserCollectionOwnProps = {
+  level: number
```
ke1k0, line 41: on line 81 please make the PropType `.isRequired` too :)

packages/ui-tree-browser/src/TreeBrowser/TreeCollection/props.ts
ke1k0: I think it is important info for the render functions that they have to be TreeNode components, because we cannot rely on the type either (we couldn't type this well yet). So I suggest using the fallback option to leave the descriptions for these props above their PropTypes, because react-docgen picks it up from there too.

packages/ui-tree-browser/src/TreeBrowser/TreeCollection/props.ts
matyasf: done

packages/ui-tree-browser/src/TreeBrowser/TreeCollection/props.ts
Brailor: I think I have a solution which shows type information for the `containerRef, renderBeforeItems, renderAfterItems` functions for the `TreeCollection` component. The issue might be with the TS utility type `Omit<>`, but this is just the guess.
This is how I made it show type info:
Since the `Omit` is not working properly:
```ts
type CollectionProps = {
  collections?: CollectionProps[]
  items?: CollectionItem[]
  expanded?: boolean
  isCollectionFlattened?: boolean
} & Omit<Collection, 'collections' | 'items'>
```

What about reversing the `Collection` type and instead of Omitting stuff from it we would create 2 seperate types, like this:
```ts
type CollectionBase = {
  id: number | string
  name: string
  descriptor?: string
  /**
   * A function that returns a reference to the underlying HTML container
   * @param el The DOM HTMLElement
   */
  containerRef?: (el: HTMLElement | null) => void
  /**
   * children of type TreeNode
   */
  renderBeforeItems?: ReactElement // TODO: Children.oneOf([TreeNode])
  /**
   * children of type TreeNode
   */
  renderAfterItems?: ReactElement // TODO: Children.oneOf([TreeNode])
}
type Collection = CollectionBase & {
  items?: number[]
  collections?: (number | string)[]
}
```

And then, we can reuse the `CollectionBase` for the `CollectionProps`: 
```ts
type CollectionProps = {
  collections?: CollectionProps[]
  items?: CollectionItem[]
  expanded?: boolean
  isCollectionFlattened?: boolean
} & CollectionBase
```
Doing this way we have proper type info for the `Tree.Collection`:
Screenshot 2022-01-20 at 10 32 41

matyasf: great, this worked :)

Brailor (approved):

ke1k0 (approved):

PR #870, opened 2022-01-06
fix(ui-range-input): fixed empty label when `displayValue={true}` but…

ke1k0:
… has no `value`/`defaultValue`

Closes: INSTUI-3356

If the `displayValue={true}` (by default) but no `value` or `defaultValue` was set, and empty value
label (ContextView) was displayed. Now it won't display the label and give a warning message.


matyasf (approved):
looks good!

HerrTopi (approved):

PR #871, opened 2022-01-07
chore(ui-drawer-layout): add TS types to `ui-drawer-layout` package

ke1k0:
Closes: INSTUI-3046

### DIFF
--- a/packages/ui-drawer-layout/src/DrawerLayout/DrawerContent/__tests__/DrawerContent.test.tsx
+++ b/packages/ui-drawer-layout/src/DrawerLayout/DrawerContent/__tests__/DrawerContent.test.tsx
@@ -24,6 +24,7 @@
 import React from 'react'
 
 import { expect, mount, stub, wait } from '@instructure/ui-test-utils'
+
 import { DrawerContent } from '../index'
 
 import { DrawerContentLocator } from '../DrawerContentLocator'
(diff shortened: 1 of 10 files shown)


packages/ui-drawer-layout/src/DrawerLayout/DrawerContent/__tests__/DrawerContent.test.tsx
```diff
@@ -38,24 +39,21 @@ describe('<DrawerContent />', async () => {
     expect(drawerContent).to.exist()
   })
 
-  it('should not have a transition class if `shouldTransition` is set to false', async () => {
+  it('should not transition on mount, just on update', async () => {
```
ke1k0, line 42: I refactored this test because it wasn't changed when we changed it's logic (twice since then actually).

packages/ui-drawer-layout/src/DrawerLayout/DrawerContent/index.tsx
```diff
@@ -37,124 +38,109 @@ import generateComponentTheme from './theme'
...
-    // @ts-expect-error ts-migrate(6133) FIXME: 'el' is declared but its value is never read.
-    contentRef: (el: any) => {},
-    // @ts-expect-error ts-migrate(6133) FIXME: 'node' is declared but its value is never read.
-    onSizeChange: (node: any) => {},
     role: 'region'
   }
 
-  state: DrawerLayoutContentState = {
-    shouldTransition: false
-  }
```
ke1k0, line 71: Refactored the. whole handling of "shouldTransition" because it used to be a prop that we no longer use (even in V7), and since an earlier refactor the state isn't used anymore either, it is directly passed to makeStyles.

packages/ui-drawer-layout/src/DrawerLayout/DrawerTray/props.ts
```diff
@@ -35,38 +35,132 @@ import type {
...
+
+  /**
+   * Called when the `<DrawerLayout.Tray />` is opened
+   */
+  onOpen?: (transitionType?: TransitionType) => void
+
+  /**
+   * Called when the `<DrawerLayout.Tray />` is closed
+   */
+  onClose?: (transitionType?: TransitionType) => void
```
ke1k0, line 73: These args are not necessarily needed here but the Transition component provides them anyway.

Brailor (approved):

packages/ui-drawer-layout/src/DrawerLayout/DrawerContent/index.tsx
```diff
@@ -37,124 +38,109 @@ import generateComponentTheme from './theme'
...
     return this.ref
   }
-  _resizeListener = null
 
-  _debounced = null
-  _timeouts = []
+  private _resizeListener?: ResizeObserver
+
+  private _debounced?: Debounced
+  private _timeouts = []
```
matyasf, line 76: this is also unused,its always an empty array

ke1k0: Oh, nice catch, haven't noticed! Thanks!

matyasf (approved):

PR #877, opened 2022-01-14
chore(ui-table): type ui-table package (INSTUI-3083)

HerrTopi:
Closes: INSTUI-3083

### DIFF
--- a/packages/ui-table/src/Table/Head/index.tsx
+++ b/packages/ui-table/src/Table/Head/index.tsx
@@ -23,7 +23,7 @@
  */
 
 /** @jsx jsx */
-import { Component, Children } from 'react'
+import React, { Component, Children } from 'react'
 
 import {
   omitProps,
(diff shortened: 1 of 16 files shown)


packages/ui-table/src/Table/Head/index.tsx
```diff
@@ -64,16 +67,23 @@ class Head extends Component<TableHeadProps> {
   }
 
   get isSortable() {
-    const [row] = Children.toArray(this.props.children)
+    const [row] = Children.toArray(
+      this.props.children
+    ) as React.ComponentElement<TableRowProps, Row>[]
     let sortable = false
 
     if (row) {
```
matyasf, line 75: This code is a bit weird, as I see its always true

matyasf (approved):
looks good!

Brailor (approved):
I'm not sure how to handle it at the moment, but this is how the documentation looks like for the `Table.Body`:
Screenshot 2022-01-21 at 12 44 57

I guess it is because of this type:

```ts

type TableBodyOwnProps = {
  hover?: boolean
  isStacked?: boolean
  headers?: TableRowProps['headers']
  /**
   * `Table.Row`
   */
  children?: React.ReactNode
}
```

Maybe this is something that is worth noting down.

PR #878, opened 2022-01-17
chore(ui-checkbox): Type UI checkbox (INSTUI-3039)

HerrTopi:
Closes: INSTUI-3039

### DIFF
--- a/packages/ui-checkbox/src/Checkbox/CheckboxFacade/props.ts
+++ b/packages/ui-checkbox/src/Checkbox/CheckboxFacade/props.ts
@@ -36,6 +36,9 @@ type CheckboxFacadeOwnProps = {
   focused?: boolean
   hovered?: boolean
   size?: 'small' | 'medium' | 'large'
+  /**
+   * Visual state showing that child checkboxes are a combination of checked and unchecked
+   */
   indeterminate?: boolean
 }
 
@@ -54,9 +57,6 @@ const propTypes: PropValidators<PropKeys> = {
   focused: PropTypes.bool,
   hovered: PropTypes.bool,
   size: PropTypes.oneOf(['small', 'medium', 'large']),
-  /**
-   * Visual state showing that child checkboxes are a combination of checked and unchecked
-   */
   indeterminate: PropTypes.bool
 }
 
--- a/packages/ui-checkbox/src/Checkbox/index.tsx
+++ b/packages/ui-checkbox/src/Checkbox/index.tsx
@@ -89,8 +89,8 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
 
     this._defaultId = uid('Checkbox')
   }
-  private _defaultId: string
-  private _input?: HTMLInputElement | null
+  private readonly _defaultId: string
+  private _input: HTMLInputElement | null = null
 
   ref: Element | null = null
 
@@ -100,7 +100,7 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
 
   componentDidMount() {
     if (this._input) {
-      this._input.indeterminate = this.props.indeterminate || false
+      this._input.indeterminate = this.props.indeterminate!
     }
 
     this.props.makeStyles?.()
--- a/packages/ui-checkbox/src/Checkbox/index.tsx
+++ b/packages/ui-checkbox/src/Checkbox/index.tsx
@@ -23,7 +23,7 @@
  */
 
 /** @jsx jsx */
-import { Component } from 'react'
+import React, { Component } from 'react'
 import keycode from 'keycode'
 
 import { FormFieldMessages } from '@instructure/ui-form-field'
@@ -43,7 +43,7 @@ import { ToggleFacade } from './ToggleFacade'
 import generateStyle from './styles'
 
 import { propTypes, allowedProps } from './props'
-import type { CheckboxProps } from './props'
+import type { CheckboxProps, CheckboxState } from './props'
 
 import type {
   CheckboxFacadeTheme,
@@ -55,11 +55,12 @@ import type {
 category: components
 tags: toggle, switch
 ---
+@tsProps
 **/
 
 @withStyle(generateStyle, null)
 @testable()
-class Checkbox extends Component<CheckboxProps> {
+class Checkbox extends Component<CheckboxProps, CheckboxState> {
   static readonly componentId = 'Checkbox'
 
   static propTypes = propTypes
@@ -74,23 +75,22 @@ class Checkbox extends Component<CheckboxProps> {
     labelPlacement: 'end'
   }
 
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'props' implicitly has an 'any' type.
-  constructor(props) {
+  constructor(props: CheckboxProps) {
     super(props)
 
     this.state = {
       focused: false,
-      hovered: false
+      hovered: false,
+      checked:
+        typeof props.checked === 'undefined'
+          ? !!props.defaultChecked
+          : undefined
     }
 
-    if (typeof props.checked === 'undefined') {
-      // @ts-expect-error ts-migrate(2339) FIXME: Property 'checked' does not exist on type 'Readonl... Remove this comment to see the full error message
-      this.state.checked = !!props.defaultChecked
-    }
-
-    // @ts-expect-error ts-migrate(2339) FIXME: Property '_defaultId' does not exist on type 'Chec... Remove this comment to see the full error message
     this._defaultId = uid('Checkbox')
   }
+  private _defaultId: string
+  private _input?: HTMLInputElement | null
 
   ref: Element | null = null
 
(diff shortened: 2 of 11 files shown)


packages/ui-checkbox/src/Checkbox/index.tsx
```diff
@@ -138,66 +131,52 @@ class Checkbox extends Component<CheckboxProps> {
...
-  handleKeyDown = (e) => {
-    if (
-      this.props.variant === 'toggle' &&
-      // @ts-expect-error ts-migrate(2339) FIXME: Property 'return' does not exist on type 'CodesMap... Remove this comment to see the full error message
-      (e.keyCode === keycode.codes.enter || e.keyCode === keycode.codes.return)
-    ) {
-      // @ts-expect-error ts-migrate(2339) FIXME: Property '_input' does not exist on type 'Checkbox... Remove this comment to see the full error message
-      this._input.click()
+  handleKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
+    if (this.props.variant === 'toggle' && e.keyCode === keycode.codes.enter) {
```
Brailor, line 135: The previous version had:
```ts
if (e.keyCode === keycode.codes.enter || e.keyCode === keycode.codes.return) {...}
```
isn't this a regression?

HerrTopi: I think not. According to TS, return isn't found on keycodes.codes. Return and Enter is an alias for '13'.

https://github.com/timoxley/keycode/blob/c282ff62a4ff05a523d6026a66e83a4892e7f6e7/index.js#L77
https://github.com/timoxley/keycode/blob/c282ff62a4ff05a523d6026a66e83a4892e7f6e7/index.js#L134

packages/ui-checkbox/src/Checkbox/props.ts
```diff
@@ -40,17 +40,39 @@ type CheckboxOwnProps = {
   label: React.ReactNode
   id?: string
   value?: string | number
+  /**
+   * object with shape: `{
+   * text: PropTypes.node,
+   * type: PropTypes.oneOf(['error', 'hint', 'success', 'screenreader-only'])
+   *   }`
+   */
```
ke1k0, line 48:
```suggestion
  /**
   * Array of objects with shape: `{
   *   text: ReactNode,
   *   type: One of: ['error', 'hint', 'success', 'screenreader-only']
   *   }`
   */
```

(this is the format we use in other input components too)

packages/ui-checkbox/src/CheckboxGroup/props.ts
```diff
@@ -35,14 +35,15 @@ import type { FormMessage } from '@instructure/ui-form-field'
 import type { PropValidators } from '@instructure/shared-types'
 
 import { Checkbox } from '../Checkbox'
+import type { CheckboxProps } from '../Checkbox/props'
 
 type CheckboxGroupOwnProps = {
-  children?: React.ReactNode // TODO: oneOf([Checkbox])
+  children?: React.ComponentElement<CheckboxProps, Checkbox>[]
```
ke1k0, line 41: Sadly this is not a perfect substitute for 'ChildrenPropTypes.oneOf([Checkbox])' so let's leave the comment here until we find the best way to types this.

packages/ui-checkbox/src/CheckboxGroup/props.ts
```diff
@@ -35,14 +35,15 @@ import type { FormMessage } from '@instructure/ui-form-field'
...
   name: string
   description: React.ReactNode
-  defaultValue?: any[]
-  value?: any // TODO: controllable(PropTypes.array)
-  onChange?: (...args: any[]) => any
+  defaultValue?: (string | number)[]
+  value?: (string | number)[]
```
ke1k0, line 45: Please leave the comment here (for controllable)

packages/ui-checkbox/src/CheckboxGroup/index.tsx
```diff
@@ -59,21 +61,18 @@ class CheckboxGroup extends Component<CheckboxGroupProps> {
...
-        value: props.defaultValue
+        value: props.defaultValue || []
       }
     }
 
-    // @ts-expect-error ts-migrate(2339) FIXME: Property '_messagesId' does not exist on type 'Che... Remove this comment to see the full error message
     this._messagesId = uid('CheckboxGroup-messages')
   }
-
+  private _messagesId: string
```
ke1k0, line 75:
```suggestion
  private readonly _messagesId: string
```

packages/ui-checkbox/src/CheckboxGroup/index.tsx
```diff
@@ -59,21 +61,18 @@ class CheckboxGroup extends Component<CheckboxGroupProps> {
...
-  constructor(props) {
-    // @ts-expect-error ts-migrate(2554) FIXME: Expected 1-2 arguments, but got 0.
-    super()
+  constructor(props: CheckboxGroupProps) {
+    super(props)
 
     if (typeof props.value === 'undefined') {
       this.state = {
-        value: props.defaultValue
+        value: props.defaultValue || []
```
ke1k0, line 69: If you give this a default value, then a lot of checks are not necessary in the code. Can you please double-check if this has any side-effects? If not, then we can simplify the code a bit.

packages/ui-checkbox/src/CheckboxGroup/index.tsx
```diff
@@ -111,40 +109,46 @@ class CheckboxGroup extends Component<CheckboxGroupProps> {
...
         : [...this.props.value]
     }
   }
 
   renderChildren() {
     const { children, name, size, disabled, readOnly } = this.props
 
-    // @ts-expect-error ts-migrate(6133) FIXME: 'index' is declared but its value is never read.
-    return Children.map(children, (child, index) => {
+    return Children.map(children, (child) => {
```
ke1k0, line 125: Just FYI: In my PR (https://github.com/instructure/instructure-ui/pull/881) I made a lot of changes in these 2 utils and had to refactor the codes in a lot of components. Depending which PR is merged first, we gonna have to resolve the conflicts here.

packages/ui-checkbox/src/CheckboxGroup/index.tsx
```diff
@@ -111,40 +109,46 @@ class CheckboxGroup extends Component<CheckboxGroupProps> {
...
+            ) > -1,
           onChange: this.handleChange,
-          // @ts-expect-error ts-migrate(2533) FIXME: Object is possibly 'null' or 'undefined'.
-          width: child.props.width || 'auto',
-          // @ts-expect-error ts-migrate(2339) FIXME: Property '_messagesId' does not exist on type 'Che... Remove this comment to see the full error message
-          'aria-describedby': this.hasMessages && this._messagesId
+          width:
+            (child as React.ComponentElement<CheckboxProps, Checkbox>).props
+              .width || 'auto',
+          'aria-describedby': this.hasMessages ? this._messagesId : ''
```
ke1k0, line 151: If you default to `''` does it still render on the DOM element? As far as I know if you default to `undefined` it doesn't render at all, not sure about `''`.

packages/ui-checkbox/src/Checkbox/index.tsx
```diff
@@ -74,23 +75,22 @@ class Checkbox extends Component<CheckboxProps> {
...
+      checked:
+        typeof props.checked === 'undefined'
+          ? !!props.defaultChecked
+          : undefined
     }
 
-    // @ts-expect-error ts-migrate(2339) FIXME: Property '_defaultId' does not exist on type 'Chec... Remove this comment to see the full error message
     this._defaultId = uid('Checkbox')
   }
+  private _defaultId: string
```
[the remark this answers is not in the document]
HerrTopi: Yeah, why not?

packages/ui-checkbox/src/Checkbox/index.tsx
```diff
@@ -99,28 +99,22 @@ class Checkbox extends Component<CheckboxProps> {
   }
 
   componentDidMount() {
-    // see https://github.com/facebook/react/issues/1798
-    // @ts-expect-error ts-migrate(2339) FIXME: Property '_input' does not exist on type 'Checkbox... Remove this comment to see the full error message
-    this._input.indeterminate = this.props.indeterminate
+    if (this._input) {
+      this._input.indeterminate = this.props.indeterminate || false
```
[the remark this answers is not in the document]
HerrTopi: Indeed. I just don't really like the warnings associated with the '!' :D

packages/ui-checkbox/src/CheckboxGroup/index.tsx
HerrTopi: It is only used in one place, so the simplification would only be "|| []"->"!". I think setting the defaultValue prop to a preset default (even if its an empty array) is counterintuitive, since it should be user set or undefined

packages/ui-checkbox/src/CheckboxGroup/index.tsx
HerrTopi: I'm not sure, but 'undefined' makes it clearer

packages/ui-checkbox/src/Checkbox/CheckboxFacade/props.ts
```diff
@@ -36,6 +36,9 @@ type CheckboxFacadeOwnProps = {
   focused?: boolean
   hovered?: boolean
   size?: 'small' | 'medium' | 'large'
+  /**
+   * Visual state showing that child checkboxes are a combination of checked and unchecked
```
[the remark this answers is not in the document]
HerrTopi: I think it's not really needed, but yeah, let's include for consistency's sake

packages/ui-checkbox/src/CheckboxGroup/index.tsx
ke1k0: I meant that if the `this.state.value` has a default of `[]`, then it can never be undefined, so there are some unnecessary checks in the code. For example the `get value() {}` getter can be refactored a bit and the first line of `handleChange` too.

ke1k0 (approved):

PR #879, opened 2022-01-18
Add deterministic Id generation

Brailor:
This PR aims to implement deterministic id generation for components.

The current way of (randomly generate ids) assigning ids to components is not suitable for doing server side rendering (SSR), because once we are rendering components in 2 places (server - client) the same React Component tree won't result in the same DOM structure, thus the [react hydration](https://reactjs.org/docs/react-dom.html#hydrate) process won't be able to identify the matching components.

We can achieve deterministic id generation by instead of randomly generate ids, we can instead keep track of each rendered instance and then hash the name of the instance with the count it has been rendered in the tree.

A new property on the `InstUISettingsProvider` component has been introduced called: `instanceMapCounter` (open to change) which enables to set the counter map from outside. In Next.js applications users will **have** to generate a new counter map for every request to keep the client and server side in sync.


docs/guides/server-side-rendering.md
```diff
@@ -0,0 +1,65 @@
+---
+title: Server side rendering (SSR)
+category: Guides
+order: 1
+---
+
+# SSR with Next.js
+
+This document will demonstrate how to achive Server Side Rendering (SSR) with [Next.js](https://nextjs.org/) using InstUI components.
```
matyasf, line 9: * achieve

docs/guides/server-side-rendering.md
```diff
@@ -0,0 +1,65 @@
...
+#### Initial steps
+
+- install InstUI related depenecies:
+
+```sh
+yarn add @instructure/emotion @instructure/ui-react-utils
+```
+
+> Note!
+> You should not install the InstUI meta pacakge(`@instructure/ui`) since it has components that are not yet Server Side Renderable out of the box.
```
matyasf, line 22: is this because of CodeEditor?

packages/ui-utils/src/hash.ts
```diff
@@ -64,6 +64,45 @@ function executeHash(input: string): string {
...
+function hashInstance(instance: string, map: Map<string, number>) {
+  if (!map.has(instance)) {
+    map.set(instance, 0)
+  } else {
+    map.set(instance, map.get(instance)! + 1)
+  }
+
+  const instanceCount = map.get(instance)!
+  const valueToHash = `${instance}_${instanceCount}`
+  const hashed = cyrb53(valueToHash)
```
matyasf, line 97: why do we need to hash this? Inst just `instanceName_count` enough to make it unique and deterministic?

packages/ui-utils/src/hash.ts
```diff
@@ -64,6 +64,45 @@ function executeHash(input: string): string {
...
+  h2 =
+    Math.imul(h2 ^ (h2 >>> 16), 2246822507) ^
+    Math.imul(h1 ^ (h1 >>> 13), 3266489909)
+
+  return (
+    (h2 >>> 0).toString(16).padStart(8, '0') +
+    (h1 >>> 0).toString(16).padStart(8, '0')
+  )
+}
+function hashInstance(instance: string, map: Map<string, number>) {
```
matyasf, line 88: Instead of passing a Map here, isn't it enough to pass the instance type? Then keep here a static map that keeps count of things, and this function could return the current instance number/name

Brailor: Sadly that is not enough because it will result in the client and server becoming out of sync the time a browser refresh happens.
This code will run in 2 environments in case of SSR.
1. server side
2. client side

At the beginning of everything the server and client will be sync.
Now imagine the following scenario, we have an app containing only a `Button` component:
1. server renders the component ->  server :`Button` counter: 1
2. client picks up and hydrates the app, thus rendering the same `Button` -> client: `Button` counter: 1
3. client does a whole browser refresh -> every static file (including .js) will have to be re-requested, parsed, and run and that means the `Button` counter will be again 1
4. meanwhile on the server we didn't do a restart meaning the server side counter will be: 2
5. client and server become out of sync

So in order to be sync all the time we have to reset the counter on each request on the server side, and I think this is the most elegant way to do that. (using `React.Context`)

packages/ui-expandable/src/Expandable/index.tsx
```diff
@@ -22,22 +22,25 @@
...
 const toggleExpanded = ({ expanded }: { expanded: boolean }) => ({
   expanded: !expanded
 })
 
+//@ts-expect-error check out this later
+@withSSR()
 /**
 ---
-category: components/utilities
+category: components
```
ke1k0, line 40: Why was the category changed?

packages/ui-expandable/src/Expandable/index.tsx
```diff
@@ -22,22 +22,25 @@
...
 import type { ExpandableProps, ExpandableState } from './props'
 
+import { withSSR } from '@instructure/ui-react-utils'
+
 const toggleExpanded = ({ expanded }: { expanded: boolean }) => ({
   expanded: !expanded
 })
 
+//@ts-expect-error check out this later
+@withSSR()
```
ke1k0, line 37: These decorators usually come between the annotation and the class declaration, don't they?

packages/ui-menu/src/Menu/index.tsx
```diff
@@ -54,11 +55,12 @@ category: components
 ---
 @tsProps
 **/
+@withSSR()
 @withStyle(generateStyle, generateComponentTheme)
 @testable()
 class Menu extends Component<MenuProps> {
   static readonly componentId = 'Menu'
-
+  // static contextType = SSRContext
```
ke1k0, line 63: these was probably left here by accident

packages/ui-react-utils/src/SSRContext/withSSR.tsx
```diff
@@ -0,0 +1,62 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+import React, { ComponentClass, forwardRef, useContext } from 'react'
+import hoistNonReactStatics from 'hoist-non-react-statics'
+
+import { SSRContext } from './SSRContext'
+import { decorator } from '@instructure/ui-decorator'
+
+const withSSR = decorator((ComposedComponent: ComponentClass) => {
```
ke1k0, line 30: I think we should add some description for it so that it appears on the docs page. Might be useful to give some explanation and guide on why it was made and how it is used (for further InstUI team generations and for anyone interested).

ke1k0: Also, should/could we write some tests for it?

packages/ui-utils/src/hash.ts
matyasf: thanks for the explanation!

packages/ui-utils/src/hash.ts
Brailor: You are correct, I've removed it, thanks!

docs/guides/server-side-rendering.md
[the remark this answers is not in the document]
Brailor: Changed, thanks!

docs/guides/server-side-rendering.md
```diff
@@ -0,0 +1,65 @@
...
+>
+> You should install only the packages you need!
+
+- in your Next.js application create - if it does not already exists - a file named `_app.js` inside the `pages` directory. This is a special file in Next.js because it allows you to override/control component initilaztion. Read more about it in the [Next.js docs](https://nextjs.org/docs/advanced-features/custom-app).
```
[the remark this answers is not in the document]
Brailor: Changed, thanks!

docs/guides/server-side-rendering.md
```diff
@@ -0,0 +1,65 @@
...
+}
+```
+
+That is it! You should be ready to go!
+
+#### Notes explaining what is going on with the above code
+
+Keeping track of instances is a way we achive deterministic `id` generation for components. Since components are being rendered on 2 places (server - client), it is crucial to have the same `ids` attached to the same instances on the client as on the server side.
+
+To achive deterministic `id` generation InstUI internally will keep track of every component instance, and will use this information to hash the instance name and the instance count to generate `ids` deterministically.
```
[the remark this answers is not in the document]
Brailor: Changed, thanks!

packages/ui-alerts/src/Alert/index.tsx
```diff
@@ -166,7 +167,7 @@ class Alert extends Component<AlertProps> {
     const liveRegion = this.getLiveRegion()
     if (liveRegion) {
       // @ts-expect-error ts-migrate(2339) FIXME: Property 'srid' does not exist on type 'Alert'.
-      this.srid = uid('Alert')
+      this.srid = hashInstance('Alert', this.props.instanceMapCounter)
```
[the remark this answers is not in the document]
Brailor: If we do that, will be the `instanceMapCounter` being shown to the end user as a suggestion?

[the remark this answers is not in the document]
ke1k0: It will, but so do `styles` and `makeStyles` too. But that's why I put 2 warnings in withStyle as well to warn ppl that it shouldn't be used directly :) https://github.com/instructure/instructure-ui/blob/a2cfa0cd8bf43d6072f1b352bbd6e43542081be8/packages/emotion/src/withStyle.tsx#L167

docs/guides/server-side-rendering.md
Brailor: Fixed

docs/guides/server-side-rendering.md
Brailor: Added note about CodeEditor

packages/ui-expandable/src/Expandable/index.tsx
Brailor: By mistake, fixed

packages/ui-expandable/src/Expandable/index.tsx
Brailor: Yeah, right, fixed it

packages/ui-react-utils/src/SSRContext/withSSR.tsx
Brailor: Added description with tests

packages/ui-alerts/src/Alert/index.tsx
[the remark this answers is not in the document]
Brailor: Thanks, done pretty much the same

packages/ui-file-drop/src/FileDrop/index.tsx
```diff
@@ -102,10 +103,13 @@ class FileDrop extends Component<FileDropProps, FileDropState> {
...
     super(props)
     // @ts-expect-error ts-migrate(2322) FIXME: Type 'string' is not assignable to type 'null'.
-    this.defaultId = uid('FileDrop')
-    ;(this as any).messagesId = uid('FileDrop-messages')
+    this.defaultId = props.deterministicId!
+    this.messagesId = generateId(
+      'FileDrop-messages',
+      //@ts-expect-error props.instanceMapCounter
+      props.instanceMapCounter
+    )
```
ke1k0, line 111: I'm a bit confused about how this util is used now. I some components you use `generateId` and some other places the `this.props.deterministicId` and here it is both :) I think I get it, the `deterministicId` is supposed to give back the id with the componentId and you can generate extra id-s with the other util? 

How about making this prop accept a function too? So by default `props.deterministicId =>  'FileDrop_1'`, but it would also accept `props.deterministicId('FileDrop-messages') =>  'FileDrop-messages_1'`? If you have a better idea, that is fine too, just would be nice to clear the confusion and the 2 different usages :)

packages/ui-react-utils/src/DeterministicIdContext/DeterministicIdProvider.tsx
```diff
@@ -0,0 +1,54 @@
...
+}>
+
+const defaultContextValue = generateInstanceMapCounter()
+
+/**
+ * ---
+ * category: components/utilities
+ * ---
+ * This is utility component for wrapping components with `DeterministicIdContext.Provider`
+ * See detailed documentation about how to use it: [InstUISettingsProvider](https://instructure.design/#InstUISettingsProvider)
```
ke1k0, line 39:
```suggestion
 * See detailed documentation about how to use it: [InstUISettingsProvider](/#InstUISettingsProvider)
```
Please use relative links, it through me to the released site when I tested it :)

ke1k0: Also, please add an example to the InstUISettingsProvider page about the usage if you say that it has detailed information :)

packages/ui-react-utils/src/DeterministicIdContext/withDeterministicId.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { generateId } from '@instructure/ui-utils'
+
+import type { DeterministicIdProviderValue } from './DeterministicIdProvider'
+import type { InstUIComponent } from '@instructure/shared-types'
+import { warn } from '@instructure/console'
+
+type WithDeterministicIdProps = {
+  instanceMapCounter?: DeterministicIdProviderValue
+  deterministicId?: string
+}
```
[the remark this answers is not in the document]
Brailor: Yes, technically they are always there (since the decorator provides them), but if I change it to non-optional, then on the user side it will complain if the user does not provide those props to the component.
That is why it would be a good idea to have 2 separate props type, one for our internal usage and one for the user facing props.

packages/ui-file-drop/src/FileDrop/index.tsx
Brailor: The idea was that the default behaviour would be passed down as a simple prop (`deterministicId`) when the component would just use essentially it's componentId for id generation.
For every other cases it would have control over how to generate id.

But I agree this is not the most elegant way of doing it.
Sadly I don't see how you suggestion might work, how to pass down a string in 1 case, and a function in another?
What about `generateId` would passed down as a function and it would accept an optional parameter - which would be defaulted to `instanceName`=`component.componentId` - and it would generate based on this parameter?

packages/ui-react-utils/src/DeterministicIdContext/withDeterministicId.tsx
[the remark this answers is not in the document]
ke1k0: Oooooh you are right, forgot about that! Nvm then :) Having to props type sound interesting, although I'm not sure how that would work.. might be worth investigating.

packages/ui-file-drop/src/FileDrop/index.tsx
ke1k0: Sure that sounds good too!

packages/ui-react-utils/src/DeterministicIdContext/DeterministicIdProvider.tsx
Brailor: I completely forgot about the relative links, thanks!
I made the suggested changes.

packages/ui-file-drop/src/FileDrop/index.tsx
Brailor: Great, should be fixed then!

packages/emotion/src/InstUISettingsProvider/index.tsx
```diff
@@ -111,17 +119,20 @@ function InstUISettingsProvider({
   }
 
   return (
-    <ThemeProvider theme={getTheme(theme)}>
-      <TextDirectionContext.Provider value={finalDir}>
-        {children}
-      </TextDirectionContext.Provider>
-    </ThemeProvider>
+    <DeterministicIdContextProvider instanceMapCounter={instanceMapCounter!}>
```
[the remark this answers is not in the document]
Brailor: Yeah, I think I misunderstood your first comment, now it should be correct.

docs/guides/server-side-rendering.md
```diff
@@ -0,0 +1,65 @@
...
+export default function MyApp(props) {
+  const { Component, pageProps } = props;
+
+  return (
+    <>
+      <Head>
+        <meta name="viewport" content="initial-scale=1, width=device-width" />
+      </Head>
+                                               {/* This is the important step */}
+      <InstUISettingsProvider instanceMapCounter={generateInstanceMapCounter()}>
```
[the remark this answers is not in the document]
Brailor: Right, changed it everywhere, thanks.

packages/shared-types/src/CommonTypes.ts
```diff
@@ -39,3 +39,10 @@ export type UIElement =
  * screen reader
  */
 export type LiveRegion = (() => Element) | Element[] | Element
+
+interface InstUIBaseComponent {
+  componentId: string
+  allowedProps: string[]
+}
+export interface InstUIComponent extends React.ComponentClass<any, any>, InstUIBaseComponent { }
```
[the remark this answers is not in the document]
Brailor: Hmm, I added it, but now I'm not even sure this is a correct thing to do, since `decorator` expects a plain `React.ComponentClass` which won't have these `componentId` and `allowedProps` static variables on them, so we have to cast every time we use this inside on our decorator implementations.

It would be nice to somehow tell the users with typings that they *need* to have these specific static variables on their components in order to use `withStyle, withDeterministicId`.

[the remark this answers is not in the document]
ke1k0: Isn't the issue simply that the extra props are all required? If I set them as optional, I don't get the errors, because then it is compatible with ComponentClass.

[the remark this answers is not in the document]
Brailor: Yeah, changing them to be optional solves the issue with the compiler, but I have a feeling this is just a workaround for now.
Maybe our component should extend the `InstUIComponent` interface instead of `React.Component`, but then again the decorators must be general enough to accept any kind of `React.ComponentClass` so I'm not sure to be honest. :D
Anyways, changed the properties to be optional.

docs/guides/server-side-rendering.md
```diff
@@ -0,0 +1,65 @@
...
+- install InstUI related dependencies:
+
+```sh
+yarn add @instructure/emotion @instructure/ui-react-utils
+```
+
+> Note!
+> You should not install the InstUI meta package(`@instructure/ui`) since it has components that are not yet Server Side Renderable out of the box.
+> The `@instructure/ui-code-editor` currently relies on browser specific APIs and requires that your choice of bundler/transpiler to touch the source code of this component, so effectively it is not in a usable state for out of box usage with Next.js.
+> You should install only the packages you need!
```
matyasf, line 24: Didn't we decided to fix CodeEditor (and accept that it might break Jest tests)?

packages/ui-react-utils/src/DeterministicIdContext/withDeterministicId.tsx
```diff
@@ -0,0 +1,101 @@
...
+type WithDeterministicIdProps = {
+  deterministicId?: (instanceName?: string) => string
+}
+/**
+ * This decorator is used to enable the decorated class to use the `DeterministicIdContext` which is needed
+ * for deterministic id generation.
+ *
+ * The context is there for the users to pass an `instanceCounterMap` Map which is then used
+ * in the child components to deterministically create ids for them based on the `instanceCounterMap`.
+ * Read more about it here: [SSR guide](/#server-side-rendering)
```
matyasf, line 48: I'd put here the whole link, otherwise it will break when opened via an IDE popup

matyasf (approved):
looks good!

ke1k0 (approved):

docs/guides/server-side-rendering.md
Brailor: Yes, I removed it

packages/ui-react-utils/src/DeterministicIdContext/withDeterministicId.tsx
Brailor: Sure, added the whole link

PR #881, opened 2022-01-19
Modify type for matchComponentTypes and safeCloneElement

ke1k0:


packages/ui-react-utils/src/safeCloneElement.ts
```diff
@@ -36,11 +36,9 @@ import { createChainedFunction } from '@instructure/ui-utils'
  * @param props Props of the element
  * @param children
  */
-function safeCloneElement<
-  P extends Record<string, any> & { style?: any } & ClassAttributes<P>
->(
+function safeCloneElement<P extends Record<string, any> & { style?: any }>(
   element: ReactElement<P> & { ref?: any },
-  props: P,
+  props: P & { ref?: any } & ClassAttributes<any>,
```
matyasf, line 41: Instead of `ClassAttributes` you can use here `Attributes`, since you define `ref` as `any`:
`props: P & { ref?: any } & Attributes`

packages/ui-react-utils/src/matchComponentTypes.ts
```diff
@@ -37,10 +37,10 @@ import { ComponentType, ReactElement, ReactNode } from 'react'
  * @param {Array} types an array of React components
  * @returns {Boolean} true if the component matches at least one of the types
  */
-function matchComponentTypes(
+function matchComponentTypes<Type extends ReactElement = ReactElement>(
   componentInstance: ReactNode,
   types: string[] | ComponentType[] | ComponentType<any>[] = []
-) {
+): componentInstance is Type {
```
matyasf, line 43: nice idea :)

matyasf (changes requested):
looks good except for the small fix in safeCloneElement

packages/ui-react-utils/src/safeCloneElement.ts
ke1k0: @matyasf  I ended up refactoring it again, and I think this time I got it finally :D I managed to infer the props and ref types from the element, and can finally return the original element type too.... Can you please re-check it?

matyasf (approved):
looks good!

packages/ui-react-utils/src/safeCloneElement.ts
```diff
@@ -22,11 +22,27 @@
...
 
+type GetProps<E extends ReactElement = ReactElement> =
+  E extends FunctionComponentElement<infer P>
+    ? P & { ref?: E['ref'] }
+    : E extends ComponentElement<infer P, any>
+    ? P & { ref?: E['ref'] }
+    : E extends DOMElement<infer P, any>
+    ? P & { ref?: E['ref'] }
+    : Record<string, any> & { ref?: any }
+
```
Brailor, line 45: I haven't tried it out (and this is just nit picking really) but can't we simplify this type expression to something like this?:

```ts
type GetProps<E extends ReactElement = ReactElement> =
  E extends
  | FunctionComponentElement<infer P>
  | ComponentElement<infer P, any>
  | DOMElement<infer P, any>
  ? P & { ref?: E['ref'] }
  : Record<string, any> & { ref?: any }
```

ke1k0: Yeah, I remember experimenting with sth like this but didn't work for some reason... now I tried it and worked fine :D Se I changed it, thanks!

Brailor (approved):

PR #882, opened 2022-01-21
INSTUI-3388 - changes in options for drilldown

ke1k0:


packages/ui-options/src/Options/Item/index.tsx
```diff
@@ -112,6 +114,9 @@ class Item extends Component<OptionsItemProps> {
     const ElementType = getElementType(Item, this.props, () => as!)
     const passthroughProps = omitProps(this.props, Item.allowedProps)
 
+    const childrenContent = callRenderProp(children)
```
ke1k0, line 117: This was actually a bugfix, because the children prop allows a function but we just passed the function as children and it broke. It probably never came up as an issue since no one was using functions for the children prop.

packages/ui-options/src/Options/README.md
```diff
@@ -130,7 +130,11 @@ class Example extends React.Component {
         width="300px"
         shadow="above"
       >
-        <Options onKeyDown={this.handleKeyDown} tabIndex="0">
+        <Options
+          onKeyDown={this.handleKeyDown}
+          onMouseOut={(e) => this.setState({ highlighted: -1 })}
+          tabIndex="0"
+        >
```
ke1k0, line 137: Just added this logic to the docs examples because the hover state was stuck if you hovered over it once.

Brailor (approved):

HerrTopi (approved):

PR #883, opened 2022-01-25
fix(emotion): make theme componentOverrides apply to Buttons

ke1k0:
The `componentOverrides` block of `InstUISettingsProvider` theme prop was not applying to Button and
other button subtypes. This happened because most of the Buttons don't have styling on their own and
just pass the themeOverrides to the underlying BaseButton component. This worked well with the local
`themeOverride={}` prop, but the `withStyle` decorator never passed them to the component. This
wasn't needed most of the components that use the theme in their style generation, but is needed for
components that just pass the themeOverrides on.

### DIFF
--- a/packages/emotion/src/withStyle.tsx
+++ b/packages/emotion/src/withStyle.tsx
@@ -220,6 +220,10 @@ const withStyle = decorator(
           {...props}
           makeStyles={makeStyleHandler}
           styles={styles}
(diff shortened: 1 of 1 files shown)


HerrTopi (approved):
Whoopsie, it was an oversight from our side

Brailor (approved):

PR #886, opened 2022-01-26
feat(emotion): themeOverride prop to accept function

ke1k0:
The `themeOverride` prop now accepts a function with the current component theme as the parameter.


Brailor (approved):
Looks good!
But we seriously need separate documentation for the theme overrides (user facing) and how __we__ are using emotion to style our components (internal components), because it is indeed hard to distinguish between the 2 in the current form.

matyasf (approved):

HerrTopi (approved):

docs/guides/using-theme-overrides.md
```diff
@@ -0,0 +1,418 @@
+---
+title: Using theme overrides
+category: Guides
+order: 4
+---
+
+## Using theme overrides
+
```
matyasf, line 8: looks nice! I'd add some text to the "theming" page too that links here. I'd also add here some introductory text what this document is about, something like:

This document details how you can customize the Instructure UI components by tweaking their theme variables. While this gives you a level of flexibility on the look and feel of the components you should be aware of 2 things:
- The default theme variables are tested to have high enough contrast ratios for WCAG conformance, if you are making changes it's you job, that your app stays WCAG conformant.
- You cannot fully customize the look of components, just to a certain degree. This is intentional, Instructure UI is a design system geared towards the Canvas look and feel, not a generic UI component library.

matyasf (changes requested):
Looks good! I'd add some extra text to the "Themes" page, I think its the first page users will see when they look for infos about theming. Something like this:

Instructure UI ships with 3 built-in themes ([instructure], [canvas], [canvas-high-contrast]). You can choose which theme to use with [InstUISettingsProvider]. You can read about how to customize themes or apply different ones to parts of you application in [Using Theme Overrides].
Finally if you are interested in depth on how InstUI's theming engine works and how can you make your own components that use it read the documentation of the [emotion] package.

docs/guides/using-theme-overrides.md
ke1k0: I added the info, have a look again please :)

matyasf (approved):
looks good!

PR #888, opened 2022-01-27
chore: update TS version to `4.5.5`

Brailor:
Closes: INSTUI-3395


packages/ui-file-drop/src/FileDrop/index.tsx
```diff
@@ -169,7 +169,7 @@ class FileDrop extends Component<FileDropProps, FileDropState> {
       list = (this.props as any).shouldAllowMultiple ? list : [list[0]]
     }
     if (shouldEnablePreview) {
-      return list.map((file) =>
+      return list.map((file: any) =>
```
Brailor, line 172: This threw a new error, but since this component is not typed I just left it as `any`

ke1k0 (approved):

PR #889, opened 2022-01-27
INSTUI-3387 ts type Method, Nav, NavToggle

NamNguyenInstructure:


packages/__docs__/src/Methods/index.tsx
```diff
@@ -23,17 +23,15 @@
...
 import { compileMarkdown } from '../compileMarkdown'
-
-class Methods extends Component {
-  static propTypes = {
-    methods: PropTypes.array.isRequired
-  }
-
+import type { MethodsProps } from './props'
+import { propTypes, allowedProps } from './props'
+class Methods extends Component<MethodsProps> {
```
ke1k0, line 32: Please leave an empty line above the class declaration for better readability :)

packages/__docs__/src/Nav/props.ts
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+import type { PropValidators } from '@instructure/shared-types'
+import PropTypes from 'prop-types'
+type NavOwnProps = {
+  docs: Record<string, any>
+  sections: Record<string, any>
+  themes?: Record<string, any>
+  icons?: Record<string, any>
+  selected?: string
```
ke1k0, line 32: Actually these could be typed a bit more precisely. As far as I can tell from the code, the Nav gets its props from the `docs-data.json` (it is in the `packages/__docs__/__build__` folder). If you copy the code from the json to a json formatter (e.g.: https://codebeautify.org/jsonviewer) then you can browse the tree easily and you can see how the data looks like. Question is: is it worth typing this json? Let's talk about it on the stand-up!

packages/__docs__/src/Nav/index.tsx
```diff
@@ -34,9 +33,11 @@ import { View } from '@instructure/ui-view'
 import { capitalizeFirstLetter } from '@instructure/ui-utils'
 
 import { NavToggle } from '../NavToggle'
-
-class Nav extends Component {
-  constructor(props) {
+import type { NavProps, NavState } from './props'
+class Nav extends Component<NavProps, NavState> {
```
ke1k0, line 37: Please leave an empty line above the class declaration for better readability :)

packages/__docs__/src/NavToggle/index.tsx
```diff
@@ -23,32 +23,29 @@
...
-class NavToggle extends Component {
-  static propTypes = {
-    summary: PropTypes.string.isRequired,
-    variant: PropTypes.oneOf(['section', 'category']),
-    children: PropTypes.node
-  }
-
+import type { NavToggleProps } from './props'
+import { propTypes, allowedProps } from './props'
+class NavToggle extends Component<NavToggleProps> {
```
ke1k0, line 34: Please leave an empty line above the class declaration for better readability :)

packages/__docs__/src/NavToggle/index.tsx
```diff
@@ -23,32 +23,29 @@
...
+  static propTypes = propTypes
+  static allowedProps = allowedProps
   static defaultProps = {
     variant: 'section',
     children: null
   }
+  private _toggle: any
```
ke1k0, line 41: You can tell its type from the ref: 
```suggestion
  private _toggle: ToggleDetails | null = null
```

And in the render method:

```js
       <ToggleDetails
          ref={(c: ToggleDetails) => {
            this._toggle = c
          }}
          fluidWidth
          {...props}
          summary={summaryContent}
        />
```

packages/__docs__/src/NavToggle/index.tsx
```diff
@@ -23,32 +23,29 @@
...
 
   focus() {
     this._toggle.focus()
   }
 
   render() {
-    const { summary, variant, ...props } = this.props
+    const { summary, variant, iconPosition, ...props } = this.props
```
ke1k0, line 48: You shouldn't remove `iconPosition` from `...props`, because it has to be passed to the `ToggleDetails`. It complains that the props are not compatible, but that's because you have to set NavToggle's prop to the same type as ToggleDetail's `iconPosition`prop.

packages/__docs__/src/NavToggle/props.ts
```diff
@@ -0,0 +1,65 @@
...
+import type { PropValidators } from '@instructure/shared-types'
+import PropTypes from 'prop-types'
+
+type NavToggleOwnProps = {
+  summary: string
+  variant?: 'section' | 'category'
+  children?: React.ReactNode
+  onToggle?: (...args: any[]) => any
+  expanded?: boolean
+  iconPosition?: string
```
ke1k0, line 34: It has to be the same type as `ToggleDetails`'s  `iconPosition` prop.

packages/__docs__/src/NavToggle/props.ts
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+import type { PropValidators } from '@instructure/shared-types'
+import PropTypes from 'prop-types'
+
+type NavToggleOwnProps = {
+  summary: string
+  variant?: 'section' | 'category'
+  children?: React.ReactNode
+  onToggle?: (...args: any[]) => any
```
ke1k0, line 32: It has to be the same type as `ToggleDetails`'s `onToggle` prop. Actually it is easier if you set all props that are passed to ToggleDetails as e.g.: `ToggleDetailsProps['onToggle']` so they will stay in sync.

ke1k0 (changes requested):

packages/__docs__/src/NavToggle/index.tsx
NamNguyenInstructure: can you explain about `null = null` part?

ke1k0: So the type of _toggle is **either** `ToggleDetails` **or** `null`:
```js
private _toggle: ToggleDetails | null
```

But because it is either one of these types, it cannot be `undefined`, and it is undefined if you don't give it an initial value. 
```js
private _toggle // = undefined
```

And since it cannot initially be `ToggleDetails` type, we set it to null
```js
private _toggle: ToggleDetails | null = null
```

ke1k0 (approved):

HerrTopi (approved):

PR #892, opened 2022-01-31
fix(ui-code-editor): change dynamically imported `CodeMirror` language modes

Brailor:
Closes: INSTUI-3394

our `CodeEditor` component dynamically imported `CodeMirror` language options
which was not
transpiled at all by `babel` which ment that we shipped
code that is not usable in browser
environments out of the box.
this change removes those dynamic `require`s and instead statically
imports them.
this might cause issues for users that uses this component and use jest in order to
test their components
I added notes on how to fix those jest related issues


ke1k0 (approved):
I suggest merging this PR right before a release so we don't forget to add the info to the release notes :)

PR #893, opened 2022-01-31
fix(ui-portal): fix `Portal` not being SSR-able

Brailor:
`Portal` tried to access the `document` in its lifecylce methods, which is not suitalbe in SSR, so I
added a check in every lifecycle to bail out if we are in a server environment.

### DIFF
--- a/packages/ui-portal/src/Portal/index.tsx
+++ b/packages/ui-portal/src/Portal/index.tsx
@@ -27,6 +27,7 @@ import ReactDOM from 'react-dom'
 
 import { passthroughProps } from '@instructure/ui-react-utils'
 import { textDirectionContextConsumer } from '@instructure/ui-i18n'
+import { canUseDOM } from '@instructure/ui-dom-utils'
 
 import { propTypes, allowedProps } from './props'
 import type { PortalNode, PortalProps, PortalState } from './props'
@@ -55,6 +56,10 @@ class Portal extends Component<PortalProps, PortalState> {
   constructor(props: PortalProps) {
     super(props)
 
+    if (!canUseDOM) {
+      return
+    }
+
     this.state = {
       mountNode: this.findMountNode(props)
     }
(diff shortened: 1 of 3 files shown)


matyasf (approved):
looks good, just a few concerns:
- could this cause a mismatch in the deterministic ID generation? e.g. something renders in the browser and gets the ID incremented, while on the server render is not called, thus IDs wont match?
- There are a few other instances where we access the DOM, should not these be fixed too? What I've found: `Alert.createScreenreaderAlert()`, `ui-dom-utils/findFocusable/visible(), isVisible.ts, ui-truncate-text/measureText()`

Brailor:
> looks good, just a few concerns:
> 
> * could this cause a mismatch in the deterministic ID generation? e.g. something renders in the browser and gets the ID incremented, while on the server render is not called, thus IDs wont match?
> * There are a few other instances where we access the DOM, should not these be fixed too? What I've found: `Alert.createScreenreaderAlert()`, `ui-dom-utils/findFocusable/visible(), isVisible.ts, ui-truncate-text/measureText()`

Regarding concern 1:
- It is possible to construct such prop combination where this could cause issues, eg:
```jsx
<Portal open>
 <Some-Component-That-Uses-IDs />
</Portal>
```
- the issue here is that the `Portal` will try to render it's children, but because on the server side that would be an error we bail out, but on the client this is legit code, thus mismatch could happen
- I'm not sure how to fix this

2:
- we should look into those, to be honest I didn't test every component, with every prop combination, that is a little bit more work

packages/ui-portal/src/Portal/index.tsx
```diff
@@ -55,6 +56,10 @@ class Portal extends Component<PortalProps, PortalState> {
   constructor(props: PortalProps) {
     super(props)
 
+    if (!canUseDOM) {
+      return
+    }
+
```
ke1k0, line 62: It probably would be nice to throw some kind of warning here so they know why Portals (and Modals, Popovers, Tooltips etc...) don't render. Also might be worth adding info about it in the SSR docs.

Brailor: I added it to the guide: 
https://github.com/instructure/instructure-ui/pull/879/files#diff-663a30594b4684ce916d76763d1f44e4aeeae00840593c32ad246483435af7a3R21-R24

Brailor: Also added the warning to the render method instead of the constructor.

ke1k0 (approved):

PR #896, opened 2022-02-02
INSTUI-3402: Docs: Fix Codepen links

matyasf:
The issue is that if we send layout="top" to Codepen it ignores every option and just shows an empty codepen. Remove this option and add escape quotes because Codepen docs recommend it.

### DIFF
--- a/packages/__docs__/src/CodePenButton/index.tsx
+++ b/packages/__docs__/src/CodePenButton/index.tsx
@@ -47,21 +47,21 @@ class CodePenButton extends Component<CodePenButtonProps> {
       js: `const render = (el) => { ReactDOM.render(el, document.getElementById('app')) }\n\n${code}`,
       private: true,
       editors: '001',
-      html:
-        '<div id="app"></div><div id="flash-messages"></div><div id="nav"></div>',
-      layout: 'top',
+      html: '<div id="app"></div><div id="flash-messages"></div><div id="nav"></div>',
       css_prefix: 'autoprefixer',
       js_pre_processor: 'babel',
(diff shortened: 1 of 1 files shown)


Brailor (commented):
Something is still not correct, now I can see the loaded source code inside codepen, but the examples are not rendered:
Screenshot 2022-02-02 at 11 57 47
Screenshot 2022-02-02 at 11 57 58

I tried it with different browsers and it is the same everywhere.

Brailor:
> Something is still not correct, now I can see the loaded source code inside codepen, but the examples are not rendered: Screenshot 2022-02-02 at 11 57 47 Screenshot 2022-02-02 at 11 57 58
> 
> I tried it with different browsers and it is the same everywhere.

Hmm, after I logged in to CodePen it is working fine
Screenshot 2022-02-02 at 12 06 27

Did we have to be logged in before? I think that is not something that was required and if it is now, maybe we should move away from CodePen?

matyasf:
hmm something is really off with Codepen. It works when I'm logged out in Chrome. It does not in FF and Chrome incognito. Also in these I cannot create a new pen... maybe we should move away from Codepen?

ke1k0:
Can confirm... Same happens for me too, it loads well now if I am/was logged in, but not in incognito and other browsers where I've never logged in before.

Might it be a "feature not bug" from codepen's point of view? They might be trying to trick ppl into making accounts like this... or it's just a bug. Before we change anything I would give codepen a bug ticket first because this feature used to work before, right?

ke1k0 (approved):

PR #897, opened 2022-02-02
feat(ui-select): add `optionsMaxHeight` prop to Select

ke1k0:
The new prop controls the maxHeight of the options list and if set, overrides the
`visibleOptionsCount` prop.

Closes: INSTUI-3405


matyasf (approved):
looks good!

PR #906, opened 2022-02-05
INSTUI 3374 upgrade dependencies and remove unused ones

matyasf:
also simplify the clean script, now its 10x faster


packages/ui-a11y-content/src/AccessibleContent/index.tsx
```diff
@@ -55,7 +55,8 @@ class AccessibleContent extends Component<AccessibleContentProps> {
 
   render() {
     const { alt, children, ...props } = this.props
-    const ElementType = getElementType(AccessibleContent, this.props)
+    // any cast is needed to prevent Expression produces a union type that is too complex to represent errors
+    const ElementType = getElementType(AccessibleContent, this.props) as any
```
matyasf, line 59: i have not idea why these errors kept popping up, when I added a `@ts-expect-error` another popped up. And I did not upgrade TypeScript... I guess these super complex props really push TS to the limits.

Brailor: Yeah, I encountered the same thing the other day, somehow we have to fix this type.

packages/command-utils/package.json
```diff
@@ -29,6 +29,6 @@
   },
   "sideEffects": false,
   "devDependencies": {
-    "@types/which": "^2.0.0"
+    "@types/which": "^2"
```
matyasf, line 32: I've made these type of changes so `yarn upgrade-interactive` shows minor version changes. Also this is a more precise notation of what dependency we require, e.g. in most cases we are OK with someLibrary version 2.x, not just someLibrary version greater than 2.3.9, less than 3

packages/ui-scripts/src/build/clean.ts
```diff
@@ -22,21 +22,22 @@
...
+    'lib',
+    'tokens',
+    '.babel-cache',
+    '.cache',
+    'types'
+  ]
+  for (const dir of dirs) {
+    if (fs.existsSync(dir)) {
+      fs.rmSync(dir, { force: true, recursive: true, maxRetries: 2 })
+    }
```
matyasf, line 41: an optimization I wanted to do since a long time. Does the same thing, does not use a dependency and is 10x faster

HerrTopi (approved):
Seems fine to me and seems to be working. Great job

packages/ui-scripts/src/build/clean.ts
Brailor: Great! Although we don't use this script anymore to clean packages, it is great to remove unneeded deps.

Brailor (approved):

PR #911, opened 2022-02-08
INSTUI 3398 ts type params playground

NamNguyenInstructure:


packages/__docs__/src/Document/index.tsx
```diff
@@ -44,6 +44,7 @@ import { ComponentTheme } from '../ComponentTheme'
 import { Heading } from '../Heading'
 import { propTypes, allowedProps } from './props'
 import type { DocumentProps, DocType, SingleChildrenType } from './props'
+import type { ViewOwnProps } from '@instructure/ui-view'
```
ke1k0, line 47: please put an empty line between the imports and `@withStyle`

packages/__docs__/src/Document/index.tsx
```diff
@@ -71,7 +72,10 @@ class Document extends Component<
     this.props.makeStyles?.()
   }
 
-  handleDetailsTabChange = (_event: Event, { index }: any) => {
+  handleDetailsTabChange = (
+    _event: React.MouseEvent<ViewOwnProps> | React.KeyboardEvent<ViewOwnProps>,
+    { index }: any
+  ) => {
```
ke1k0, line 78: Since this has to be the exact same as the `onRequestTabChange` prop of Tabs, you can use it's type:

```js
handleDetailsTabChange: TabsProps['onRequestTabChange'] = (
    _event,
    { index }
  ) => {
    this.setState({
      selectedDetailsTabIndex: index
    })
  }
```

packages/__docs__/src/Params/index.tsx
```diff
@@ -23,17 +23,14 @@
...
-
-class Params extends Component {
-  static propTypes = {
-    params: PropTypes.array.isRequired,
-    layout: PropTypes.string
-  }
+import type { ParamsProps } from './props'
+import { propTypes } from './props'
+class Params extends Component<ParamsProps> {
+  static propTypes = propTypes
```
ke1k0, line 33: Empty line please

```suggestion
import type { ParamsProps } from './props'
import { propTypes } from './props'

class Params extends Component<ParamsProps> {
  static propTypes = propTypes
```

packages/__docs__/src/Params/props.ts
```diff
@@ -0,0 +1,46 @@
...
+import React from 'react'
+
+type ParamsOwnProp = {
+  params: {
+    name: string
+    type: { name: string }
+    defaultvalue: string
+    description: string
+  }[]
+  layout?: string
```
ke1k0, line 35: Actually this can be traced back all the way to the App component. The `layout` state variable is set here, then passed to Document, down to Params. The `updateLayout` method sets this, so you can see what values it can get: 

```js
updateLayout = (matches) => {
    let layout = 'small'

    if (matches.length > 0) {
      if (matches.includes('medium') && matches.length === 1) {
        layout = 'medium'
      } else if (matches.includes('large') && matches.length === 2) {
        layout = 'large'
      } else if (matches.includes('x-large')) {
        layout = 'x-large'
      }
    }

    this.setState({ layout })
  }
```

So you know that layout is one of these: 
```js
layout?: 'small' | 'medium' | 'large' | 'x-large'
````

File: packages/__docs__/src/Params/props.ts [hunk 5]
```diff
@@ -0,0 +1,46 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+import type { PropValidators } from '@instructure/shared-types'
+import PropTypes from 'prop-types'
+import React from 'react'
+
+type ParamsOwnProp = {
+  params: {
```
ke1k0, line 29: This is the `params` prop of the `doc` that gets passed, so you can use that type here: `params: DocType['params']`

packages/__docs__/src/Params/props.ts
```diff
@@ -0,0 +1,46 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+import type { PropValidators } from '@instructure/shared-types'
+import PropTypes from 'prop-types'
+import React from 'react'
```
ke1k0, line 26: This can be removed, unused import

packages/__docs__/src/Playground/index.tsx
```diff
@@ -58,26 +59,10 @@ const fullScreenIconPath = (
...
-      'checkerboard-inverse',
-      'light',
-      'inverse',
-      'none'
-    ]),
-    readOnly: PropTypes.bool
-  }
-
+class Playground extends Component<PlaygroundProps, PlaygroundState> {
+  _fullScreenButton: IconButton | null | undefined
```
ke1k0, line 63: We try to type the variables in a way that they cannot be both `null` and `undefined`. You can do this if you set the default value to `null`:

`_fullScreenButton: IconButton | null = null`

(Also, please move it under the static props, it's hard to find hiding up here :) )

packages/__docs__/src/Playground/index.tsx
```diff
@@ -100,18 +85,23 @@ class Playground extends Component {
   }
 
   componentDidMount() {
-    this.props.makeStyles()
+    this.props.makeStyles?.()
   }
 
-  componentDidUpdate(prevProps, prevState, snapshot) {
+  componentDidUpdate(prevProps: PlaygroundProps, prevState: PlaygroundState) {
```
ke1k0, line 91: Unused prop error:
```suggestion
  componentDidUpdate(_prevProps: PlaygroundProps, prevState: PlaygroundState) {
```

packages/__docs__/src/Playground/index.tsx
```diff
@@ -215,8 +211,16 @@ class Playground extends Component {
...
-          <div css={styles.playground}>
+        {({
+          library,
+          themeKey,
+          themes
+        }: {
+          library: Record<string, any>
+          themeKey: string
+          themes: Record<string, any>
+        }) => (
```
ke1k0, line 222: These are coming from the context (AppContext), but that part of the code is not typed yet. Please add a TODO:

`// TODO: update type once AppContext is typed`

packages/__docs__/src/Playground/props.ts
```diff
@@ -0,0 +1,94 @@
...
+import type { ComponentStyle, WithStyleProps } from '@instructure/emotion'
+import type {
+  PropValidators,
+  Border,
+  Typography
+} from '@instructure/shared-types'
+import PropTypes from 'prop-types'
+
+type PlaygroundOwnProps = {
+  styles?: PlaygroundStyle
```
ke1k0, line 34: `styles` don't have to be typed, because the `withStyle` decorator adds it. Please remove it from the props here.

packages/__docs__/src/Playground/theme.ts
```diff
@@ -21,16 +21,17 @@
...
  * Generates the theme object for the component from the theme and provided additional information
  * @param  {Object} theme The actual theme object.
  * @return {Object} The final theme object with the overrides and component variables
  */
-const generateComponentTheme = (theme) => {
+const generateComponentTheme = (theme: Theme): PlaygroundTheme => {
   const { borders, typography, key: themeName } = theme
 
-  const themeSpecificStyles = {}
+  const themeSpecificStyles: Record<string, any> = {}
```
ke1k0, line 34: This can actually be deleted because this doesn't do anything :)

ke1k0 (changes requested):

ke1k0 (approved):

HerrTopi (approved):

PR #913, opened 2022-02-08
refactor: update docs, storybook, karma unit tests to Webpack 5

matyasf:
What has changed and needs to be tested:
- docs app (in watch mode too)
- storybook (in watch mode too)
- Karma unit tests (in watch mode too)


packages/__examples__/.storybook/webpack.config.js
```diff
@@ -40,23 +39,20 @@ module.exports = ({ config, mode }) => {
...
+    config.devtool = false
   }
-
-  console.log(`Building Storybook with React version ${React.version}...`)
-
-  if (parseFloat(React.version) < 16) {
-    console.error(
-      'Storybook requires React > 16. Run `yarn install:react:16` before running `yarn start:examples`.'
-    )
-  }
```
matyasf, line 59: I've removed this check, no one has React 15 installed from the devs

packages/__examples__/package.json
```diff
@@ -11,7 +11,7 @@
   "homepage": "https://instructure.github.io/instructure-ui/",
   "bugs": "https://github.com/instructure/instructure-ui/issues",
   "scripts": {
-    "prestart": "yarn build:examples && yarn bundle",
+    "prestart": "yarn bundle",
```
matyasf, line 14: this was a bug, it was building its JSON twice

packages/ui-eslint-config/package.json
```diff
@@ -21,7 +21,6 @@
     "@typescript-eslint/parser": "^4.22.0",
     "eslint-config-prettier": "^8",
     "eslint-plugin-compat": "^3",
-    "eslint-plugin-ejs": "^0.0.2",
```
matyasf, line 24: I've removed .ejs file linting. We have like 5 ejs files in the repo (and most of them in the deprecated template apps); this plugin was not updated since 5 years and has a rarely occuring crash that takes hours to debug (was hit by it during this upgrade)

packages/ui-karma-config/lib/index.js
```diff
@@ -179,13 +180,19 @@ module.exports = function makeConfig({ coverageDirectory, coverageThreshold }) {
...
-
-        //TODO: this is probably a hack and webpack 5 will remove these features
-        //investigate how to do this properly
-        node: { fs: 'empty', module: 'empty' },
+        devtool: NO_DEBUG ? false : 'inline-source-map',
+        plugins: [
+          // fix Buffer/process is not defined errors:
+          new webpack.ProvidePlugin({ process: 'process/browser' }),
+          new webpack.ProvidePlugin({ Buffer: ['buffer', 'Buffer'] })
+        ],
```
matyasf, line 188: Webpack 4 was shipped with stubs for some node.js only APIs, these need to be added manually in Webpack 5

package.json
```diff
@@ -74,7 +74,8 @@
   "license": "MIT",
   "resolutions": {
     "react": "^17.0.0",
-    "react-dom": "^17.0.0"
+    "react-dom": "^17.0.0",
+    "@storybook/react/webpack": "^5"
```
matyasf, line 78: Storybook was a huge pain to get working with Webpack 5. They supply these builder and manager, but it has Webpack 4 as a dependency. This causes `yarn` to get confused because it has to choose which webpack to run when we run it via `ui-scripts`, so it needs to be specified here (and in some other places just to be sure :/ ).
Sotrybook plans to remove its Webpack 4 dependency with v6, so then we can remove this code

packages/__examples__/.storybook/webpack.config.js
```diff
@@ -40,23 +39,20 @@ module.exports = ({ config, mode }) => {
       }
     }
   }
+  // Disable linting for Storybook builds, it fails the build when it tries to
+  // lint its autogenerated files.
+  baseConfig.module.rules = rules.filter(
+    (rule) => rule.loader !== 'eslint-loader'
+  )
```
matyasf, line 46: Our webpack config does linting, I think we should remove it, it leads to some weird scenarios. For example it tries to link autogenerated files by Storybook

matyasf:
> The Netlify app does not load at all, see the console for the error.

oops I missed to check this, should be fixed

packages/__docs__/buildScripts/samplemedia/placeholder-image.js
```diff
@@ -22,6 +22,7 @@
  * SOFTWARE.
  */
 
+// TODO raw-loader is deprecated in Webpack 5, use https://webpack.js.org/guides/asset-modules/
```
ke1k0, line 25: Can you please make tickets for the TODOs you made in this ticket? At least for those that are important.

ke1k0 (commented):
I tested `yarn bootstrap`, `yarn dev`, `yarn start`, `yarn dev:examples`, `yarn test:patchset`. The last one has a failed test ("Menu with a trigger should apply offset values to Popover"), that might be bacause of the "await unmount()" logic in the test.

Btw, is the PR still WIP? if not, can you please remove it from the name of the PR? :)

packages/ui-menu/src/Menu/__tests__/Menu.test.tsx
```diff
@@ -334,10 +334,10 @@ describe('<Menu />', async () => {
 
     it('should apply offset values to Popover', async () => {
       const getTransforms = (transform: string) => {
-        const transformValues = transform.match(/([\d])\w*/g)!.slice(-2)
+        const transformValues = new DOMMatrixReadOnly(transform)
         return {
-          transformX: parseInt(transformValues[0]),
-          transformY: parseInt(transformValues[1])
+          transformX: Math.floor(transformValues.m41),
+          transformY: Math.floor(transformValues.m42)
```
matyasf, line 340: This test had 2 issues:
- The regexp was wrong non fractional values (e.g. it was parsing 5, 6.4 as [6,4]. This issue was hidden by:
- For some reason on local machines the transform's Y value is higher by 0.5 pixel compared to the value in CI. I guess it's some low level Chrome bug, the only difference between the 2 systems is that CI is running on Linux, local dev is OSX I've gone back until 8.14 and could reproduce the issue... Added a Math.floor to ignore subpixel values

packages/__docs__/buildScripts/samplemedia/placeholder-image.js
matyasf: done

packages/__docs__/src/index.js
```diff
@@ -29,6 +29,11 @@ import { App } from './App'
 import { canvas } from '@instructure/ui-themes'
 import { InstUISettingsProvider } from '@instructure/emotion'
 import '../globals'
+import { Buffer } from 'buffer'
```
Brailor, line 32: Importing `Buffer` from `buffer` is actually a node.js module, is this intended? Because you installed a package called `buffer`  which specifically says that if you want to import the `Buffer` from that package you should import it as such:
```js
import { Buffer } from 'buffer/' // the '/' is needed to differentiate between node's 'buffer' module and the 'buffer' package
```
See:
https://github.com/feross/buffer#usage

Actually, I don't even think we need this import and the `buffer` package altogether since you provide `Buffer` with webpack:  https://github.com/instructure/instructure-ui/pull/913/files#diff-a728496508f687dee8416140d4a823a213effc80ff956540f63262e43db14ef7R43

https://webpack.js.org/plugins/provide-plugin/

packages/ui-karma-config/lib/index.js
```diff
@@ -179,14 +180,13 @@ module.exports = function makeConfig({ coverageDirectory, coverageThreshold }) {
...
-        //investigate how to do this properly
-        node: { fs: 'empty', module: 'empty' },
-        resolve: {
-          extensions: baseWebpackConfig.resolve.extensions
-        },
+        devtool: NO_DEBUG ? false : 'inline-source-map',
+        plugins: [
+          // fix Buffer/process is not defined errors:
+          new webpack.ProvidePlugin({ process: 'process/browser' }),
+          new webpack.ProvidePlugin({ Buffer: ['buffer', 'Buffer'] })
```
Brailor, line 187: Can't we simply call this only one time?
```js
new webpack.ProvidePlugin({
    process: 'process/browser',
    Buffer: ['buffer', 'Buffer']
  }),
```

packages/ui-webpack-config/config/plugins.js
```diff
@@ -40,6 +38,8 @@ const envVars = {
 }
 
 module.exports = [
-  new webpack.EnvironmentPlugin(envVars),
-  DEBUG ? new WatchMissingNodeModulesPlugin(path.resolve('node_modules')) : null
-].filter(Boolean)
+  // fix Buffer/process is not defined errors:
+  new webpack.ProvidePlugin({ process: 'process/browser' }),
```
Brailor, line 42: Can't we simply call this only one time?
```js
new webpack.ProvidePlugin({
    process: 'process/browser',
    Buffer: ['buffer', 'Buffer']
  }),
```

packages/__docs__/src/index.js
Brailor: I removed the buffer import here, and it seems to be working fine.

matyasf: oh yeah, this was left here by accident, the provide plugin is enough in `uii-webpack-config/config/plugins.js`

packages/ui-karma-config/lib/index.js
```diff
@@ -115,12 +116,6 @@ module.exports = function makeConfig({ coverageDirectory, coverageThreshold }) {
       coverageIstanbulReporter,
       browsers,
 
-      client: {
-        chai: {
-          includeStack: true
-        }
-      },
```
matyasf, line 122: this was doing here nothing, this is not the way to configure chai

packages/ui-test-utils/src/utils/expect.ts
```diff
@@ -31,6 +31,10 @@ import promised from 'chai-as-promised'
 import assertions from './assertions'
 
 const init = (chai: Chai.ChaiStatic) => {
+  // eslint-disable-next-line no-param-reassign
+  chai.config.truncateThreshold = 999
+  // eslint-disable-next-line no-param-reassign
+  chai.config.includeStack = true
```
matyasf, line 37: this is how I could configure chai. Now it wont truncate errors, so we will get nice errors instead of stuff like "abc has the value of a.... but expecteted true". It cuts off at 40 characters by default :/

packages/ui-webpack-config/config/plugins.js
matyasf: yes, done

packages/ui-menu/src/Menu/__tests__/Menu.test.tsx
```diff
@@ -424,15 +424,15 @@ describe('<Menu />', async () => {
         expect(await accessible()).to.be.true()
       })
 
-      it('should meet standards when menu is open', async () => {
-        await mount(
+      it.only('should meet standards when menu is open', async () => {
```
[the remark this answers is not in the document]
matyasf: yes, another mistake :)

packages/ui-karma-config/lib/index.js
matyasf: yes, done

Brailor (approved):

PR #914, opened 2022-02-09
chore: fix `deterministicId` being passed to html elements

Brailor:
This PR also fixes the TS reference errors popped up on master.

### DIFF
--- a/packages/ui-react-utils/src/omitProps.ts
+++ b/packages/ui-react-utils/src/omitProps.ts
@@ -29,7 +29,7 @@
  * Return an object with the remaining props after the given props are omitted.
  *
  * Automatically excludes the following props:
- * 'theme', 'children', 'className', 'style', 'styles', 'makeStyles', 'themeOverride'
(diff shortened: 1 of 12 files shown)


HerrTopi (approved):

packages/ui-react-utils/src/omitProps.ts
```diff
@@ -67,7 +67,7 @@ const omit = <T>(originalObject: T, keysToOmit: string[]) => {
       key === 'styles' ||
       key === 'makeStyles' ||
       key === 'themeOverride' ||
-      key === 'instanceCounterMap'
```
ke1k0, line 70: Shouldn't it be both? We shouldn't accidentally pass `instanceCounterMap` either, or?

Brailor: No, since `instanceCounterMap` is not passed down to any component anymore, just the `deterministicId` function.
https://github.com/instructure/instructure-ui/blob/master/packages/ui-react-utils/src/DeterministicIdContext/withDeterministicId.tsx#L39-L76

ke1k0: oh cool, thanks :)

PR #920, opened 2022-02-15
INSTUI-3421 Fix tests and run them in random order

matyasf:
- Menu: wrap a11y tests so they dont fail randomly
- Select,SimpleSelect: ignore false positive test
- test sandbox: Fix some rare race condition that can result in components not properly unmounting
- We should not test for propTypes errors. PropTypes caches these so they are displayed only once, so for example if 2 tests are checking for something like "id is required in Button" then the second test will fail.
- there should be no code in "describe" blocks (except constants), test code should go into it() functions
- make axe errors nicer, now it displays the issue, its remedy and the violating code in one block (previously it was displaying the offending DOM element before the test results, making it really hard to read).


packages/ui-link/src/Link/__tests__/Link.test.tsx
```diff
@@ -188,9 +188,8 @@ describe('<Link />', async () => {
 
   describe('with `as` prop', async () => {
     describe('with `onClick`', async () => {
-      const onClick = stub()
```
matyasf, line 191: we should not put code in the `describe` block. When I ran tests via WebStorm these threw an exception that was somehow swallowed when running via console.

packages/ui-test-sandbox/src/utils/reactComponentWrapper.tsx
```diff
@@ -146,18 +146,17 @@ class ReactComponentWrapper {
   }
 
   unmount() {
-    return new Promise((resolve, reject) => {
```
matyasf, line 149: I dont knpow why this was a Promise, there is no async code here

packages/ui-test-sandbox/src/utils/sandbox.ts
```diff
@@ -139,44 +131,38 @@ class Sandbox {
...
-
     setAttributes(document.documentElement, this._attributes.document)
     setAttributes(document.body, this._attributes.body)
-
+    const newRecords = this._observer.takeRecords()
+    for (const newRecord of newRecords) {
+      this._addedNodes = this._addedNodes.concat(
+        Array.from(newRecord.addedNodes)
+      )
+    }
```
matyasf, line 157: This could fix some dangling DOM elements not unmounted. Although adding this here did not seem to have any effect :/

packages/ui-test-utils/src/utils/expect.ts
```diff
@@ -33,8 +33,6 @@ import assertions from './assertions'
 const init = (chai: Chai.ChaiStatic) => {
   // eslint-disable-next-line no-param-reassign
   chai.config.truncateThreshold = 999
-  // eslint-disable-next-line no-param-reassign
-  chai.config.includeStack = true
```
matyasf, line 37: this was just some totally useless stack trace, displaying lines in Karma and Mocha

packages/ui-test-sandbox/src/utils/reactComponentWrapper.tsx
Brailor: I guess this is some kind of hacky solution to wait 1 tick of the event loop. Kinda the same if you were to do a `setTimeout(.., 0)` call.

packages/ui-test-utils/src/utils/generateA11yTests.tsx
```diff
@@ -78,8 +81,16 @@ export function generateA11yTests<Props>(
...
               const element = within(subject.getDOMNode())
-
-              expect(await element.accessible()).to.be.true()
+              let axeOptions = undefined
+              if (ariaRulesToIgnore) {
+                axeOptions = { ignores: ariaRulesToIgnore }
+              }
+              const result = await element.accessible(axeOptions)
+              if (result !== true) {
+                // \n is needed because chai cuts the log in two with a
```
Brailor, line 90: Just some nitpicking:
```js
if(!result)
```

Brailor (approved):
Looks nice!

HerrTopi (approved):

PR #921, opened 2022-02-15
test: add tests for SSR

Brailor:
Add simple test cases for testing whether our components can be rendered on the server side with node.js (thus the `testEnvironment='node'`).

### DIFF
--- a/packages/ui/jest.config.js
+++ b/packages/ui/jest.config.js
@@ -23,5 +23,6 @@
  */
 module.exports = {
   testEnvironment: 'node',
-  rootDir: '__tests__'
+  rootDir: '__tests__',
+  restoreMocks: true
 }
--- a/packages/ui/tsconfig.json
+++ b/packages/ui/tsconfig.json
@@ -1,7 +1,4 @@
 {
   "extends": "../../tsconfig.json",
-  "compilerOptions": {
-    "resolveJsonModule": true,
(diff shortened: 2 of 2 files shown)


packages/ui/__tests__/ui.spec.js
```diff
@@ -0,0 +1,270 @@
...
+  }
+}
+
+describe('Testing every imported component for SSR', () => {
+  beforeAll(() => {
+    // since we don't provide every needed prop propTypes will complain when
+    // we render the components, so in order to not to spam the console
+    // we supress these warnings/errors
+    jest.spyOn(console, 'warn').mockImplementation(() => {})
+    jest.spyOn(console, 'error').mockImplementation(() => {})
```
matyasf, line 253: I think you need to call `mockRestore` after the test has run so it does not influence other tests https://jestjs.io/docs/mock-function-api#mockfnmockrestore

matyasf (changes requested):
I think mocks need to be restored

packages/ui/__tests__/ui.spec.js
Brailor: Thanks, I changed the `restoreMocks` config to be true.

matyasf (approved):
looks good :)

matyasf (approved):

PR #922, opened 2022-02-17
Update yarn to version 3

Brailor:
This PR updates yarn from yarn classic to yarn modern (yarn berry), while keeping the `node_modules` folder, thus not using the following yarn modern features:
- [plug & play](https://yarnpkg.com/features/pnp)
- [zero installs](https://yarnpkg.com/features/zero-installs)


packages/ui-view/package.json
```diff
@@ -19,7 +19,8 @@
     "build": "ui-build --modules es,cjs",
     "build:watch": "yarn ts:check --watch & ui-build --watch",
     "build:types": "tsc -p tsconfig.build.json",
-    "ts:check": "tsc -p tsconfig.build.json --noEmit --emitDeclarationOnly false"
+    "ts:check": "tsc -p tsconfig.build.json --noEmit --emitDeclarationOnly false",
+    "hello": "echo hello from view!"
```
matyasf, line 23: this was left here by accident :)

Brailor: Oops, should be fixed

matyasf (approved):
I could run everything in the console, bit if I click to run a command in a sub-package in Webstorm I get some errors like this:
```
/Users/matyas.szabo/.nvm/versions/node/v16.13.2/bin/node /Users/matyas.szabo/CODE/instructure-ui/node_modules/yarn/bin/yarn.js run test:vrt
command not found: ui-test
```

I guess its some error in WebStorm, because in console it works.

package.json
```diff
@@ -38,23 +32,20 @@
...
     "generate:package": "instui create package",
     "commit": "ui-scripts --commit",
-    "install:react": "ui-scripts --install-react",
-    "install:packages": "lerna bootstrap --reject-cycles",
     "bootstrap": "node scripts/bootstrap.js",
     "bootstrap:docs": "npm-run-all clean:modules install:packages clean build",
     "build": "lerna run build --stream",
     "build:watch": "lerna run build:watch --stream",
     "build:docs": "lerna run bundle --stream --scope docs-app",
-    "build:types": "lerna run build:types --stream",
```
ke1k0, line 52: Too bad, I was using this one (not sure why but I prefer it over build:ts, probably because the errors are more user-friendly)

package.json
```diff
@@ -38,23 +32,20 @@
...
     "update:package:list": "lerna run generate:package:list --stream --scope @instructure/instui-config",
-    "commit:package:list": "git add **/package-list.json",
+    "commit:package:list": "git add packages/instui-config/package-lists/**/package-list.json",
     "generate:component": "instui create component",
     "generate:package": "instui create package",
     "commit": "ui-scripts --commit",
-    "install:react": "ui-scripts --install-react",
-    "install:packages": "lerna bootstrap --reject-cycles",
     "bootstrap": "node scripts/bootstrap.js",
     "bootstrap:docs": "npm-run-all clean:modules install:packages clean build",
```
ke1k0, line 40: this is using `clean:modules` and `install:packages`, but those were removed.

package.json
ke1k0: can you please leave this? :)

package.json
```diff
@@ -78,37 +69,28 @@
...
-      "slack_channel": "#instui",
-      "jira_host": "instructure.atlassian.net",
-      "jira_project_id": "17900",
-      "jira_project_key": "INSTUI",
-      "npm_scope": "@instructure:registry=https://registry.npmjs.org/",
-      "gh_pages_branch": "gh-pages",
-      "gh_pages_dir": "packages/__docs__/__build__",
-      "gh_pages_cname": "instructure.design",
-      "changelog_url": "https://instructure.design/#CHANGELOG"
-    },
```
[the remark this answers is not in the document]
ke1k0: I can't see any changes related to these comments yet. Can you please check if these can be deleted? Also there's connected info in that readme.

package.json
ke1k0: Can you please check this too?

ke1k0 (commented):
I added some notes, and also (I might be doing sth wrong, but) `yarn` fails for me when I try to run it :/

package.json
Brailor: Sadly this is not really possible with yarn modern, because it takes package dependencies more seriously.
What that mean is since the sub packages (like ui-view and such) don't have `typescript` listed as their dependency, they can't call `tsc` from themselves, thus we have to call `tsc` where `typescript` is actually listed as a dependency: from the root.

However I renamed the new way of building types to the same thing as the previous one: `yarn build:types` .
This is also much better way of building them because it won't rebuild everything when you change only 1 package, it actually can determine what packages needs to be re-build and it can do it efficiently vs the old way of always rebuilding everything.

package.json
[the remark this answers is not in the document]
Brailor: I removed settings which we surely don't use, I added back those where I was not sure.
Also why should I touch the `ui-scripts` package's readme? That is a public package and I didn't modify any of the code init.

package.json
Brailor: I'm not sure what you mean, nothing is using those scripts anymore.

ke1k0: When I commented this, the `bootstrap:docs` script was net yet removed and it was still using them, but not valid anymore :)

PR #931, opened 2022-03-01
fix(emotion): fix text direction not working on simple html elements

Brailor:


HerrTopi (approved):

packages/emotion/src/InstUISettingsProvider/index.tsx
```diff
@@ -27,14 +27,26 @@ import { TextDirectionContext } from '@instructure/ui-i18n'
...
+} from '@instructure/ui-react-utils'
 import type { DeterministicIdProviderValue } from '@instructure/ui-react-utils'
+import type { AsElementType } from '@instructure/shared-types'
 
 type InstUIProviderProps = {
   theme?: ThemeOrOverride
-  dir?: 'ltr' | 'rtl' // TODO allow "auto" too
   instanceCounterMap?: DeterministicIdProviderValue
-}
+} & (
```
ke1k0, line 40: For me TS doesn't underline if I add the `as` prop without the `dir` :(

packages/emotion/src/InstUISettingsProvider/index.tsx
```diff
@@ -136,12 +150,25 @@ function InstUISettingsProvider({
       </ThemeProvider>
     </DeterministicIdContextProvider>
   )
+
+  if (dir) {
+    const Element = getElementType(InstUISettingsProvider, { as })
+    providers = <Element dir={finalDir}>{providers}</Element>
+  } else if (as && process.env.NODE_ENV !== 'production') {
+    console.warn(
+      "The 'as' property should be used in conjunction with the 'dir' property!"
```
ke1k0, line 159: It might be overkill but we could write a test for this warning too.

Brailor: You are correct, I added a test case for this.

packages/emotion/src/InstUISettingsProvider/index.tsx
Brailor: Should be fixed:
Screenshot 2022-03-24 at 10 23 48

ke1k0 (approved):

PR #932, opened 2022-03-02
Remove unused env vars and linting in Webpack

matyasf:


packages/template-app/template/build-scripts/webpack.js.ejs
```diff
@@ -27,9 +27,6 @@ const childProcess = require('child_process')
 const {
   NODE_ENV,
   DEBUG,
-  UNMANGLED_CLASS_NAMES,
-  DISABLE_SPEEDY_STYLESHEET,
-  USE_WEBPACK_CSS_LOADERS,
```
matyasf, line 32: these are unused since InstUI v7

packages/ui-webpack-config/config/module/rules.js
```diff
@@ -77,22 +73,4 @@ const rules = [
   }
 ]
 
-if (!noLint) {
-  // TODO make this diabled by default or rather do not lint here
-  rules.unshift({
```
matyasf, line 82: Webpack does not need to lint, we're linting on commit

ke1k0 (approved):

Brailor (approved):

PR #933, opened 2022-03-03
Hackweek/regression testing

Brailor:
This PR adds regression testing capabilities to our snapshot release workflow, so we can be more confident that we don't accidentally release a version that might have some regressions in it which we could not detect with our regular karma/jest test suites.

It uses the [testcafe](https://testcafe.io/) end-to-end testing framework to test a simple application (placed in the `./regression-test` directory) which just simply renders every one of our component with the necessary props. (examples taken from the performance metric tool)

This simple app will always use the latest released snapshot version, [prepare.js](https://github.com/instructure/instructure-ui/pull/933/files#diff-09d6ff0adefbe4d7de37e53a40c2207a48490773b1df99e886b7469318b1f321) will ensure that the correct snapshot version is retrieved and applied for the `package.json` before we install the app dependencies .

So the final workflow is:
```mermaid
flowchart LR
    id1(PR) -->| merge to master|id2(Release triggered)-->id3(New snapshot version released)-->id4(Use this snapshot for regression testing)

```
_Note_: The regression testing workflow will only run for snapshot releases, not for semver changing releases.

The regression testing suite is fairly simple at this point, it only does a couple of things:
1. checks whether our components renders correctly
2. checks whether the browser console shows errors

To further increase our confidence that we don't ship bad code the regression test suite will run in chrome and firefox.
The app tested by the regression test suite is bundled with different bundlers (webpack@v4 and esbuild for now), what this mean is that we will test the app in 4 different scenarios:
- chrome with esbuild
- firefox with esbuild
- chrome with webpack@v4
- firefox with webpack@v4


regression-test/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "@instructure/regression-test",
+  "description": "InstUI based app used for regression tests.",
+  "version": "0.0.1",
+  "private": true,
+  "scripts": {
+    "preinstall": "node prepare.js",
```
Brailor, line 7: This will ensure that we are always using the latest available snapshot version

matyasf (approved):
looks good, let's use it!

HerrTopi (approved):

PR #939, opened 2022-03-11
INSTUI-3423 v8 breaking changes codemods

matyasf:
these are the codemods one must run after they updated their dependencies to InstUI v8. Does the following:

 * `<ApplyTheme />` -> `<InstUISettingsProvider />`
 * `Menu.theme` -> `Menu.componentId` (for every InstUI component)
 * Renames every InstUI components `theme={}` prop to `themeOverride={}`

Also this PR includes a fix to the v7 `List` codemod, that was not executing when `List` was imported from `@instructure/ui-list`.
Finally it makes it possible to run the codemods on TypeScript files (until now Prettier was crashing on TS files.)


packages/ui-codemods/__testfixtures__/updateV7Props/list.input.js
```diff
@@ -1,12 +1,12 @@
 /* eslint-disable */
-import { List } from '@instructure/ui-lists'
+import { List } from '@instructure/ui-list'
```
matyasf, line 2: this was a bug in a v7 codemod

packages/ui-codemods/lib/utils/formatSource.ts
```diff
@@ -26,14 +26,18 @@ import prettier from 'prettier'
 
 export default function formatSource(source: string, sourcePath: string) {
   let options = null
-
+  const extension = sourcePath.split('.').pop()
+  let parser = 'babel'
+  if (extension === 'ts' || extension === 'tsx') {
+    parser = 'typescript'
+  }
```
matyasf, line 33: This makes it possible to run the mods on TypeScript files

PR #942, opened 2022-03-16
INSTUI-3364 - a11y for drilldown

ke1k0:


matyasf:
The navigation with the left/right arrows is a bit clunky, I've found the folowing issues with it (the "video settings" example is a good one to test these):
- When I press the left arrow key when on the outermost menu, the menu closes and the focus gets lost, nothing is focused. When I exit by pressing ESC or clicking outside the focus goes back to the trigger correctly.
- if I enter/exit a submenu with left/right arrows the whole menu is focused. It would be much nicer to focus on the first item in the list IMO (unless there is nothing focusable). This is also buggy, sometimes it causes a jump to the top/bottom of the page, for example: Click the "Select members" example, use the arrow keys to navigate inside "Consortiums" (so the border is selected) and press the left arrow key

matyasf (changes requested):
see my comment above

Brailor (commented):
Here is my findings:

- if you enter a sub-menu where there are a lot of options listed and you press and hold the down/up arrow, then the scrolling is not 'smooth', meaning you don't know where are you currently in the scrolling until you stop holding the arrow. After you stop holding the arrow the scrollbar just jumps into place.
- with VO (the only screen reader I tested this) if you open the video settings example, then on the "Comments" checkbox the state of the (ticked/unticked) is never read out
- in case of the Drilldown is not a Popover, then you can't toggle the checkbox with keyboard only, so nothing happens (6th example) to the checkbox when you press enter/space. When the Drilldown is a Popover you can toggle the checkbox with enter/space.
- the 5th example throws an error and stops rendering when navigating into its sub-menu:
Screenshot 2022-03-21 at 12 11 09

ke1k0:
> Here is my findings:
> 
> * if you enter a sub-menu where there are a lot of options listed and you press and hold the down/up arrow, then the scrolling is not 'smooth', meaning you don't know where are you currently in the scrolling until you stop holding the arrow. After you stop holding the arrow the scrollbar just jumps into place.
> * with VO (the only screen reader I tested this) if you open the video settings example, then on the "Comments" checkbox the state( ticked/unticked) of the checkbox is never read out
> * in case of the Drilldown is not a Popover, then you can't toggle the checkbox with keyboard only, so nothing happens (6th example) to the checkbox when you press enter/space. When the Drilldown is a Popover you can toggle the checkbox with enter/space.
> * the 5th example throws an error and stops rendering when navigating into its sub-menu:
> 
> 

1. I couldn't reproduce this one, can you show me please?
2. Yes, well, it is a special case, it shouldn't be handled in drilldown itself, more like on the user side. I edited the example, should read a better description now.
3. That was due to the 2nd checkbox example being outdated/unfinished. :) I removed that example, the video settings checkbox is the one that it is supposed to look like.
4. Thats an outdated example too and will be deleted/edited later, but I fixed the error for now.

Brailor (approved):

matyasf (approved):
while it behaves strangely on my machine (see my commend above) others could not reproduce this bug.

PR #957, opened 2022-03-24
Drilldown storybook examples

matyasf:
This PR takes a bit unusual approach to write Storybook tests for Drilldown. Since it has so many props and its sub-components should be tested too I've hand written most of the test cases. To do this I had to resort to a trick, I pass around fake props that specify the individual examples. This way almost every functionality is tested, the test cases are easy to identify, we dont have millions of examples.

### DIFF
--- a/packages/ui-drilldown/src/Drilldown/__examples__/Drilldown.examples.tsx
+++ b/packages/ui-drilldown/src/Drilldown/__examples__/Drilldown.examples.tsx
@@ -28,17 +28,15 @@ import type { StoryConfig } from '@instructure/ui-test-utils'
 
 import { Drilldown } from '../index'
 import type { DrilldownProps } from '../props'
-// eslint-disable-next-line import/no-extraneous-dependencies
-import { Button } from '@instructure/ui-buttons'
 
 export default {
   sectionProp: 'trigger',
   maxExamplesPerPage: 8,
   propValues: {
     trigger: [
-      <Button margin="medium" key={1}>
-        hello
-      </Button>,
+      <button key={1} style={{ margin: '5px' }}>
+        a drilldown trigger
+      </button>,
       undefined
     ],
     // these are non-existing props, we just pass them to getComponentProps()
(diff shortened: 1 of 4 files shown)


packages/ui-test-utils/src/utils/generateComponentExamples.ts
```diff
@@ -28,16 +28,54 @@ import { generatePropCombinations } from './generatePropCombinations'
...
+  /**
+   * Specifies the total max number of examples. Default: 500
+   */
   maxExamples?: number
-  propValues?: Partial<Record<keyof Props, any[]>>
+  /**
+   * An object with keys that correspond to the component props. Each key has a
+   * corresponding value array. This array contains possible values for that prop.
+   */
+  propValues?: Partial<Record<keyof Props | string, any[]>>
```
matyasf, line 49: I had to add `| string` to be able to pass custom props

Brailor (approved):
Looks good! Just 1 question tough:
Can't we somehow give the examples better titles? It looks little funky right now in certain cases:
Screenshot 2022-03-29 at 15 26 09

packages/ui-drilldown/src/Drilldown/__examples__/Drilldown.examples.tsx
```diff
@@ -30,33 +30,227 @@ import { Drilldown } from '../index'
 import type { DrilldownProps } from '../props'
 
 export default {
-  // sectionProp: 'shape',
-  propValues: {},
-  filter: () => {
-    return false
+  sectionProp: 'trigger',
```
ke1k0, line 33: I wouldn't use the trigger as a section prop, in fact I wouldn't use a section prop at all. I would rather add a few custom example cases where it has a `trigger`, and only test the `disabled`, `withArrow` and `defaultShow` props on them (maybe the `placement` too), because those seem to be the only ones relevant (all of the rest is handled inside the drilldown and will look the same as the flat version).

packages/ui-test-utils/src/utils/generateComponentExamples.ts
```diff
@@ -58,7 +96,7 @@ export type ExamplesPage<Props> = {
 export type Example<Props> = {
   Component: ComponentType
   componentProps: Partial<Props>
-  exampleProps: Partial<Props>
+  exampleProps: Record<string, any> // actually Partial<ViewProps>
```
ke1k0, line 99: Shouldn't we change the type to `Partial<ViewProps>`then? Or it throws errors then?

matyasf: Nope, it creates a circular dependency

PR #958, opened 2022-03-24
fix(ui-text-input): fix TextInput focusring issue with popover

HerrTopi:
Possible explanation:
https://github.com/facebook/react/issues/17894#issuecomment-591643059

### DIFF
--- a/packages/ui-text-input/src/TextInput/index.tsx
+++ b/packages/ui-text-input/src/TextInput/index.tsx
@@ -31,7 +31,7 @@ import {
   passthroughProps,
   withDeterministicId
 } from '@instructure/ui-react-utils'
-import { isActiveElement } from '@instructure/ui-dom-utils'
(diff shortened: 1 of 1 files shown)


packages/ui-text-input/src/TextInput/index.tsx
```diff
@@ -97,6 +97,10 @@ class TextInput extends Component<TextInputProps, TextInputState> {
   }
   componentDidMount() {
     this.props.makeStyles?.(this.makeStyleProps())
+    document
+      .getElementById(this.id)
+      //@ts-expect-error TODO
+      ?.addEventListener('focus', this.handleFocus)
```
ke1k0, line 103: Nice catch :) I would recommend using our own addEventListener util, and add a comment above it with the issue link and an explanation for the workaround. Also, please always remove listeners on unmount in order to prevent memory leaks. I found a similar workaround in RangeInput, you can use that as a guide :)

PR #960, opened 2022-03-25
Remove reactDom.findDOMNode from `testable.ts`

matyasf:
`testable.ts` was throwing warnings about reactDOM.findDOMNode being deprecated because it was using it. Replacing it with our findDOMNode uncovered lots of small issues where we did not return the root DOM node with elementRef, these are fixed now.


packages/ui-karma-config/lib/index.js
```diff
@@ -197,6 +197,14 @@ module.exports = function makeConfig({ coverageDirectory, coverageThreshold }) {
...
       webpackMiddleware: { stats: 'errors-only' },
 
+      client: {
+        mocha: {
+          // change the default test timeout to 5 secs. This must be higher
+          // than our async test timeouts (currently 1.9 secs) because if this
+          // happens first we do not get any sensible error message.
+          timeout: 5000
+        }
+      },
```
matyasf, line 207: this is a small fix, that makes our tests return error messages when the `await find...` calls in tests do not return in time. Their timeout was set to 1900ms, which Mocha's default is 2000ms. And when Mocha timeouts first it returns no details about was timed out.

packages/ui-testable/src/testable.ts
```diff
@@ -83,32 +84,28 @@ const testable =
...
+              let node: Element | undefined = undefined
               if (this._testableUnmounted) {
                 return
               }
               try {
                 // Use this.DOMNode for components that render as non-native Portals...
                 node =
-                  (this as ComponentWithDOMNode).DOMNode || findDOMNode(this)
+                  (this as ComponentWithDOMNode).DOMNode ||
+                  (findDOMNode(this) as Element)
```
matyasf, line 95: this is the change that removed reactDOM.findDOMNode and uncovered all the fixed errors.

packages/ui-buttons/src/CloseButton/index.tsx
```diff
@@ -113,10 +113,13 @@ class CloseButton extends Component<CloseButtonProps> {
...
-      <span {...passthroughProps(props)} css={styles?.closeButton}>
+      <span
+        {...passthroughProps(props)}
+        css={styles?.closeButton}
+        ref={this.handleRef}
+      >
         <BaseButton
           renderIcon={IconXSolid}
-          elementRef={this.handleRef}
           interaction={this.interaction}
```
matyasf, line 123: This is the general bug pattern that was uncovered by switching from React's deprecated to our version of `findDOMNode`: `ref` was not pointing to the top DOM element in the component

ke1k0: Just on first glance: Looking at this makes me worried that fixes like this (moving the whole `handleRef` up) might cause breaking changes :( When we added the `ref`-s, most components already had their `elementRef` in place. Even for CloseButton, if you check older versions (e.g. in V7), you can see that the elementRef was always giving back the BaseButton, not the wrapper. If someone relies on this in their tests for example, than it is a breaking change for them. So the `elementRef`s should stay where they were imho.

matyasf: hmmm then some of the tests will fail and we might not be able to fix them. An example: If we leave the `elementRef` of `ToggleButton` as is (see this changeset) then the `should display a tooltip without hover/focus when isShowingTooltip is true` will fail. This is because we find the tooltip like this:

```
const tooltip = await button.findTooltipContent()
```
this code does calls (after some loops) `locator(ToggleButton.selector,...).find()`. `ToggleButton.selector` is added to the component by `testable.ts`, that uses our `findDOMNode` to find out where to add it. 
And since our `findDOMNode` return the value of our `ref` prop, it wont find the tooltip, since its one level up in the DOM.

TL/DR: locators search in the DOM subtree returned by our `findDOMNode` so if its not top level it wont see whats above.

So we can fix it the following ways:
1. We move the refs up as in this commit. We can treat the previous solution as a bug since it worked like in this commit before we added the `ref`s (the values I added here are the same that `ReactDOM.findDOMNode` would return). 
2. We introduce some new prop to these components that points to the top of them and somehow access it after its mounted.
3. We rewrite the tests to get the parent of the node, e.g. call `button.getParentNode()`

What do you think? I'd prefer 1), maybe 2)

matyasf: As we discussed it I'll put back `elementRef`s to their original place and just move `ref`s

packages/ui-menu/src/Menu/index.tsx
```diff
@@ -492,11 +496,11 @@ class Menu extends Component<MenuProps> {
           if (typeof popoverRef === 'function') {
             popoverRef(el)
           }
+          this.ref = el ? el.ref : null
```
ke1k0, line 499: Popover has an elementRef prop, I used it for Drilldown too. It is easier because the Popover's ref gives back the popover itself: 

```js
        elementRef={(element) => {
          // setting ref for "Popover" version, the popover root
          // (if there is no trigger, we set it in handleDrilldownRef)
          this.handleRef(element)
        }}
        ref={(el) => {
          this._popover = el
          if (typeof popoverRef === 'function') {
            popoverRef(el)
          }
        }}
```

packages/ui-navigation/src/Navigation/NavigationItem/index.tsx
```diff
@@ -67,7 +75,7 @@ class NavigationItem extends Component<NavigationItemProps> {
     this.props.makeStyles?.({ minimized: this.props.minimized })
   }
 
-  renderLink() {
+  renderLink(addRef: boolean) {
```
ke1k0, line 78: It took some time for me to understand this. How about putting hasTooltip into a getter so you can use it here and in the render too? For first read it is not obvious what the "addRef" depends on.

packages/ui-navigation/src/Navigation/NavigationItem/props.ts
```diff
@@ -32,6 +32,7 @@ import type {
 import type { WithStyleProps, ComponentStyle } from '@instructure/emotion'
 
 type NavigationItemOwnProps = {
+  elementRef?: (el: Element | null) => void
```
ke1k0, line 35: Please add new props to the `allowedProps` array too

packages/ui-pill/src/Pill/index.tsx
```diff
@@ -95,6 +95,7 @@ class Pill extends Component<PillProps, PillState> {
   }
 
   renderPill(
+    needsRef: boolean,
```
ke1k0, line 98: Same as in NavigationItem: I don't think "needsRef" is self-explanatory enough, I would use `this.state.truncated` here too (and/or some comment/explanation above it)

packages/ui-select/src/Select/index.tsx
```diff
@@ -135,15 +136,14 @@ class Select extends Component<SelectProps> {
...
+    this.inputRef && this.inputRef.focus()
   }
 
   get _input() {
     console.warn(
       '_input property is deprecated and will be removed in v9, please use ref instead'
     )
-
-    return this.ref
+    return this.inputRef
```
ke1k0, line 146: Technically, if you look back in the git history, this `_input` was doing the exact same thing as the new `inputRef`, so I would just revert it to the original and remove this getter + warning. Also, a name like `inputRef` indicates a component prop that returns with the input element, not a private prop on the component (e.g.: DateInput):

```js
  /**
   * Provides a ref to the underlying input element.
   */
  inputRef?: (element: HTMLInputElement | null) => void
```

packages/ui-tooltip/src/Tooltip/props.ts
```diff
@@ -55,6 +55,10 @@ type TooltipRenderChildren = (
 ) => React.ReactNode
 
 type TooltipOwnProps = {
+  /**
+   * provides a reference to the underlying html root element
+   */
+  elementRef?: (element: Element | null) => void
```
ke1k0, line 61: Please add new props to the `allowedProps` array too

HerrTopi (approved):
Nice work, if you correct the commented stuff, you can merge it :)

packages/ui-pill/src/Pill/index.tsx
matyasf: done

packages/ui-navigation/src/Navigation/NavigationItem/props.ts
matyasf: done

packages/ui-navigation/src/Navigation/NavigationItem/index.tsx
matyasf: hmm I like it better this way, getters are hard to read IMO

packages/ui-menu/src/Menu/index.tsx
matyasf: OK, I've done it this way

packages/ui-tooltip/src/Tooltip/props.ts
matyasf: done

ke1k0 (approved):
Has a minor conflict in the package json, but otherwise approved, thanks! :)

PR #964, opened 2022-03-29
INSTUI-3063-TS-typing-ui-navigation

NamNguyenInstructure:
Closes: INSTUI-3021

### DIFF
--- a/packages/ui-navigation/src/AppNav/Item/index.tsx
+++ b/packages/ui-navigation/src/AppNav/Item/index.tsx
@@ -44,7 +44,7 @@ import generateStyle from './styles'
 import generateComponentTheme from './theme'
 import type { AppNavItemProps } from './props'
 import { allowedProps, propTypes } from './props'
-import type { ViewOwnProps, ViewProps } from '@instructure/ui-view'
+import type { ViewOwnProps } from '@instructure/ui-view'
 /**
 ---
 parent: AppNav
--- a/packages/ui-navigation/src/AppNav/Item/index.tsx
+++ b/packages/ui-navigation/src/AppNav/Item/index.tsx
@@ -44,12 +44,13 @@ import generateStyle from './styles'
 import generateComponentTheme from './theme'
 import type { AppNavItemProps } from './props'
 import { allowedProps, propTypes } from './props'
-
+import type { ViewOwnProps, ViewProps } from '@instructure/ui-view'
 /**
 ---
 parent: AppNav
 id: AppNav.Item
 ---
+@tsProps
 @module Item
 **/
 @withStyle(generateStyle, generateComponentTheme)
@@ -62,14 +63,15 @@ class Item extends Component<AppNavItemProps> {
 
   static defaultProps = {
     children: null,
-    // @ts-expect-error ts-migrate(6133) FIXME: 'event' is declared but its value is never read.
-    onClick: function (event) {},
+    onClick: function (_event: React.MouseEvent<ViewProps>) {},
     isSelected: false,
     cursor: 'pointer',
     isDisabled: false
   } as const
 
   ref: Element | null = null
+  handleFocus?: React.FocusEventHandler<ViewOwnProps>
+  handleBlur?: React.FocusEventHandler<ViewOwnProps>
 
   componentDidMount() {
     this.props.makeStyles?.()
@@ -89,8 +91,7 @@ class Item extends Component<AppNavItemProps> {
     }
   }
 
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'e' implicitly has an 'any' type.
-  handleClick = (e) => {
+  handleClick = (e: React.MouseEvent<ViewProps>) => {
     const { isDisabled, onClick } = this.props
 
     if (isDisabled) {
@@ -132,9 +133,7 @@ class Item extends Component<AppNavItemProps> {
         display="flex"
         position="relative"
         borderRadius="medium"
-        // @ts-expect-error ts-migrate(2339) FIXME: Property 'handleFocus' does not exist on type 'Ite... Remove this comment to see the full error message
         onFocus={this.handleFocus}
-        // @ts-expect-error ts-migrate(2339) FIXME: Property 'handleBlur' does not exist on type 'Item... Remove this comment to see the full error message
         onBlur={this.handleBlur}
         cursor={isDisabled ? 'not-allowed' : cursor}
         css={this.props.styles?.item}
--- a/packages/ui-navigation/src/AppNav/Item/index.tsx
+++ b/packages/ui-navigation/src/AppNav/Item/index.tsx
@@ -63,7 +63,7 @@ class Item extends Component<AppNavItemProps> {
 
   static defaultProps = {
     children: null,
-    onClick: function (_event: React.MouseEvent<ViewProps>) {},
+    onClick: function (_event: React.MouseEvent<Element>) {},
     isSelected: false,
     cursor: 'pointer',
     isDisabled: false
@@ -91,7 +91,7 @@ class Item extends Component<AppNavItemProps> {
     }
   }
 
-  handleClick = (e: React.MouseEvent<ViewProps>) => {
+  handleClick = (e: React.MouseEvent<Element>) => {
     const { isDisabled, onClick } = this.props
 
     if (isDisabled) {
--- a/packages/ui-navigation/src/AppNav/Item/index.tsx
+++ b/packages/ui-navigation/src/AppNav/Item/index.tsx
@@ -63,7 +63,6 @@ class Item extends Component<AppNavItemProps> {
 
   static defaultProps = {
     children: null,
-    onClick: function () {},
     isSelected: false,
     cursor: 'pointer',
     isDisabled: false
--- a/packages/ui-navigation/src/AppNav/Item/index.tsx
+++ b/packages/ui-navigation/src/AppNav/Item/index.tsx
@@ -44,7 +44,7 @@ import generateStyle from './styles'
 import generateComponentTheme from './theme'
 import type { AppNavItemProps } from './props'
 import { allowedProps, propTypes } from './props'
-import type { ViewOwnProps } from '@instructure/ui-view'
+
 /**
 ---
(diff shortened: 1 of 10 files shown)


packages/ui-navigation/src/AppNav/index.tsx
```diff
@@ -45,6 +52,7 @@ import { allowedProps, propTypes } from './props'
 /**
 ---
 category: components
+@tsProps
```
matyasf, line 55: @tsProps needs to be outside of the YAML part (`--- some YAML ---`)

packages/ui-navigation/src/AppNav/Item/props.ts
```diff
@@ -31,72 +31,77 @@ import type {
...
   /**
    * Content to display after the renderLabel text, such as a badge
    */
-  renderAfter: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
+  renderAfter?: React.ReactNode | ((...args: any[]) => any)
   /**
    * The visual to display (ex. an Image, Logo, Avatar, or Icon)
    */
-  renderIcon: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
+  renderIcon?: React.ReactNode | ((...args: any[]) => any)
```
matyasf, line 49: these are all entered as a parameter for `callRenderProp`. This is quite a complex type but from other places one can see that it either accepts a React component or a function that returns a React component: `React.ReactNode | (() => React.ReactNode)`

packages/ui-navigation/src/AppNav/Item/index.tsx
```diff
@@ -89,8 +91,7 @@ class Item extends Component<AppNavItemProps> {
     }
   }
 
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'e' implicitly has an 'any' type.
-  handleClick = (e) => {
+  handleClick = (e: React.MouseEvent<Element>) => {
```
matyasf, line 94: This should be `React.MouseEvent<ViewProps>`

packages/ui-navigation/src/AppNav/Item/index.tsx
```diff
@@ -62,14 +63,15 @@ class Item extends Component<AppNavItemProps> {
 
   static defaultProps = {
     children: null,
-    // @ts-expect-error ts-migrate(6133) FIXME: 'event' is declared but its value is never read.
-    onClick: function (event) {},
+    onClick: function (_event: React.MouseEvent<Element>) {},
```
matyasf, line 66: you can delete the parameter if you dont use it, it looks a bit nicer.

ke1k0 (changes requested):

packages/ui-navigation/src/AppNav/index.tsx
```diff
@@ -187,32 +189,31 @@ class AppNav extends Component<AppNavProps> {
...
-                  ? item.props.onClick
-                  : null
+                (item as Item).props.onClick && !(item as Item).props.href
+                  ? (item as Item).props.onClick
+                  : () => {}
               }
               key={index}
             >
-              {callRenderProp(item.props.renderLabel)}
+              {callRenderProp((item as Item).props.renderLabel)}
```
matyasf, line 216: I'd make a new variable instead of using so many casts:
```
          const appNavItem = item as Item
          return (
            <Menu.Item
              href={
                appNavItem.props.href ? appNavItem.props.href : undefined
              }
              onClick={
                appNavItem.props.onClick && !appNavItem.props.href
                  ? appNavItem.props.onClick
                  : () => {}
              }
              key={index}
            >
              {callRenderProp(appNavItem.props.renderLabel)}
            </Menu.Item>
          )
```

packages/ui-navigation/src/AppNav/props.ts
```diff
@@ -41,15 +41,52 @@ import type {
...
   margin?: Spacing
+  /**
+   * Provides a reference to the underlying nav element
+   */
   elementRef?: (element: Element | null) => void
+  /**
+   * Customize the text displayed in the menu trigger when links overflow
+   * the overall nav width.
+   */
   renderTruncateLabel?: React.ReactNode | ((...args: any[]) => any)
```
matyasf, line 76: See my comment above, this also comes from callRenderProp

packages/ui-navigation/src/AppNav/index.tsx
ke1k0: If it has the correct type (I commented on this too) then the casts are not needed at all

packages/ui-navigation/src/Navigation/NavigationItem/index.tsx
```diff
@@ -75,15 +75,14 @@ class NavigationItem extends Component<NavigationItemProps> {
...
       <ElementType
         {...props}
         href={href}
         onClick={onClick}
         css={this.props.styles?.navigationItem}
-        aria-label={this.props.minimized ? label : undefined}
+        aria-label={this.props.minimized ? (label as string) : undefined}
         // @ts-expect-error TODO: fix TS2590: Expression produces a union type that is too complex to represent.
-        ref={(element) => {
+        ref={(element: Element | null) => {
```
matyasf, line 85: Something is wrong here in the original code. If we allow `label` to be a React node then the screenreader will read garbage for the `aria-label`. We should only allow `string` for the label prop.

This would be a breaking change, but luckily no one uses this component currently.

matyasf (changes requested):
Nice work, but it needs some smaller changes, see my comments.

packages/ui-navigation/src/AppNav/Item/index.tsx
```diff
@@ -62,8 +63,7 @@ class Item extends Component<AppNavItemProps> {
 
   static defaultProps = {
     children: null,
-    // @ts-expect-error ts-migrate(6133) FIXME: 'event' is declared but its value is never read.
-    onClick: function (event) {},
+    onClick: function () {},
```
ke1k0, line 66: You can remove onClick from the defaultProps, there is no point of it having a default anymore.

(It doesn't make any difference in the code, and the reason why we added them to defaultProps before was to be able to show the function signature -> those are rendered from the TS types now so no need to keep them in defaultProps)

packages/ui-navigation/src/AppNav/Item/props.ts
```diff
@@ -31,72 +31,77 @@ import type {
...
-
-type AppNavItemStyle = ComponentStyle<'item' | 'label'>
-
-const propTypes: PropValidators<PropKeys> = {
   /**
    * The text to display. If the `icon` prop is used, label text must be wrapped
    * in `ScreenReaderContent`.
    */
-  renderLabel: PropTypes.oneOfType([PropTypes.node, PropTypes.func]).isRequired,
+  renderLabel: React.ReactNode | string | (() => React.ReactNode)
```
ke1k0, line 41: The `string` type is not needed, it is included in `React.ReactNode`

```suggestion
  renderLabel: React.ReactNode | (() => React.ReactNode)
```

packages/ui-navigation/src/AppNav/index.tsx
```diff
@@ -187,48 +185,40 @@ class AppNav extends Component<AppNavProps> {
       </li>
     )
   }
-
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'items' implicitly has an 'any' type.
-  renderMenu(items) {
+  renderMenu(items: (ReactChild | ReactFragment | ReactPortal)[]) {
```
ke1k0, line 188: The correct type here is this:

```suggestion
  renderMenu(items: React.ComponentElement<AppNavItemProps, Item>[]) {
```

We know that this component accepts only AppNav.Items as children, so these are always going to be AppNav.Item component elements. If you type it like this, you can remove the `const appNavItem = item as Item` casts from inside the method as well.

packages/ui-navigation/src/Navigation/NavigationItem/index.tsx
```diff
@@ -51,8 +52,7 @@ class NavigationItem extends Component<NavigationItemProps> {
 
   static defaultProps = {
     as: 'a',
-    // @ts-expect-error ts-migrate(6133) FIXME: 'e' is declared but its value is never read.
-    onClick: function (e, selected) {},
+    onClick: function () {},
```
ke1k0, line 55: `onClick` can be deleted here

packages/ui-navigation/src/Navigation/index.tsx
```diff
@@ -63,24 +65,24 @@ class Navigation extends Component<NavigationProps, NavigationState> {
...
     children: null,
     defaultMinimized: false,
-    // @ts-expect-error ts-migrate(6133) FIXME: 'event' is declared but its value is never read.
-    onMinimized: function (event, minimized) {},
-    // @ts-expect-error ts-migrate(6133) FIXME: 'e' is declared but its value is never read.
-    onClick: function (e) {}
+    onMinimized: function (
+      _event: React.SyntheticEvent,
+      _minimized: boolean
+    ) {},
```
ke1k0, line 71: `onMinimized` can be deleted here from the defaultProps

ke1k0 (changes requested):

packages/ui-navigation/src/AppNav/index.tsx
NamNguyenInstructure: I tried your suggestion but it caused an error since `renderMenu` will be called at line 259 
`{hiddenChildren.length > 0 && this.renderMenu(hiddenChildren)}` with the `hiddenChildren` is an array of `ReactChild | ReactFragment | ReactPortal` since it is the return value of `Children.toArray()`

ke1k0: Yes, because the type needs to be corrected there as well. In the render method, when you set the `childrenArray`, it has to be typed to be an array of AppNav.Item children as well, because the component only accepts AppNav.Item children. The error will disappear after that.

```ts
const childrenArray = Children.toArray(children) as React.ComponentElement<AppNavItemProps, Item>[]
```

ke1k0 (approved):

matyasf (approved):

PR #970, opened 2022-03-31
docs: add theme override info on component pages

ke1k0:
Closes: INSTUI-2954

Adds info and example for themeOverride (and theme overrides in general) under the component Default
Theme Variable section.

### DIFF
--- a/packages/__docs__/src/ComponentTheme/index.tsx
+++ b/packages/__docs__/src/ComponentTheme/index.tsx
@@ -24,16 +24,19 @@
 
 /** @jsx jsx */
 import { Component } from 'react'
+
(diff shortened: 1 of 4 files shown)


Brailor (approved):
Nice!
However I had an idea:
For some of the components the documentation page is really long, maybe we should add a table of content sort of thing at the top of the page/side of the page to ease navigation, what do you think?

ke1k0:
> Nice! However I had an idea: For some of the components the documentation page is really long, maybe we should add a table of content sort of thing at the top of the page/side of the page to ease navigation, what do you think?
> 
> Something like this: Screenshot 2022-04-06 at 10 27 53

Yeah that would be nice! I'll make a ticket for it.

PR #971, opened 2022-04-01
feat(shared-types,ui-range-input): add accessible variant for RangeIn…

ke1k0:
…put handle

Closes: INSTUI-3458

The previous thumb design is not fully accessible (doesn't have sufficiant color contrast and focus
indication), so it has become deprecated. For this feature to be non-breaking, we indtroduced a new
`thumbVariant` prop on the component. The `deprecated` variant (and connected theme variables) will
be removed in V9, please use the `accessible` variant.

### DIFF
--- a/packages/shared-types/src/ComponentThemeVariables.ts
+++ b/packages/shared-types/src/ComponentThemeVariables.ts
@@ -962,6 +962,12 @@ export type RangeInputTheme = {
   minWidth: string | 0
   handleSize: string | 0
   handleBackground: Colors['backgroundBrand']
+  handleBorderColor: Colors['borderLightest']
+  handleBorderSize: Border['widthMedium']
+  handleShadow: string | 0
+  handleFocusInset: Border['widthSmall']
+  handleFocusRingSize: Border['widthMedium']
+  handleFocusRingColor: Colors['backgroundLightest']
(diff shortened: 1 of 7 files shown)


Brailor:
I don't really understand why this change would be considered a breaking change.

If I understand correctly then the current implementation of RangeInput is buggy (does not meet a11y standards), thus this change is a bugfix which only changes visual aspects of the component - which we did already a couple of times for other components without marking the change as breaking -.

Since we can fix this bug without breaking the component APIs I don't think we should potentially delay the fix for some consumers who won't update their code after this fix.

ke1k0:
> I don't really understand why this change would be considered a breaking change.
> 
> If I understand correctly then the current implementation of RangeInput is buggy (does not meet a11y standards), thus this change is a bugfix which only changes visual aspects of the component - which we did already a couple of times for other components without marking the change as breaking -.
> 
> Since we can fix this bug without breaking the component APIs I don't think we should potentially delay the fix for some consumers who won't update their code after this fix.

@Brailor the breaking change is in the theme variables. Since there are new theme variables (and even if I'd kept the old ones, they would affect different elements/styles), we have to think of the people using theme overrides.

HerrTopi (approved):

Brailor (approved):

PR #972, opened 2022-04-04
fix(ui-simple-select): revert refactor on SimpleSelect

ke1k0:
According to a bug report, this change was causing performance issues
in SimpleSelect when combined with Formik and InstUI Tray, when
the select had 1000+ items.

Closes: INSTUI-3461

### DIFF
--- a/packages/ui-simple-select/src/SimpleSelect/index.tsx
+++ b/packages/ui-simple-select/src/SimpleSelect/index.tsx
@@ -111,12 +111,6 @@ class SimpleSelect extends Component<SimpleSelectProps, SimpleSelectState> {
     return this.ref
   }
 
-  get childrenArray() {
-    return Children.toArray(this.props.children) as Array<
-      OptionChild | GroupChild
-    >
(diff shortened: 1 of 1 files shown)


packages/ui-simple-select/src/SimpleSelect/index.tsx
```diff
@@ -169,10 +163,14 @@ class SimpleSelect extends Component<SimpleSelectProps, SimpleSelectState> {
...
 
   getFirstOption() {
+    const children = Children.toArray(this.props.children) as (
+      | OptionChild
+      | GroupChild
+    )[]
     let match: OptionChild | undefined
 
-    for (let i = 0; i < this.childrenArray.length; i++) {
-      const child = this.childrenArray[i]
```
matyasf, line 175: this was calling the getter in every iteration, I can imagine that it causes a slowdown for lots of elements.

matyasf (approved):

Brailor (approved):

PR #975, opened 2022-04-05
chore: revert using `git-cz` package to `commitzien`

Brailor:
There was a little confusion from my side when upgrading yarn, since now we have to declare every dependency that we are actually using, for example one of the commands in `ui-scripts` will call `lib-1`, and we wish to call that command from the root, then the root's `package.json` have to have `lib-1` listed as it's dependency.

This is not ideal, I guess `ui-scripts` is written in a way that is really hard to overcome this issue, I would suggest a complete overhaul of the package is needed.

The confusion was that our commit prompt `commitzien` exposes a command called `git-cz`, and since `commitzien` was not included in the root `package.json` dependency list it was failing because it could not find that command, therefore  I installed the lib called `git-cz` which is a completely different package than `commitizen`.
See relevant issue: https://github.com/streamich/git-cz/issues/61


ke1k0 (approved):
Tried it out, seems to work fine again, thanks :)

matyasf (approved):

PR #978, opened 2022-04-06
INSTUI-3411 Update dependencies

matyasf:
Updated most of our dependencies to the latest version.
Test plan:
Run as many `yarn` commands as possible: `yarn dev`, `test`, `bootstrap`,... and test the docs app and storybook that it works


scripts/bootstrap.js
```diff
@@ -23,39 +23,54 @@
...
-  // this config lets us use the exisiting shell session for the sub processes stdins and stdouts
+  // this config lets us use the existing shell session for the sub processes stdins and stdouts
   // and lets us handle the stderrs of sub processes
   // if one of the sub processes fails, then we terminate the other sub process and exit the main process
   const spawnStdIoOpts = { stdio: ['inherit', 'inherit', 'pipe'] }
-  const tsBuild = spawn('yarn', ['build:ts'], spawnStdIoOpts)
+  execSync(
+    'lerna run prepare-build --stream --scope @instructure/ui-icons',
+    opts
+  )
```
matyasf, line 38: This needs to be run before Babel and TSC build because both depend on it

scripts/clean.js
```diff
@@ -23,7 +23,7 @@
  * SOFTWARE.
  */
 
-const fsp = require('fs/promises')
+const fs = require('fs')
```
matyasf, line 26: This needs to be run in sync otherwise other jobs start while this is running

packages/ui-scripts/package.json
```diff
@@ -24,7 +24,6 @@
   "license": "MIT",
   "dependencies": {
     "@babel/cli": "^7.8.4",
-    "@codesandbox/common": "^1.0.8",
```
matyasf, line 27: This dependency was breaking the build when using Yarn 3.2.0. It was only used for a command like script to create CodeSandbox-es, which we never used, so I daleted it (it was not updated since a while)

Brailor:
On the docs page some of the components cause an error:
Screenshot 2022-04-08 at 14 11 25

This can be observed locally as well.

matyasf:
> On the docs page some of the components cause an error

Fixed this, seems we cannot upgrade react-docgen to the latest alpha

Brailor (approved):

ke1k0 (approved):
This is the only warning that I found (`yarn dev:examples`), but I confirmed that it appears on master as well :) I couldn't find any other issue.

image

PR #987, opened 2022-04-08
Update `yarn.lock` file after version bump 

Brailor:
In yarn 3 the `yarn install` in CI environments has the flag `--immutable` automatically turned on, which means it will fail when the yarn.lock file has been modified by the install. The bump script actually modifies every package's version so when you do an install after the bump the lockfile will be updated as well.

This PR changes the bump script to automatically include the lockfile change in the bump commit so it won't fail in CI anymore.


ke1k0 (approved):
Works fine, thanks!

matyasf (approved):

PR #989, opened 2022-04-08
feat(ui-link): add `role` and `forceButtonRole ` prop to Link component

ke1k0:
Closes: INSTUI-3483

Setting the new `forceButtonRole` prop to false turns off the default logic which forced the aria role `button` on Link when it has an onClick handler and is not a button element.

### DIFF
--- a/packages/ui-link/src/Link/index.tsx
+++ b/packages/ui-link/src/Link/index.tsx
@@ -65,7 +65,8 @@ class Link extends Component<LinkProps, LinkState> {
     interaction: undefined,
     color: 'link',
     iconPlacement: 'start',
-    isWithinText: true
+    isWithinText: true,
+    forceButtonRole: true
   } as const
 
   state = { hasFocus: false }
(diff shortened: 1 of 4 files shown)


packages/ui-link/src/Link/index.tsx
```diff
@@ -182,6 +182,23 @@ class Link extends Component<LinkProps, LinkState> {
     return hasVisibleChildren(this.props.children)
   }
 
+  get role() {
+    const { ariaRole, onClick } = this.props
+
+    // null will reset the role
+    if (ariaRole === null) {
+      return undefined
+    }
```
matyasf, line 191: hmm I dont like this logic, IMO we should treat `null` and `undefined` the same whenever possible (and we're doing it in all the places I know.). I'd have rather something like this:

- `undefined`/`null`/any other value: just pass it down
- `forceButtonRole`: if has onClick handler force button role, otherwise undefined

ke1k0: Yeah you are right, I wasn't very happy with it either. I refactored it and added the prop that turns this logic off.

packages/ui-link/src/Link/props.ts
```diff
@@ -67,6 +67,17 @@ type LinkOwnProps = {
    */
   as?: AsElementType
 
+  /**
+   * The ARIA role of the element.
```
matyasf, line 71: I'd add here that it can be overridden if `forceButtonRole` is true

matyasf (approved):

Brailor (approved):

PR #990, opened 2022-04-11
feat(ui-link): add `role` and `forceButtonRole ` prop to Link component

ke1k0:
Closes: INSTUI-3484

Setting the new `forceButtonRole` prop to false turns off the default logic which forced the aria role `button` on Link when it has an onClick handler and is not a button element.

### DIFF
--- a/packages/ui-link/src/Link/index.js
+++ b/packages/ui-link/src/Link/index.js
@@ -78,6 +78,15 @@ class Link extends Component {
      */
     as: PropTypes.elementType,
     /**
+     * The ARIA role of the element.
+     */
+    role: PropTypes.string,
+    /**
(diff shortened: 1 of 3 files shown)


packages/ui-link/src/Link/index.js
```diff
@@ -77,6 +77,15 @@ class Link extends Component {
      * The element type to render as (will default to `<a>` if href is provided)
      */
     as: PropTypes.elementType,
+    /**
+     * The ARIA role of the element.
```
matyasf, line 81: I'd mention here that the value defined here could be overwritten by `forceButtonRole`

matyasf (approved):

Brailor (approved):

PR #991, opened 2022-04-11
INSTUI-3457-add-version-control-codesandboxbutton

NamNguyenInstructure:


packages/__docs__/src/CodeSandboxButton/index.tsx
```diff
@@ -87,14 +87,14 @@ import 'moment/min/locales'
       js_pre_processor: 'babel',
       ...this.props.options
     }
-    const codeSandboxIconSvg =
-      'M115.498 261.088v-106.61L23.814 101.73v60.773L65.81 186.85v45.7l49.688 28.54Zm23.814.627l50.605-29.151V185.78l42.269-24.495v-60.011l-92.874 53.621v106.82Zm80.66-180.887l-48.817-28.289l-42.863 24.872l-43.188-24.897l-49.252 28.667l91.914 52.882l92.206-53.235ZM0 222.212V74.495L127.987 0L256 74.182v147.797l-128.016 73.744L0 222.212Z'
+    const getVersion = window.location.href.match(/v\d/gm)
+    const currentVersion = getVersion ? getVersion[0] : 'v8'
```
ke1k0, line 91: Come to think of it, I think the cleanest and most reliable solution here would be using the actual root package version here, so that we can guarantee the same version as the docs page. You can just simply read this from the root package json. It seem to be a better solution, because we might change the url versioning later, but the package json will always have the correct version, and we won't need to touch it when we release the new version :) What do you think @matyasf and @Brailor ?

```js
    const rootPackage = require('../../../../package.json') // root package.json
    const rootPackageVersion = rootPackage.version

    const dependencies = JSON.stringify(
      {
        dependencies: {
          '@instructure/debounce': `^${rootPackageVersion}`,
          '@instructure/ui': `${rootPackageVersion}`,
          '@instructure/ui-icons': `${rootPackageVersion}`,
          ...
```

matyasf:
as we discussed this is quite a complex topic and we will revisit it later

PR #993, opened 2022-04-11
fix(ui-date-time-input): fix Date input to fill container in `DateTim…

ke1k0:
…eInput`


matyasf (approved):
looks good, VRT gives a nice visual overview of the change. IMO this could be a breaking change, but since this is a relatively new component I doubt that it has many uses. (and we can consider the previous look a bug because it looks really off)

matyasf (approved):

Brailor (approved):

PR #996, opened 2022-04-13
INSTUI-3467 - generate table of content to docs pages

ke1k0:


Brailor (approved):
This is great!
I would prefer the TOC to be on the right side of the page, but this is just a personal preference.

ke1k0:
> This is great! I would prefer the TOC to be on the right/left side of the page and ideally it would be "sticky", so no matter how scrolled down are you on the page it would be always visible, but this is just a personal preference.

@Brailor Yes indeed, that would be better. The problem is that the page already has a side menu, and another one on the right side would take too much space away from the content. I consulted with Ákos about this, and he agreed that it wouldn't work well on smaller viewports either, and might even have some a11y issues. We talked about redesigning the whole docs page in the future, but in the meantime we agreed to place the PoC on the top of the page.

matyasf:
I'd leave out Notes and warnings, they look odd in the TOC, see for example https://996--preview-instui.netlify.app/#using-theme-overrides https://996--preview-instui.netlify.app/#server-side-rendering 

Also maybe dont display the TOC if it has less than 4 items or the page is less than say 1500 characters? it does not look nice e.g. in the Packages category https://996--preview-instui.netlify.app/#babel-plugin-transform-imports

matyasf (changes requested):
Code looks good, I just left a comment about how it should work in some cases

packages/__docs__/src/Document/index.tsx
```diff
@@ -291,8 +297,105 @@ import { ${importName} } from '${esPath}'
...
+  //             size="small"
+  //             itemSpacing="xx-small"
+  //           >
+  //             {TOC}
+  //           </List>
+  //         </ToggleDetails>
+  //       </View>
+  //     </InstUISettingsProvider>
+  //   )
+  // }
```
matyasf, line 394: You can delete this :)

matyasf (approved):
looks good except that large commented out code can be deleted

packages/__docs__/src/Document/index.tsx
ke1k0: Oh boy, forgot to clean up! :D Thanks :)

PR #997, opened 2022-04-19
INSTUI-3488 Allow to disable dates in DateTimeInput

matyasf:
One can either specify an array of ISO dates or a function that will get called for each time a date needs to be validated.

### DIFF
--- a/packages/ui-date-time-input/src/DateTimeInput/README.md
+++ b/packages/ui-date-time-input/src/DateTimeInput/README.md
@@ -5,13 +5,19 @@ describes: DateTimeInput
 A DateTimeInput component is used to enter a date-time value. It is built as a composition
 of the [DateInput](#DateInput) and [TimeSelect](#TimeSelect) components.
 
-#### Localization
+The properties of DateTimeInput prefaced with _date_ are passed to the underlying [DateInput](#DateInput)
+(e.g. _dateLabel_ is DateInput's _label_), while the properties prefaced
+with _time_ are forwarded to the underlying
+[TimeSelect](#TimeSelect). Others are either shared by both sub-components (e.g. locale), or
+are unique to DateTimeInput (e.g. description).
+
+### Localization
 
 The component is localized via its `locale` and `timezone` parameters. Both are read from `props`, `context` and from the browser's locale in this priority order. `locale` determines the language and format dates and time are displayed in (e.g. month names, AM/PM or 24-hour format) and the beginning of the week (e.g. Monday in Germany, Sunday in the U.S.) in the dropdown calendar.
 
-#### Examples
+### Examples
 
-A DateTimeInput with `columns` layout and a default value:
+#### A DateTimeInput with `columns` layout and a default value:
 
 ```js
 ---
@@ -39,7 +45,7 @@ class Example extends React.Component {
 render(<Example />)
 ```
 
-A required DateInput with `stacked` layout that warns if the value in the past:
+#### A required DateInput with `stacked` layout that warns if the value in the past:
 
 ```js
 ---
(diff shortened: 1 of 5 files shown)


packages/ui-date-time-input/src/DateTimeInput/__examples__/DateTimeInput.examples.ts
```diff
@@ -41,7 +41,8 @@ export default {
         { text: 'success message', type: 'success' }
       ]
     ],
-    interaction: ['enabled', 'disabled']
+    interaction: ['enabled', 'disabled'],
+    disabledDates: [undefined, ['2018-01-18T16:00']]
```
ke1k0, line 45: I would somehow filter these examples, because the disabled days cannot be seen when the dropdown is not open anyway, maybe only when the current date is invalid and the message is shown.

packages/ui-date-time-input/src/DateTimeInput/README.md
```diff
@@ -147,6 +147,47 @@ class Example extends React.Component {
 render(<Example />)
 ```
 
+A `DateTimeInput` with some disabled dates and a custom error message for them:
```
ke1k0, line 150: Could you please add an example for the `string[]` version as well?

Also, now the page has a few examples, I would suggest adding subtitles for the examples so that it can be seen easier what they are showcasing and the new ToC will render on them too.

packages/ui-date-time-input/src/DateTimeInput/__tests__/DateTimeInput.test.tsx
```diff
@@ -685,4 +685,42 @@ describe('<DateTimeInput />', async () => {
     })
     expect(await dateTimeInput.find(':contains(5/1/2017, 1:30 PM)')).to.exist()
   })
+
```
ke1k0, line 688: Could you please add tests for the function versions too (both the message and the`disabledDates` prop)? It would be nice to test that these dates actually get disabled in the dropdown too.

packages/ui-date-time-input/src/DateTimeInput/props.ts
```diff
@@ -28,7 +28,7 @@ import type { FormMessage } from '@instructure/ui-form-field'
 import type { InteractionType } from '@instructure/ui-react-utils'
 import { I18nPropTypes } from '@instructure/ui-i18n'
 import type { Moment } from '@instructure/ui-i18n'
-import PropTypes from 'prop-types'
+import PropTypes, { string } from 'prop-types'
```
ke1k0, line 31: This made me confused first, can we just use PropTypes.string where it is used? I'd prefer that for consistency's stake, but I guess this works too :)

ke1k0 (commented):
The feature seems to work fine, I just had some smaller comments about the docs and tests :)

HerrTopi (approved):

packages/ui-date-time-input/src/DateTimeInput/__examples__/DateTimeInput.examples.ts
matyasf: true, I just set disabledDates to undefined because we're never showing the calendar popup

packages/ui-date-time-input/src/DateTimeInput/__tests__/DateTimeInput.test.tsx
matyasf: done

packages/ui-date-time-input/src/DateTimeInput/props.ts
matyasf: done

PR #998, opened 2022-04-20
fix(ui-text-area): add missing `ref` to TextArea

ke1k0:


matyasf (approved):
nice catch :) I wonder why we missed this one.

HerrTopi (approved):

PR #1000, opened 2022-04-22
Colorpicker

HerrTopi:
**Add `ColorPicker` component**

_(instui-config,shared-types,ui-color-picker,ui-color-utils,ui-dom-utils,ui-drilldown,ui-position,ui)_

The ColorPicker is a versatile component that can be used to select colors and check their contrast ratios.

This commit also adds the ColorPicker's subcomponents as standalone components: ColorIndicator, ColorMixer, ColorPreset, ColorContrast

Epic: INSTUI-3427

### DIFF
--- a/packages/shared-types/src/ComponentThemeVariables.ts
+++ b/packages/shared-types/src/ComponentThemeVariables.ts
@@ -399,6 +399,113 @@ export type CodeEditorTheme = {
   operatorColor: Colors['textDarkest']
 }
 
+export type ColorContrastTheme = {
+  width: string
+
+  fontFamily: Typography['fontFamily']
+  fontWeight: Typography['fontWeightNormal']
+  lineHeight: Typography['lineHeightCondensed']
+  fontSize: Typography['fontSizeMedium']
+
+  statusWrapperBottomMargin: Spacing['xxSmall']
+  colorIndicatorRightMargin: Spacing['small']
+  colorPreviewBottomMargin: Spacing['small']
+  colorPreviewTopMargin: Spacing['xSmall']
+  labelBottomMargin: Spacing['xxSmall']
+
+  smallBorder: Border['widthSmall']
+
+  successColor: Colors['shamrock']
+  failureColor: Colors['crimson']
+  pickedHexColor: Colors['ash']
+}
+
+export type ColorIndicatorTheme = {
+  borderWidth: Border['widthSmall']
+  backgroundImage: string
+  backgroundSize: string
+  backgroundPosition: string
+  circleIndicatorSize: string
+  rectangleIndicatorSize: string
+  colorIndicatorBorderColor: string
+  rectangularIndicatorBorderRadius: Spacing['xxSmall']
+  rectangularIndicatorBorderWidth: Border['widthSmall']
+}
+
+export type ColorMixerPaletteTheme = {
+  indicatorBorderColor: Colors['licorice']
+  whiteColor: Colors['white']
+  colorIndicatorBorderColor: string
+
+  indicatorBorderWidth: Border['widthSmall']
+  paletteBorderRadius: Border['radiusMedium']
+  paletteBorderWidth: Border['widthSmall']
+
+  disabledOverlayZIndex: Stacking['topmost']
+}
+
+export type ColorMixerRGBAInputTheme = {
+  labelFontWeight: Typography['fontWeightBold']
+  inputContainerTopMargin: Spacing['xSmall']
+  tgbInputTopMargin: Spacing['xSmall']
+  rgbaInputTopMargin: Spacing['medium']
+}
+
+export type ColorMixerSliderTheme = {
+  indicatorBorderColor: Colors['licorice']
+
+  checkerboardBackgroundImage: string
+  checkerboardBackgroundSize: string
+  checkerboardBackgroundPosition: string
+  colorIndicatorBorderColor: string
+
+  indicatorBorderWidth: Border['widthSmall']
+  sliderBorderWidth: Border['widthSmall']
+  indicatorZIndex: Stacking['above']
+  disabledOverlayZIndex: Stacking['topmost']
+}
+
+export type ColorPickerTheme = {
+  hashMarkColor: Colors['ash']
+  warningIconColor: Colors['textWarning']
+  errorIconColor: Colors['textDanger']
+  successIconColor: Colors['textSuccess']
+  popoverSeparatorColor: Colors['tiara']
+  popoverFooterColor: Colors['porcelain']
+  checkerboardBackgroundColor: Colors['white']
+  checkerboardBackgroundImage: string
+  checkerboardBackgroundSize: string
+  checkerboardBackgroundPosition: string
+  colorIndicatorBorderColor: string
+  simpleColorContainerLeftPadding: Spacing['xSmall']
+  hashMarkContainerLineHeight: Typography['fontSizeXLarge']
+  hashMarkContainerLeftPadding: Spacing['xSmall']
+  hashMarkContainerRightPadding: Spacing['xxxSmall']
+  errorIconsRightPadding: Spacing['small']
+  successIconRightPadding: Spacing['small']
+  labelRightMargin: Spacing['xxSmall']
+  popoverContentPadding: Spacing['xxSmall']
+  popoverContentBlockBorderWidth: Border['widthSmall']
+  popoverContentBlockTopMargin: Spacing['small']
+  popoverContentBlockBottomMargin: Spacing['small']
+  popoverFooterPadding: Spacing['xSmall']
+  popoverFooterTopBorderWidth: Border['widthSmall']
+  colorMixerButtonContainerLeftMargin: Spacing['xSmall']
+}
+
+export type ColorPresetTheme = {
+  xxSmallSpacing: Spacing['xxSmall']
+  smallSpacing: Spacing['small']
+  selectedIndicatorBackgroundColor: Colors['white']
+  selectedIndicatorBorderColor: Colors['licorice']
+  popoverDividerColor: Colors['tiara']
+  smallBorder: Border['widthSmall']
+  popoverFooterColor: Colors['porcelain']
+  checkerboardBackgroundImage: string
+  checkerboardBackgroundSize: string
+  checkerboardBackgroundPosition: string
+}
+
 export type DrawerLayoutContentTheme = {
   duration: Transitions['duration']
 }
@@ -1456,6 +1563,16 @@ export interface ThemeVariables {
   CheckboxFacade: CheckboxFacadeTheme
   ToggleFacade: ToggleFacadeTheme
   CodeEditor: CodeEditorTheme
+  ColorContrast: ColorContrastTheme
+  ColorIndicator: ColorIndicatorTheme
+  ColorMixerPalette: ColorMixerPaletteTheme
+  'ColorMixer.Palette': ColorMixerPaletteTheme
+  ColorMixerRGBAInput: ColorMixerRGBAInputTheme
+  'ColorMixer.RGBAInput': ColorMixerRGBAInputTheme
+  ColorMixerSlider: ColorMixerSliderTheme
+  'ColorMixer.Slider': ColorMixerSliderTheme
+  ColorPicker: ColorPickerTheme
+  ColorPreset: ColorPresetTheme
   DrawerLayoutContent: DrawerLayoutContentTheme
   'DrawerLayout.Content': DrawerLayoutContentTheme
   DrawerLayoutTray: DrawerLayoutTrayTheme
--- a/packages/ui-color-picker/src/ColorContrast/README.md
+++ b/packages/ui-color-picker/src/ColorContrast/README.md
@@ -1,0 +1,111 @@
+---
+describes: ColorContrast
+---
+
+A component for displaying color contrast between two colors. It will perform checks according to the [WCAG 2.1 standard](https://webaim.org/articles/contrast/#ratio), determinig if a given contrast ratio is acceptable for `normal`, `large` or `graphics` texts. `normal` needs to be `4.5`, the other two `3`.
+
+### Color Contrast
+
+```js
+---
+render: false
+example: true
+---
+class Example extends React.Component {
+  render() {
+    return (
+      <ColorContrast
+        firstColor="#FF0000"
+        secondColor="#FFFF00"
+        label="Color Contrast Ratio"
+        successLabel="PASS"
+        failureLabel="FAIL"
+        normalTextLabel="Normal text"
+        largeTextLabel="Large text"
+        graphicsTextLabel="Graphics text"
+        firstColorLabel="Background"
+        secondColorLabel="Foreground"
+      />
+    );
+  }
+}
+
+render(<Example />);
+```
+
+### In-line Color setting
+
+```js
+---
+render: false
+example: true
+---
+class Example extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      selectedForeGround: "#0CBF94",
+      selectedBackGround: "#35423A",
+    };
+  }
+
+  render() {
+    return (
+      <div>
+        <ColorPreset
+          label="Background"
+          colors={[
+            "#ffffff",
+            "#0CBF94",
+            "#0C89BF",
+            "#BF0C6D",
+            "#BF8D0C",
+            "#ff0000",
+            "#576A66",
+            "#35423A",
+            "#35423F",
+          ]}
+          selected={this.state.selectedBackGround}
+          onSelect={(selectedBackGround) =>
+            this.setState({ selectedBackGround })
+          }
+        />
+        <ColorPreset
+          label="Foreground"
+          colors={[
+            "#ffffff",
+            "#0CBF94",
+            "#0C89BF",
+            "#BF0C6D",
+            "#BF8D0C",
+            "#ff0000",
+            "#576A66",
+            "#35423A",
+            "#35423F",
+          ]}
+          selected={this.state.selectedForeGround}
+          onSelect={(selectedForeGround) =>
+            this.setState({ selectedForeGround })
+          }
+        />
+        <hr style={{ width: "272px", marginLeft: 0 }} />
+        <ColorContrast
+          withoutColorPreview
+          firstColor={this.state.selectedBackGround}
+          secondColor={this.state.selectedForeGround}
+          label="Contrast Ratio"
+          successLabel="PASS"
+          failureLabel="FAIL"
+          normalTextLabel="Normal text"
+          largeTextLabel="Large text"
+          graphicsTextLabel="Graphics text"
+        />
+      </div>
+    );
+  }
+}
+
+render(<Example />);
+
+
+```
--- a/packages/ui-color-picker/src/ColorIndicator/README.md
+++ b/packages/ui-color-picker/src/ColorIndicator/README.md
@@ -1,0 +1,76 @@
+---
+describes: ColorIndicator
+---
+
+A component displaying a circle with checkerboard background capable of displaying colors
+
+### Color Indicator
+
+```js
+---
+example: true
+---
+<div style={{display:'flex', background:'white'}}>
+  <View margin='small'>
+    <ColorIndicator color=''/>
+  </View>
+  <View margin='small'>
+    <ColorIndicator color='#ff0000'/>
+  </View>
+  <View margin='small'>
+    <ColorIndicator color='#ff000088'/>
+  </View>
+  <View margin='small'>
+    <ColorIndicator color='#ff000000'/>
+  </View>
+  <View margin='small'>
+    <ColorIndicator color='yellow'/>
+  </View>
+  <View margin='small'>
+    <ColorIndicator color='rgb(155,55,82)'/>
+  </View>
+  <View margin='small'>
+    <ColorIndicator color='rgba(155,55,82,.5)'/>
+  </View>
+  <View margin='small'>
+    <ColorIndicator color='hsl(30, 100%, 50%)'/>
+  </View>
+  <View margin='small'>
+    <ColorIndicator color='hsla(30, 100%, 50%, .3)'/>
+  </View>
+  <View margin='small'>
+    <ColorIndicator color='hwb(1.5708rad 60% 0%)'/>
+  </View>
+</div>
+
+
+```
+
+### Shapes
+
+`ColorIndicator` can have a `shape` prop. It is either a `circle` which is the default, or a `rectangle`
+
+```js
+---
+example: true
+---
+<div style={{display:'flex', background:'white'}}>
+  <View margin='small'>
+    <ColorIndicator color='#ff0000'/>
+  </View>
+  <View margin='small'>
+    <ColorIndicator color='#ff0000' shape='rectangle'/>
+  </View>
+</div>
+```
+
+### Color Button Pattern
+
+```js
+---
+example: true
+---
+<IconButton screenReaderLabel="ColorIndicator button">
+  <ColorIndicator color='#ff0000'/>
+</IconButton>
+```
--- a/packages/ui-color-picker/src/ColorIndicator/index.tsx
+++ b/packages/ui-color-picker/src/ColorIndicator/index.tsx
@@ -1,0 +1,89 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015 - present Instructure, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+/** @jsx jsx */
+import { Component } from 'react'
+
+import { testable } from '@instructure/ui-testable'
+import { passthroughProps } from '@instructure/ui-react-utils'
+import { withStyle, jsx } from '@instructure/emotion'
+
+import generateStyle from './styles'
+import generateComponentTheme from './theme'
+
+import type { ColorIndicatorProps } from './props'
+import { propTypes, allowedProps } from './props'
+
+/**
+---
+category: components
+---
+@tsProps
+**/
+@withStyle(generateStyle, generateComponentTheme)
+@testable()
+class ColorIndicator extends Component<ColorIndicatorProps> {
+  static propTypes = propTypes
+  static allowedProps = allowedProps
+  static readonly componentId = 'ColorIndicator'
+
+  static defaultProps = {
+    color: '',
+    shape: 'circle'
+  }
+
+  ref: HTMLDivElement | null = null
+
+  componentDidMount() {
+    this.props.makeStyles?.()
+  }
+
+  componentDidUpdate() {
+    this.props.makeStyles?.()
+  }
+
+  handleRef = (el: HTMLDivElement | null) => {
+    const { elementRef } = this.props
+
+    this.ref = el
+
+    if (typeof elementRef === 'function') {
+      elementRef(el)
+    }
+  }
+
+  render() {
+    const { color, elementRef, ...props } = this.props
+    return (
+      <div
+        {...passthroughProps(props)}
+        ref={this.handleRef}
+        css={this.props.styles?.colorIndicator}
+      />
+    )
+  }
+}
+
+export default ColorIndicator
+export { ColorIndicator }
--- a/packages/ui-color-picker/src/ColorIndicator/props.ts
+++ b/packages/ui-color-picker/src/ColorIndicator/props.ts
@@ -1,0 +1,68 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015 - present Instructure, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+import PropTypes from 'prop-types'
+
+import type { WithStyleProps, ComponentStyle } from '@instructure/emotion'
+import type {
+  OtherHTMLAttributes,
+  PropValidators,
+  ColorIndicatorTheme
+} from '@instructure/shared-types'
+
+type ColorIndicatorOwnProps = {
+  /**
+   * Valid CSS color string. E.g.: #555, rgba(55,55,55,1). It can accept empty strings
+   */
+  color?: string
+  /**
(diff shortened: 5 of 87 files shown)


packages/ui-color-picker/src/ColorIndicator/README.md
```diff
@@ -0,0 +1,27 @@
...
+### Color Indicator
+
+```js
+---
+example: true
+---
+<ColorIndicator color='#ff0000'/>
+
+
+```
```
matyasf, line 16: I'd add here an example with transparency too.

packages/ui-color-picker/src/ColorIndicator/styles.ts
```diff
@@ -0,0 +1,63 @@
...
+const generateStyle = (
+  componentTheme: ColorIndicatorTheme,
+  props: ColorIndicatorProps
+): ColorIndicatorStyle => {
+  const { color } = props
+
+  return {
+    colorIndicator: {
+      label: 'colorIndicator'
+    },
```
matyasf, line 47: This seems to serve to use to me.

packages/ui-color-picker/src/ColorIndicator/styles.ts
```diff
@@ -0,0 +1,63 @@
...
+    checkerBoardWithColor: {
+      width: '1.5rem',
+      height: '1.5rem',
+      borderRadius: '1.5rem',
+      boxSizing: 'border-box',
+      borderWidth: componentTheme.smallBorder,
```
matyasf, line 53: I'd rename `componentTheme.smallBorder` to something more descriptive, like `borderWidth`

packages/ui-color-picker/src/ColorIndicator/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  componentDidMount() {
+    this.props.makeStyles?.()
+  }
+
+  componentDidUpdate() {
+    this.props.makeStyles?.()
+  }
+
+  handleRef = (el: Element | null) => {
```
matyasf, line 70: this is not called in anywhere here, so `ref` will be always null

packages/ui-color-picker/src/ColorIndicator/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+    const { color, elementRef, ...props } = this.props
+
+    return (
+      <div
+        {...passthroughProps(props)}
+        css={this.props.styles?.checkerBoardWithColor}
+        style={{
+          borderColor: this.calcBlendedColor(
+            hexToRgb(colorIndicatorBorderColor),
+            hexToRgb(isValid(color || 'none') ? color || 'none' : '#fff')
```
matyasf, line 102: I think this can be simplified: `tinyColor`'s `Color()` allows undefined values (you need to allow this in our `isvalid` method), so you can just write `  hexToRgb(isValid(color) ? color : '#fff')`

packages/ui-color-picker/src/ColorIndicator/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+### Color Button Pattern
+
+```js
+---
+example: true
+---
+<IconButton>
+  <ColorIndicator color='#ff0000'/>
+</IconButton>
```
matyasf, line 26: This example generates an error in the console about `ScreenReaderLabel` missing

packages/ui-color-picker/src/ColorContrast/README.md
```diff
@@ -0,0 +1,111 @@
+---
+describes: ColorContrast
+---
+
+A component for displaying `ColorContrast` between two colors. It will perform checks according to the WCAG 2.1 standard, determinig if a given contrast ratio is acceptable for `normal` `large` or `graphics` texts. `normal` needs to be `4.5`, the other two `3`
```
matyasf, line 5: I'd rewrite this is bit to:

```
A component for displaying color contrast between two colors. It will perform checks according to the [WCAG 2.1 standard](https://webaim.org/articles/contrast/#ratio), determinig if a given contrast ratio is acceptable for `normal`, `large` or `graphics` texts. `normal` needs to be `4.5`, the other two `3`.
```

packages/ui-color-picker/src/ColorIndicator/README.md
HerrTopi: Good idea. Done

packages/ui-color-picker/src/ColorIndicator/styles.ts
HerrTopi: True. The other class is renamed to colorIndicator

packages/ui-color-picker/src/ColorIndicator/index.tsx
HerrTopi: I remember that there was an edge-case where I had to use this logic, but I can't remember what was it, nor can I reproduce, so for the time being I removed it. Maybe it was fixed later, but I never thought about it

packages/ui-color-picker/src/ColorIndicator/README.md
HerrTopi: Added

packages/ui-color-picker/src/ColorPreset/README.md
```diff
@@ -0,0 +1,110 @@
...
+        />
+      </div>
+    );
+  }
+}
+
+render(<Example />);
+
+
+```
```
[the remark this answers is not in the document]
HerrTopi: Oh, really. Popover works funky with disabled. Thanks, fixed

packages/ui-color-picker/src/ColorPreset/index.tsx
```diff
@@ -0,0 +1,305 @@
...
+import { View } from '@instructure/ui-view'
+import { Tooltip } from '@instructure/ui-tooltip'
+import { Popover } from '@instructure/ui-popover'
+import { Text } from '@instructure/ui-text'
+import { IconAddLine, IconCheckDarkSolid } from '@instructure/ui-icons'
+import {
+  colorTohex8,
+  hexToRgb
+} from '@instructure/ui-color-utils/src/conversions'
+import { isValid } from '@instructure/ui-color-utils/src/isValid'
```
matyasf, line 41: needs default import

packages/ui-color-picker/src/ColorPreset/index.tsx
```diff
@@ -0,0 +1,305 @@
...
+      openAddNew: false,
+      newColor: { r: 51, g: 99, b: 42, a: 1 }
+    }
+  }
+  static defaultProps = {
+    disabled: false
+  }
+  ref: Element | null = null
+
+  handleRef = (el: Element | null) => {
```
matyasf, line 71: this is unused

packages/ui-color-picker/src/ColorPreset/index.tsx
```diff
@@ -0,0 +1,305 @@
...
+      isShowingContent={this.state.openAddNew}
+      onShowContent={() => {
+        if (this.props.disabled) return
+        this.setState({ openAddNew: true })
+      }}
+      onHideContent={() => {
+        this.setState({ openAddNew: false })
+      }}
+      on="click"
+      screenReaderLabel="Popover Dialog Example"
```
matyasf, line 140: default text here

matyasf (changes requested):
please look at my comments for changes

packages/ui-color-picker/src/ColorContrast/README.md
```diff
@@ -0,0 +1,111 @@
...
+          ]}
+          selected={this.state.selectedForeGround}
+          onSelect={(selectedForeGround) =>
+            this.setState({ selectedForeGround })
+          }
+        />
+        <hr style={{ width: "272px", marginLeft: 0 }} />
+        <ColorContrast
+          withoutColorPreview
+          firstColor={this.state.selectedBackGround}å
```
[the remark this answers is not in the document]
HerrTopi: I don't even know how to produce that character :D

packages/shared-types/src/ComponentThemeVariables.ts
```diff
@@ -399,6 +399,88 @@ export type CodeEditorTheme = {
...
+  smallSpacing: Spacing['small']
+
+  smallBorder: Border['widthSmall']
+
+  successColor: Colors['shamrock']
+  failureColor: Colors['crimson']
+  pickedHexColor: Colors['ash']
+}
+
+export type SliderTheme = {
```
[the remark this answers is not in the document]
HerrTopi: Good idea. Changed them (:

packages/ui-color-picker/src/ColorIndicator/props.ts
```diff
@@ -0,0 +1,64 @@
...
+  ToProp,
+  OtherHTMLAttributes,
+  PropValidators
+} from '@instructure/shared-types'
+
+type ColorIndicatorOwnProps = {
+  /**
+   * Valid CSS color string. E.g.: #555, rgba(55,55,55,1)
+   */
+  color: string
```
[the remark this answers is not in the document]
HerrTopi: Well, maybe. Changed it, overall it seems to be the better solution

packages/ui-color-picker/src/ColorIndicator/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+A component displaying a circle with checkerboard background capable of displaying colors
+
+### Color Indicator
+
+```js
+---
+example: true
+---
+<ColorIndicator color='#ff0000'/>
```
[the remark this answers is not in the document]
HerrTopi: I provided a lot of examples, good idea

packages/ui-color-picker/src/ColorIndicator/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+      <div
+        {...passthroughProps(props)}
+        ref={elementRef}
+        css={this.props.styles?.colorIndicator}
+        style={{
+          borderColor: this.calcBlendedColor(
+            hexToRgb(colorIndicatorBorderColor),
+            hexToRgb(isValid(color) ? color : '#fff')
+          )
+        }}
```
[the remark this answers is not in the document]
HerrTopi: Yes, we can

packages/ui-color-picker/src/ColorIndicator/styles.ts
```diff
@@ -0,0 +1,61 @@
...
+  props: ColorIndicatorProps
+): ColorIndicatorStyle => {
+  const { color } = props
+
+  return {
+    colorIndicator: {
+      label: 'colorIndicator',
+      width: '1.5rem',
+      height: '1.5rem',
+      borderRadius: '1.5rem',
```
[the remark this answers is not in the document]
HerrTopi: Yeah, in this way, they can override it if necessary

packages/ui-color-picker/src/ColorPreset/README.md
```diff
@@ -0,0 +1,110 @@
...
+---
+class Example extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      selected: "",
+      colors: [
+        "#ffffff",
+        "#0CBF94",
+        "#0C89BF00",
```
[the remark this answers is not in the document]
HerrTopi: If we follow the logic of the calculation of the border color, it shouldn't. It is an oversight from design. The colorIndicator has the same behaviour.
The main problem arises from continuous change: when you start to make the color more and more transparent, the border has to be more and more transparent as well, if not, the contrast will pop more and more (and that isn't what design described).

packages/ui-color-picker/src/ColorIndicator/theme.ts
```diff
@@ -0,0 +1,52 @@
...
+    backgroundSize: '.5rem .5rem',
+    backgroundPosition: '0 0, 0 .25rem, .25rem -0.25rem, -0.25rem 0px'
+  }
+
+  return {
+    ...componentVariables
+  }
+}
+
+export const colorIndicatorBorderColor = '#384A5899'
```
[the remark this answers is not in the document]
HerrTopi: It shouldn't be overridable. This is the desired bordercolor for colorindicators across themes.

packages/ui-color-picker/src/ColorPreset/index.tsx
```diff
@@ -0,0 +1,301 @@
...
+      ${(c2.g * c2Alpha) / alpha + (c1.g * c1Alpha * (1 - c2Alpha)) / alpha},
+      ${(c2.b * c2Alpha) / alpha + (c1.b * c1Alpha * (1 - c2Alpha)) / alpha},
+      ${c2.a}
+      )`
+  }
+
+  onMenuItemSelected =
+    (color: string) =>
+    (
+      _e: React.MouseEvent<Element, MouseEvent>,
```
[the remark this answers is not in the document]
HerrTopi: It doesn't throws it to me. I think React should be in scope without importing

packages/ui-color-picker/src/ColorPreset/props.ts
```diff
@@ -0,0 +1,147 @@
...
+   * Label text of the component
+   */
+  label?: string
+
+  /**
+   * If set, a `plus` button will appear for the preset. Those components whose corresponding keys aren't provided (`colorMixer` or `colorContrast`)
+   * will not be rendered.
+   * The `onPresetChange` function gets called when a color gets added or removed from the preset list.
+   * It will be called with the new list of colors
+   */
```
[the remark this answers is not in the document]
HerrTopi: It's not pretty per se, but does the job of explaining what it does. I think in this case, function over from would be the preferred approach. It already has an example, but still. This is the documentation of the prop

packages/ui-color-utils/src/conversions.ts
```diff
@@ -22,20 +22,79 @@
...
 }
 
-export { color2hex }
+/**
+ * Transforms any `TinyColor` to 8 length HEX (alpha included)
+ * @param {ColorInputWithoutInstance} any color representation from `TinyColor`
+ * @returns {String} An 8 length hex string like `#FF0000FF`
+ * @module colorTohex8
+ */
+const colorTohex8 = (color: ColorInputWithoutInstance): string => {
```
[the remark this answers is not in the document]
HerrTopi: These are the same. I guess you wanted some change :D

packages/ui-color-utils/src/conversions.ts
```diff
@@ -22,20 +22,79 @@
...
+import type { ColorInputWithoutInstance } from 'tinycolor2'
+
+/**
+ * Converts any valid `TinyColor` colors to hex string
  * @param {String} rgb a color string
  * @returns {String} a hex string like `#FF0000`
+ * @module color2hex
  */
-function color2hex(rgb: string): string {
+const color2hex = (rgb: ColorInputWithoutInstance): string => {
```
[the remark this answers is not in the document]
HerrTopi: No can do. This is an older util, so it is used elsewhere unfortunately

packages/ui-color-utils/src/conversions.ts
```diff
@@ -22,20 +22,79 @@
...
+ * @returns {String} An 8 length hex string like `#FF0000FF`
+ * @module colorTohex8
+ */
+const colorTohex8 = (color: ColorInputWithoutInstance): string => {
+  return Color(color).toHex8String().toUpperCase()
+}
+
+/**
+ * Transforms any `TinyColor` to RGBA object ( {r:number, g:number, b:number, a:number} )
+ * also exported as `hexToRgb` for backward compatiblity reasons
```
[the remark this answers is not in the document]
HerrTopi: For consistency and backward compatibility

packages/ui-color-utils/src/conversions.ts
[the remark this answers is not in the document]
HerrTopi: So it isn't the same :D colorTohex8 uppercase 'H' instead of lowercase 'h'

packages/ui-color-picker/src/ColorIndicator/theme.ts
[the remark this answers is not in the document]
HerrTopi: After consideration, I can make this overridable

packages/ui-color-picker/src/ColorPreset/README.md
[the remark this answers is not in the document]
HerrTopi: Must be discussed with design

packages/ui-color-picker/src/ColorPicker/index.tsx
```diff
@@ -0,0 +1,556 @@
...
+          </Tooltip>
+        </span>
+      </div>
+    ) : (
+      label
+    )
+  }
+
+  stripAlphaIfNeeded = (hex: string) =>
+    hex.slice(-2) === 'FF' ? hex.slice(0, -2) : hex
```
ke1k0, line 361: Just for future-proofing purpose, I would add the 8 char check, just in case it would be used on a 6char hex :D

```suggestion
  stripAlphaIfNeeded = (hex: string) =>
    hex.length === 8 && hex.slice(-2) === 'FF' ? hex.slice(0, -2) : hex
```

packages/ui-color-picker/src/ColorPicker/props.ts
```diff
@@ -0,0 +1,269 @@
...
+  /**
+   * TODO It's only mandatory IF not in simple View
+   * Sets the ScreenReaderLabel for the popover
+   */
+  popverScreenReaderLabel?: string
+  /**
+   * TODO It's only mandatory IF not in simple View
+   * Sets the ScreenReaderLabel for the popover Button
+   */
+  popverButtonScreenReaderLabel?: string
```
ke1k0, line 139:
```suggestion
  popoverButtonScreenReaderLabel?: string
```

packages/ui-color-picker/src/ColorPicker/index.tsx
```diff
@@ -0,0 +1,556 @@
...
+      }}
+      onHideContent={() => {
+        this.setState({ openColorPicker: false })
+      }}
+      on="click"
+      screenReaderLabel={this.props.popverScreenReaderLabel || ''}
+      shouldContainFocus
+      shouldReturnFocus
+      shouldCloseOnDocumentClick
+      offsetY="16px"
```
ke1k0, line 385: should be `rem`

packages/ui-color-picker/src/ColorPicker/index.tsx
```diff
@@ -0,0 +1,556 @@
...
+      mountNode={() => document.getElementById('main')}
```
ke1k0, line 386: The `mountNode` prop here doesn't make any sense, because there is no `#main` element here.  Actually it might be dangerous if the users page has an element like that in the DOM and the popover renders at the wrong place. It should be removed here and in ColorPreset too.

packages/ui-color-picker/src/ColorPicker/index.tsx
```diff
@@ -0,0 +1,556 @@
...
+          () =>
+            this.setState({
+              openColorPicker: false,
+              mixedColor: this.state.hexCode
+            })
+        )}
+    </div>
+  )
+  renderDefaultPopoverContent = () => (
+    <React.Fragment>
```
ke1k0, line 422:
```suggestion
    <>
```

packages/ui-color-picker/src/ColorPicker/props.ts
```diff
@@ -0,0 +1,269 @@
...
+} from '@instructure/shared-types'
+
+type ContrastStrength = 'min' | 'mid' | 'max'
+type MessageType = Array<{
+  type: 'success' | 'hint' | 'error' | 'screenreader-only'
+  text: string
+}>
+
+type ColorPickerOwnProps = {
+  /**
```
ke1k0, line 41: This is another long description for a type, I would move the details to the example and link it here

packages/ui-color-picker/src/ColorPicker/theme.ts
```diff
@@ -0,0 +1,64 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+import type { Theme } from '@instructure/ui-themes'
+import { ColorPickerTheme } from '@instructure/shared-types'
```
ke1k0, line 26:
```suggestion
import type { ColorPickerTheme } from '@instructure/shared-types'
```

packages/ui-color-picker/src/ColorPicker/theme.ts
```diff
@@ -0,0 +1,64 @@
...
+const generateComponentTheme = (theme: Theme): ColorPickerTheme => {
+  const { colors, borders, spacing, typography } = theme
+
+  const componentVariables = {
+    hashMarkColor: colors?.ash,
+    warningIconColor: colors?.fire,
+    errorIconColor: colors?.crimson,
+    successIconColor: colors?.shamrock,
+    popoverSeparatorColor: colors?.tiara,
+    popoverFooterColor: colors?.porcelain,
```
ke1k0, line 42: These "named" colors are only used directly in illustrations and stuff like that. For everything else we have to use the "functional colors" like `errorIconColor: colors?.textDanger`.

packages/ui-color-picker/src/ColorPicker/theme.ts
```diff
@@ -0,0 +1,64 @@
...
+    smallBorder: borders?.widthSmall,
+    smallSpacing: spacing?.small,
+    xSmallSpacing: spacing?.xSmall,
+    xxSmallSpacing: spacing?.xxSmall,
+    xxxSmallSpacing: spacing?.xxxSmall,
+    xLargeFontSize: typography?.fontSizeXLarge,
```
ke1k0, line 48: please use more specific variable names

packages/ui-color-picker/src/ColorMixer/RGBAInput/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+    return (
+      <div css={styles?.RGBAInput}>
+        {label && <div css={styles?.label}>{label}</div>}
+        <div css={styles?.inputContainer}>
+          <span css={styles?.rInput}>
+            <TextInput
+              disabled={disabled}
+              value={`${this.state.value.r}`}
+              onChange={(e) => this.handleChange('r', e)}
+              renderLabel=""
```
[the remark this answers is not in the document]
HerrTopi: Let's talk about implementation

packages/ui-color-picker/src/ColorMixer/Slider/index.tsx
```diff
@@ -0,0 +1,189 @@
...
+    this.props.onChange(this.calcValueFromSliderPosition(newPosition))
+  }
+
+  render() {
+    return (
+      <View
+        disabled={this.props.disabled}
+        position="relative"
+        width={this.props.width}
+        height={this.props.height}
```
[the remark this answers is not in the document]
HerrTopi: Even better: they don't even need these props to work properly. Removed from here and ColorPalette as well

packages/ui-color-picker/src/ColorPicker/index.tsx
```diff
@@ -0,0 +1,556 @@
...
+  }
+  get isDefaultPopover() {
+    return this.renderMode === 'defaultPopover'
+  }
+  get isCustomPopover() {
+    return this.renderMode === 'customPopover'
+  }
+  renderCircle() {
+    return <div css={this.props.styles?.colorCircle} />
+  }
```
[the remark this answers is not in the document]
HerrTopi: Indeed. Note: styles.colorCircle was unecessary because of this. Also deleted

packages/ui-color-picker/src/ColorPicker/index.tsx
HerrTopi: Victim of Copy/pasta

packages/ui-color-picker/src/ColorPicker/index.tsx
```diff
@@ -0,0 +1,556 @@
...
+      <div css={this.props.styles?.colorPicker} ref={this.handleRef}>
+        <TextInput
+          {...passthroughProps(props)}
+          isRequired={isRequired}
+          disabled={disabled}
+          renderLabel={() => this.renderLabel()}
+          display="inline-block"
+          width={width}
+          placeholder={placeholderText}
+          themeOverride={{ padding: '0 0.75rem 0 0' }}
```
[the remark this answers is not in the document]
HerrTopi: Actually, broken padding did kinda what I needed, so if I just override it to empty string (`themeOverride={{ padding: '' }}`), it will set the padding to 0, which is ok

packages/ui-color-picker/src/ColorPicker/index.tsx
```diff
@@ -0,0 +1,556 @@
...
+          renderAfterInput={this.renderAfterInput()}
+          renderBeforeInput={this.renderBeforeInput()}
+          value={this.state.hexCode}
+          onChange={(event, value) => this.handleOnChange(event, value)}
```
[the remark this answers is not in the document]
HerrTopi: No, it won't, because I need to use the class scope (`this`) in the `handleOnChange` function, but it will have a different `this` if you don't use this arrow function or bind the correct `this` to the function 
https://reactjs.org/docs/faq-functions.html

packages/ui-color-picker/src/ColorPicker/props.ts
```diff
@@ -0,0 +1,269 @@
...
+   * If 'value' is set, this must be set. It'll be called on every change
+   */
+  onChange?: (value: string) => void
+
+  /**
+   * Placeholder for the inputfield
+   */
+  placeholderText: string
+  /**
+   * TODO It's only mandatory IF not in simple View
```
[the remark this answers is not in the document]
HerrTopi: I removed it for the time being. The whole conditionally mandatory stuff should be discussed by the team

packages/ui-color-picker/src/ColorPicker/index.tsx
```diff
@@ -0,0 +1,556 @@
...
+      tooltip,
+      value,
+      width,
+      withAlpha,
+      ...props
+    } = this.props
+    return (
+      <div css={this.props.styles?.colorPicker} ref={this.handleRef}>
+        <TextInput
+          {...passthroughProps(props)}
```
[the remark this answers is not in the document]
HerrTopi: Indeed, good catch. I moved  it to the div

packages/ui-color-picker/src/ColorPreset/styles.ts
```diff
@@ -0,0 +1,94 @@
...
+    },
+    presetRect: {
+      width: '100%',
+      height: '100%',
+      borderRadius: componentTheme.xxSmallSpacing,
+      boxSizing: 'border-box',
+      borderStyle: 'solid',
+      backgroundImage: componentTheme.checkerboardBackgroundImage,
+      backgroundSize: componentTheme.checkerboardBackgroundSize,
+      backgroundPosition: componentTheme.checkerboardBackgroundPosition
```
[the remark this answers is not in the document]
HerrTopi: It is a good idea. I did the refactor and looks much better. Thanks for the suggestion

packages/ui-color-picker/src/ColorPreset/README.md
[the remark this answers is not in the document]
HerrTopi: We found a good solution (setting the maximum transparency of the border to 60%), so rolling with it for the time being

packages/ui-color-picker/src/ColorPreset/index.tsx
```diff
@@ -0,0 +1,301 @@
...
+
+  componentDidMount() {
+    this.props.makeStyles?.()
+  }
+
+  componentDidUpdate() {
+    this.props.makeStyles?.()
+  }
+
+  calcBlendedColor = (c1: RGBAType, c2: RGBAType) => {
```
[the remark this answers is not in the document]
HerrTopi: After a couple of refactors, This moved to colorIndicator's style.ts and it is used only there.
There is another, similar method in colorContrast, but it fundamentally different from this, so no need to merge them

packages/ui-color-picker/src/ColorPicker/README.md
```diff
@@ -0,0 +1,399 @@
...
+---
+
+The `ColorPicker` is a versatile component that can be used to select colors and check their contrast ratios. It has 2 modes of operation:
+
+- In the simple, color input mode, it lets the user enter hex codes. It will display the color, validate the hex and check the contrast.
+- In the more complex, color picker mode, the same functionality is available as for the simpler mode, but there is the option to pick a color from a visual color mixer component or from any other method added in a popover.
+
+The component can be either `uncontrolled` or `controlled`. If the `onChange` and `value` props are used, it will behave in a `controlled` manner, otherwise `uncontrolled`.
+
+### ColorPicker with default popover content
```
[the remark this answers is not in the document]
HerrTopi: The second padding should indeed be weird. It aims to demonstrate that the user can design this as they want

packages/ui-color-picker/src/ColorPreset/index.tsx
```diff
@@ -246,18 +237,25 @@ class ColorPreset extends Component<ColorPresetProps, ColorPresetState> {
...
-      <Menu.Item value="select">Select</Menu.Item>
-      <Menu.Item value="remove">Remove</Menu.Item>
-    </Menu>
+      <Drilldown.Page withoutHeaderSeparator id="root" renderTitle={color}>
+        <Drilldown.Option value="select" id="select">
+          Select
+        </Drilldown.Option>
+        <Drilldown.Option value="remove" id="remove">
+          Remove
+        </Drilldown.Option>
```
ke1k0, line 254: Just a suggestion: the options have an own `onOptionClick` prop, probably easier to use here instead of the root `onSelect`.

packages/ui-color-utils/src/conversions.ts
```diff
@@ -91,10 +91,15 @@ const colorToHsla = (
...
-  colorTohex8,
+  colorToHex,
+  colorToHex8,
   colorToHsva,
   colorToHsla,
   colorToRGB,
-  colorToRGB as hexToRgb
+  //these are needed for backward compatibility
+  //TODO deprecate
+  colorToRGB as hexToRgb,
```
ke1k0, line 101: As far as I can tell, the `hexToRGB` is only used in the new component, so it might be enough to refactor there. Only the `color2hex` util existed here before.

packages/ui-color-utils/src/index.ts
```diff
@@ -28,3 +28,9 @@ export { lighten } from './lighten'
 export { contrast } from './contrast'
 export { isValid } from './isValid'
 export { color2hex } from './conversions'
+export { colorToHex8 } from './conversions'
+export { colorToHsva } from './conversions'
+export { colorToHsla } from './conversions'
+export { colorToRGB } from './conversions'
+export { hexToRgb } from './conversions'
```
ke1k0, line 35: these can be combined in one export, like `export {color2hex, colorToHex8, ... } from './conversions'`

packages/ui-color-utils/src/conversions.ts
```diff
@@ -22,20 +22,84 @@
...
+ * Color conversion utilities to transform between `TinyColor` colors (https://github.com/bgrins/TinyColor)
+ * @module conversions
+ */
+
+import Color from 'tinycolor2'
+import type { ColorInputWithoutInstance } from 'tinycolor2'
+
+/**
+ * Converts any valid `TinyColor` colors to hex string
  * @param {String} rgb a color string
```
ke1k0, line 38:
```suggestion
 * @param {ColorInputWithoutInstance} rgb a color string
```

packages/ui-color-picker/src/ColorIndicator/props.ts
```diff
@@ -0,0 +1,67 @@
...
+
+type ColorIndicatorStyle = ComponentStyle<'colorIndicator'>
+
+const propTypes: PropValidators<PropKeys> = {
+  color: PropTypes.string,
+  elementRef: PropTypes.func,
+  shape: PropTypes.oneOf(['circle', 'rectangle'])
+}
+
+const allowedProps: AllowedPropKeys = ['color']
```
ke1k0, line 64: I know it is a pain to keep this up-to-date (couldn't find a good typing solution for it yet), but can you please check for all components that the `allowedProps` match the props after all the refacts?

packages/ui-color-picker/src/ColorPreset/index.tsx
```diff
@@ -0,0 +1,299 @@
...
+      }}
+      onHideContent={() => {
+        this.setState({ openAddNew: false })
+      }}
+      on="click"
+      screenReaderLabel={this.props.popoverScreenReaderLabel}
+      shouldContainFocus
+      shouldReturnFocus
+      shouldCloseOnDocumentClick
+      offsetY="16px"
```
ke1k0, line 123: Use rem please. But I think it accepts number as well. Oh and please put a TODO at Position's `offsetToPx` util to fix the offset handling when it is a string (em/rem/px)

packages/ui-color-picker/src/ColorPreset/index.tsx
HerrTopi: Yeah, I tried, but it is kinda the same effort to use, so I left it as is

packages/ui-color-utils/src/conversions.ts
HerrTopi: Yeah, very true, nice catch

packages/ui-color-picker/src/ColorPreset/ColorPresetLocator.ts
```diff
@@ -0,0 +1,30 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+import { locator } from '@instructure/ui-test-locator'
+
+import { ColorPreset } from './index'
+
+export const ColorPresetLocator = locator(ColorPreset.selector, {
```
matyasf, line 28: My linter complains that I need a `// @ts-ignore selectors are not visible for TS` comment here.

matyasf (changes requested):
almost finished, very nice work!

packages/ui-color-picker/src/ColorPicker/index.tsx
```diff
@@ -0,0 +1,602 @@
...
+    }
+    this.setState({
+      showHelperErrorMessages: false,
+      hexCode: value,
+      mixedColor: `${value}`
+    })
+  }
+
+  //TODO remove any
+  handleOnPaste(event: any) {
```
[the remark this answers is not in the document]
HerrTopi: Yaaay! Nice, thanks a lot for solving this

matyasf (approved):

ke1k0 (approved):

Brailor (approved):

PR #1005, opened 2022-04-29
Inst UI 3498 add new sorting method

NamNguyenInstructure:


packages/ui-tree-browser/src/TreeBrowser/props.ts
```diff
@@ -148,6 +148,7 @@ type CollectionBase = {
 type Collection = CollectionBase & {
   items?: number[]
   collections?: (number | string)[]
+  compareFunc?: (a: any, b: any) => number
```
Brailor, line 151: Instead of defining `a` and `b` as `anys`, I think it would beneficial for the users to provide some proper type definitions for the `compareFunc`.
Here is an example what I have in mind:

```ts
// in TreeBrowser/props.ts
// define some helper types:
type CompareCollection = Collection & { type: 'collection' }
type CompareItem = CollectionItem & { type: 'item' }
// this is a sum type, so CompareObject is one of CompareCollection OR CompareItem
type CompareObject = CompareItem | CompareCollection

....
// and then use this sum type for the compareFunc:
compareFunc?: (a: CompareObject, b: CompareObject) => number
...

Doing this the user will have proper ide suggestions and autocompletion.

File: packages/ui-tree-browser/src/TreeBrowser/props.ts [hunk 2]
```diff
@@ -163,6 +164,7 @@ type CollectionProps = {
   items?: CollectionItem[]
   expanded?: boolean
   isCollectionFlattened?: boolean
+  compareFunc?: (a: any, b: any) => number
```
Brailor, line 167: Same thing applies here as above.

packages/ui-tree-browser/src/TreeBrowser/README.md
```diff
@@ -455,6 +455,65 @@ render(<Example/>)
...
+            },
+            2: { id: 2, name: "English Assignments", collections: [4], items: [] },
+            3: { id: 3, name: "Math Assignments", collections: [5], items: [2,1],
+            //  The items appear before subcollections
+            compareFunc: (a,b)=>{
+              if(a.type==="item" && b.type==="collection")
+              return -1
+              if(a.type==="collection" && b.type==="item")
+              return 1
+              return 0
```
Brailor, line 494: I would apply some code formatting here, because it looks a bit funky right now.

```js
if (a.type === "item" && b.type === "collection") {
    return -1
}
if (a.type === "collection" && b.type === "item") {
    return 1
}
return 0
```

Brailor (changes requested):

Brailor (approved):

HerrTopi (approved):

PR #1017, opened 2022-05-24
revert: "refactor(emotion): import lodash functions directly to leverage tree-shaking"

Brailor:
Reverts instructure/instructure-ui#1015 because it might end up creating a bigger bundle size for consumers that are not using the same imports (eg `import {debounce} from 'lodash/debounce'`) as we do in the library.

I.e. we might be better off if we did specific imports like import {debounce} from 'lodash/debounce'; but unless we did that everywhere then some stuff would be bringing in the individual files while other stuff would still be bringing in all of lodash, and it’s a net loss.

Discussion:
https://instructure.slack.com/archives/CNRQZ6J8P/p1653310039694569

### DIFF
--- a/packages/emotion/package.json
+++ b/packages/emotion/package.json
@@ -33,7 +33,6 @@
     "@instructure/ui-react-utils": "8.24.2",
     "@instructure/ui-themes": "8.24.2",
     "@instructure/ui-utils": "8.24.2",
-    "emotion-theming": "^11",
     "hoist-non-react-statics": "^3.3.2",
     "lodash": "^4",
     "prop-types": "^15"
@@ -41,8 +40,7 @@
   "devDependencies": {
     "@instructure/ui-babel-preset": "8.24.2",
     "@instructure/ui-test-utils": "8.24.2",
-    "@types/lodash": "^4.14.171",
-    "cpy-cli": "^3.1.1"
+    "@types/lodash": "^4.14.171"
   },
   "peerDependencies": {
     "react": ">=16.8 <=17"
--- a/packages/emotion/src/getTheme.ts
+++ b/packages/emotion/src/getTheme.ts
@@ -81,6 +81,8 @@ const getTheme =
           'No theme provided for [InstUISettingsProvider], using default `canvas` theme.'
         )
       }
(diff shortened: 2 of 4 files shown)


packages/emotion/package.json
```diff
@@ -33,16 +33,14 @@
...
-    "emotion-theming": "^11",
     "hoist-non-react-statics": "^3.3.2",
     "lodash": "^4",
     "prop-types": "^15"
   },
   "devDependencies": {
     "@instructure/ui-babel-preset": "8.24.2",
     "@instructure/ui-test-utils": "8.24.2",
-    "@types/lodash": "^4.14.171",
-    "cpy-cli": "^3.1.1"
```
Brailor, line 45: According to [`depcheck`](https://github.com/depcheck/depcheck) these dependencies were not used.

packages/emotion/src/withStyle.tsx
```diff
@@ -29,9 +29,9 @@ import type {
   RefAttributes
 } from 'react'
 
-import isEqual from 'lodash/isEqual'
 import hoistNonReactStatics from 'hoist-non-react-statics'
 
+import { deepEqual as isEqual } from '@instructure/ui-utils'
```
Brailor, line 34: Removed the `isEqual` import and usage since we have our "own" (which is just a wrapper around `fast-deep-equal`) which is supposed to be much more [performant](https://www.npmjs.com/package/fast-deep-equal#performance-benchmark).

packages/emotion/src/getTheme.ts
```diff
@@ -82,6 +81,8 @@ const getTheme =
           'No theme provided for [InstUISettingsProvider], using default `canvas` theme.'
         )
       }
+      //TODO: replace cloneDeeps with native `structuredClone` API
+      //once we hit last 2 version browser support
```
matyasf, line 85: Can you please make a ticket about this, so we dont forget it?

matyasf (approved):

packages/emotion/src/getTheme.ts
Brailor: yeah, sure, good idea

HerrTopi (approved):

PR #1019, opened 2022-05-26
fix(ui-grid): modify Ts types for GridCol

HerrTopi:
fix(ui-grid): modify Ts types for GridCol to accept prositive fractions as well

Closes: INSTUI-3513

### DIFF
--- a/packages/ui-grid/src/GridCol/index.tsx
+++ b/packages/ui-grid/src/GridCol/index.tsx
@@ -28,6 +28,7 @@ import { Component } from 'react'
 import { omitProps } from '@instructure/ui-react-utils'
 
 import { withStyle, jsx } from '@instructure/emotion'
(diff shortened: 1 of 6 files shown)


packages/ui-grid/src/GridCol/index.tsx
```diff
@@ -67,12 +68,36 @@ class GridCol extends Component<GridColProps> {
...
+            shouldWarn = true
+          }
+        })
+      }
+    }
+    if (shouldWarn) {
+      warn(false, 'Col width must be positive!')
+    }
+  }
+
```
Brailor, line 92: Just an idea:
maybe we should warn the user if the width is over 12 as well

what do you think?

Brailor (approved):

matyasf (approved):
looks good!

PR #1020, opened 2022-05-26
fix(ui-a11y-utils,ui-dialog,ui-modal): calling stale callbacks

Brailor:
Closes: INSTUI-3490

TEST PLAN:
use https://codesandbox.io/s/functional-comp-based-modal-dismiss-lph8yt as a base, then place the code in our docs app and verify that the fix works

### DIFF
--- a/packages/ui-a11y-utils/src/FocusRegion.ts
+++ b/packages/ui-a11y-utils/src/FocusRegion.ts
@@ -61,8 +61,11 @@ class FocusRegion {
     this._id = uid()
   }
 
-  updateElement(element: Element | Node) {
+  updateElement(element: Element | Node, options?: FocusRegionOptions) {
     this._contextElement = element
+    if (options) {
+      this._options = options
+    }
     if (this._keyboardFocusRegion) {
       this._keyboardFocusRegion.updateElement(element)
     }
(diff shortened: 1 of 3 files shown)


packages/ui-a11y-utils/src/FocusRegion.ts
```diff
@@ -61,8 +61,11 @@ class FocusRegion {
...
 
-  updateElement(element: Element | Node) {
+  updateElement(element: Element | Node, options?: FocusRegionOptions) {
     this._contextElement = element
+    if (options) {
+      this._options = options
+    }
     if (this._keyboardFocusRegion) {
       this._keyboardFocusRegion.updateElement(element)
     }
```
Brailor, line 71: This was the issue:
the `FocusRegion`'s `this._options` are holding the callback references (such as `onDismiss`) for `Dialog`.
For some reason these callback references were not in sync with the actual state of the `Dialog` and were holding the same reference from the first time `Dialog` was mounted in the dom, thus were calling stale functions with stale state values.
This issue for some reason was only an issue for function based components.

matyasf (approved):

HerrTopi (approved):

PR #1022, opened 2022-05-30
chore(ui-file-drop): add TS types in `FileDrop` component

NamNguyenInstructure:


packages/ui-file-drop/src/FileDrop/index.tsx
```diff
@@ -113,6 +121,12 @@ class FileDrop extends Component<FileDropProps, FileDropState> {
     ;(this.props as any).makeStyles(this.makeStyleProps())
```
matyasf, line 121: You should fix these too: `this.props.makeStyles?.`. Generally there should be no `any`s unless really necessary.

packages/ui-file-drop/src/FileDrop/index.tsx
```diff
@@ -157,34 +171,39 @@ class FileDrop extends Component<FileDropProps, FileDropState> {
   get invalid() {
```
matyasf, line 171:
```get hasMessages() {
    return (
      (this.props as any).messages && (this.props as any).messages.length > 0
    )
  }
``` 
->
```
 get hasMessages() {
    return (
      this.props.messages && this.props.messages.length > 0
    )
  }
```
why did you leave this one out?

packages/ui-file-drop/src/FileDrop/index.tsx
```diff
@@ -157,34 +171,39 @@ class FileDrop extends Component<FileDropProps, FileDropState> {
...
+    e: React.ChangeEvent | React.DragEvent,
+    shouldEnablePreview?: boolean
+  ): Array<File> {
+    let list: Array<File> = []
+    Array.from(getEventFiles(e, this.fileInputEl)).forEach((file) => {
+      const fileObj = this.convertToFile(file)
+      if (fileObj) list.push(fileObj)
+    })
     if (list.length > 1) {
       list = (this.props as any).shouldAllowMultiple ? list : [list[0]]
```
matyasf, line 189: no need for any here

matyasf (changes requested):
Please dont leave so many `any`s in the code! in 99% of the cases `this.props` does not need to be cast to `any`, I've not commented on all cases, but please fix all of them

packages/ui-file-drop/src/FileDrop/index.tsx
NamNguyenInstructure: I think "(this.props as any)" appears in the code with some intention

packages/ui-file-drop/src/FileDrop/index.tsx
```diff
@@ -304,40 +302,30 @@ class FileDrop extends Component<FileDropProps, FileDropState> {
...
     if (this.state.isFocused && keyEventIsClickButton(e)) {
-      if ((this.props as any).shouldAllowRepeats) {
-        // @ts-expect-error ts-migrate(2531) FIXME: Object is possibly 'null'.
-        this.fileInputEl.value = null
+      if (this.props.shouldAllowRepeats) {
+        this.fileInputEl!.value = null as unknown as string
       }
       // This bit of logic is necessary for MS browsers but causes unwanted warnings in Firefox
       // So we need to apply this logic only on MS browsers
       /* istanbul ignore if  */
```
matyasf, line 324: you can delete these comments

packages/ui-file-drop/src/FileDrop/utils/accepts.ts
```diff
@@ -22,7 +22,7 @@
  * SOFTWARE.
  */
 
-function accepts(file: any, acceptProp: any) {
+function accepts(file: any, acceptProp: string | string[]) {
```
matyasf, line 25: here `file` has `File` type

packages/ui-file-drop/src/FileDrop/utils/accepts.ts
```diff
@@ -45,7 +45,7 @@ function accepts(file: any, acceptProp: any) {
   return true
 }
 
-function getAcceptList(accept: any) {
+function getAcceptList(accept: string | string[]) {
   const list = Array.isArray(accept) ? accept : accept.split(',')
   return list.map((a: any) => a.trim().replace(/^\w+$/, '.$&'))
```
matyasf, line 50: you can remove the `any` type here

packages/ui-file-drop/src/FileDrop/utils/accepts.ts
```diff
@@ -45,7 +45,7 @@ function accepts(file: any, acceptProp: any) {
   return true
```
matyasf, line 45: `return acceptList.some((type: any) => {` -> `return acceptList.some((type) => {`
TS will infer (figure out) the correct type for this

matyasf (changes requested):
There are some small changes needed in `accepts.ts` otherwise nice work!

packages/ui-file-drop/src/FileDrop/index.tsx
```diff
@@ -76,20 +64,23 @@ class FileDrop extends Component<FileDropProps, FileDropState> {
...
+      _accepted: ArrayLike<DataTransferItem | File>,
+      _e: React.DragEvent
+    ) {},
+    onDropRejected: function (
+      _rejected: ArrayLike<DataTransferItem | File>,
+      _e: React.DragEvent
+    ) {},
+    onDragEnter: function (_e: React.DragEvent) {},
+    onDragOver: function (_e: React.DragEvent) {},
+    onDragLeave: function (_e: React.DragEvent) {},
```
Brailor, line 83: I think we settled on not doing this with event handler callback functions, so you can delete these from the defaultProps.

matyasf (approved):
nice work!

Brailor (approved):

PR #1026, opened 2022-06-01
refactor publish script

Brailor:


packages/ui-scripts/src/post-publish.ts
```diff
@@ -64,7 +64,7 @@ async function postPublish(
     createGitTagForRelease(releaseVersion)
 
     if (hasJiraConfig(config)) {
-      issueKeys = await getIssuesInRelease(jiraProjectKey)
```
Brailor, line 67: This is not an async function.

packages/ui-scripts/src/publish.ts
```diff
@@ -62,73 +65,140 @@ async function publish({
...
+        publishArgs.push(...['--tag', tag])
+      }
+      await runCommandAsync('npm', publishArgs, [])
+
+      await wait(500)
+
+      yield pkg
     }
   }
 }
```
Brailor, line 204: This is the main chunk of the change, this async generator function will release each package that is given to it with the correct version and will wait 500ms between releasing packages.

packages/ui-scripts/src/publish.ts
```diff
@@ -22,34 +22,37 @@
...
+      packageName: pkgJSON.name!,
+      version: pkgJSON.version!,
+      isMaintenance,
+      config: getConfig(pkgJSON)
+    })
+  } catch (err: any) {
+    error(err)
+    process.exit(1)
+  }
+})()
```
Brailor, line 55: The previous version had nested try-catch blocks and some of the catch blocks were simply swallowing the error, this new way only has 1 try-catch which means any error we might encounter during the release process will halt the whole process.

packages/ui-scripts/src/post-publish.ts
matyasf: hmm I'd get rid of this code, we're not connecting to JIRA

packages/ui-scripts/src/publish.ts
```diff
@@ -62,73 +65,140 @@ async function publish({
...
+      {
+        stdio: 'pipe'
+      }
+    )
+    const packageVersions: string[] = JSON.parse(stdout).versions
+
+    // if the package is already in the registry then don't try
+    // to publish it again
+    if (packageVersions.includes(version)) {
+      info(`📦  v${version} of ${pkg.name} is already published!`)
```
matyasf, line 191: hmm we should discuss how we handle this case. I think last time the release failed because for some reason empty packages were released. I'm leaning towards aborting with a huge error message in this case, normally we should never try to publish over a package.

Brailor: The last version had a bug in it, so it tried to release a version that already existed in the registry.
This new version will not even try to release the package that is in the registry.

packages/ui-scripts/src/post-publish.ts
Brailor: Done, since the `post-publish` effectively just called this JIRA function I've removed the command altogether.

packages/ui-scripts/src/publish.ts
Brailor: As discussed in the daily, I added the code to throw an error when this check is true.

packages/ui-scripts/src/post-publish.ts
Brailor: Since we didn't use the `post-publish` script at all I've deleted it altogether.

matyasf (approved):
looks good!

packages/pkg-utils/src/get-package.ts
```diff
@@ -22,43 +22,54 @@
...
   const pkg = readPkgUp.sync({
-    cwd: fs.realpathSync(options.cwd),
-    normalize: options.normalize
+    cwd: fs.realpathSync(opts.cwd),
+    normalize: opts.normalize
   })
 
   // for backwards compat:
-  pkg.pkg = pkg.packageJson
+  ;(pkg as PackageJson).pkg = pkg?.packageJson
```
ke1k0, line 69: Just a suggestion: you can cast the type on declaration

```ts
  const pkg = readPkgUp.sync({
    cwd: fs.realpathSync(opts.cwd),
    normalize: opts.normalize
  }) as PackageJson
```

packages/ui-scripts/package.json
```diff
@@ -18,7 +18,7 @@
   "scripts": {
     "lint": "ui-test --lint",
     "lint:fix": "ui-test --lint --fix",
-    "build:watch": "tsc -w -p tsconfig.build.json",
+    "build:watch": "tsc  -b tsconfig.build.json -w",
```
ke1k0, line 21: This might crash with Matyi's change (he merged it yesterday)

ke1k0 (approved):

packages/pkg-utils/src/get-package.ts
Brailor: Thanks, fixed

packages/ui-scripts/package.json
Brailor: Should be fixed

PR #1034, opened 2022-06-07
Update dependencies

matyasf:
- Update to deps to the latest allowed version
- Update TS to 4.7.3
- Fix some smaller type issues

### DIFF
--- a/packages/emotion/src/styleUtils/makeThemeVars.ts
+++ b/packages/emotion/src/styleUtils/makeThemeVars.ts
@@ -54,7 +54,7 @@ function makeThemeVars<P extends string, V extends Partial<ThemeVarGroup>>(
 ) {
   const themeVars: Record<string, unknown> = {}
   ;(Object.keys(vars) as Array<keyof V>).forEach((variable) => {
-    themeVars[camelize(`${prefix}-${variable}`)] = vars[variable]
+    themeVars[camelize(`${prefix}-${String(variable)}`)] = vars[variable]
   })
   return themeVars as PrefixedThemeVars<P, V>
 }
--- a/packages/ui-babel-preset/lib/index.js
+++ b/packages/ui-babel-preset/lib/index.js
@@ -36,7 +36,7 @@ module.exports = function (
   const envPresetConfig = opts.node ? getNodeEnvConfig() : getWebEnvConfig(opts)
 
   const presets = [
-    require('@babel/preset-typescript').default,
+    [require('@babel/preset-typescript').default, { allowDeclareFields: true }],
     [require('@babel/preset-env').default, envPresetConfig],
     [require('@babel/preset-react').default, { useBuiltIns: true }]
   ]
@@ -89,7 +89,6 @@ module.exports = function (
     require('@babel/plugin-transform-destructuring').default,
     [require('@babel/plugin-proposal-decorators').default, { legacy: true }], // must run before plugins that set displayName!
     require('./babel-plugin-add-displayname-for-react'),
-    require('@babel/plugin-proposal-class-properties').default,
     require('@babel/plugin-proposal-export-default-from').default,
     [
       require('@babel/plugin-transform-runtime').default,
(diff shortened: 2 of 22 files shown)


packages/emotion/src/styleUtils/makeThemeVars.ts
```diff
@@ -54,7 +54,7 @@ function makeThemeVars<P extends string, V extends Partial<ThemeVarGroup>>(
 ) {
   const themeVars: Record<string, unknown> = {}
   ;(Object.keys(vars) as Array<keyof V>).forEach((variable) => {
-    themeVars[camelize(`${prefix}-${variable}`)] = vars[variable]
+    themeVars[camelize(`${prefix}-${String(variable)}`)] = vars[variable]
```
matyasf, line 57: TS 4.7 has a bit stricter checks in this case, its needed when `variable` is `Symbol` type

packages/ui-babel-preset/lib/index.js
```diff
@@ -89,7 +89,6 @@ module.exports = function (
     require('@babel/plugin-transform-destructuring').default,
     [require('@babel/plugin-proposal-decorators').default, { legacy: true }], // must run before plugins that set displayName!
     require('./babel-plugin-add-displayname-for-react'),
-    require('@babel/plugin-proposal-class-properties').default,
```
matyasf, line 92: this is no longer needed, because its part of `babel/preset-env`

packages/ui-date-time-input/src/DateTimeInput/index.tsx
```diff
@@ -64,8 +64,7 @@ class DateTimeInput extends Component<DateTimeInputProps, DateTimeInputState> {
     dateFormat: 'LL' // Localized date with full month, e.g. "August 6, 2014"
   } as const
 
-  context!: React.ContextType<typeof ApplyLocaleContext>
-
+  declare context: React.ContextType<typeof ApplyLocaleContext>
```
matyasf, line 67: This is the new recommended syntax for `context` (which is required for React 18)

packages/ui-options/src/Options/Item/props.ts
```diff
@@ -146,5 +146,5 @@ const allowedProps: AllowedPropKeys = [
   'children'
 ]
 
-export type { OptionsItemProps, OptionsItemStyle }
+export type { OptionsItemProps, OptionsItemStyle, ItemProps }
```
ke1k0, line 149: Please prefix all variable names that are exported from the package with the component's name. Since `OptionsItemProps` is already taken, I would call this `OptionsItemMainProps` or sth like this.

packages/ui-i18n/src/bidirectional.tsx
```diff
@@ -32,6 +32,7 @@ import hoistNonReactStatics from 'hoist-non-react-statics'
 // see https://github.com/microsoft/TypeScript/issues/4881
 export type BidirectionalProps = {
   dir?: 'ltr' | 'rtl'
+  children?: React.ReactNode
```
ke1k0, line 35: React's `PropsWithChildren` type can be used too:

```tsx
export type BidirectionalProps = PropsWithChildren<{
  dir?: 'ltr' | 'rtl'
}>
```

packages/ui-react-utils/src/safeCloneElement.ts
```diff
@@ -38,7 +38,7 @@ type GetProps<E extends ReactElement = ReactElement> = E extends
   | FunctionComponentElement<infer P>
   | ComponentElement<infer P, any>
   | DOMElement<infer P, any>
-  ? P & { ref?: E['ref'] }
+  ? P & { ref?: any }
```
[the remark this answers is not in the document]
matyasf: I've spent like 30 mins to come up with something better, but could not :/

packages/ui-options/src/Options/Item/props.ts
matyasf: fixed

packages/ui-i18n/src/bidirectional.tsx
matyasf: fixed, thanks, I did not knew about this type

ke1k0 (approved):

packages/ui-react-utils/src/safeCloneElement.ts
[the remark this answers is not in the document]
ke1k0: Thanks for trying :)

HerrTopi (approved):

PR #1035, opened 2022-06-08
chore(ui-editable): add TS type in `Editable` component

NamNguyenInstructure:
chore(ui-editable): add TS type in `InplaceEdit` component

### DIFF
--- a/packages/ui-editable/src/Editable/props.ts
+++ b/packages/ui-editable/src/Editable/props.ts
@@ -21,19 +21,89 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-
+import React from 'react'
 import PropTypes from 'prop-types'
 
 import type { PropValidators } from '@instructure/shared-types'
 
-type EditableOwnProps = {
-  mode: 'view' | 'edit'
-  onChangeMode: (...args: any[]) => any
-  children?: (...args: any[]) => any
-  render?: (...args: any[]) => any
-  value?: any
-  onChange?: (...args: any[]) => any
+type GetContainerProps = (props?: Record<string, any>) => {
+  ref: React.RefCallback<any>
+  onMouseOver: (event: React.MouseEvent) => void
+  onMouseOut: (event: React.MouseEvent) => void
+  onMouseDown?: (event: React.MouseEvent) => void
+  onKeyUp?: (event: React.KeyboardEvent) => void
   readOnly?: boolean
+} & Record<string, any>
+
+type GetViewerProps = (props?: Record<string, any>) => {
+  mode: 'view' | 'edit'
+  readOnly?: boolean
+} & Record<string, any>
+
+type GetEditorProps = (props?: Record<string, any>) => {
+  mode: 'view' | 'edit'
+  onBlur: () => void
+  editorRef: (el: HTMLElement | null) => void
+  readOnly?: boolean
+} & Record<string, any>
+
+type GetEditButtonProps = (props?: Record<string, any>) => {
+  onClick: () => void
+  onFocus: () => void
+  onBlur: () => void
+  isVisible: boolean
+  buttonRef: (el: Element) => void
+  readOnly?: boolean
+} & Record<string, any>
+
+type EditableRenderProps = {
+  mode: string
+  getContainerProps: GetContainerProps
+  getViewerProps: GetViewerProps
+  getEditorProps: GetEditorProps
+  getEditButtonProps: GetEditButtonProps
+}
+
+type EditableOwnProps = {
+  /**
+   * If `'view'`: the view component is rendered,
+   * if `'edit'`: the edit component is rendered
+   */
+  mode: 'view' | 'edit'
+  /**
+   * Called when the component's mode changes.
+   */
+  onChangeMode: (newMode: string) => void
+  /**
+   * Function that you can supply that will return the children of this component.
+   * It has one parameter has the following fields:
+   * - mode: `view` or `edit`, depending on whether the view or the editor should be rendered.
+   * - getContainerProps(props) - Props to be spread onto the container element
+   * - getEditorProps(props) - Props to be spread onto the editor element
+   * - getEditButtonProps(props) - Props to be spread onto the edit button element
+   */
+  children?: (props: EditableRenderProps) => React.ReactNode
+  /**
+   * Identical to children
+   */
+  render?: (props: EditableRenderProps) => React.ReactNode
+  /**
+   * The current value.
+   * The value is managed by the consuming app, but we need to tell Editable
+   * it's changed or it won't re-render
+   */
+  value?: any
+  /**
+   * Called when Editable switches from edit to view mode and the value has changed.
+   */
+  onChange?: (value: any) => void
+  /**
+   * The mode is fixed as 'view'
+   */
+  readOnly?: boolean
+  /**
+   * provides a reference to the underlying html root element (container)
+   */
   elementRef?: (element: Element | null) => void
 }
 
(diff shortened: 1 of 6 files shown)


packages/ui-editable/src/Editable/props.ts
```diff
@@ -26,66 +26,102 @@ import PropTypes from 'prop-types'
...
+  onClick: () => void
+  onFocus: () => void
+  onBlur: () => void
+  isVisible: boolean
+  buttonRef: (el: Element) => void
+  readOnly?: boolean
+} & Record<string, any>
 
-type EditableProps = EditableOwnProps
+type renderProps = {
```
ke1k0, line 59: Since this is exported from the component, we have to prefix it, so please rename it to `EditableRenderProps`.

packages/ui-editable/src/InPlaceEdit/index.tsx
```diff
@@ -125,13 +126,18 @@ class InPlaceEdit extends Component<InPlaceEditProps> {
...
     label,
     ...buttonProps
+  }: {
+    isVisible: boolean
+    readOnly?: boolean
+    label: string
+    elementRef?: (el: Element | null) => void
+    onClick: () => void
+    onFocus: () => void
+    onBlur: () => void
```
ke1k0, line 140: The props will be passed to an IconButton component, so you can type it in a way that it accepts `Partial<IconButtonProps>`, and those extra 3 props we use in our own method.

```suggestion
    }: Partial<IconButtonProps> & {
    label: string
    isVisible: boolean
    readOnly?: boolean
```

packages/ui-editable/src/InPlaceEdit/index.tsx
```diff
@@ -150,14 +156,15 @@ class InPlaceEdit extends Component<InPlaceEditProps> {
...
     getViewerProps,
-    // @ts-expect-error ts-migrate(7031) FIXME: Binding element 'getEditorProps' implicitly has an... Remove this comment to see the full error message
     getEditorProps,
-    // @ts-expect-error ts-migrate(7031) FIXME: Binding element 'getEditButtonProps' implicitly ha... Remove this comment to see the full error message
     getEditButtonProps
+  }: {
+    getContainerProps: renderProps['getContainerProps']
+    getViewerProps: renderProps['getViewerProps']
+    getEditorProps: renderProps['getEditorProps']
+    getEditButtonProps: renderProps['getEditButtonProps']
```
ke1k0, line 167: You can simplify this:

```tsx
  renderAll = ({
    getContainerProps,
    getViewerProps,
    getEditorProps,
    getEditButtonProps
  }: renderProps) => { ... }
```

packages/ui-editable/src/InPlaceEdit/index.tsx
```diff
@@ -166,12 +173,16 @@ class InPlaceEdit extends Component<InPlaceEditProps> {
...
+    const { ['ref']: omitRef, ...omitRefFromContainerProps } =
+      getContainerProps()
     return (
       <Flex
         display={this.props.inline ? 'inline-flex' : 'flex'}
         direction={flexDir}
         justifyItems={justifyItems}
-        {...getContainerProps()}
+        {...omitRefFromContainerProps}
+        {...(omitRef as unknown as (el: Flex) => void)}
```
ke1k0, line 185: I think this TS error shows that the `ref` is indeed not typed correctly in the `getContainerProps`, and it should be fixed instead. I would revert this change and type the `ref` like this:

```tsx
type containerProps = (props?: Record<string, any>) => {
  ref: React.RefCallback<any>
  ...
}
```

packages/ui-editable/src/InPlaceEdit/props.ts
```diff
@@ -26,39 +26,26 @@ import PropTypes from 'prop-types'
...
 
-type InPlaceEditOwnProps = {
-  renderViewer: (...args: any[]) => any
-  renderEditor: (...args: any[]) => any
-  renderEditButton: (...args: any[]) => any
-  mode: 'view' | 'edit'
-  onChangeMode: (...args: any[]) => any
-  value?: any
-  onChange?: (...args: any[]) => any
+type ExtendeRenderEditButton = {
```
ke1k0, line 31:
```suggestion
type ExtendedRenderEditButton = {
```

packages/ui-editable/src/InPlaceEdit/props.ts
```diff
@@ -26,39 +26,26 @@ import PropTypes from 'prop-types'
...
+type InPlaceEditOwnProps = {
   /**
    * Function to render the view mode component.
    * It is the consumer's responsibility to provide the
    * current value or children.
    *
    * @returns {element} the viewer DOM sub-tree.
    */
-  renderViewer: PropTypes.func.isRequired,
+  renderViewer: () => JSX.Element
```
ke1k0, line 48: These `JSX.Element`s should be `React.ReactNode`s

packages/ui-editable/src/Editable/props.ts
```diff
@@ -77,16 +77,20 @@ type EditableOwnProps = {
...
    * @param {Boolean} mode
+   *
    * @param {Function} renderProps.getContainerProps - Props to be spread onto the container element
+   *
    * @param {Function} renderProps.getEditorProps - Props to be spread onto the editor element
+   *
    * @param {Function} renderProps.getEditButtonProps - Props to be spread onto the edit button element
    */
-  children?: (props: renderProps) => JSX.Element
+  children?: (props: EditableRenderProps) => JSX.Element
```
ke1k0, line 89: `children` has to be the same as `render` 

```tsx
children?: (props: EditableRenderProps) => React.ReactNode
```

packages/ui-editable/src/Editable/props.ts
```diff
@@ -77,16 +77,20 @@ type EditableOwnProps = {
   onChangeMode: (newMode: string) => void
   /**
    * @param {Object} renderProps
+   *
    * @param {Boolean} mode
```
ke1k0, line 81: `* @param {Boolean} renderProps.mode`

packages/ui-editable/src/InPlaceEdit/index.tsx
```diff
@@ -38,6 +39,7 @@ import generateStyle from './styles'
 
 import { propTypes, allowedProps } from './props'
 import type { InPlaceEditProps } from './props'
+import type { EditableRenderProps } from '../Editable/props'
```
ke1k0, line 42: @tsProps is missing from this component too

ke1k0 (changes requested):

packages/ui-editable/src/Editable/props.ts
```diff
@@ -21,71 +21,111 @@
...
+   *
    * @param {Function} renderProps.getContainerProps - Props to be spread onto the container element
+   *
    * @param {Function} renderProps.getEditorProps - Props to be spread onto the editor element
+   *
    * @param {Function} renderProps.getEditButtonProps - Props to be spread onto the edit button element
    */
```
matyasf, line 88: this looks quite ugly in the docs with the new TS props, I'd refactor it to:
```
  /**
   * Function that you can supply that will return the children of this component.
   * It has one parameter has the following fields:
   * - mode: `view` or `edit`, depending on whether the view or the editor should be rendered.
   * - getContainerProps(props) - Props to be spread onto the container element
   * - getEditorProps(props) - Props to be spread onto the editor element
   * - getEditButtonProps(props) - Props to be spread onto the edit button element
   */
```

matyasf: (same for the other fields, please remove any JSDoc annotations (like `@param`)and use Markdown if needed, since they look ugly

matyasf (changes requested):

matyasf (changes requested):
`InPlaceEdit/props.ts` has still lots of JSDoc in the comments (`@props`, `@returns`,..) rewrite these, so it looks nice in the docs app.

Also you can delete `componentWillUnmount() {}` from `Editable/intex.ts`

Othewise it looks good, nice work!

packages/ui-editable/src/Editable/props.ts
```diff
@@ -71,8 +71,10 @@ type EditableOwnProps = {
...
   mode: 'view' | 'edit'
   /**
-   * Called when the component's mode changes
-   * @param {string} new_mode
+   * Called when the component's mode changes.
+   *
+   * Parameter:
+   * - newMode: string
```
ke1k0, line 77: I think you can remove the JsDoc of the parameter here, because the TS type shows it well enough, that it has a string `newMode` param:
_____________________________________
image
_____________________________________

packages/ui-editable/src/Editable/props.ts
```diff
@@ -96,7 +98,9 @@ type EditableOwnProps = {
   value?: any
   /**
    * Called when Editable switches from edit to view mode and the value has changed.
-   * @param {any} value
+   *
+   * Parameter:
+   * - value: any
```
ke1k0, line 103: Same here, the TS type is enough.

ke1k0:
When you are finished with the PR, please squash the commits into 1 commit in the end, so we can get rid of the lot of `fix PR` commit messages

ke1k0 (approved):

matyasf (approved):
looks good!

PR #1036, opened 2022-06-08
refactor: fix scripts when running inside a package

matyasf:
This PR fixes 'command not found' errors when running a script inside a package's folder from the console.
It uses yarn's '-T' flag which instructs yarn to look for the script in the root of the monorepo. https://yarnpkg.com/cli/run#options-T%2C-top-level

To test:
- run as many of our scripts as possible
- run scripts from inside a package's folder
- run scripts via your IDE (e.g. WebStorm's green play buttons in the package JSON)

The alternative of this solution would have been in include `ui-scripts` and `typescript` as a devDependency in each of our packages. I've decided against this solution because it would make upgrading Typescript a pain.


Brailor (approved):

ke1k0 (approved):
Seems to work :) Thanks!

PR #1037, opened 2022-06-14
INSTUI-3096: add TS type to `Tray` component

NamNguyenInstructure:


ke1k0 (changes requested):

packages/ui-tray/src/Tray/index.tsx
```diff
@@ -57,23 +58,9 @@ class Tray extends Component<TrayProps> {
...
-    onClose: (event) => {},
-    // @ts-expect-error ts-migrate(6133) FIXME: 'event' is declared but its value is never read.
-    onDismiss: (event) => {},
-    onEnter: () => {},
-    onEntering: () => {},
-    onEntered: () => {},
-    onExit: () => {},
-    onExiting: () => {},
-    onExited: () => {},
     mountNode: null,
```
ke1k0, line 61: You can remove the props that are set to `null` from `defaultProps`, they are not needed anymore.

packages/ui-tray/src/Tray/index.tsx
```diff
@@ -194,8 +179,7 @@ class Tray extends Component<TrayProps> {
         {portalIsOpen && (
           <Transition
             in={open}
-            // @ts-expect-error ts-migrate(2769) FIXME: No overload matches this call.
-            type={this.transition}
+            type={this.transition as TransitionType}
```
ke1k0, line 182: It is better to type the getter, it just needs a little help so it knows what it returns:

```ts
get transition(): TransitionType {
    return `slide-${this.direction}`
  }
```

packages/ui-tray/src/Tray/index.tsx
```diff
@@ -109,12 +103,12 @@ class Tray extends Component<TrayProps> {
...
   }
 
-  get transition() {
+  get transition(): TransitionType {
     return `slide-${this.direction}`
   }
 
   get defaultFocusElement() {
-    return this.props.defaultFocusElement || (() => this._closeButton)
+    return this.props.defaultFocusElement
```
ke1k0, line 112: Now that the getter is only returning the prop, there is no point to use a getter, just directly access the prop where you need it.

PR #1039, opened 2022-06-21
feat(ui-responsive): add `display` prop to Responsive

ke1k0:
Closes: INSTUI-3541

The Responsive component is rendered as a `<div>`, so it has `display="block"` by default, but can
be changed with the new `display` prop.


packages/ui-responsive/src/Responsive/index.js
```diff
@@ -193,7 +206,7 @@ class Responsive extends Component {
       renderFunc = children || render
     }
     return (
-      <div>
+      <div style={{ display }}>
```
ke1k0, line 209: Not very elegant, but I decided to go with inline style, because it is not worth making the component themeable for this, especially in V7.

Brailor (approved):

PR #1047, opened 2022-06-27
docs(ui-drilldown,ui-editable,ui-position): remove `experimental` tag…

ke1k0:
… from the docs app

We've found that the "experimental" flag is misleading on the components, prevents users from using
the component, and are usually left there for long time. We've decided to remove it from our
documentation.

### DIFF
--- a/packages/__docs__/buildScripts/DataTypes.ts
+++ b/packages/__docs__/buildScripts/DataTypes.ts
@@ -42,7 +42,6 @@ type PackagePathData = {
 
 // YAML descriptions exported with gray-matter
 type YamlMetaInfo = {
-  experimental?: boolean
   category?: string
   describes?: string
   description: string
--- a/packages/__docs__/src/App/index.tsx
+++ b/packages/__docs__/src/App/index.tsx
@@ -465,13 +465,6 @@ class App extends Component<AppProps, AppState> {
         }
       >
         {this.renderThemeSelect()}
-        {currentData.experimental && (
-          <div>
-            <Pill color="info" margin="small 0">
-              Experimental
-            </Pill>
-          </div>
-        )}
         <Section id={currentData.id} heading={heading}>
           <Document
             doc={{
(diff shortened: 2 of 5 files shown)


Brailor:
I would not remove the capability to display experimental things in the docs, I would only remove the experimental flag from the components themself. (Editable, Drilldown)

matyasf:
> I would not remove the capability to display experimental things in the docs, I would only remove the experimental flag from the components themself. (Editable, Drilldown)

hmm I would take it out since we're not using it and are not planning to. What is an experimental component? Its released but should not be used in prod? Then why dont we test to more to make it not experimental?

ke1k0:
> > I would not remove the capability to display experimental things in the docs, I would only remove the experimental flag from the components themself. (Editable, Drilldown)
> 
> hmm I would take it out since we're not using it and are not planning to. What is an experimental component? Its released but should not be used in prod? Then why dont we test to more to make it not experimental?

Also, the current behaviour is pretty "black box", so if this every comes up as a request later, we can bring it back or come up with a better solution (probably connecting it with the decorator too).

Brailor:
> > I would not remove the capability to display experimental things in the docs, I would only remove the experimental flag from the components themself. (Editable, Drilldown)
> 
> hmm I would take it out since we're not using it and are not planning to. What is an experimental component? Its released but should not be used in prod? Then why dont we test to more to make it not experimental?

This experimental flag is not only useful for components, lets say we publish the docs generator as an npm package and then people could use it basically for anything they would like to have documentation for.

I can see a use-case for it, lets say you are a library author who publishes some API in a beta version and would like to somehow showcase that the API is not completely ready but it can be used with caution, also indicating that feedback about said API would be welcomed.

But this is kinda far fetched into the future, the reality is that probably noone will use it, so I'm up for deleting it and adding it back if it is needed by anyone ever.

Brailor (approved):

PR #1052, opened 2022-07-01
chore(ui-source-code-editor): update codemirror dependencies

ke1k0:
Closes: INSTUI-3510


matyasf (approved):
looks good. One question: For me voiceOver was reading what was unselected in the CodeEditor before too, what has changed?

ke1k0:
> looks good. One question: For me voiceOver was reading what was unselected in the CodeEditor before too, what has changed?

The change here is that when you select some text and delete it, the SR has to read it was deleted, the current live version doesn't (just reads the selection itself).

HerrTopi (approved):

PR #1054, opened 2022-07-04
INSTUI-3494 : support React 18

matyasf:
Note that it's still not running sometimes in parallel mode because we're still using `ReactDOM.render` in a few places (e.g. Alert)

To test:
Set React's version to 17 in the main package JSON's `resolutions` field and run some tests

### DIFF
--- a/packages/shared-types/package.json
+++ b/packages/shared-types/package.json
@@ -23,7 +23,9 @@
     "access": "public"
   },
   "dependencies": {
-    "prop-types": "^15",
-    "react": "^17.0.0"
+    "prop-types": "^15"
+  },
+  "peerDependencies": {
+    "react": ">=16.8 <=18"
   }
 }
--- a/packages/ui-flex/src/Flex/index.tsx
+++ b/packages/ui-flex/src/Flex/index.tsx
@@ -88,7 +88,8 @@ class Flex extends Component<FlexProps> {
   }
 
   renderChildren(children: FlexProps['children']) {
-    return Children.map(children, (child) => {
+    // TODO this might fail if children are () => ReactNode
+    return Children.map(children as React.ReactNode, (child) => {
       if (!child) {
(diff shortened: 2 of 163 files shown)


packages/shared-types/package.json
```diff
@@ -23,7 +23,9 @@
     "access": "public"
   },
   "dependencies": {
-    "prop-types": "^15",
-    "react": "^17.0.0"
```
matyasf, line 27: This was a bug, we should never add React directly as a dependency

packages/ui-flex/src/Flex/index.tsx
```diff
@@ -88,7 +88,8 @@ class Flex extends Component<FlexProps> {
   }
 
   renderChildren(children: FlexProps['children']) {
-    return Children.map(children, (child) => {
+    // TODO this might fail if children are () => ReactNode
```
matyasf, line 91: React 18 has better types for some things, if it uncovered a potential bug I've left a comment.

packages/ui-react-utils/src/callRenderProp.ts
```diff
@@ -43,21 +34,7 @@ import React, {
...
-        P,
-        ClassicComponent<P, ComponentState>,
-        ClassicComponentClass<P>
-      >
-    | ComponentClass<P>
-    | ReactNode
-    | (() => ReactNode),
-  props: P = {} as P
-) {
+function callRenderProp<P>(value: Renderable<P>, props: P = {} as P) {
```
matyasf, line 37: Extracting this type resulted in lots of small changes. This was needed because React 18 has better checks and we could broaden the type in cases where a prop value ended up as a parameter for this function.

HerrTopi (approved):
Seems nice. Tested as suggested

Brailor (approved):

PR #1062, opened 2022-07-12
refactor(ui-webpack-config): use `require.resolve` instead of raw strings

Brailor:
This change is needed in order to use our webpack config in the docs-generator project. The current behavior does not allow to use the config when the docs-generator is installed as a dependency of another project, this change fixes that.


packages/__examples__/.storybook/webpack.config.js
```diff
@@ -23,27 +23,24 @@
...
   // https://github.com/storybookjs/storybook/issues/12864
   const rules = baseConfig.module.rules
   for (const rule of rules) {
     if (rule.use) {
       for (let i = 0; i < rule.use.length; i++) {
-        if (rule.use[i].loader === 'thread-loader') {
+        if (
+          rule.use[i].loader &&
+          rule.use[i].loader.includes('thread-loader')
+        ) {
```
Brailor, line 38: Since the config now includes the full resolved path to `thread-loader` instead of it being a single string, this filtering logic had to be synchronized.

HerrTopi (approved):

matyasf (approved):

PR #1067, opened 2022-07-20
fix(ui-webpack-config): `iconExample` is missing from SVGIcon and Inl…

NamNguyenInstructure:
…ineSVG docs pages


ke1k0 (approved):
Looks good, thanks!

matyasf (approved):

PR #1070, opened 2022-07-20
INSTUI-3575 : fix ReactDOM.render warnings in docs

matyasf:
Update syntax to the Ract 18 one.

Also fix "The resource
https://instui-docs.s3.us-east-2.amazonaws.com/hero2.jpg was preloaded using link preload but not
used within a few seconds from the window's load event." warnings

### DIFF
--- a/packages/__docs__/src/Hero/index.tsx
+++ b/packages/__docs__/src/Hero/index.tsx
@@ -348,6 +348,7 @@ class Hero extends Component<HeroProps> {
             <Img
               src="https://instui-docs.s3.us-east-2.amazonaws.com/hero2.jpg"
               display="block"
+              rel="prefetch"
               constrain="cover"
               overlay={{
                 color: styles?.backgroundColor as string,
--- a/packages/__docs__/src/index.js
+++ b/packages/__docs__/src/index.js
@@ -23,16 +23,15 @@
  */
 
 import React from 'react'
-import ReactDOM from 'react-dom'
+import { createRoot } from 'react-dom/client'
 
 import { App } from './App'
 import { canvas } from '@instructure/ui-themes'
 import { InstUISettingsProvider } from '@instructure/emotion'
 import '../globals'
 
-ReactDOM.render(
+createRoot(document.getElementById('app')).render(
   <InstUISettingsProvider theme={canvas}>
     <App />
-  </InstUISettingsProvider>,
-  document.getElementById('app')
+  </InstUISettingsProvider>
 )
--- a/packages/__docs__/src/Preview/index.tsx
+++ b/packages/__docs__/src/Preview/index.tsx
@@ -24,17 +24,13 @@
 
 /** @jsx jsx */
 import React, { Component } from 'react'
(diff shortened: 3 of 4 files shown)


packages/__docs__/src/Hero/index.tsx
```diff
@@ -348,6 +348,7 @@ class Hero extends Component<HeroProps> {
             <Img
               src="https://instui-docs.s3.us-east-2.amazonaws.com/hero2.jpg"
               display="block"
+              rel="prefetch"
```
matyasf, line 351: This fixes the "The resource  was preloaded using link preload but not used within a few seconds" warnings.

packages/__docs__/src/index.js
```diff
@@ -23,16 +23,15 @@
...
 import { InstUISettingsProvider } from '@instructure/emotion'
 import '../globals'
 
-ReactDOM.render(
+createRoot(document.getElementById('app')).render(
   <InstUISettingsProvider theme={canvas}>
     <App />
-  </InstUISettingsProvider>,
-  document.getElementById('app')
+  </InstUISettingsProvider>
```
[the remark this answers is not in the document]
Brailor: Also maybe it would be a good idea to turn the docs-app into development mode when we run it with `yarn dev`.

packages/__docs__/src/index.js
```diff
@@ -23,16 +23,15 @@
...
 import React from 'react'
-import ReactDOM from 'react-dom'
+import { createRoot } from 'react-dom/client'
 
 import { App } from './App'
 import { canvas } from '@instructure/ui-themes'
 import { InstUISettingsProvider } from '@instructure/emotion'
 import '../globals'
 
-ReactDOM.render(
```
ke1k0, line 33: I can still see warnings, because the Alert component still has `ReactDOM.render` in it.

matyasf: yep, this just fixes them in the docs. One in `Preview` was really annoying, it was triggering the warning on every code change in the code editor.
This is more tricky to fix because:   
- we need to import `createRoot` asynchronously like here: https://github.com/instructure/instructure-ui/blob/master/packages/ui-test-sandbox/src/utils/reactComponentWrapper.tsx#L27 This might require other code changes too. 
- Removing a root can trigger sometimes scary sounding warnings, I have yet to figure out how to deal with them properly: https://stackoverflow.com/questions/73043828/how-to-unmount-something-created-with-createroot-properly 

So I'd do this in a separate ticket. Also AFAIK this just renders a very small subtree in the legacy slow mode, so it's not a big practical issue.

packages/__docs__/src/index.js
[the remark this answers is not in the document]
matyasf: Its already in dev mode. I've tried to turn on strict mode, but it causes other warnings and duplicate DOM elements to appear, fixing these would be out of scope for this PR. I've made a new ticket for this.

ke1k0 (approved):

Brailor (approved):

PR #1071, opened 2022-07-20
INSTUI-3576 fix TS type and propType for liveRegion prop

matyasf:
`liveRegion` can be a function that returns an array of DOM elements. Also its propType was
erroneously allowing React elements when it should only allow native DOM elements triggering false
prop validation errors.

### DIFF
--- a/packages/ui-drawer-layout/src/DrawerLayout/DrawerTray/props.ts
+++ b/packages/ui-drawer-layout/src/DrawerLayout/DrawerTray/props.ts
@@ -206,7 +206,7 @@ const propTypes: PropValidators<PropKeys> = {
   liveRegion: PropTypes.oneOfType([
     PropTypes.arrayOf(PropTypes.elementType),
     PropTypes.elementType,
-    PropTypes.number
+    PropTypes.func
   ]),
   onDismiss: PropTypes.func,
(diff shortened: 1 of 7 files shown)


packages/ui-drawer-layout/src/DrawerLayout/DrawerTray/props.ts
```diff
@@ -204,9 +204,9 @@ const propTypes: PropValidators<PropKeys> = {
   mountNode: PropTypes.oneOfType([element, PropTypes.func]),
   defaultFocusElement: PropTypes.oneOfType([PropTypes.element, PropTypes.func]),
   liveRegion: PropTypes.oneOfType([
-    PropTypes.arrayOf(PropTypes.element),
-    PropTypes.element,
-    PropTypes.func
+    PropTypes.arrayOf(PropTypes.elementType),
+    PropTypes.elementType,
+    PropTypes.number
```
ke1k0, line 209: This `number` type seems to be unintentional :)

matyasf: yep, fixed

ke1k0 (approved):

HerrTopi (approved):

PR #1075, opened 2022-07-25
Add new TopNavBar component skeleton

ke1k0:


packages/instui-cli/lib/commands/create.js
```diff
@@ -47,6 +47,7 @@ exports.builder = (yargs) => {
     })
   }
 
+  // TODO: remove create app script and any related code
```
matyasf, line 50: We should create a ticket about this. IMO it's not a breaking change, we told people that we only have the "no breaking change without major version release" policy only for UI code, not for configuration or build scripts.

packages/template-component/package.json
```diff
@@ -16,6 +16,7 @@
   "dependencies": {
     "@babel/runtime": "^7.13.10",
     "@instructure/emotion": "8.27.0",
+    "@instructure/shared-types": "8.27.0",
     "@instructure/ui-react-utils": "8.27.0",
     "@instructure/ui-testable": "8.27.0",
     "prop-types": "^15"
```
matyasf, line 22: the `ui-color-utils` is not needed as a devDependency on line 26 AFAIK

packages/instui-cli/lib/commands/create.js
```diff
@@ -63,35 +64,50 @@ exports.builder = (yargs) => {
   )
 
   const packageCommand = 'package'
+  const packageFormatInstructions = ' (in kebab-case, e.g. my-package)'
   yargs.command(
     packageCommand,
-    'Create an Instructure UI package. DEPRECATED, this will be deleted in InstUI 9',
+    'Create an Instructure UI package. FOR INTERNAL USE ONLY!',
```
HerrTopi, line 70: I am not sure if we truly want to have this internally only. I think anyone could and should use this, just don't assume that it won't have breaking changes

packages/instui-cli/lib/commands/create.js
```diff
@@ -63,35 +64,50 @@ exports.builder = (yargs) => {
...
+      })
     }
   )
 
   const componentCommand = 'component'
+  const componentFormatInstructions = ' (in PascalCase, e.g. NumberInput)'
   yargs.command(
     componentCommand,
-    'Create an Instructure UI component. DEPRECATED, this will be deleted in InstUI 9',
+    'Create an Instructure UI component. FOR INTERNAL USE ONLY!',
```
HerrTopi, line 94: I am not sure if we truly want to have this internally only. I think anyone could and should use this, just don't assume that it won't have breaking changes

HerrTopi (approved):

packages/template-component/package.json
ke1k0: It is used in the `theme.test.ts.ejs`

packages/instui-cli/lib/commands/create.js
ke1k0: INSTUI-3580

packages/instui-cli/lib/commands/create.js
ke1k0: We agreed on the standup that we should leave it "internal use only" for now, because it would require a lot more of setup, documentation and guides to make it easy-to-use for other teams too.

packages/ui-top-nav-bar/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "private": true,
+  "name": "@instructure/ui-top-nav-bar",
+  "version": "8.27.0",
+  "description": "A UI component library made by Instructure Inc.",
```
matyasf, line 5: This should be a more specific description. Something like "A responsive top navigation bar with light and dark theme"

matyasf (approved):
looks good!

packages/ui-top-nav-bar/package.json
ke1k0: Ofc, this is just the skeleton for the folder placeholder, I'll do the rest on the feature branch :)

PR #1080, opened 2022-08-05
chore(ui-color-picker): fix color palette's indicator

NamNguyenInstructure:


packages/ui-color-picker/src/ColorMixer/ColorPalette/index.tsx
```diff
@@ -93,6 +93,17 @@ class ColorPalette extends Component<ColorPaletteProps, ColorPaletteState> {
   componentWillUnmount() {
     this.removeEventListeners()
   }
+  get padding() {
```
NamNguyenInstructure, line 96: I add one `padding` property to shrink the area where the indicator can move, this will make the component look nicer

packages/ui-color-picker/src/ColorMixer/ColorPalette/index.tsx
```diff
@@ -148,17 +160,17 @@ class ColorPalette extends Component<ColorPaletteProps, ColorPaletteState> {
   applyBoundaries(x: number, y: number) {
     let newXPosition = x
     let newYPosition = y
-    if (x > this.paletteWidth) {
-      newXPosition = this.paletteWidth
+    if (x > this.paletteWidth - this.padding) {
```
NamNguyenInstructure, line 163: Because now we have the padding so instead of moving from 0 to `props.width` or `props.height`, the indicator now will move from `padding` to `props.width-padding`

packages/ui-color-picker/src/ColorMixer/ColorPalette/styles.ts
```diff
@@ -60,10 +60,10 @@ const generateStyle = (
       borderWidth: componentTheme.indicatorBorderWidth,
       borderColor: componentTheme.indicatorBorderColor,
       top: `${
-        state?.colorPosition?.y - px(`${props.indicatorRadius / 16}rem`)
+        state?.colorPosition?.y - 1 - px(`${props.indicatorRadius / 16}rem`)
```
NamNguyenInstructure, line 63: Because the indicator has a border so we need to subtract one more pixel to place the center of the indicator in the right position

packages/ui-color-picker/src/ColorMixer/ColorPalette/index.tsx
```diff
@@ -102,17 +113,18 @@ class ColorPalette extends Component<ColorPaletteProps, ColorPaletteState> {
     return this._paletteRef!.getBoundingClientRect().height
   }
   calcSaturation = (position: number) =>
-    Math.round((position / this.paletteWidth) * 100) / 100
+    Math.round(((position - this.padding) / this.movableWidth) * 100) / 100
```
NamNguyenInstructure, line 116: Now the movable area of the indicator is smaller, we need to adjust the color calculation formula

packages/ui-color-picker/src/ColorMixer/ColorPalette/index.tsx
```diff
@@ -63,6 +63,8 @@ class ColorPalette extends Component<ColorPaletteProps, ColorPaletteState> {
   private _paletteRef: HTMLDivElement | null = null
   private _mouseMoveListener?: { remove(): void }
   private _mouseUpListener?: { remove(): void }
+  //@ts-expect-error Property 'window' does not exist on type 'ColorPalette'
+  private _paletteOffset = (getFontSize(this.window) / 16) * 2
```
ke1k0, line 67: I would use the `px()` util, then you don't need the ts-expect-error. This util gives back the px value of '0.125rem', depending on the current document font size.

There is another problem: this value is still hardcoded. As far as I understand, this value comes from somewhere, (I assume it is "2 times the indicator border width", but you are more familiar with this). These are usually theme variables, meaning they change if someone overrides them. Try to make it dependent on the theme variable connected to this logic. Let me know if you need help with it. 

```suggestion
  private _paletteOffset = px('0.125rem')
```

matyasf (approved):
looks good!

ke1k0 (approved):
I would've liked better if the offset was a calculated value, but now at least it can be overridden as a theme variable, which is good for now

PR #1087, opened 2022-08-24
chore: remove Google Analytics

NamNguyenInstructure:


packages/__docs__/src/index.html
```diff
@@ -108,32 +108,6 @@
     </style>
 
     <meta name="description" content="Instructure UI" />
-
```
ke1k0, line 111: I think this code is connected to the analytics too: 

```tsx
trackPage(page: string) {
    let title = this._defaultDocumentTitle
    if (page !== 'index') {
      title = `${page} - ${this._defaultDocumentTitle}`
    }

    document.title = title!

    const ga = (window as any).ga
    if (ga) {
      ga('set', 'page', page)
      ga('set', 'title', title)
      ga('send', 'pageview')
    }
  }
```

ke1k0 (approved):
Seems good!

PR #1088, opened 2022-08-24
fix: fix propType for liveRegion

matyasf:
Note that this requires a polyfill for Element if you do server-side-rendering.


ke1k0 (approved):
Seems good, just please remind us to manually put the warning in the change log and release notes when we release it :)

NamNguyenInstructure (approved):

PR #1089, opened 2022-08-25
React 18 fixes

Brailor:


packages/__docs__/src/App/index.tsx
```diff
@@ -123,36 +124,12 @@ class App extends Component<AppProps, AppState> {
...
-            themeKey: Object.keys(docsData.themes)[0]
-          },
-          this.scrollToElement
-        )
-      })
-      .catch((error) => {
-        // eslint-disable-next-line no-console
-        console.error('Unable to load docs data :(\n' + error)
-      })
   }
```
Brailor, line 127: These had to be moved out from the constructor since these fetch calls are side effects and we have to be able to cancel them.

packages/__docs__/src/index.js
```diff
@@ -31,7 +31,9 @@ import { InstUISettingsProvider } from '@instructure/emotion'
...
 
 createRoot(document.getElementById('app')).render(
-  <InstUISettingsProvider theme={canvas}>
-    <App />
-  </InstUISettingsProvider>
+  <React.StrictMode>
+    <InstUISettingsProvider theme={canvas}>
+      <App />
+    </InstUISettingsProvider>
+  </React.StrictMode>
```
Brailor, line 38: Although the docs will be much slower in dev mode (because of the double mounting of `React.StrictMode`), we will able to catch potential bugs much sooner.

packages/ui-drilldown/src/Drilldown/index.tsx
```diff
@@ -990,14 +1010,6 @@ class Drilldown extends Component<DrilldownProps, DrilldownState> {
...
     }
 
-    if (this.getPageChildById(id)) {
-      warn(
-        false,
-        `Duplicate id: "${id}"! Make sure all options have unique ids, otherwise they won't be rendered.`
-      )
-      return null
-    }
-
```
Brailor, line 1000: Because of this check and of the double mounting the drilldown did not render any options in `React.StrictMode` since the child with the corresponding `id` would be already in the "cache" and thus this logic would be true.
I moved up the logic for checking duplicate id-ed children to the `renderList` function.

packages/ui-alerts/src/Alert/index.tsx
```diff
@@ -23,7 +23,7 @@
  */
 /** @jsx jsx */
 import React, { Component } from 'react'
-import ReactDOM from 'react-dom'
+import { createRoot, Root } from 'react-dom/client'
```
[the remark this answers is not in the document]
Brailor: I ended up refactoring the `Alert` component to not to use `ReactDOM.render`/`React.createRoot.render` API, instead it will render the screen reader content using `React.Portal`. Doing this made the component much easier to understand without any breaking changes.

packages/ui-alerts/src/Alert/__tests__/Alert.test.tsx
```diff
@@ -174,7 +174,10 @@ describe('<Alert />', async () => {
         </Alert>
       )
 
-      expect(subject.getDOMNode()).to.not.exist()
+      const root = document.querySelector('[data-ui-test-utils]')!
+
+      expect(root.children.length).to.equal(0)
+      expect(liver.children.length).to.equal(1)
     })
```
Brailor, line 181: I had to change this test because the component will return the result of `React.createPortal` when the `screenReaderOnly` prop is true and thus the `subject.getDOMNode()` would return the rendered content in the Portal.
Instead it will check whether the root of the React tree is empty or not.

ke1k0 (approved):
Seems good!

matyasf (changes requested):
Looks good except that the docs app crashes in the Controlled TimeSelect example when I enter a non-existent variable value

packages/ui-test-sandbox/src/utils/reactComponentWrapper.tsx
```diff
@@ -26,7 +26,7 @@ import React from 'react'
 import ReactDOM from 'react-dom'
 const createRoot = async () => {
   // eslint-disable-next-line import/no-unresolved
-  return await import('react-dom/client')
+  return import('react-dom/client')
```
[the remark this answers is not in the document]
Brailor: Probably, i'm not 100% sure tough

Brailor:
> Looks good except that the docs app crashes in the Controlled TimeSelect example when I enter a non-existent variable value

Should be fixed

packages/__docs__/src/Preview/index.tsx
```diff
@@ -48,14 +47,22 @@ class Preview extends Component<PreviewProps, PreviewState> {
...
-  private _root?: Root
+  static getDerivedStateFromProps(props: PreviewProps, state: PreviewState) {
+    if (props.error) {
+      return {
+        ...state,
+        error: props.error
+      }
+    }
+    return null
+  }
```
Brailor, line 58: If we get an error from the `PreviewErrorBoundary`, then map that error into the state.

packages/__docs__/src/Preview/index.tsx
```diff
@@ -154,5 +142,35 @@ class Preview extends Component<PreviewProps, PreviewState> {
...
+          error: this.state.error
+        })
+      }
+      return child
+    })
+
+    return children
+  }
+}
+
```
Brailor, line 174: This component had to be added since the code coming from `compileAndRenderExample` could contain invalid code and that invalid code would cause an error during rendering inside the `Preview` component. To prevent that from happening, this `PreviewErrorBoundary` will 'catch' the error, then passes down to the `Preview` and then `Preview` can handle the error as usual.

PR #1094, opened 2022-09-01
fix(ui-tabs): tab bar is not scrollable on the left side

NamNguyenInstructure:


matyasf (changes requested):
Functionally it works OK with mouse clicks but it would be necessary to **have the same gradient on the left side as on the right if there are still elements overflowing on the left side**. Imagine that the user is navigated to a page where there is a tab bar with such a selection that there are elements overflowing to the left, in this case the user has no visual indicator that there are more elements.

Also the navivation is buggy with keyboard arrows, but this should be another ticket and PR

packages/ui-tabs/src/Tabs/index.tsx
```diff
@@ -113,7 +113,9 @@ class Tabs extends Component<TabsProps, TabsState> {
     super(props)
 
     this.state = {
```
NamNguyenInstructure, line 115: I added two states to specify the appearance of the overlay

packages/ui-tabs/src/Tabs/index.tsx
```diff
@@ -228,18 +230,11 @@ class Tabs extends Component<TabsProps, TabsState> {
       this._tabListPosition &&
       typeof this._tabList.scrollTo === 'function' // test for scrollTo support
     ) {
```
NamNguyenInstructure, line 232: Now we add the overlay on both ends so direction is no longer needed.

packages/ui-tabs/src/Tabs/styles.ts
```diff
@@ -131,6 +131,21 @@ const generateStyle = (
         background: `linear-gradient(to right, ${componentTheme.scrollFadeColor} 0%, rgba(255, 255, 255, 0) 100%)`
       }
     },
+    leftScrollOverlay: {
+      label: 'tabs__scrollOverlay',
+      height: `calc(100% - (${componentTheme.tabVerticalOffset} + 0.25rem))`,
```
ke1k0, line 136: It is not necessary, but I recommend refactoring the common overlay styles into a const (that are the same for both the left and right overlay), because then you don't have to update it in 2 places later.

packages/ui-tabs/src/Tabs/styles.ts
```diff
@@ -116,7 +116,7 @@ const generateStyle = (
       })
     },
 
-    scrollOverlay: {
+    rightScrollOverlay: {
       label: 'tabs__scrollOverlay',
```
ke1k0, line 120: The label and the property name have to match so it can be easily identifiable in the rendered code. Also, since we are writing bidirectional components, I would recommend naming them with "start" instead of "left" and "end" instead of "right".

```suggestion
    endScrollOverlay: {
      label: 'tabs__endScrollOverlay',
```

ke1k0 (changes requested):

packages/ui-tabs/src/Tabs/__tests__/Tabs.test.tsx
```diff
@@ -448,18 +448,17 @@ describe('<Tabs />', async () => {
       const subject = await mount(<Example width="150px" />)
       const tabs = await TabsLocator.find()
 
-      expect(await tabs.find('[class$="-tabs__scrollOverlay"]')).to.exist()
-      expect(await tabs.find('[class$="-tabs__scrollSpacer"]')).to.exist()
+      expect(await tabs.find('[class$="-tabs__endScrollOverlay"]')).to.exist()
+
+      const tabPanels = await tabs.findAllTabPanels()
+      tabPanels[2].click()
```
NamNguyenInstructure, line 454: I'm not sure how to test the scrolling event so I choose to click  the middle tab

ke1k0 (approved):

PR #1102, opened 2022-09-19
docs: provide link to the Github theme variables

NamNguyenInstructure:
Closes: INTUI-3548

### DIFF
--- a/packages/__docs__/buildScripts/utils/getPathInfo.ts
+++ b/packages/__docs__/buildScripts/utils/getPathInfo.ts
@@ -23,6 +23,7 @@
  */
 
 import path from 'path'
+import fs from 'fs'
 import type { LibraryOptions, PackagePathData } from '../DataTypes'
 
 export function getPathInfo(
@@ -30,7 +31,7 @@ export function getPathInfo(
   projectRoot: string,
   library: LibraryOptions
 ): PackagePathData {
-  return {
+  const pathInfo: PackagePathData = {
     relativePath: path.relative(projectRoot, resourcePath),
     extension: path.extname(resourcePath),
     srcPath: relativePath(resourcePath, projectRoot),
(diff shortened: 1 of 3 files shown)


packages/__docs__/src/Document/index.tsx
```diff
@@ -199,6 +200,28 @@ class Document extends Component<DocumentProps, DocumentState> {
...
+
+    let legacySrcUrl
+
+    if (legacyGitBranch) {
+      legacySrcUrl = themeUrl.replace('/master/', `/${legacyGitBranch}/`)
+    }
+
+    return (
+      <View as="div" margin="0 0 x-large 0">
+        You can check these variables at{' '}
```
ke1k0, line 219: Text suggestion: `See which global theme variables are mapped to the component here: [link]`

packages/__docs__/src/Document/index.tsx
```diff
@@ -199,6 +200,28 @@ class Document extends Component<DocumentProps, DocumentState> {
...
+        <Link href={legacySrcUrl || themeUrl}>{themePath}</Link>
```
ke1k0, line 220: I would also use this link in the block below on the "default theme variables" text.

Screenshot 2022-09-20 at 10 10 45

ke1k0 (changes requested):

matyasf (approved):
nice!

ke1k0 (commented):
Please merge the 2 commits into one (generally it is not nice to have "fix pr" commits littering the git history), and I think the commit type should be "docs" instead of "refactor".

packages/__docs__/buildScripts/utils/getPathInfo.ts
```diff
@@ -23,14 +23,15 @@
...
+import fs from 'fs'
 import type { LibraryOptions, PackagePathData } from '../DataTypes'
 
 export function getPathInfo(
   resourcePath: string,
   projectRoot: string,
   library: LibraryOptions
 ): PackagePathData {
-  return {
+  const ans: PackagePathData = {
```
ke1k0, line 34: I'd prefer more semantic names here, like `const: packagePathData: PackagePathData = { ... }`

ke1k0 (changes requested):
Just had a small comment, otherwise looks good!

ke1k0 (approved):

PR #1104, opened 2022-09-21
feat(ui-time-select): allow non-step input value with new prop

matyasf:
When setting `allowNonStepInput` to `true` one can enter any valid time value. It needs to be typed exactly (except for upper/lowercase)

### DIFF
--- a/packages/ui-time-select/src/TimeSelect/README.md
+++ b/packages/ui-time-select/src/TimeSelect/README.md
@@ -5,6 +5,7 @@ describes: TimeSelect
 `TimeSelect` component is a higher level abstraction of [Select](#Select) specifically for selecting time values. The list of possible values can be configured via the component's props. If you need to select times and dates, be sure to check out the documentation around [Time and Date patterns](#TimeDate).
 
 ### Uncontrolled
+
 For the most basic implementations, `TimeSelect` can be uncontrolled. If desired, the `defaultValue` prop can be used to set the initial selection.
 
 ```javascript
@@ -15,11 +16,13 @@ render: true
 <TimeSelect
   renderLabel="Choose a time"
   onChange={(e, { value }) => console.log(value)}
+  onHideOptions={(e)=> console.log("hide opts")}
   // defaultValue={new Date().toISOString()}
 />
 ```
 
 ### Controlled
+
 To use `TimeSelect` controlled, simply provide the `value` prop an ISO string. The `onChange` callback provides the ISO value of the corresponding option that was selected. Use this value to update the state.
 
 ```javascript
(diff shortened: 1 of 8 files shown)


packages/ui-time-select/src/TimeSelect/README.md
```diff
@@ -15,11 +16,13 @@ render: true
 <TimeSelect
   renderLabel="Choose a time"
   onChange={(e, { value }) => console.log(value)}
+  onHideOptions={(e)=> console.log("hide opts")}
```
ke1k0, line 19: Is this needed?

packages/ui-time-select/src/TimeSelect/props.ts
```diff
@@ -205,6 +205,23 @@ type TimeSelectOwnProps = {
    * property.
    */
   timezone?: string
+  /**
+   * Whether to allow the user to enter non-step divisible values in the input field.
+   * Note that even if this is set to `false` one can enter non-step divisible values programatically.
+   * The user will need to enter the value exactly (ecxept for lower/uppercase) as specified by the `format` prop  for
```
ke1k0, line 211:
```suggestion
   * The user will need to enter the value exactly (except for lower/uppercase) as specified by the `format` prop  for
```

packages/ui-time-select/src/TimeSelect/index.tsx
```diff
@@ -330,17 +330,20 @@ class TimeSelect extends Component<TimeSelectProps, TimeSelectState> {
...
       highlightedOptionId: newOptions.length > 0 ? newOptions[0].id : undefined,
       isShowingOptions: true
     })
-
     if (!this.state.isShowingOptions) {
       this.props.onShowOptions?.(event)
     }
+    this.props.onInputChange?.(
+      event,
+      value,
```
ke1k0, line 344: Somehow we need to pass the date/time value as well, not just the text value. It will be needed, in case someone wants to use the new prop in a controlled mode.

ke1k0 (commented):
Seems to work fine! I had a few comments, and also, can you please squash the commits a bit?

packages/ui-time-select/src/TimeSelect/README.md
matyasf: I think its useful because it shows users the order of the events firing

packages/ui-time-select/src/TimeSelect/index.tsx
matyasf: I've added a new field thats an ISO datetime string

NamNguyenInstructure (approved):

ke1k0 (approved):
please resolve the conflict and I would squash the commits to keep the git history tidy, but other then it seems okay

PR #1105, opened 2022-09-23
fix(ui-text-input): remove focusring when receiving disabled prop

HerrTopi:
Closes: INSTUI-3638

### DIFF
--- a/packages/ui-text-input/src/TextInput/index.tsx
+++ b/packages/ui-text-input/src/TextInput/index.tsx
@@ -140,6 +140,15 @@ class TextInput extends Component<TextInputProps, TextInputState> {
       })
     }
 
+    if (
+      getInteraction({ props: prevProps }) !== 'disabled' &&
+      getInteraction({ props: this.props }) === 'disabled'
+    ) {
+      this.setState({
+        hasFocus: false
+      })
+    }
+
     this.props.makeStyles?.(this.makeStyleProps())
   }
 


packages/ui-text-input/src/TextInput/index.tsx
```diff
@@ -140,6 +140,17 @@ class TextInput extends Component<TextInputProps, TextInputState> {
       })
     }
 
+    if (
+      getInteraction({ props: { interaction: prevProps.interaction } }) !==
+        'disabled' &&
+      getInteraction({ props: { interaction: this.props.interaction } }) ===
+        'disabled'
+    ) {
```
ke1k0, line 148: IMO this should be either `props.interaction === 'disabled'`, if you want to check for the props only, or `getInteraction(props) === 'disabled'`, if you want to check globally. This util also checks for the html `disabled` prop (both `<TextInput disabled {...} />` and `<TextInput interaction="disabled" {...} />`), so I would pass the whole props object.

packages/ui-text-input/src/TextInput/index.tsx
```diff
@@ -140,6 +140,17 @@ class TextInput extends Component<TextInputProps, TextInputState> {
...
+      this.setState({
+        hasFocus: false
```
ke1k0, line 150: I would also tell Omar that putting the focus back on the TextInput after the overlay is gone should happen on their side, and it can be easily achieved with the public `focus()` method on the component.

```tsx
  // Get the ref in the render method:
  render() {
    return (
      <div position="relative">
        <Select
          ref={(textInput)=> { this._input = textInput }}
          {...}
        </Select>
     </div>
    )
  }

  // use the focus method after the overlay is gone
  componentDidUpdate(prevState) {
    if (
      prevState.isOverlayActive !== this.state.isOverlayActive &&
      this.state.isOverlayActive === true
    ) {
      setTimeout(() => {
        this.setState({ isOverlayActive: false }, () => {
          this._input.focus()
          console.log("active element after", document.activeElement);
        });
      }, 2000);
    }
  }
```

HerrTopi: Yeah, I will mention this, but it seems, that he doesn't have an issue with the focus management, only with the bug.

packages/ui-text-input/src/TextInput/index.tsx
HerrTopi: Yeah, I might as well. I think the whole interaction util is overkill, but since the code has it everywhere when accessing value from `interaction` prop, I'd keep it for consistency's sake.

ke1k0 (approved):

Brailor (approved):

PR #1112, opened 2022-10-13
fix(shared-types,ui-color-picker): css problems

NamNguyenInstructure:


ke1k0 (changes requested):
When I checked the preview, I saw that the popover of the ColorPreset component still looks kinda weird. It has an empty scrollbar too (because of the `overflow: scroll` instead of `overflow: auto`), and the spacing between the blocks doesn't seem nice either.
image

NamNguyenInstructure:
Hmm, that's weird because when I check it on my computer, it's ok. I have tested with Chrome, Firefox, and Safari. Do you have free time between 2 and 4? We can find out why.
Screenshot 2022-10-17 at 11 29 27

ke1k0:
Not the ColorPicker, but the ColorPreset, here: https://1112--preview-instui.netlify.app/#ColorPreset

HerrTopi (approved):

ke1k0 (dismissed):

ke1k0 (commented):
Seems good! My only request is that please merge the 2 commits, because like this they will appear like 2 separate changes in the Changelog

ke1k0 (approved):

PR #1126, opened 2022-11-04
Lodash tree-shaking via babel plugin

jacekk:
Fix similar to this one: https://github.com/instructure/instructure-ui/pull/1015 , but solved in a different way.
I am totally surprised it was reverted in https://github.com/instructure/instructure-ui/pull/1017 , and I cannot figure out who (and why) decided to for such a step **backward**.

Direct one-by-one imports are the most effective way to make the bundle slim when it comes to lodash.
The other way is to use a babel plugin. Ref: https://www.blazemeter.com/blog/import-lodash-libraries
This PR uses the plugin.

---

In my case, I am using `@instructure/emotion` and couple simple UI components from this library.

**BEFORE:**

before--emotion-fix

---

**AFTER:**

after--babel-fix

---

A change similar to #1015 would result in a similar bundle size:

after--emotion-fix

but I guess someone decided to not maintain one-by-one imports.

### DIFF
--- a/packages/ui-babel-preset/lib/index.js
+++ b/packages/ui-babel-preset/lib/index.js
@@ -101,7 +101,8 @@ module.exports = function (
       }
     ],
     require('@babel/plugin-syntax-dynamic-import').default,
-    require('babel-plugin-transform-undefined-to-void')
+    require('babel-plugin-transform-undefined-to-void'),
+    require('babel-plugin-lodash')
   ])
 
   if (process.env.NODE_ENV === 'production') {
--- a/packages/ui-babel-preset/package.json
+++ b/packages/ui-babel-preset/package.json
@@ -36,6 +36,7 @@
     "babel-loader": "^8.1.0",
     "babel-plugin-dynamic-import-node": "^2.3.0",
     "babel-plugin-istanbul": "^6.0.0",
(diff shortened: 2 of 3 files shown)


Brailor:
Hi @jacekk!

Thanks for the pull request!

The reason we decided to revert the PR you mentioned is not because we did not want to maintain the one-by-one imports, but because we had concerns that the change might negatively affect the bundle size of Canvas.

According to my knowledge, no experiments had been made to test this theory in practice, so I think we reverted the PR a little bit prematurely.

It would be great to merge this PR, or do the one-by-one imports manually because it *can* reduce bundle sizes in certain cases (I did a little [experiment](https://github.com/Brailor/test-lodash-import) myself and saw great reduction). 

The concern is still the bundle size of Canvas, it would be great to somehow test it out what happens when we use the one-by-one import approach.

@cvkline @eschiebel, can we somehow collaborate and make this PR happen?

Brailor:
Hi @jacekk!

Just a quick update:

I realized that the instui project only utilizes `lodash` in a couple of places internally (and also did not use that many functions from it), so I decided to get rid of it entirely.
Here is the PR about that change: #1128 

Once it is reviewed and merged, I will create a release for it.

Brailor:
Closing this since #1128 has been merged.

PR #1128, opened 2022-11-15
Remove `lodash` dependency from component packages

Brailor:
This PR removes our dependency on `lodash` for consumer facing packages (emotion, ui-source-code-editor), this could be achieved by using our internal utility functions instead of relying on `lodash`. 

I created 2 applications with the same code, but with different @instrucutre/ui dependencies:
- one with `lodash` still used internally:
Screenshot 2022-11-17 at 15 39 56

- another one with this PR applied:
Screenshot 2022-11-17 at 15 40 05

### DIFF
--- a/packages/emotion/src/getTheme.ts
+++ b/packages/emotion/src/getTheme.ts
@@ -21,10 +21,9 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-import { merge, cloneDeep } from 'lodash'
 import { canvas } from '@instructure/ui-themes'
 import { ThemeRegistry } from '@instructure/theme-registry'
-import { isBaseTheme } from '@instructure/ui-utils'
+import { isBaseTheme, mergeDeep } from '@instructure/ui-utils'
 
 import type { BaseTheme } from '@instructure/shared-types'
 
@@ -81,11 +80,9 @@ const getTheme =
           'No theme provided for [InstUISettingsProvider], using default `canvas` theme.'
         )
       }
-      //TODO: replace cloneDeeps with native `structuredClone` API
-      //once we hit last 2 version browser support
-      currentTheme = cloneDeep(globalTheme || canvas)
+      currentTheme = globalTheme || canvas
     } else {
-      currentTheme = cloneDeep(ancestorTheme)
+      currentTheme = ancestorTheme
     }
 
     const themeName = currentTheme.key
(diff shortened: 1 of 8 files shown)


packages/emotion/src/getTheme.ts
```diff
@@ -81,11 +80,9 @@ const getTheme =
           'No theme provided for [InstUISettingsProvider], using default `canvas` theme.'
         )
       }
-      //TODO: replace cloneDeeps with native `structuredClone` API
-      //once we hit last 2 version browser support
-      currentTheme = cloneDeep(globalTheme || canvas)
+      currentTheme = globalTheme || canvas
     } else {
-      currentTheme = cloneDeep(ancestorTheme)
+      currentTheme = ancestorTheme
```
Brailor, line 85: These `cloneDeep`'s were unnecessary because we don't modify the source objects.

packages/ui-babel-preset/lib/index.js
```diff
@@ -176,7 +176,8 @@ function getWebEnvConfig(opts) {
       browsers: require('@instructure/browserslist-config-instui')
     },
     useBuiltIns: 'entry',
-    corejs: 3,
+    // this version has to match the version in package.json
+    corejs: '3.26.1',
```
Brailor, line 180: According to the [babel docs](https://babeljs.io/docs/en/babel-preset-env#corejs) this has to be a string.

packages/ui-source-code-editor/src/SourceCodeEditor/index.tsx
```diff
@@ -469,9 +469,9 @@ class SourceCodeEditor extends Component<SourceCodeEditorProps> {
 
     if (this.direction === 'rtl' && !rtlMoveVisually) {
       // we need to clone the original so that it doesn't get overridden
-      return merge(
-        cloneDeep(defaultKeymap),
-        rtlHorizontalArrowKeymap
+      return mergeDeep(
+        defaultKeymap as unknown as Record<string, unknown>,
+        rtlHorizontalArrowKeymap as unknown as Record<string, unknown>
```
Brailor, line 474: The original `merge` is actually does not work the way we expect, thus creating a bug in this logic, for which I created a separate bug ticket.

So I think this is fine to merge and fix the bug in another PR.

ke1k0: I would still at least change it to `[...defaultKeymap, ...rtlHorizontalArrowKeymap]`, since we realized that they are arrays, and mergeDeep accepts objects :) And then leave a comment to fix the broken logic

ke1k0 (approved):

HerrTopi (approved):

PR #1139, opened 2022-12-21
feat(ui-drilldown): allow controlled functionality when selecting options

Brailor:
Closes: INSTUI-3685

users now are allowed to set a newly added `selected` property on `Drilldown.Option` which allows them to control the

selection logic from outside of the component fix `selectableType=single`allowing de-selecting options
TEST PLAN:
run the test cases, verify that
[this](https://codesandbox.io/s/flamboyant-glitter-2dt3bg?file=/index.js) use case is working as expected

### DIFF
--- a/packages/ui-drilldown/src/Drilldown/index.tsx
+++ b/packages/ui-drilldown/src/Drilldown/index.tsx
@@ -57,7 +57,6 @@ import { withStyle, jsx } from '@instructure/emotion'
 
 import { DrilldownSeparator } from './DrilldownSeparator'
 import { DrilldownOption } from './DrilldownOption'
-import type { DrilldownOptionValue } from './DrilldownOption/props'
 import { DrilldownGroup } from './DrilldownGroup'
 import type { DrilldownGroupProps, GroupChildren } from './DrilldownGroup/props'
 import { DrilldownPage } from './DrilldownPage'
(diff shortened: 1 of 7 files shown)


ke1k0:
I just tried the preview page, and the multi-select example throws an error when I try to navigate to a subpage: 
image

Also, the single-select fix has to be a separate commit so that it shows up in the changelog.

packages/ui-drilldown/src/Drilldown/index.tsx
```diff
@@ -1050,6 +1058,8 @@ class Drilldown extends Component<DrilldownProps, DrilldownState> {
       tabIndex: -1
     }
 
+    // console.log({ optionProps })
```
ke1k0, line 1061: not needed

ke1k0 (changes requested):
I commented on the PR itself:
- the multi-select example throws error
- he single-select fix has to be a separate commit

packages/ui-drilldown/src/Drilldown/index.tsx
Brailor: Removed

ke1k0 (approved):
Seems good, thanks for separating the commits. I approved the PR, but if you have the time, it would be nice to add an example on the docs page to show this new controlled behaviour (or shall we try to keep it secret? XD)

Brailor:
> Seems good, thanks for separating the commits. I approved the PR, but if you have the time, it would be nice to add an example on the docs page to show this new controlled behaviour (or shall we try to keep it secret? XD)

For now I would keep it a "secret" until we are sure that this is working as expected.

HerrTopi (approved):

PR #1145, opened 2023-01-09
INSTUI-3701 fix(ui-modal): fix screenreader focus order in Chrome

matyasf:
Screenreader was skipping modal content in Chrome because we made a View focusable. This fix makes the View unfocusable


ke1k0 (changes requested):
Sadly I could reproduce the error in Firefox, removing this line disables tabbing in the modal. :(

Brailor (approved):

ke1k0 (approved):

PR #1147, opened 2023-01-13
fix(ui-a11y-utils): fix type for removedNodes and addedNodes

NamNguyenInstructure:


packages/ui-a11y-utils/src/ScreenReaderFocusRegion.ts
```diff
@@ -115,22 +119,25 @@ class ScreenReaderFocusRegion {
...
+          // Node has been removed from the DOM, make sure it is
+          // removed from our list of hidden nodes as well
+
+          if (removedNode.tagName.toLowerCase() !== 'iframe') {
+            this.restoreNode(removedNode as Element)
+          }
+          const iframeBodies = this.parseIframeBodies(removedNode as Element)
+          iframeBodies.forEach((iframeBody) => {
+            this.restoreNode(iframeBody)
+          })
```
ke1k0, line 138: The `as Element` castings are not necessarily anymore, because the new `isElement` type guard you wrote tells TypeScript that these are indeed Elements.

```suggestion
      this.hideNodes(addedNodes.filter(isElement))

      removedNodes.filter(isElement).forEach(
        (removedNode) => {
          // Node has been removed from the DOM, make sure it is
          // removed from our list of hidden nodes as well

          if (removedNode.tagName.toLowerCase() !== 'iframe') {
            this.restoreNode(removedNode)
          }
          const iframeBodies = this.parseIframeBodies(removedNode)
          iframeBodies.forEach((iframeBody) => {
            this.restoreNode(iframeBody)
          })
```

ke1k0 (changes requested):

ke1k0 (approved):

Brailor (approved):

HerrTopi (approved):

PR #1152, opened 2023-02-02
fix(ui-color-picker): axe-check tests failing

NamNguyenInstructure:
modify the aria-role of the color indicator, re-organize imports, re-write locator

### DIFF
--- a/packages/ui-color-picker/src/ColorMixer/ColorPalette/index.tsx
+++ b/packages/ui-color-picker/src/ColorMixer/ColorPalette/index.tsx
@@ -31,6 +31,9 @@ import type { HSVType } from '@instructure/ui-color-utils'
 
 import { View } from '@instructure/ui-view'
 import type { ViewOwnProps } from '@instructure/ui-view'
+import { px } from '@instructure/ui-utils'
+import { withDeterministicId } from '@instructure/ui-react-utils'
+import { ScreenReaderContent } from '@instructure/ui-a11y-content'
 
 import shallowCompare from '../utils/shallowCompare'
 
@@ -39,7 +42,6 @@ import type { ColorPaletteProps, ColorPaletteState } from './props'
 
 import generateStyle from './styles'
 import generateComponentTheme from './theme'
-import { px } from '@instructure/ui-utils'
 
 /**
 ---
@@ -47,6 +49,7 @@ private: true
 ---
 @tsProps
 **/
+@withDeterministicId()
 @withStyle(generateStyle, generateComponentTheme)
 class ColorPalette extends Component<ColorPaletteProps, ColorPaletteState> {
   static propTypes = propTypes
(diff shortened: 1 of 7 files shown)


packages/ui-color-picker/src/ColorMixer/ColorPalette/index.tsx
```diff
@@ -234,8 +238,11 @@ class ColorPalette extends Component<ColorPaletteProps, ColorPaletteState> {
         tabIndex={this.props.disabled ? undefined : 0}
         onKeyDown={(e) => this.handleKeyDown(e)}
         onMouseDown={(e) => this.handlePaletteMouseDown(e)}
-        aria-label={this.props.navigationExplanationScreenReaderLabel}
+        id={this._id}
       >
+        <ScreenReaderContent>
+          {this.props.navigationExplanationScreenReaderLabel}
+        </ScreenReaderContent>
```
ke1k0, line 245: So this is the part we have to redo: remove the `ScreenReaderContent` and put the `aria-label` back. 

The tests will throw the error again, so we agreed with the a11y team to ignore these errors. This is the way to do it:

```js
// in `generateA11yTests`
generateA11yTests(Component, ComponentExamples, [
  'name-of-the-aria-error'
])

// in `accessible`
expect(
    await component.accessible({ ignores: ['name-of-the-aria-error'] })
  ).to.be.true()
})
```

ke1k0 (changes requested):

ke1k0:
BTW next time you can just reopen a closed pull request :) If you have no permission, let us know. :)

packages/ui-color-picker/src/ColorPreset/index.tsx
```diff
@@ -272,6 +272,7 @@ class ColorPreset extends Component<ColorPresetProps, ColorPresetState> {
               />
             </div>
           )}
+          <ScreenReaderContent>{`Color: ${color}`}</ScreenReaderContent>
```
ke1k0, line 275: This makes the screen reader read two labels (from the tooltip plus this one). Since this another case where the screen reader reads it well, because we solved it with a workaround, and the a11y check is not smart enough to see it, I would just ignore the `button-name` error in the a11y tests.

ke1k0 (changes requested):

packages/ui-color-picker/src/ColorPreset/index.tsx
```diff
@@ -38,6 +38,7 @@ import { Text } from '@instructure/ui-text'
 import { Drilldown } from '@instructure/ui-drilldown'
 import type { DrilldownProps } from '@instructure/ui-drilldown'
 import { IconAddLine, IconCheckDarkSolid } from '@instructure/ui-icons'
+import { ScreenReaderContent } from '@instructure/ui-a11y-content'
```
ke1k0: You need to remove this import because it is not used anymore. Also, you forgot to ignore the `button-name` rule in the a11y test where they break.

ke1k0 (changes requested):

ke1k0 (approved):

HerrTopi (approved):

PR #1154, opened 2023-02-02
feat(ui-time-select): mountNode property

jacekk:
For components like `Select` and `Popover` custom `mountNode` property can be passed. `TimeSelect` does not allow to set it, despite the fact that it is using those components under the hood.

This is quite necessary for one of scenarios in our Impact app.


ke1k0 (approved):

Brailor (approved):
Thanks for the PR, it looks good!

PR #1157, opened 2023-02-13
INSTUI-3728: Update dependencies

NamNguyenInstructure:


matyasf:
Usually we dont change version numbers where its not necessary (e.g. we have just `^5` instead of say `^5.3.4`, as I see these were put in by the new yarn update dependencies plugin. There are pros and cons for this new approach:
Pros:  
- We can use `yarn upgrade-interactive`.     
- We are more sure that something works (e.g. what if v5.1.x is a buggy package, we will not know if we began using it with 5.2.1).   

Cons:   
- Update dependencies updates tons of files.  
- We are forcing users to update some dependencies even if it might not be necessary for them.  

What do you think @HerrTopi @Brailor @ke1k0 ?

matyasf (approved):

ke1k0 (changes requested):
Please merge the commits into one `chore: update dependencies` commit.

ke1k0 (approved):

PR #1165, opened 2023-02-24
refactor(ui-navigation): remove raw-loader

NamNguyenInstructure:


matyasf (approved):

ke1k0 (approved):
Next time I would call this a "chore" type commit instead of "refactor", but otherwise it's okay

PR #1171, opened 2023-03-10
refactor(ui-link): add border radius for `Link`

NamNguyenInstructure:


packages/ui-link/src/Link/styles.ts
```diff
@@ -50,6 +50,7 @@ const generateStyle = (
     fontWeight: componentTheme.fontWeight,
     transition: 'outline-color 0.2s',
     verticalAlign: 'baseline',
+    borderRadius: '0.25rem',
```
ke1k0, line 53: I would say that it looks better with 0.125rem radius, it is more consistent with the CondensedButton too. Also, don't hardcode it, make it come from the theme's small border radius variable.
image

ke1k0 (changes requested):

packages/shared-types/src/ComponentThemeVariables.ts
```diff
@@ -619,6 +619,7 @@ export type ImgTheme = {
 }
 
 export type LinkTheme = {
+  borderRadius: Border['radiusSmall']
```
ke1k0, line 622: Since this is applied to the focus ring, I would group it with the other "focusOutline" variables, and call it `focusOutlineBorderRadius`. It is always a good practice to name the variables in a way that it is obvious what they are for. 

Also, please merge the commits into one. :)

ke1k0 (approved):

matyasf (approved):

PR #1178, opened 2023-03-21
fix(ui-text-input): fix padding calculation for before/after element

ke1k0:
Closes: INSTUI-3748

This is a workaround for an edge-case, when the TextInput's parent is hidden on load, so the element is not visible either. In this case the computed width is going to be 'auto', so we assume it has width so that the padding won't be removed. This fix affects all TextInput based components, e.g.: Select, SimpleSelect, DateSelect, etc.

### DIFF
--- a/packages/ui-text-input/src/TextInput/index.tsx
+++ b/packages/ui-text-input/src/TextInput/index.tsx
@@ -289,6 +289,14 @@ class TextInput extends Component<TextInputProps, TextInputState> {
     const computedStyle = getComputedStyle(element)
     const { width, paddingInlineStart, paddingInlineEnd } = computedStyle
 
+    if (width === 'auto') {
+      // This is a workaround for an edge-case, when the TextInput's parent
+      // is hidden on load, so the element is not visible either.
+      // In this case the computed width is going to be 'auto', so we assume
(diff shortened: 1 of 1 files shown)


HerrTopi (approved):

NamNguyenInstructure (approved):

packages/ui-text-input/src/TextInput/index.tsx
```diff
@@ -289,6 +289,14 @@ class TextInput extends Component<TextInputProps, TextInputState> {
     const computedStyle = getComputedStyle(element)
     const { width, paddingInlineStart, paddingInlineEnd } = computedStyle
 
+    if (width === 'auto') {
```
raman-savitski-instructure, line 292: Can we add here `(width === 'auto' || width === '')` ?
In my case on complex page its empty string instead of "auto" and same problem still exist.

ke1k0: Hi @raman-savitski-instructure ! I've made the change in https://github.com/instructure/instructure-ui/pull/1185

raman-savitski-instructure: @ke1k0 thanks a lot for this quick fix!

PR #1182, opened 2023-03-29
INSTUI-3756 fix(ui-date-time-input): dateTimeInput does not trigger invalidDateTimeMessage on blur

matyasf:
note that there is a slight visual change, now if you move with the arrows the input text wont change.
Also now triggering the invalid date message is more consistent, it triggers on empty inputs if `isRequired` is true and does not if its false.

### DIFF
--- a/packages/ui-date-time-input/src/DateTimeInput/README.md
+++ b/packages/ui-date-time-input/src/DateTimeInput/README.md
@@ -65,6 +65,7 @@ class Example extends React.Component {
   onChange = (e, isoDate) => {
     let messages = []
     if (!isoDate) {
+      // this happens if an invalid date is entered
       this.setState({ messages })
       return
     }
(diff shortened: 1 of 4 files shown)


packages/ui-date-time-input/src/DateTimeInput/README.md
```diff
@@ -63,9 +63,14 @@ class Example extends React.Component {
   }
 
   onChange = (e, isoDate) => {
+    let messages = []
```
ke1k0, line 66: Now the example is in "invalid mode" when the page loads. Is that a good thing? I would assume it is a valid pattern that on load the input can be empty before the user touches it, it is weird that it already loads with error. Imho...

matyasf: hmm the error message is definitely wrong in this case, I'll fix that. Also you are right, usually required form fields only start to complain about not having a value after you moved the focus out of them, I'll try to fix it.

ke1k0 (approved):
Seems to work fine. If I really wanted to nitpick, I would modify the example that it gives a different error message when leaving it blank ("Input is required..." instead of "Invalid..."), but other than that it works :)

HerrTopi (approved):

PR #1183, opened 2023-04-03
fix(ui-motion): transition's behavior conflict

NamNguyenInstructure:


packages/ui-motion/src/Transition/BaseTransition/index.ts
```diff
@@ -119,7 +119,9 @@ class BaseTransition extends React.Component<
     }
 
     if (prevProps.in !== this.props.in) {
-      this.startTransition(this.props.in, true)
+      if (_prevState.transitioning)
+        setTimeout(() => this.startTransition(this.props.in, true), 300)
```
matyasf, line 123: please try to solve this without `setTimeout`. e.g. what if this is called multiple times in a single frame? or its the delay is set to much more than 300ms? A proper way would be to cancel the ongoing transition and start the new one requested.

matyasf (changes requested):

matyasf (approved):
looks good! I've used this code to test it: 
```
  constructor (props) {
    super(props)

    this.state = {
      open: false
    }
  }

  handleButtonClick = () => {
    if (this.state.open) {
      // Delay the re-opening
      setTimeout(() => {
        this.setState({ open: true })
        console.log('Re-opened')
      }, 10)
    }
    this.setState(function (state) {
      return { open: !state.open }
    })
  }

  render () {
    return (
      <div>
        <Button onClick={this.handleButtonClick}>
          {this.state.open ? 'Close' : 'Open'} the Modal
        </Button>
        <Modal
          open={this.state.open}
          onDismiss={() => { this.setState({ open: false }) }}
          size="auto"
          label="Modal Dialog: Hello Media"
          shouldCloseOnDocumentClick
          variant="inverse"
        >
          <Modal.Header>
            <Flex>
              <Flex.Item shouldGrow shouldShrink>
                <Heading level="h2"><TruncateText>A small image</TruncateText></Heading>
              </Flex.Item>
              <Flex.Item>
                <CloseButton
                  color="primary-inverse"
                  placement="end"
                  offset="small"
                  onClick={this.handleButtonClick}
                  screenReaderLabel="Close"
                />
              </Flex.Item>
            </Flex>
          </Modal.Header>
          <Modal.Body padding="none">
            <Img
              src={placeholderImage(500, 250)}
              display="block"
              margin="0 auto"
            />
          </Modal.Body>
          <Modal.Footer>
            <Button onClick={this.handleButtonClick} withBackground={false} color="primary-inverse" type="submit">Submit</Button>
          </Modal.Footer>
        </Modal>
      </div>
    )
  }
}

render(<Example />)
```

this is all kinds of buggy in the current V8 but works nice in this PR

PR #1184, opened 2023-04-06
feat(ui-source-code-editor): add `height` and `width` props to SourceCodeEditor

ke1k0:
Closes: INSTUI-3757

### DIFF
--- a/packages/ui-source-code-editor/src/SourceCodeEditor/props.ts
+++ b/packages/ui-source-code-editor/src/SourceCodeEditor/props.ts
@@ -179,6 +179,16 @@ type SourceCodeEditorOwnProps = {
    */
   attachment?: 'bottom' | 'top'
 
+  /**
+   * The height of the editor.
+   */
+  height?: string
+
+  /**
+   * The width of the editor.
+   */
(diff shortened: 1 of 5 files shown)


matyasf (approved):
looks good!

packages/ui-source-code-editor/src/SourceCodeEditor/props.ts
```diff
@@ -179,6 +179,10 @@ type SourceCodeEditorOwnProps = {
    */
   attachment?: 'bottom' | 'top'
 
+  height?: string
```
HerrTopi, line 182: I think we should provide a small explanation of the prop, even if it's trivial what it does.

For example in the select component we write for the width the following: "The width of the text input."

HerrTopi (approved):

NamNguyenInstructure (approved):

packages/ui-source-code-editor/src/SourceCodeEditor/props.ts
ke1k0: I've added the descriptions.

PR #1192, opened 2023-04-20
DateTimeInput: Allow the user to enter non-step divisible time values

matyasf:
Closes: INSTUI-3766

This exposes the `allowNonStepInput` prop from the underlying `TimeSelect`, it should behave the same way

### DIFF
--- a/packages/ui-date-time-input/src/DateTimeInput/README.md
+++ b/packages/ui-date-time-input/src/DateTimeInput/README.md
@@ -45,7 +45,7 @@ class Example extends React.Component {
 render(<Example />)
 ```
 
-#### A required DateInput with `stacked` layout that warns if the value in the past:
+#### A required DateInput with `stacked` layout that warns if the value in the past that allows the user to enter any time value:
 
 ```js
 ---
@@ -95,6 +95,7 @@ class Example extends React.Component {
             value={this.state.value}
             invalidDateTimeMessage="Invalid date!"
             messages={this.state.messages}
+            allowNonStepInput={true}
             isRequired
           />
         </div>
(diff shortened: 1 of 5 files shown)


NamNguyenInstructure (approved):

packages/ui-date-time-input/src/DateTimeInput/index.tsx
```diff
@@ -515,7 +515,8 @@ class DateTimeInput extends Component<DateTimeInputProps, DateTimeInputState> {
       colSpacing,
       isRequired,
       interaction,
-      renderWeekdayLabels
+      renderWeekdayLabels,
+      allowNonStepInput
```
ke1k0, line 519: I would add `allowNonStepInput: false` to the defaultProps.

packages/ui-date-time-input/src/DateTimeInput/README.md
```diff
@@ -45,7 +45,7 @@ class Example extends React.Component {
 render(<Example />)
 ```
 
-#### A required DateInput with `stacked` layout that warns if the value in the past:
+#### A required DateInput with `stacked` layout that warns if the value in the past that allows the user to enter any time value:
```
ke1k0, line 48: This sentence doesn't really make sense :) Can you please rewrite it a bit? Also, it might be too long for a title, you could add more detailed info in a paragraph below.

packages/ui-date-time-input/src/DateTimeInput/index.tsx
matyasf: fixed

ke1k0 (approved):

PR #1196, opened 2023-04-25
INSTUI-2832 refactor(cz-lerna-changelog,ui-token-scripts): convert to TS and ES6 imports

matyasf:
`cz-lerna-changelog`: (makes the commit type chooser on `yarn commit`)Converted to TS. AFAIK this project cannot be a pure ES6 project, because this is a plugin for commitizen which only accepts commonJS plugins. Its a TS project now and commitizen uses its commonJS transpilation.

`ui-token-scripts`: (makes the icons)Converted from commonJS to ES6 imports.


packages/ui-token-scripts/lib/commands/generate-tokens.js
```diff
@@ -22,81 +22,80 @@
...
-  try {
-    handleGenerateTokens({
+    await handleGenerateTokens({
       themeKey,
       styleDictionarySource,
       outputPath,
       sourcePath
     })
-  } catch (e) {
-    error(e)
```
matyasf, line 100: removing these catchall handlers fixes the bug that this script failing was not failing the whole build

packages/cz-lerna-changelog/src/index.ts
```diff
@@ -0,0 +1,131 @@
...
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore not typed
+import buildCommit from 'cz-customizable/lib/build-commit'
+import autocomplete from 'inquirer-autocomplete-prompt'
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore not typed
+import { getChangedPackages, getPackages } from '@instructure/pkg-utils'
+import { makeDefaultQuestions } from './make-default-questions'
+import { autoCompleteQuestions } from './autocomplete-questions'
+
```
matyasf, line 38: typing is far from perfect here, the goal was to get rid of commonJS imports instead of perfect typing

HerrTopi (approved):

PR #1197, opened 2023-04-26
INSTUI-3773 clean up `ui-scripts`

matyasf:
- remove not needed functions
- convert every command to `yargs`
- add delay to all batch `npm` commands to prevent errors
- removed deprecated commands that just resulted in a warning

to test: run the changed commands


package.json
```diff
@@ -88,10 +87,6 @@
     "yarn": ">=3"
   },
   "config": {
-    "ui-scripts": {
-      "npm_scope": "@instructure:registry=https://registry.npmjs.org/",
-      "changelog_url": "https://instructure.design/#CHANGELOG"
-    },
```
matyasf, line 94: this is the config of ui-scripts. changelog_url is unused now, I've moved the other to a constant where its used.

package.json
```diff
@@ -46,8 +46,7 @@
     "clean": "yarn node scripts/clean.js",
     "export:icons": "lerna run export --stream --scope @instructure/ui-icons",
     "bump": "ui-scripts bump",
-    "link:packages": "ui-scripts --link-packages",
-    "release": "ui-scripts --publish",
+    "release": "ui-scripts publish",
```
matyasf, line 49: i had to remove `--` from params because `yargs` doesnt like them

packages/ui-scripts/package.json
```diff
@@ -18,32 +18,29 @@
   },
```
matyasf, line 18: I actually really dislike that `ui-scripts` has 3 differently named binaries. What do you think of keeping just one (e.g. `ui-test --randomize` will be `ui-scripts test --randomize`) ? (in a later PR)

HerrTopi: I think this is a very good idea. Lets have a task about it!

matyasf: 👍 ill create one

HerrTopi (approved):

PR #1202, opened 2023-05-03
feat(shared-types,ui-flex): add `gap` prop to Flex

NamNguyenInstructure:


packages/ui-flex/src/Flex/props.ts
```diff
@@ -76,6 +76,13 @@ type FlexOwnProps = {
    */
   margin?: Spacing
 
+  /**
+   * Valid values are `0`, `none`, `auto`, `xxx-small`, `xx-small`, `x-small`,
+   * `small`, `medium`, `large`, `x-large`, `xx-large`. Apply these values via
+   * familiar CSS-like shorthand. For example: `margin="small large"`.
```
HerrTopi, line 82: Copy - pasta mistake: margin is left there instead of gap.

packages/ui-flex/src/Flex/styles.ts
```diff
@@ -85,6 +86,11 @@ const generateStyle = (
     'row-reverse': 'row-reverse'
   }
 
+  // gap css prop
+  const getGapValue = (gap: FlexProps['gap'], theme: FlexTheme) => {
```
HerrTopi, line 90: If there is only a return in the arrow function, leave it out:

`const getGapValue = (gap: FlexProps['gap'], theme: FlexTheme) => getShorthandPropValue('Flex', theme, gap, 'gap')`

More consistent and easier to read

packages/ui-flex/src/Flex/README.md
```diff
@@ -46,6 +46,76 @@ example: true
...
+  <Flex withVisualDebug direction="column-reverse" gap="small">
+    <Flex.Item>One</Flex.Item>
+    <Flex.Item>Two</Flex.Item>
+    <Flex.Item>Three</Flex.Item>
+    <Flex.Item>Four</Flex.Item>
+  </Flex>
+</div>
+```
+
+You can also set the gap between rows and columns by using the `gap` property. Make sure that the `wrap` property is set to `wrap` or `wrap-reverse`.
```
HerrTopi, line 85: "Make sure that the `wrap` property is set to `wrap` or `wrap-reverse`."

Please add a warning in the index file if gap is provided without the proper wrap prop

HerrTopi (approved):

matyasf (approved):

PR #1203, opened 2023-05-03
INSTUI-3772 Refactor(ui-time-select): rewrite logic if allowNonStepInput is true

matyasf:
The autocomplete behaviour is now different if `allowNonStepInput` is `true`:   
- if the user types in 1 or 2 letters it offers the step inputs as choices. 
- if more it will show all the minutes

Now one can enter a value with the following methods:   
1. User clicks/presses enter on an option in the dropdown and focuses away
2. User enters a valid time manually and focuses away

To test:
test all the examples of `TimeSelect` and `DateTimeInput`


HerrTopi (commented):
It should work with the second too if allowNonStepInput is set to true (as we discussed)

NamNguyenInstructure (commented):
Aside from the issue Topi mentioned, everything else seems to be fine.

HerrTopi (approved):
Nice work!

PR #1204, opened 2023-05-08
chore: disable webpack dev server client overlay

NamNguyenInstructure:


matyasf (approved):

HerrTopi (commented):
Are you sure this is the right approach? I checked the history of this file, and we never had this line here, nevertheless, it was working correctly.
Won't we just mask a problem this way?

matyasf:
> Are you sure this is the right approach? I checked the history of this file, and we never had this line here, nevertheless, it was working correctly. Won't we just mask a problem this way?

It could be because I've updated some Webpack dependencies

matyasf:
> Are you sure this is the right approach? I checked the history of this file, and we never had this line here, nevertheless, it was working correctly. Won't we just mask a problem this way?

It could be because I've updated some Webpack dependencies

matyasf:
> Are you sure this is the right approach? I checked the history of this file, and we never had this line here, nevertheless, it was working correctly. Won't we just mask a problem this way?

It could be because I've updated some Webpack dependencies

matyasf:
> Are you sure this is the right approach? I checked the history of this file, and we never had this line here, nevertheless, it was working correctly. Won't we just mask a problem this way?

It could be because I've updated some Webpack dependencies

matyasf:
> Are you sure this is the right approach? I checked the history of this file, and we never had this line here, nevertheless, it was working correctly. Won't we just mask a problem this way?

It could be because I've updated some Webpack dependencies

NamNguyenInstructure:
After reviewing the old commits, I discovered that the error occurred when we update the Webpack from 5.75 to 5.78

matyasf:
hmm so its a webpack "feature" then, perhaps "overlay" was false until now? I like the fix this way as it is now

HerrTopi (approved):

PR #1214, opened 2023-05-18
fix(ui-select,ui-truncate-text): disable pointer events for internal …

HerrTopi:
…parts of the component

Closes: INSTUI-3780


matyasf (changes requested):
Sadly this change breaks links inside `TruncateText`, e.g. compare https://1214--preview-instui.netlify.app/#TruncateText/#single-line vs https://instructure.design/#TruncateText/#single-line

PR #1215, opened 2023-05-20
fix(ui-popover): fix TruncateText in Popover not accepting click events

matyasf:


HerrTopi (approved):
Nice work! Thanks for helping (:

Issue #1181, opened 2023-03-28
ui-menu focus element that is not yet positioned correctly


czeslaaw:
### Background Information
We are in the middle of rewriting the Impact dashboard to Instructure UI. When opening ui-menu the page scrolls to the top-left corner around 80% of the time.

**Package Version(s)**:
8.36.0
React 18.2.0

**Browser**:
Chrome 111.0.5563.110

**OS:**
MacOS 12.6.3

**Device:**
Mac

**Component**:
ui-menu

### Describe the Bug
There seems to be some kind of race condition between positioning dropdown menu and focusing on it. It happens 80% of time. Trying to debug it makes it happened less often 😈 

https://user-images.githubusercontent.com/1939060/228291426-3d71b12e-b0ec-415b-a517-1bca0267e561.mov

https://user-images.githubusercontent.com/1939060/228291560-2911d003-b280-4ad2-8cfc-daac76065c4d.mov

### Steps To Reproduce
Minimal code to reproduce:
```ts
import { Button } from '@instructure/ui-buttons';
import { Menu } from '@instructure/ui-menu';
import React from 'react';

export const App: React.FC = () => {
	const triggerButton = <Button>Trigger</Button>;

	return (
		<>
			<p style={{ fontSize: '150px' }}>Somthing big to require scrolling for button</p>
			<Menu trigger={triggerButton}>
				<Menu.Item>Foo</Menu.Item>
				<Menu.Item>Bar</Menu.Item>
				<Menu.Item>Baz</Menu.Item>
			</Menu>
		</>
	);
};
```

### Expected Behavior
ui-menu not scrolling

**Current Workaround(s):**
Could not find any

**Products Affected:**
Impact

**Are you willing to submit a PR to fix?**
- [ ] Yes, I'm willing to submit a PR

**Requested Priority:**
Normal

ke1k0:
Hi @czeslaaw ! Thank you for the bug report, we will have a look at it soon.

czeslaaw:
What is the status of this issue?

czeslaaw:
When I cloned this project I couldn't reproduce it but it is reproducible on this minimal code:

index.jsx
```jsx
import { Button } from "@instructure/ui-buttons";
import { Menu } from "@instructure/ui-menu";
import React from "react";
import ReactDOM from "react-dom/client";

export const App = () => {
  const triggerButton = <Button>Trigger</Button>;

  return (
    <>
      <p style={{ fontSize: "200px" }}>
[...]

PR #1217, opened 2023-05-31
bug(ui-dialog,ui-a11y-utils): fix focusing not yet positioned elements

czeslaaw:
Fix for this issue: https://github.com/instructure/instructure-ui/issues/1181

# How to reproduce

* Remove `React.StrictMode` tag from `packages/__docs__/src/index.js`
* Replace content of `packages/__docs__/src/App/index.tsx` with:
```jsx
import { Button } from '@instructure/ui-buttons'
import { Menu } from '@instructure/ui-menu'
import React from 'react'

export const App: React.FC = () => {
  const triggerButton = <Button>Trigger</Button>

  return (
    <>
      <p style={{ fontSize: '150px' }}>
        Somthing big to require scrolling for button
      </p>
      {new Array(100).fill(null).map((_, index) => (
        <Menu key={index} trigger={triggerButton}>
          {new Array(12).fill(null).map((_, index) => (
            <Menu.Item key={index}>{index}</Menu.Item>
          ))}
        </Menu>
      ))}
    </>
  )
}
```
* Trigger any button a few times, eventually it should scroll to the top of the page.

### DIFF
--- a/packages/ui-a11y-utils/src/KeyboardFocusRegion.ts
+++ b/packages/ui-a11y-utils/src/KeyboardFocusRegion.ts
@@ -143,7 +143,14 @@ class KeyboardFocusRegion {
 
   focusDefaultElement() {
     if (this.defaultFocusElement) {
-      ;(this.defaultFocusElement as HTMLElement).focus()
+      const focusElement = this.defaultFocusElement as HTMLElement
+      const positionContainer = focusElement.closest('[data-position-content]')
+
+      if (!positionContainer) {
+        focusElement.focus()
+      } else {
+        this.tryFocusOnPositionedElement(positionContainer, focusElement)
+      }
     } else {
       if (this.shouldContainFocus) {
         // Blur the active element to place focus on the document body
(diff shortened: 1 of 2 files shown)


czeslaaw:
@matyasf @ke1k0 @NamNguyenInstructure Please review :)

packages/ui-a11y-utils/src/KeyboardFocusRegion.ts
```diff
@@ -160,6 +167,45 @@ class KeyboardFocusRegion {
...
+      focusElement.focus()
+    } else {
+      this._raf.push(
+        requestAnimationFrame(() => {
+          this.tryFocusOnPositionedElement(
+            positionContainer,
+            focusElement,
+            tries - 1
+          )
+        })
```
czeslaaw, line 204: I tried `MutationObserver` and `IntersectionObserver` but this simple solution works the best

packages/ui-dialog/src/Dialog/index.tsx
```diff
@@ -99,20 +96,11 @@ class Dialog extends Component<DialogProps> {
...
-        // final position, making the page jump.
-        this._timeouts.push(
-          setTimeout(() => {
-            this._focusRegion = FocusRegionManager.activateRegion(
-              this.contentElement as Element,
-              {
-                ...options
-              }
-            )
-          }, 0)
```
czeslaaw, line 115: Since react 18 rendering can be interrupted so a simple `setTimeout` doesn't guarantee that DOM will be completely up to date.

packages/ui-a11y-utils/src/KeyboardFocusRegion.ts
```diff
@@ -160,6 +167,45 @@ class KeyboardFocusRegion {
     }
   }
 
+  tryFocusOnPositionedElement(
+    positionContainer: Element,
+    focusElement: HTMLElement,
+    tries = 3
```
czeslaaw, line 173: Arbitrary number. I tried this on a very beefy page and it never run out of tries.

matyasf:
thanks for the PR, we'll look into it

NamNguyenInstructure (approved):
It's great! Thank you for your work.

HerrTopi (approved):
The issue seems to be gone. Thanks a lot for the fix, much appreciated.

matyasf (approved):
looks good

matyasf:
thanks a lot for this fix!

PR #1219, opened 2023-06-12
fix(ui-a11y-utils): dialog reopens when clicking on the trigger

NamNguyenInstructure:


packages/ui-a11y-utils/src/FocusRegion.ts
```diff
@@ -139,7 +139,7 @@ class FocusRegion {
           addEventListener(doc, 'mousedown', this.captureDocumentClick, true)
         )
         this._listeners.push(
-          addEventListener(doc, 'mousedown', this.handleDocumentClick)
+          addEventListener(doc, 'click', this.handleDocumentClick)
```
NamNguyenInstructure, line 142: When the trigger is clicked, two event handlers are triggered: the "toggle" event of the Popover and the "handleDismiss" event of the FocusRegion. Because the "mousedown" handler is triggered before the "click" event, the Dialog is closed and reopened when the event handler of Popover's trigger is fired. By changing the FocusRegion to "click", handlers are now executed in the order they were registered: toggle -> handleDismiss.

balzss:
this is a revert of an old bugfix that introduced a new bug. closing this PR and including the revert in the new fix

PR #1222, opened 2023-06-14
fix(ui-a11y-utils, ui-modal): popover keeps reopen when clicking on t…

NamNguyenInstructure:
…he trigger

### DIFF
--- a/packages/ui-a11y-utils/src/FocusRegion.ts
+++ b/packages/ui-a11y-utils/src/FocusRegion.ts
@@ -136,10 +136,10 @@ class FocusRegion {
 
       if (this._options.shouldCloseOnDocumentClick) {
         this._listeners.push(
-          addEventListener(doc, 'mousedown', this.captureDocumentClick, true)
+          addEventListener(doc, 'click', this.captureDocumentClick, true)
         )
         this._listeners.push(
-          addEventListener(doc, 'mousedown', this.handleDocumentClick)
+          addEventListener(doc, 'click', this.handleDocumentClick)
         )
 
         Array.from(doc.getElementsByTagName('iframe')).forEach((el) => {
--- a/packages/ui-dialog/src/Dialog/__tests__/Dialog.test.tsx
+++ b/packages/ui-dialog/src/Dialog/__tests__/Dialog.test.tsx
@@ -119,7 +119,7 @@ describe('<Dialog />', async () => {
       expect(dialog.containsFocus()).to.be.true()
     })
 
-    await within(dialog.getOwnerDocument().documentElement).mouseDown()
+    await within(dialog.getOwnerDocument().documentElement).click()
 
     await wait(() => {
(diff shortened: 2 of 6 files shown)


packages/ui-a11y-utils/src/FocusRegion.ts
```diff
@@ -136,10 +136,10 @@ class FocusRegion {
 
       if (this._options.shouldCloseOnDocumentClick) {
         this._listeners.push(
-          addEventListener(doc, 'mousedown', this.captureDocumentClick, true)
```
NamNguyenInstructure, line 139: Revert the mouse events

packages/ui-dialog/src/Dialog/__tests__/Dialog.test.tsx
```diff
@@ -119,7 +119,7 @@ describe('<Dialog />', async () => {
       expect(dialog.containsFocus()).to.be.true()
     })
```
NamNguyenInstructure, line 121: I reverted the tests back to their original state.

packages/ui-modal/src/Modal/index.tsx
```diff
@@ -211,7 +240,7 @@ class Modal extends Component<ModalProps, ModalState> {
         open
         label={label}
         defaultFocusElement={this.defaultFocusElement}
-        shouldCloseOnDocumentClick={shouldCloseOnDocumentClick}
```
NamNguyenInstructure, line 214: `shouldCloseOnDocumentClick` is set to `false` to disable the logic of the `FocusRegion`

packages/ui-modal/src/Modal/index.tsx
```diff
@@ -182,6 +198,20 @@ class Modal extends Component<ModalProps, ModalState> {
...
   }
 
+  captureDocumentClick: React.MouseEventHandler<any> = (event) => {
+    const { target } = event
+    const _dialogElement = findDOMNode(this._content) as Element
+    this._preventCloseOnDocumentClick =
+      event.button !== 0 || contains(_dialogElement, target as Node)
+  }
+
+  onClick: React.MouseEventHandler<any> = (event) => {
```
NamNguyenInstructure, line 208: This event handler checks if `shouldCloseOnDocumentClick` is `true` and the mouse pointer is pressed inside the popped-up Dialog.

packages/ui-modal/src/Modal/index.tsx
```diff
@@ -182,6 +198,20 @@ class Modal extends Component<ModalProps, ModalState> {
     )
   }
 
+  captureDocumentClick: React.MouseEventHandler<any> = (event) => {
```
NamNguyenInstructure, line 201: Check if the mouse is pressed down inside the dialog.

balzss:
closing this since the fix would introduce new bugs

PR #1223, opened 2023-06-19
refactor(ui-dom-utils): update `findFocusable` logic

NamNg1121:


packages/ui-dom-utils/src/findFocusable.ts
```diff
@@ -83,12 +83,7 @@ function findFocusable(
 
 function hidden(element: Element | Node) {
   const cs = getComputedStyle(element)
-  return (
```
NamNguyenInstructure, line 86: This refers to a [ticket](https://instructure.slack.com/archives/C0JCJ63TR/p1682957023840189) on the Slack channel. After researching the issue, I discovered that the developers who were implementing the feature had left. Therefore, I decided to remove the feature and test how it affected the components. Overall, everything was fine in the manual test. I suggest writing unit tests.

matyasf:
The code of findFocusable was lifted from some ancient JQuery UI helper. For some unknown reason it
treats an element as non-focusable if its CSS display prop is not inline and its size is 0 or less.
It seems to work fine without this criteria, and we could not find any reason why this clause was
added in the first place.

matyasf (approved):

HerrTopi (approved):

PR #1224, opened 2023-06-21
INSTUI-3784 Merge ui-scripts binaries into one

balzss:
[INSTUI-3784](https://instructure.atlassian.net/jira/software/c/projects/INSTUI/boards/1223?modal=detail&selectedIssue=INSTUI-3784)

[INSTUI-3784]: https://instructure.atlassian.net/browse/INSTUI-3784?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/ui-scripts/package.json
```diff
@@ -17,8 +17,8 @@
     "ui-test": "./lib/test/index.js"
```
matyasf, line 17: The `ui-build` and `ui-test` are no longer needed.

packages/ui-scripts/lib/build/babel.js
```diff
@@ -29,83 +29,89 @@ import { fileURLToPath } from 'url'
...
+          DEBUG ? `DEBUG=1` : false
+        ])
+        .filter(Boolean)
     }
 
-    modules = arg.split(',')
+    let modules = ['es']
 
-    if (modules.some((mod) => !['es', 'cjs'].includes(mod))) {
-      throw new Error(`Invalid --modules argument: '${arg}'`)
```
matyasf, line 85: I think the original solution is better here: Only allow `es` or `cjs` as values and error out when the `modules` argument is missing. You could use http://yargs.js.org/docs/#api-reference-choiceskey-choices and http://yargs.js.org/docs/#demandOption

packages/ui-scripts/lib/build/index.js
```diff
@@ -23,20 +23,25 @@
...
+  // eslint-disable-next-line no-console
+  console.log(
+    'This command is deprecated! Please `use ui-scripts clean` instead'
+  )
 } else {
-  babel()
+  // eslint-disable-next-line no-console
+  console.log(
+    'This command is deprecated! Please `use ui-scripts build` instead'
+  )
```
matyasf, line 46: Luckily we dont need this part, we've communicated a while ago that auxiliary scripts like these are for internal use only, users should not use them and can change without notice.

packages/ui-scripts/lib/test/index.js
```diff
@@ -23,14 +23,15 @@
...
+  console.log(
+    'This command is deprecated! Please `use ui-scripts lint` instead'
+  )
 } else {
-  karma()
+  // eslint-disable-next-line no-console
+  console.log(
+    'This command is deprecated! Please use `ui-scripts test` instead'
+  )
 }
```
matyasf, line 37: again, not needed.

matyasf (changes requested):
you should change these in `template-component` and in `template-package too.
See my comments, otherwise nice work!

packages/ui-scripts/lib/build/babel.js
balzss: I think the original solution defaulted to `es` if there were no `--modules` argument. Should I keep it like that or use `demandOptions`? I will also add the yargs `choices` option

matyasf: oh you're right, it had some wonky double check for the `modules` argument. We can default to `es`, there is also a `yargs` option for default :)

HerrTopi (approved):
After the requested changes, it seems nice. Good work

matyasf (approved):

PR #1228, opened 2023-07-04
fix(ui-a11y-utils): Drilldown keeps reopening when clicking on the trigger

balzss:
should fix the bug introduced with [this pr](#1194 )

Closes: INSTUI-3790

test plan 1:
- open a modal
- click on modal and drag the mouse outside and release
- the modal shouldn't close

test plan 2:
- open drilldown in popover
- click on the button again
- the popover should close


matyasf (approved):
nice work!

HerrTopi (approved):

PR #1232, opened 2023-07-14
docs(ui-select,ui-simple-select): add docs about behaviour in native HTML forms

matyasf:


HerrTopi (approved):
Should we add this to V7 as well? Maybe/possibly not, so I'll approve this

balzss (approved):

PR #1235, opened 2023-07-20
voiceover in safari doesn't read grouped select options

balzss:
INSTUI-3810

test plan:

- open the new example in safari
- turn on voiceover
- open the select menu and use arrow keys or mouse to navigate options
- voiceover should announce every option when highlighted


matyasf (approved):
nice work!

PR #1240, opened 2023-07-24
add jest + react testing library to the project

balzss:
- jest + react testing library has been added to the project without removing anything from the old tests
- you can add new tests in the `__new-tests__` directory inside a package
- you can run jest from the root with the `yarn test:new` command


balzss:
closing in favor of #1246

PR #1241, opened 2023-07-24
INSTUI-3823 Get InstUI build on Windows

matyasf:
This PR gets `yarn bootstrap` and `yarn dev` working on windows. `yarn test` also runs, but its quite flaky -- some tests fail, some others just throw an error.
This is accomplished by using `cross-spawn` via its API to spawn `babel`, `karma` and `webpack` instead of `concurrently` and `execa` via command line.
On OSX nothing should change except `babel` being a bit more verbose.

To test:
On OSX run our usual build/test commands, especially with parameters (e.g. `yarn test --scope..`).  
On Windows run `bootstrap`, `dev`. If you get tons of popups change the default program to run `js` files to `node.exe`

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,7 @@
     "@types/react-dom": "^18.2.7",
     "chalk": "^4.1.2",
     "commitizen": "^4.3.0",
+    "cross-spawn": "^7.0.3",
     "danger": "^11.2.6",
     "esbuild": "^0.18.12",
     "eslint": "^8.44.0",
@@ -80,8 +81,7 @@
     "netlify-cli": "13.2.2",
     "npm-run-all": "^4.1.5",
     "typescript": "5.1.6",
-    "webpack": "^5.88.1",
-    "cross-spawn": "^7.0.3"
+    "webpack": "^5.88.1"
   },
   "engines": {
     "node": ">=14",
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "generate:component": "yarn workspace @instructure/instui-cli instui create component",
     "generate:package": "yarn workspace @instructure/instui-cli instui create package",
     "commit": "ui-scripts commit",
-    "bootstrap": "time yarn node scripts/bootstrap.js",
+    "bootstrap": "yarn node scripts/bootstrap.js",
     "build": "lerna run build --stream",
     "build:watch": "lerna run build:watch --stream",
     "build:docs": "lerna run bundle --stream --scope docs-app",
@@ -80,7 +80,8 @@
     "netlify-cli": "13.2.2",
     "npm-run-all": "^4.1.5",
     "typescript": "5.1.6",
-    "webpack": "^5.88.1"
+    "webpack": "^5.88.1",
+    "cross-spawn": "^7.0.3"
   },
   "engines": {
     "node": ">=14",
--- a/packages/__docs__/buildScripts/utils/getClientProps.mts
+++ b/packages/__docs__/buildScripts/utils/getClientProps.mts
@@ -22,7 +22,7 @@
  * SOFTWARE.
  */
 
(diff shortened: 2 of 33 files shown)


packages/command-utils/lib/index.js
```diff
@@ -42,93 +42,78 @@ function error(...args) {
...
+    const commandList = Array.isArray(cl) ? cl : [cl]
+    for (const command of commandList) {
+      promises.push(runCommandAsync(command.bin, command.args, command.envVars))
     }
-  })
-
-  let result = { status: 1 }
-
-  try {
-    result = runCommandSync(`${resolveBin('concurrently')}`, args)
```
matyasf, line 107: This was the biggest issue. This is an ugly hack to allow synchronous parallel execution in node (= execute multiple scripts simultaneously synchronously). This was accomplished by running `concurrently` (a lib to execute multiple scripts at the same time) synchronously. This had multiple issues:
- we needed to pass arguments and env variables in a huge string like `concurrently "ENV_VAR1=A, ENV_VAR2=B scriptA --arg1 abc --arg2 bcd" ENV_VAR3=3 scriptB --arg1=123` that is very hard to read and is not Windows compatible. 
- We were running `concurrently` not with native code but with a lib called `execa`. Dontknow why
- All these scripts executing scripts made IDE debuggers get lost unable to place breakpoints
- some of them also mess with `stdout` causing formatting like colors to be lost.

This solution is pure native code, easy to read with a price that `runCommandsConcurrently` became async (we never made use of its synchronous nature)

package.json
```diff
@@ -36,7 +36,7 @@
     "generate:component": "yarn workspace @instructure/instui-cli instui create component",
     "generate:package": "yarn workspace @instructure/instui-cli instui create package",
     "commit": "ui-scripts commit",
-    "bootstrap": "time yarn node scripts/bootstrap.js",
```
matyasf, line 39: `time` does not exist on Windows, we should use a node API in a later PR

HerrTopi:
It seems that `yarn bootstrap`, and `yarn dev` works, but the docs app missing most of its pages. I'd bet it's a path issue. This probably works differently on mac and windows

packages/__docs__/buildScripts/utils/getClientProps.mts
```diff
@@ -22,7 +22,7 @@
  * SOFTWARE.
  */
 
-import type { LibraryOptions, ParsedDoc, ProcessedFile } from '../DataTypes'
+import type { LibraryOptions, ParsedDoc, ProcessedFile } from '../DataTypes.mjs'
```
matyasf, line 25: This rewrite was needed because I thought that an issue would be fixed by upgrading react-docgen... it was not, and it required that our TS files are emitted with ESM imports. This needed a change in the extension otherwise they'd be treated as commonJS files.

packages/ui-navigation/src/AppNav/props.ts
```diff
@@ -80,7 +80,7 @@ type AppNavOwnProps = {
    * the navigation changes. Passes in the `visibleItemsCount` as
    * a parameter.
    */
-  onUpdate?: ({ visibleItemsCount }: { visibleItemsCount: number }) => void
+  onUpdate?: (visibleItemsCount: { visibleItemsCount: number }) => void
```
matyasf, line 83: this was needed because react-docgen was failing with the old code, see https://github.com/reactjs/react-docgen/issues/827

HerrTopi (approved):
Yaaay, it works! Fantastic work, thanks a lot!

balzss (approved):
works fine on mac as well (`bootstrap`, `dev`, `test`, `build`)

packages/ui-navigation/src/AppNav/props.ts
danez: This was fixed in react-docgen@6.0.3.

PR #1242, opened 2023-07-26
fix(ui-flex): documentation wording

camray:
Previous wording was a bit confusing


balzss (approved):
I agree, this is better. Thanks for the PR!

HerrTopi (approved):

PR #1248, opened 2023-08-08
fix: <Select/> width prop only applies to the input inside

balzss:
INSTUI-3820

test plan:
- add `width="8rem"` to a `Select` or `SimpleSelect` component
- click to the right of the component -> the dropdown shouldn't open

### DIFF
--- a/packages/ui-select/src/Select/index.tsx
+++ b/packages/ui-select/src/Select/index.tsx
@@ -673,6 +673,13 @@ class Select extends Component<SelectProps> {
     const highlightedOptionId = this.highlightedOptionId
     const selectedOptionId = this.selectedOptionId
 
+    // props.width should apply to the whole component, not just the input in it
+    const extendedStyle = {
+      ...(styles?.select as object),
+      width: this.props.width || '100%',
+      display: 'inline-block'
+    }
+
     return (
       <Selectable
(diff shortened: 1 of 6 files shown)


packages/ui-select/src/Select/index.tsx
```diff
@@ -673,6 +673,13 @@ class Select extends Component<SelectProps> {
     const highlightedOptionId = this.highlightedOptionId
     const selectedOptionId = this.selectedOptionId
 
+    // props.width should apply to the whole component, not just the input in it
+    const extendedStyle = {
+      ...(styles?.select as object),
+      width: this.props.width || '100%',
+      display: 'inline-block'
+    }
```
matyasf, line 681: CSS should be in `styles.ts`. This should be in the `select` object there.

matyasf (changes requested):
see my comments

packages/ui-select/src/Select/index.tsx
balzss: good point, moved it there

matyasf (approved):

matyasf (commented):
Can you please check Chromatic? It shows visual changes, this should not happen, this'd be a breaking change.

matyasf (approved):

joyenjoyer (approved):

PR #1251, opened 2023-08-16
fix: Add id as allowed prop for ColorPicker

judikdavid:
We need to allow ColorPicker to get it's own id
through props to be able to use with a form library like Formik which requires this prop.


packages/ui-color-picker/src/ColorPicker/props.ts
```diff
@@ -199,6 +199,11 @@ type ColorPickerOwnProps = {
    */
   tooltip?: React.ReactNode
 
+  /**
+   * The id of the text input.
```
matyasf, line 203: I'd add ` One is generated if not supplied.`

matyasf (changes requested):

packages/ui-color-picker/src/ColorPicker/props.ts
judikdavid: Added it, thank you.

matyasf (approved):

balzss (approved):

PR #1256, opened 2023-08-17
refactor(command-utils): fix runCommandAsync

matyasf:
We treated it as something that returns a Promise when it does not. This fixes it by wrapping it in a Promise
TEST PLAN:
Change some command temporarly (e.g. ui-scripts tag) so it uses RunCommandAsync and output its return value, e.g.
```
try {
    const res = await runCommandAsync("ls", [], {}, {stdio: 'pipe', stderr: 'pipe'})
    console.log(res)
  } catch (err) {
    console.log("RESULT", err)
  }
```

### DIFF
--- a/packages/__docs__/src/Returns/index.tsx
+++ b/packages/__docs__/src/Returns/index.tsx
@@ -29,7 +29,8 @@ import { Table } from '@instructure/ui-table'
 import { compileMarkdown } from '../compileMarkdown'
 
 import type { ReturnsProps } from './props'
-import { JSDocFunctionReturns } from '../../buildScripts/DataTypes'
+// eslint-disable-next-line import/no-unresolved
+import type { JSDocFunctionReturns } from '../../buildScripts/DataTypes'
 
 class Returns extends Component<ReturnsProps> {
   renderRows() {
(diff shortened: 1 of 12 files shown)


packages/ui-scripts/lib/test/lint.js
```diff
@@ -51,7 +51,8 @@ export default {
     if (jspaths.length) {
       commands['eslint'] = getCommand('eslint', [
         ...jspaths,
-        '--ext .js,.jsx,.ts,.tsx',
+        '--ext',
+        '.js,.jsx,.ts,.tsx',
```
matyasf, line 55: This was causing an error when running `lint`, I have no idea why it did not came up until now.

packages/__docs__/src/Returns/index.tsx
```diff
@@ -29,7 +29,8 @@ import { Table } from '@instructure/ui-table'
 import { compileMarkdown } from '../compileMarkdown'
 
 import type { ReturnsProps } from './props'
-import { JSDocFunctionReturns } from '../../buildScripts/DataTypes'
+// eslint-disable-next-line import/no-unresolved
+import type { JSDocFunctionReturns } from '../../buildScripts/DataTypes'
```
matyasf, line 33: The linter has found a couple of errors, this is an ugly workaround. Its getting confused because the docs build scripts are compiled to ES6 modules and the docs app to commonJS. IMO we should not bother much with this, rather move the docs app to a modern build system like ESBuild

joyenjoyer (approved):

PR #1257, opened 2023-08-18
chore(ui-scripts): don't run git hooks from ui-scripts

balzss:
INSTUI-3833

### test plan:

- `yarn bump` should run and create a new commit


matyasf (approved):
looks good, nice work :) 
Maybe It'd nicer to merge this with the env vars, so setting them does not overwrite it.

HerrTopi (approved):

balzss:
> looks good, nice work :) Maybe It'd nicer to merge this with the env vars, so setting them does not overwrite it.

I was thinking what **if we want to** override it but that probably won't be the case

PR #1259, opened 2023-08-18
fix: fix `document is undefined` SSR errors

matyasf:
There were some parts of the code where we try to access 'document', and this will raise errors in an SSR environment. This patch fixes these cases. Note that there might be other ones, we should have some SSR testing suite

TEST PLAN:
make a next.js app with the instructions in their site. Add components and run our utils in the render function, e.g. px(), TopNavBar

CLOSES: InstUI-3816


balzss (approved):
tested it and it's working fine. test method:

- created a default next.js app
- usd `npm link` to add `@instructure/ui-utils`
- imported the `px(..)` util from `ui-utils`
- it worked

HerrTopi (approved):

PR #1262, opened 2023-08-22
Inst UI 3832 responsive breakpoints

joyenjoyer:
Add support for multiple breakpoints in media queries.

**Test plan:**

1. Go to the Responsive component page of the InstUI docs app
2. Add extra breakpoints (`minWidth`, `maxWidth`) into the query part of the Responsive component
3. Check if errors are popping up & newly added breakpoints work as intended to CSS specifications & the appropriate part of the HTML is getting rendered

### DIFF
--- a/packages/ui-responsive/src/Responsive/README.md
+++ b/packages/ui-responsive/src/Responsive/README.md
@@ -81,3 +81,48 @@ example: true
   }}
 </Responsive>
 ```
+
+### Defining multiple breakpoints
+
+Setting up multiple media query breakpoints is supported. You can use 4 breakpoints at the same time: `minWidth`, `maxWidth`, `minHeight`,
+`maxHeight`. In the following example, the use of two breakpoints, the `minWidth` and `maxWidth`, are used to render the component with the `medium` label.
+
+**Note:** Overlapping breakpoints for height are not fully supported.
+
+```js
+---
+example: true
+---
+<Responsive
+  match="media"
+  query={{
+    small: { maxWidth: 400 },
+    medium: { minWidth: 401, maxWidth: 599 },
+    large: { minWidth: 600}
+  }}
+>
+  {(props, matches) => {
+    if (matches.includes('large')) {
+      return (
+        <Billboard
+          message="Large breakpoint"
+          hero={<IconUserLine />}
+          style={{ border: '10px solid #aaa' }}
+        />
+      )
+    } else if (matches.includes('medium') && !matches.includes('large')) {
+      return (
+        <Byline description="Medium breakpoint">
+          <Avatar name="Alexander Hamilton" />
+        </Byline>
+      )
+    } else {
+      return (
+        <Pill color="primary">
+          Small breakpoint
+        </Pill>
+      )
+    }
+  }}
+</Responsive>
+```
--- a/packages/ui-responsive/src/Responsive/__tests__/Responsive.test.tsx
+++ b/packages/ui-responsive/src/Responsive/__tests__/Responsive.test.tsx
@@ -190,4 +190,35 @@ describe('<Responsive />', async () => {
 
     expect(responsiveDiv.style.display).to.equal('inline-flex')
   })
+
+  it('should provide correct props in case of multiple breakpoints in query', async () => {
+    const renderSpy = spy()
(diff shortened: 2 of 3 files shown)


matyasf:
Just some preliminary feedback:

Please edit https://github.com/instructure/instructure-ui/blob/master/packages/ui-responsive/src/Responsive/README.md so it includes an example with multiple breakpoints.

balzss:
small thing but it might make sense to write the test with the new frameworks (jest + react testing library)

packages/ui-responsive/src/Responsive/__tests__/Responsive.test.tsx
```diff
@@ -190,4 +190,35 @@ describe('<Responsive />', async () => {
...
+      medium: { options: [1, 2, 3], icons: { edit: true, flag: false } },
+      large: { margin: 'small', label: 'hello world', describedBy: 'fakeId' }
+    }
+    await mount(
+      <div style={{ width: 500, height: 400 }}>
+        <Responsive
+          props={props}
+          query={{
+            small: { maxWidth: 300, maxHeight: 200 },
+            medium: { maxWidth: 800, maxHeight: 500 },
```
aaronshaf, line 207: wouldn't a height of 199 match both small and medium here?

joyenjoyer: Yes, both would be matched for a heigth of 199, furthermore a warning would be emitted if there were properties with the same keys. In this testcase (width: 500, height: 400), the medium labelled props should be picked up.

joyenjoyer:
Done.



> Just some preliminary feedback:
> 
> Please edit https://github.com/instructure/instructure-ui/blob/master/packages/ui-responsive/src/Responsive/README.md so it includes an example with multiple breakpoints.

packages/ui-responsive/src/Responsive/__tests__/Responsive.test.tsx
aaronshaf: Wouldn't both small and medium props be picked up in some cases at a height of 199?

matyasf: yes, they would. But this is not an issue, one can write queries with overlapping intervals. A warning (`The prop ${prop} is defined at 2 or more breakpoints...`) will be emitted if one defines a the same prop for overlapping intervals. So one can write something like:
```json
props=
  {small: {border:"black"}}
  {medium: {margin:"small"}}
query=
  {small: {minHeight: 200}
  {medium: {minHeight: 300}
```
And if `height` is above 299 pixels both props (`border` and `margin`) are passed to the render method without an issue. 
But if one defines overlapping query intervals where the same props are defined will result in a warning because the prop value is ambiguous, for example:
```json
props:
  {small: {border:"black"}}
  {medium: {border:"red"}}
query:
  {small: {minHeight: 200}
  {medium: {minHeight: 300}
```
This will result in a warning because InstUI cannot decide what value the `border` should have above 299 pixels.

matyasf:
Nvm there is a warning, just for some reason Chrome automatically sets a console filter now

joyenjoyer:
Please review the code again @matyasf @aaronshaf.

matyasf (approved):
looks good!

packages/ui-responsive/src/Responsive/README.md
```diff
@@ -81,3 +81,48 @@ example: true
   }}
 </Responsive>
 ```
+
+### Defining multiple breakpoints
+
+Setting up multiple media query breakpoints are supported. You can use 4 breakpoints at the same time: `minWidth`, `maxWidth`, `minHeight`,
```
HerrTopi, line 87: I think the correct grammar is:
```suggestion
Setting up multiple media query breakpoints is supported. You can use 4 breakpoints at the same time: `minWidth`, `maxWidth`, `minHeight`,
```

HerrTopi (approved):

packages/ui-responsive/src/Responsive/README.md
joyenjoyer: Yes, thank you!

PR #1263, opened 2023-08-23
fix(ui-source-code-editor): fix broken codemirror syntax highlight and add unit test

balzss:
added unit test for syntax highlight

Closes: INSTUI-3831

TEST PLAN:
check example code in docs app and see if syntax highlight works


matyasf (approved):
nice work! It'd be really good to figure out sometime why this tends to break

HerrTopi (approved):

PR #1264, opened 2023-08-23
chore: increase chromatic threshold to avoid false positives

balzss:
INSTUI-3837

this seems to work but I'm not 100%, my suggestion is to merge this and get back to the issue if we still have false positives in chromatic


HerrTopi (approved):
Seems good. Let's see if this fixes it

joyenjoyer (approved):

PR #1265, opened 2023-08-23
Account for absent position property on element

aaronshaf:


matyasf:
We'll fix it here: https://github.com/instructure/instructure-ui/pull/1266 The code looks good, but we use conventional commits (which was broken for a few commits perhaps this is why the git hook was not triggered)

PR #1268, opened 2023-08-24
move from yarn to npm

balzss:
INSTUI-3737

- CI breaks when `@netlify/esbuild-linux-64` is missing([support ticket #1](https://answers.netlify.com/t/deploy-fails-with-failed-to-find-package-netlify-esbuild-linux-64-on-the-file-system/97041), [support ticket #2](https://answers.netlify.com/t/github-actions-install-error-failed-to-find-package-netlify-esbuild-linux-64/91035)) -> works if it's manually installed from the workflow file

### Test plan

> Please note: this is a huge PR that touches a lot of our tooling so testing/reviewing should be done with extra care

- check out the PR and try running all of our usual scripts with npm (`npm install`, `npm run bootstrap`, `npm run dev`)
- try building the docs app
- try running tests
- check the docs: all (most) mentions of `yarn` should be replaced
- make a fresh clone and see if the usual scripts still run
- if anything else to test comes to your mind, please leave a comment about it


matyasf:
There are some leftovers from yarn:
`.yarnrc.yml` and the `.yarn` folder, can you delete these?

.github/workflows/pr-validation.yml
```diff
@@ -11,22 +11,25 @@ jobs:
         name: Install Node 16
         with:
           node-version: '16'
-          cache: 'yarn'
+          cache: 'npm'
+      - name: Install netlify-cli
+        run: npm install --save-dev @nx/nx-linux-x64-gnu
```
matyasf, line 16: This installs `nx` not netlify

balzss: `nx` also needs to be installed in the CI, I just labelled it wrong. I'll fix it

balzss:
> There are some leftovers from yarn: `.yarnrc.yml` and the `.yarn` folder, can you delete these?

@matyasf sure!

package.json
```diff
@@ -14,12 +14,12 @@
     "packages/*"
   ],
   "scripts": {
-    "prestart": "yarn bootstrap",
-    "start": "yarn workspace docs-app start",
-    "start:watch": "yarn workspace docs-app start:watch ",
-    "dev": "yarn start:watch",
+    "prestart": "npm run bootstrap",
```
matyasf, line 17: This is behaving differently from before. when I run `npm start dev` it runs this one too, so it will take ages.

```
matyas.szabo@c02dg48pmd6t instructure-ui % npm start dev

> instructure-ui@8.43.0 prestart
> npm run bootstrap


> instructure-ui@8.43.0 bootstrap
> node scripts/bootstrap.js

cleaning packages...
lerna notice cli v7.1.3
clean finished
lerna notice filter including "@instructure/ui-icons"
lerna info filter [ '@instructure/ui-icons' ]

> @instructure/ui-icons:prepare-build
...
```

Same for `npm run start:examples`

balzss: for the first one I think you should do `npm run dev` not `npm start dev`

balzss: for the second, it runs bootstrap because we have this line in package.json: `"prestart:examples": "npm run bootstrap"`

balzss: so this seems to be intended

HerrTopi: All the scripts including `npm run dev` is pretty fast for me. Arguably faster than before

matyasf (approved):
nice work!

PR #1270, opened 2023-08-31
docs: fix playground reset on resize

HerrTopi:
Closes: INSTUI-3838

TEST PLAN:
Open up the docs app. Change something in the playground. Resize the screen. check if the playground reset or not

### DIFF
--- a/packages/__docs__/src/Description/index.tsx
+++ b/packages/__docs__/src/Description/index.tsx
@@ -32,11 +32,18 @@ import type { DescriptionProps } from './props'
 class Description extends Component<DescriptionProps> {
   static propTypes = propTypes
   static allowedProps = allowedProps
+  private compiledMarkdown: any[] | null = null
+
+  componentDidMount() {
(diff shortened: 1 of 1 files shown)


joyenjoyer (approved):

packages/__docs__/src/Description/index.tsx
```diff
@@ -32,11 +32,18 @@ import type { DescriptionProps } from './props'
 class Description extends Component<DescriptionProps> {
   static propTypes = propTypes
   static allowedProps = allowedProps
+  private compiledMarkdown: any[] | null = null
+
+  componentDidMount() {
```
matyasf, line 37: shouldt this be in `componentDidUpdate`? If the props change then it wont be updated.

HerrTopi: Unfortunately the content of the compiledMarkdown can not be stored as state, so it cant be rewritten with componentDidUpdate. 
Luckily the props can never change, so for the time being, this solution should work fine

matyasf (approved):

PR #1272, opened 2023-09-04
Update dependencies

HerrTopi:


packages/ui-codemods/lib/utils/formatSource.ts
```diff
@@ -24,15 +24,15 @@
 
 import prettier from 'prettier'
 
-export default function formatSource(source: string, sourcePath: string) {
+export default async function formatSource(source: string, sourcePath: string) {
```
matyasf, line 27: if you make this async then everything that calls it needs to be awaited. This would make codemods async which is a large task (if even possible), so I'd not upgrade prettier in this package

matyasf (changes requested):

joyenjoyer (approved):

matyasf (approved):

PR #1275, opened 2023-09-06
fix(ui-responsive): round dimensions for query matcher

joyenjoyer:
Closes: INSTUI-3785

If dimensions are fractions, integer breakpoints are not going to match with them. Rounding width and height values solves this issue.

### **Test plan:**

1. Set up breakpoints in a Responsive component (or use the example code added in a followup)
2. Resize width and check if props object is not empty


joyenjoyer:
[example.txt](https://github.com/instructure/instructure-ui/files/12537004/example.txt)

balzss (approved):
looks good to me :)

HerrTopi (approved):

PR #1290, opened 2023-09-19
fix(ui-tree-browser): improve TreeBrowser keyboard only navigation

csmaty:
In the current implementation, when the `TreeBrowser` component is placed as the last focusable element within a modal, attempting to navigate forward using the Tab button disrupts the tab navigation cycle.

This happens because `stopPropagation` is called on all keydown events by default in `handleKeyDown`.

## Before change:

https://github.com/instructure/instructure-ui/assets/15729147/7c7db4ca-4c3c-48fd-b537-23a3219c2e83

## After change:

https://github.com/instructure/instructure-ui/assets/15729147/0fb3af21-b83b-4b5b-8927-d1407608d473


joyenjoyer (approved):
Nice work.

HerrTopi (approved):
Nice catch and fix. Thank you!

Issue #1286, opened 2023-09-13
SimpleSelect KO and SR opening issues


akos0horvath:
### Background Information

**Package Version(s)**:
ui-simple-select v8.43.1

**Browser**:
Chrome

**OS:**
Win10, Win11, MacOS Ventura

**Device:**
Lenovo Thinkpad, Macbook Pro

**Component**:
SimpleSelect

**Screen Reader:**
NVDA, VoiceOver

### Describe the Issue
SimpleSelect does not open with 'space' key on Win10 & Win11 when NVDA is on.
SimpleSelect does not open with control+option+space with Chrome with VoiceOver on Mac.
Probably two separate issues.

### Steps To Reproduce
On Windows:
1. Turn on NVDA
2. Open Chrome
3. Go to https://instructure.design/#SimpleSelect
4. Press space on a SimpleSelect component => nothing happens.
5. (turn off NVDA, press space => it opens the dropdown as expected)

On Mac:
1. Turn on VoiceOver
2. Open Chrome
3. Go to https://instructure.design/#SimpleSelect
4. Press control+option+space on a SimpleSelect component => nothing happens.

### Expected Behavior
SimpleSelect should open on 'space' key on Windows regardless of NVDA is enabled or not.
SimpleSelect should open on control+option+space with VoiceOver with Chrome (it works with Safari).

### Additional Information

**Requested Priority:**
High

matyasf:
Thanks for the report, we are looking into it

joyenjoyer:
Hey @akos0horvath! The [latest release](https://github.com/instructure/instructure-ui/releases/tag/v8.44.0)  contains the fix for this issue. Let us know if there is any issue still.

akos0horvath:
thank you @joyenjoyer , all good now!

PR #1291, opened 2023-09-19
Inst UI 3858 select dropdown

joyenjoyer:
**Test plan**
Please open up the issue [here](https://github.com/instructure/instructure-ui/issues/1286) and go through the steps to reproduce section for all of the Windows/Mac Safari/Chrome combinations.


packages/ui-utils/src/index.ts
```diff
@@ -21,9 +21,6 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-export { Browser } from './Browser'
-export { isEdge } from './isEdge'
-export { isIE11 } from './isIE11'
```
matyasf, line 26: Removing Browser and isIE11 would be a breaking change. Please add them back and add to its documentation that its deprecated (use @deprecated) and will be removed in v9

packages/ui-utils/src/getBrowser.ts
```diff
@@ -0,0 +1,63 @@
...
+const isEdge = () => {
+  return getBrowser().name === 'Edge'
+}
+
+const isIE = () => {
+  return getBrowser().name === 'IE'
+}
+
+const isChrome = () => {
+  return getBrowser().name === 'Chrome'
```
matyasf, line 56: This part is tricky. As I see from UAParser's code the browser's name can be a lot of things that uses the same engine. It has a also an "engine" field, but that one does not seem to differentiate Webkit for Chrome and Safari.  So this code is fine, but we should not that it would not detect e.g. Chromium. Maybe I'd add 2-3 of the most popular alternative browsers

matyasf (approved):

HerrTopi (approved):

PR #1293, opened 2023-09-20
feat(grid): add stretch option to vAlign

lalii05:
My goal with this PR is to be able to stretch grid cells, so that if one cell is taller then the others then they can all get the same height.

### DIFF
--- a/packages/ui-grid/src/Grid/props.ts
+++ b/packages/ui-grid/src/Grid/props.ts
@@ -44,7 +44,7 @@ type GridOwnProps = {
   colSpacing?: 'none' | 'small' | 'medium' | 'large'
   rowSpacing?: 'none' | 'small' | 'medium' | 'large'
   hAlign?: 'start' | 'center' | 'end' | 'space-around' | 'space-between'
-  vAlign?: 'top' | 'middle' | 'bottom'
+  vAlign?: 'top' | 'middle' | 'bottom' | 'stretch'
   startAt?: GridBreakpoints
   visualDebug?: boolean
 }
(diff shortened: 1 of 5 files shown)


packages/ui-grid/src/Grid/props.ts
```diff
@@ -70,7 +70,7 @@ const propTypes: PropValidators<PropKeys> = {
     'space-around',
     'space-between'
   ]),
-  vAlign: PropTypes.oneOf(['top', 'middle', 'bottom']),
+  vAlign: PropTypes.oneOf(['top', 'middle', 'bottom', 'stretch']),
```
HerrTopi, line 73: This stretch option needs to be added to the ts type above as well

packages/ui-grid/src/GridCol/props.ts
```diff
@@ -95,7 +95,7 @@ const propTypes: PropValidators<PropKeys> = {
     'space-around',
     'space-between'
   ]),
-  vAlign: PropTypes.oneOf(['top', 'middle', 'bottom']),
+  vAlign: PropTypes.oneOf(['top', 'middle', 'bottom', 'stretch']),
```
HerrTopi, line 98: This stretch option needs to be added to the ts type above as well

packages/ui-grid/src/GridRow/props.ts
```diff
@@ -72,7 +72,7 @@ const propTypes: PropValidators<PropKeys> = {
     'space-around',
     'space-between'
   ]),
-  vAlign: PropTypes.oneOf(['top', 'middle', 'bottom']),
+  vAlign: PropTypes.oneOf(['top', 'middle', 'bottom', 'stretch']),
```
HerrTopi, line 75: This stretch option needs to be added to the ts type above as well

HerrTopi (changes requested):

joyenjoyer (approved):

PR #1297, opened 2023-09-26
fix(ui-color-picker): add a padding to colorpicker button

joyenjoyer:
Closes: INSTUI-3867

Adding a padding to color picker button calculated by the text input children element's label height and margin.

This should be replaced in INSTUI v9 with a better solution.

**Test plan**
Add 

```      
renderMessages={() => [{ type: "error", text: "Long error message" }]}
```
to a Color Picker as an attribute and check if the input field and the botton stays aligned.

### DIFF
--- a/packages/ui-color-picker/src/ColorPicker/index.tsx
+++ b/packages/ui-color-picker/src/ColorPicker/index.tsx
@@ -116,7 +116,8 @@ class ColorPicker extends Component<ColorPickerProps, ColorPickerState> {
       hexCode: '',
       showHelperErrorMessages: false,
       openColorPicker: false,
-      mixedColor: ''
+      mixedColor: '',
+      labelHeight: 0
     }
   }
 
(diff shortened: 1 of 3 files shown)


packages/ui-color-picker/src/ColorPicker/index.tsx
```diff
@@ -591,14 +602,18 @@ class ColorPicker extends Component<ColorPickerProps, ColorPickerState> {
           themeOverride={{ padding: '' }}
           renderAfterInput={this.renderAfterInput()}
           renderBeforeInput={this.renderBeforeInput()}
+          inputContainerRef={this.setLabelHeight}
```
HerrTopi, line 605: Wouldn't it be better to save the ref on the class and calculate the height on the style instead of using state? It would be more consistent for with the codebase, for sure and would make more sense setting the ref on inputContainerRef, instead of calculating the label hight. Is there anything preventing us doing that way?

HerrTopi (changes requested):

HerrTopi (approved):

matyasf (approved):

PR #1299, opened 2023-09-27
chore(ui-utils): add chromium based browser to browser check

joyenjoyer:
Closes: INSTUI-3868


matyasf (approved):

balzss (approved):
isn't ms edge also chromium based?

joyenjoyer:
> isn't ms edge also chromium based?

Good point, I added it.

PR #1309, opened 2023-10-09
feat(ui-spinner): add delay prop and functionality to the Spinner

HerrTopi:
Closes: INSTUI-3884

### DIFF
--- a/packages/ui-spinner/src/Spinner/__new-tests__/Spinner.test.tsx
+++ b/packages/ui-spinner/src/Spinner/__new-tests__/Spinner.test.tsx
@@ -1,0 +1,52 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015 - present Instructure, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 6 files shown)


packages/ui-spinner/src/Spinner/index.tsx
```diff
@@ -61,6 +61,7 @@ class Spinner extends Component<SpinnerProps> {
 
   ref: Element | null = null
   private readonly titleId?: string
+  private delayTimeout?: NodeJS.Timeout
```
joyenjoyer, line 64: Does this work with SSR?

HerrTopi: I see no reason why it shouldn't.

balzss (approved):

packages/ui-spinner/src/Spinner/__new-tests__/Spinner.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+import React from 'react'
+import { render, waitFor } from '@testing-library/react'
+
+import '@testing-library/jest-dom/extend-expect'
+import Spinner from '../index'
+
+describe('<Spinner />', () => {
+  describe('with the delay prop', () => {
+    it('should delay rendering', async () => {
+      const { container } = render(<Spinner delay={1000} />)
```
matyasf, line 33: I'd use a smaller delay (say 300ms) to not slow down our test suite

packages/ui-spinner/src/Spinner/index.tsx
```diff
@@ -148,6 +164,10 @@ class Spinner extends Component<SpinnerProps> {
       </View>
     )
   }
+
+  render() {
+    return this.state.shouldRender ? this.renderSpinner() : null
```
matyasf, line 169: I would still render a transparent rectangle when`shouldRender` is false to prevent "pop in" and potential realignment of the UI. Maybe render the svg with `visibility: hidden`?

matyasf (changes requested):

packages/ui-spinner/src/Spinner/index.tsx
HerrTopi: This would defeat the purpose of the delay, since the goal of the feature is to not disrupt the UI if the load time is small

matyasf (approved):

PR #1310, opened 2023-10-10
INSTUI-3877 fix(ui-date-time-input): clear TimeSelect value when DateInput value is cleared

matyasf:
TEST PLAN:
make a DateTimeInput with `isRequired = false`. When you clear the date value the time value and the message should clear.
When isRequired = true it should display an invalid date message.


balzss (approved):
looks and works fine. are those chromatic errors false positives? I don't see any visual changes...

joyenjoyer (approved):
nice job

PR #1316, opened 2023-10-11
fix(ui-tray): remove double mount/unmount in ui-tray

### LINKED ISSUE #1315: Tray open/close triggers multiple needless re-mounts/unmounts of tray contents
### Background Information

**Package Version(s)**:
ui-tray 8.45.0

**Browser**:
Chrome

**Component**:
Tray

### Describe the Bug

When a <Tray>'s open is prop toggled from false to true, the inner contents are unmounted then mounted twice. It renders and mounts a trivial component a total of three times. This should only happen once.

LIkewise, when open is toggled from true to false, the inner contents are unmounted, then re-mounted, then unmounted again. This should only happen once.

### Steps To Reproduce
Reproduced here:

https://codesandbox.io/s/ecstatic-river-lh3px3

See the console in codesandbox

### Expected Behavior

Trivial component should only mount once upon Tray toggling from closed to open.

Trivial component should only unmount once upon Tray toggling from open to closed.

**Products Affected:**

Canvas. Perf. If using data hooks to load data in contents of Tray, can sometimes trigger multiple needless data fetches.

**Requested Priority:**

High

aaronshaf:
closes #1315

### DIFF
--- a/packages/ui-tray/src/Tray/index.tsx
+++ b/packages/ui-tray/src/Tray/index.tsx
@@ -201,25 +201,21 @@ class Tray extends Component<TrayProps> {
               css={this.props.styles?.tray}
               ref={contentRef}
             >
-              {this.state.transitioning ? (
-                children
-              ) : (
-                <Dialog
-                  as="div"
-                  label={label}
-                  defaultFocusElement={defaultFocusElement}
-                  open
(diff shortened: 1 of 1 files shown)


aaronshaf:
how do I re-trigger?

PR #1327, opened 2023-10-18
feat(ui-tabs): add active property to tabs

joyenjoyer:
Closes: INSTUI-3876

Add support for multiple tabs->one tab panel. This makes using React Router easier with less boilerplate code.

Test plan: check if the new example in the Tabs page works fine.

### DIFF
--- a/packages/ui-tabs/src/Tabs/__new-tests__/Tabs.test.tsx
+++ b/packages/ui-tabs/src/Tabs/__new-tests__/Tabs.test.tsx
@@ -1,0 +1,128 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015 - present Instructure, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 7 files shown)


matyasf (approved):
Looks good. While I'm not a fan of this new API (its confusing, e.g. why am I able to specify children to all tabs when 1 will be rendered anyway?) I guess we cant do more without breaking changes. In the future we should switch to a similar structure as material-ui IMO https://mui.com/material-ui/react-tabs/#experimental-api

balzss (approved):

packages/ui-tabs/src/Tabs/__new-tests__/Tabs.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+    expect(container).toBeInTheDocument()
+    expect(tabContent).toBeInTheDocument()
+
+    const childContent = screen.queryByText('Child')
+
+    expect(childContent).toBeNull()
+  })
+
+  it('should render the same content in second tab when selected', () => {
```
joyenjoyer, line 89: I added a test as you requested. @matyasf

PR #1328, opened 2023-10-18
Inst UI 3879 convert UI dialog to the new test format

git-nandor:


balzss (changes requested):
please use the latest versions of the testing library packages. you can do this by deleting the related lines from the package.json and install them with `npm i -D @testing-library/jest-dom @testing-library/react @testing-library/user-event`

balzss (approved):

joyenjoyer (approved):

PR #1329, opened 2023-10-18
support ReactTestUtils in decorators that use forwardRef

balzss:
INSTUI-3893

### DIFF
--- a/packages/emotion/src/__tests__/withStyle.test.tsx
+++ b/packages/emotion/src/__tests__/withStyle.test.tsx
@@ -24,6 +24,8 @@
 
 /** @jsx jsx */
 import React from 'react'
+import ReactDOM from 'react-dom'
+import ReactTestUtils from 'react-dom/test-utils'
 import PropTypes from 'prop-types'
 
 import { expect, match, mount, stub, within } from '@instructure/ui-test-utils'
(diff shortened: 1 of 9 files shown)


packages/emotion/src/__tests__/withStyle.test.tsx
```diff
@@ -136,6 +138,28 @@ describe('@withStyle', async () => {
...
+    render() {
+      return (
+        <div>
+          <ThemeableComponent />
+        </div>
+      )
+    }
+  }
+
+  it.only('can be found and tested with ReactTestUtils', async () => {
```
matyasf, line 151: remove `.only`

packages/emotion/src/__tests__/withStyle.test.tsx
```diff
@@ -136,6 +138,28 @@ describe('@withStyle', async () => {
...
+    const rootNode = document.createElement('div')
+    document.body.appendChild(rootNode)
+
+    // eslint-disable-next-line react/no-render-return-value
+    const rendered = ReactDOM.render(<WrapperComponent />, rootNode)
+    ReactTestUtils.findRenderedComponentWithType(
+      rendered as any,
+      (ThemeableComponent as any).originalType
+    )
```
matyasf, line 160: This test has no assertion. Or will it throw an error if its not found?

matyasf (changes requested):

packages/emotion/src/__tests__/withStyle.test.tsx
balzss: it will and then the test fails

joyenjoyer (approved):
nice work

packages/emotion/src/__tests__/withStyle.test.tsx
balzss: done

matyasf (approved):

PR #1349, opened 2023-11-04
docs: refactor markdown rendering and add switching …

HerrTopi:
…feature to examples


docs/guides/writing-docs.md
```diff
@@ -0,0 +1,358 @@
...
+<Button>InstUI button</Button>
+```
+
+**_Note:_** you can use any instUI components in the examples
+
+### 3. example
+
+The most complex type is the `type: example`. It will render as the `embed` did, but it also provides access to the code, which is editable and the changes reflect on the rendered view immediately.
+
+There are to ways these examples can be written.
```
balzss, line 83: typo, should be `two`

package.json
```diff
@@ -117,5 +117,9 @@
   },
   "browserslist": [
     "extends @instructure/browserslist-config-instui"
-  ]
+  ],
+  "dependencies": {
+    "marked-react": "^2.0.0",
+    "uuid": "^9.0.1"
+  }
```
matyasf, line 124: is there any reason why these need to be in the root package.json?

HerrTopi: Nope :D

docs/contributor-docs/api-guidelines.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: API Guidelines
-category: Getting Started
-order: 5
+category: Contributor Guides
+order: 4
```
matyasf, line 4: I really like the new sidebar ordering, its much more readable for endusers IMO

docs/contributor-docs/writing-docs.md
```diff
@@ -0,0 +1,386 @@
...
+
+# How To Write Docs
+
+This page documents how to add and maintain documentation.
+
+## Overview
+
+The documentation site is generated from the **source code** and the `.md` files. There are two types of `.md` files: `README.md` and `named-md-files.md`, under the `docs folder`.
+
+The **source code** is parsed with [JSDoc](https://jsdoc.app/) and it inprocides `type` information for the docs.
```
matyasf, line 15: its also parsed by `react-docgen` for React components.

matyasf (changes requested):
When I try to open a functional example in CodeSandbox I get and error: `{"error":"Unable to process params for /define - please double check them. See https://codesandbox.io/docs/importing#create-wizard for more details."}` 
Also see my other comments.

packages/__docs__/src/Playground/index.tsx
```diff
@@ -188,24 +200,62 @@ class Playground extends Component<PlaygroundProps, PlaygroundState> {
     )
   }
 
-  render() {
-    const { styles } = this.props
-    const { code, fullscreen, rtl } = this.state
-    const PreviewComponent = ({ themeKey, themes }: AppContextType) => (
+  renderPreview = (code: any, themeKey: any, themes: any) => {
```
matyasf, line 203: Can you please type these?

packages/__docs__/src/Playground/index.tsx
```diff
@@ -188,24 +200,62 @@ class Playground extends Component<PlaygroundProps, PlaygroundState> {
...
           fullscreen={fullscreen}
           rtl={rtl}
           themeKey={themeKey}
           themes={themes}
         />
       </PreviewErrorBoundary>
     )
+  }
+
+  renderTabPanel = (themeKey: any, themes: any) => {
```
matyasf, line 225: Can you please type these?

matyasf (changes requested):
There is an empty gap between the sample rendering and the edit, codesandbox etc buttons that was not there in the original, please remove it
image
Also there are 2 cases with `any` typing

balzss (approved):
do we need to display the class/function tab all the time? might be better to only show when users open the source code of the component. otherwise it looks good

matyasf (approved):
looks good!

PR #1357, opened 2023-11-08
docs: add storybook example guide to guides

HerrTopi:
Closes: INSTUI-3931


matyasf (changes requested):
The doc is there, but its out of date. If we include it please make these changes:

- it should be noted in the doc that this is for internal development, not something for end users.
- run through the doc and check for outdated parts, e.g. `npm run generate:sketch` does not exists anymore

joyenjoyer (approved):

PR #1374, opened 2023-11-22
fix(shared-types,ui-heading): change h1 heading weight

joyenjoyer:
Closes: INSTUI-3934

**Test plan:** check if h1 level has font weight of 400


packages/ui-heading/src/Heading/theme.ts
```diff
@@ -53,7 +53,7 @@ const generateComponentTheme = (theme: Theme): HeadingTheme => {
     lineHeight: typography?.lineHeightFit,
```
balzss, line 53: I think you need to change this as well

balzss (approved):

HerrTopi (approved):

PR #1400, opened 2023-12-08
feat(ui-drilldown,ui-top-nav-bar): add shouldCloseOnClick

joyenjoyer:
Closes: INSTUI-3911

This property only works in smallViewPort mode.

**Test:** to make testing easier, you can use the code from the test case, which I added. Put it into a playground and observe, what happens with the menu with different value's of the `shouldCloseOnClick` prop.

**Expected behaviors for values:**

- `always`: the `Tray` should always close, when an option with `always` is selected
- `never`: the `Tray` should never close, when an option with `never` is selected
- `auto`: the `Tray` should close, if `href` property is added to the option, otherwise it should not!

### DIFF
--- a/packages/ui-drilldown/src/Drilldown/DrilldownOption/props.ts
+++ b/packages/ui-drilldown/src/Drilldown/DrilldownOption/props.ts
@@ -44,6 +44,8 @@ type DrilldownOptionValue = string | number | undefined
 
 type RenderContentVAlign = 'start' | 'center' | 'end'
 
+type ShouldCloseOnClick = 'auto' | 'always' | 'never'
+
 type DrilldownOptionVariant = Exclude<OptionsItemProps['variant'], 'selected'>
 
 type RenderContentProps = {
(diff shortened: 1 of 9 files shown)


packages/ui-drilldown/src/Drilldown/DrilldownOption/props.ts
```diff
@@ -200,7 +205,11 @@ const propTypes: PropValidators<PropKeys> = {
   descriptionRole: PropTypes.string,
   onOptionClick: PropTypes.func,
   defaultSelected: PropTypes.bool,
-  elementRef: PropTypes.func
+  elementRef: PropTypes.func,
+  shouldCloseOnClick: PropTypes.oneOfType([
```
balzss, line 209: is there a reason this can't be simply just a boolean?

matyasf: IMO this should be an enum, it has 3 values. While its possible with a boolean, it could be easily a source of bugs when e.g. `undefined` is auto cast to boolean

balzss: so you suggest something like `'auto' | 'always' | 'never'`?

matyasf: yes

balzss (approved):

HerrTopi (approved):

PR #1402, opened 2023-12-12
chore: add cypress component testing

balzss:
INSTUI-3958

test:

- run component tests locally in headless mode
- run component tests in non-headless mode
- check [documentation](https://1402--preview-instui.netlify.app/#integration-testing)


tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "./tsconfig.build.json",
   "compilerOptions": {
     "baseUrl": "./packages",
+    "types": ["cypress"],
```
matyasf, line 5: Putting this here makes WebStorm go a bit crazy. When I open one of the files in the `cypress` folder it complains that its not under any tsconfig, and then outputs hundreds of errors not finding built-in things of the DOM API like `Node` or `Array`.
I think it would be better to add a tsconfig file to this folder too and inherit from this one.

matyasf (changes requested):

tsconfig.json
matyasf: And to import types from Cypress explicitly instead of adding them globally

balzss: I don't even think that part is needed anymore. I was trying to solve some issue but that wasn't the right solution and I just left it there accidentally. Can you please check if it works for you now?

docs/testing/integration-testing.md
```diff
@@ -0,0 +1,79 @@
+---
+title: Integration testing
+category: Testing
+order: 5
+---
+
+# Integration testing
```
matyasf, line 7: I would not call these integration tests. Its rather something like read-world component testing

cypress/component/Popover.cy.tsx
```diff
@@ -0,0 +1,92 @@
...
+  Button,
+  CloseButton,
+  FormFieldGroup,
+  TextInput
+} from '../../packages/ui'
+
+import '../support/component'
+import 'cypress-real-events'
+
+const PopoverExample = () => {
```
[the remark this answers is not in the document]
balzss: my reasoning was that unlike unit tests, we should only have complex patterns tested like this and not all of our components. and these complex tests could include multiple components at the same time (say a Popover and a Drilldown) so it would be hard to know where to put them

docs/testing/integration-testing.md
balzss: good idea, that's more clear

cypress/component/Popover.cy.tsx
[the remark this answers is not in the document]
matyasf: ok, im sold

matyasf (approved):

HerrTopi (approved):

PR #1414, opened 2024-01-17
feat(ui-date-time-input): add initialTimeForNewDate prop to datetimei…

joyenjoyer:
…nput

Closes: INSTUI-3966

**Test:** 

1. set `initialTimeForNewDate` on DateTimeInput 
2. make sure `defaultValue` prop is not set
3. pick a date
4. observe if the chosen `initialTimeForNewDate` value appears in the TimeSelect input field

### DIFF
--- a/packages/ui-date-time-input/src/DateTimeInput/index.tsx
+++ b/packages/ui-date-time-input/src/DateTimeInput/index.tsx
@@ -40,6 +40,7 @@ import {
 } from '@instructure/ui-icons'
 import type { DateTimeInputProps, DateTimeInputState } from './props'
 import { propTypes, allowedProps } from './props'
+import { error } from '@instructure/console'
 
 /**
 ---
@@ -133,6 +134,23 @@ class DateTimeInput extends Component<DateTimeInputProps, DateTimeInputState> {
             .date(this.state.iso.date())
             .month(this.state.iso.month())
             .year(this.state.iso.year())
+        }
+        if (this.props.initialTimeForNewDate && !this.state?.timeSelectValue) {
+          const hour = Number(this.props.initialTimeForNewDate.slice(0, 2))
+          const minute = Number(this.props.initialTimeForNewDate.slice(3, 5))
+          if (isNaN(hour) || isNaN(minute)) {
+            error(
+              false,
+              `[DateTimeInput] initialTimeForNewDate prop is not in the correct format. Please use HH:MM format.`
+            )
+          } else if (hour < 0 || hour > 23 || minute > 59 || minute < 0) {
+            error(
+              false,
+              `[DateTimeInput] 0 <= hour < 24 and 0 <= minute < 60 for initialTimeForNewDate prop.`
+            )
+          } else {
+            parsed.hour(hour).minute(minute)
+          }
         }
         const newTimeSelectValue = parsed.toISOString()
         if (this.isDisabledDate(parsed)) {
--- a/package-lock.json
+++ b/package-lock.json
@@ -48201,6 +48201,8 @@
         "@instructure/ui-test-utils": "8.52.0",
         "@instructure/ui-themes": "8.52.0",
         "@testing-library/jest-dom": "^6.1.4",
+        "@testing-library/react": "^14.0.0",
+        "@testing-library/jest-dom": "^6.1.4",
         "@testing-library/react": "^14.0.0"
       },
       "peerDependencies": {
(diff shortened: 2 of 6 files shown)


packages/ui-date-time-input/src/DateTimeInput/index.tsx
```diff
@@ -134,6 +135,23 @@ class DateTimeInput extends Component<DateTimeInputProps, DateTimeInputState> {
...
         }
+        if (this.props.initialTimeForNewDate && !this.state.timeSelectValue) {
+          const hour = Number(this.props.initialTimeForNewDate.slice(0, 2))
+          const minute = Number(this.props.initialTimeForNewDate.slice(3, 5))
+          if (isNaN(hour) || isNaN(minute)) {
+            error(
+              false,
+              `[DateTimeInput] initialTimeForNewDate prop is not in the correct format. Please use HH:MM format.`
+            )
+          } else if (hour > 23 || minute > 59) {
```
matyasf, line 146: to be thorough I'd also add here a check for negative numbers. I'm sure that moment's parse-all parser would interpret them which is not good for us

matyasf (changes requested):
if I enter both defaultValue and initialTimeForNewDate the component crashes with `TypeError: Cannot read properties of undefined (reading 'timeSelectValue')` it should disregard `initialTimeForNewDate` and output a warning that it does nothing

packages/ui-date-time-input/src/DateTimeInput/index.tsx
```diff
@@ -134,6 +135,35 @@ class DateTimeInput extends Component<DateTimeInputProps, DateTimeInputState> {
             .month(this.state.iso.month())
             .year(this.state.iso.year())
         }
+        if (
+          !this.props.defaultValue &&
+          this.props.initialTimeForNewDate &&
+          !this.state?.timeSelectValue
+        ) {
+          const hour = Number(this.props.initialTimeForNewDate.slice(0, 2))
```
HerrTopi, line 143: I think you should validate this with Moment and parse it after so you can extract the hour and minutes safely and reliably

joyenjoyer: This is a simple parse of two numbers here, I added plenty of validation here and in the props check. I consider this enough, thanks for your advice though.

We would like to change Moment for something more future proof as soon as possible.

HerrTopi (approved):

matyasf (changes requested):
looks better now, but if there is a `defaultValue` then its value is ignored. To reproduce use this code:

```
  render () {
    return (
      <div style= { {height: '10rem',width: '40em'}}>
        <DateTimeInput
          description="Pick a date and time"
          datePlaceholder="Choose a date"
          dateRenderLabel="Date"
          timeRenderLabel="Time"
          invalidDateTimeMessage="Invalid date!"
          prevMonthLabel='Previous month'
          nextMonthLabel='Next month'
          defaultValue="2018-01-18T13:30"
          layout="columns"
          initialTimeForNewDate="23:43"
        />
    </div>)
  }
}
render(<Example />)
```

delete the date value, click outside of the component, and then enter a valid date. It will use `12:00` instead of `23:43` as date value

joyenjoyer:
> looks better now, but if there is a `defaultValue` then its value is ignored. To reproduce use this code:
> 
> ```
> class Example extends React.Component {
>   render () {
>     return (
>       <div style= { {height: '10rem',width: '40em'}}>
>         <DateTimeInput
>           description="Pick a date and time"
>           datePlaceholder="Choose a date"
>           dateRenderLabel="Date"
>           timeRenderLabel="Time"
>           invalidDateTimeMessage="Invalid date!"
>           prevMonthLabel='Previous month'
>           nextMonthLabel='Next month'
>           defaultValue="2018-01-18T13:30"
>           layout="columns"
>           initialTimeForNewDate="23:43"
>         />
>     </div>)
>   }
> }
> render(<Example />)
> ```
> 
> delete the date value, click outside of the component, and then enter a valid date. It will use `12:00` instead of `23:43` as date value

This is what we discussed in a meeting and what you requested in the comment above the last one:

> if I enter both defaultValue and initialTimeForNewDate the component crashes with `TypeError: Cannot read properties of undefined (reading 'timeSelectValue')` **it should disregard `initialTimeForNewDate` and output a warning that it does nothing**



If both props are present now, `initialTimeForNewDate` is ignored. A warning message is thrown too:

`Warning: [DateTimeInput] if defaultValue is set, initialTimeForNewDate will be ignored. 
logMessage @ console.js:53`

Let me know if there is a misunderstanding!

matyasf (approved):

PR #1416, opened 2024-01-22
feat(ui-source-code-editor): add custom search panel

balzss:
INSTUI-3961

test plan:

- go to the search example at the SourceCodeEditor in docs
- hit cmd+f -> the search panel should open
- start typing -> matches should be highlighted
- hit enter on a search -> it should jump to the next occurrence of the word
- try the up and down arrows -> it should jump to the prev/next occurrence of the word
- hit escape -> the search panel should close
- check documentation for the search panel
- try the `searchConfig` prop

### DIFF
--- a/packages/ui-source-code-editor/src/SourceCodeEditor/SearchPanel.tsx
+++ b/packages/ui-source-code-editor/src/SourceCodeEditor/SearchPanel.tsx
@@ -1,0 +1,147 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015 - present Instructure, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+import React, { useState } from 'react'
+import { createRoot } from 'react-dom/client'
+import {
+  setSearchQuery,
+  search,
+  findNext,
+  findPrevious,
+  SearchQuery,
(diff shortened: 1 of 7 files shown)


packages/ui-source-code-editor/src/SourceCodeEditor/SearchPanel.tsx
```diff
@@ -0,0 +1,136 @@
...
+    findNext(view)
+  }
+
+  const handleFindPrev = () => {
+    handleHighlightSearch(searchQueryStr)
+    findPrevious(view)
+  }
+
+  return (
+    <div className="search-panel">
```
matyasf, line 87: is this class name needed by CodeMirror? If not then it needs to be removed. Also is this `div` needed? its already rendered to a `div`

packages/ui-source-code-editor/src/SourceCodeEditor/SearchPanel.tsx
```diff
@@ -0,0 +1,136 @@
...
+      <TextInput
+        inputRef={(r) => r?.focus()}
+        size="small"
+        display="inline-block"
+        width="20rem"
+        placeholder="Find in code..."
```
matyasf, line 93: We never hardcode text in our components, they are almost always used by a i18n library. Please expose this to a required prop.

packages/ui-source-code-editor/src/SourceCodeEditor/SearchPanel.tsx
```diff
@@ -0,0 +1,136 @@
...
+        onKeyUp={handleKeyUp}
+        renderBeforeInput={<IconSearchLine size="x-small" />}
+        renderAfterInput={
+          <span>
+            <IconButton
+              size="small"
+              withBorder={false}
+              withBackground={false}
+              onClick={handleFindNext}
+              screenReaderLabel="Next result"
```
matyasf, line 105: hardcoded text

packages/ui-source-code-editor/src/SourceCodeEditor/SearchPanel.tsx
```diff
@@ -0,0 +1,136 @@
...
+export default function customSearch() {
+  return search({
+    createPanel: (view) => {
+      const dom = document.createElement('div')
+      dom.style.padding = '8px'
+      const root = createRoot(dom)
+      root.render(<SearchPanel view={view} />)
+      return { dom }
+    },
+    top: true // this is not working if the `createPanel` prop is present. might be a bug on codemirror's side
```
matyasf, line 134: hmm in this case should not we remove it? Codemirror might fix the bug and suddenly search panels will appear at the top..

matyasf (changes requested):

packages/ui-source-code-editor/src/SourceCodeEditor/SearchPanel.tsx
balzss: in that case should the search panel be disabled by default to limit the number of required props? and instead of a boolean, the prop could be the search panel config

packages/ui-source-code-editor/src/SourceCodeEditor/SearchPanel.tsx
balzss: good point, I will remove it

packages/ui-source-code-editor/src/SourceCodeEditor/SearchPanel.tsx
balzss: no, I was using it for debugging. I'll remove it

packages/ui-source-code-editor/src/SourceCodeEditor/SearchPanel.tsx
matyasf: hmm good questions. I would not disable it by default, its better that nothing. I also like the idea of a search panel config object, if its `null`/`undefined` its not rendered, and has these texts that are mandatory

matyasf (approved):

joyenjoyer (approved):

HerrTopi (approved):

PR #1424, opened 2024-02-02
fix(ui-editable): fix the Esc key event propagation inside the Editab…

judikdavid:
…le component

Previously when you hit Esc in an InPlacEdit component inside a Modal, the Modal would close.

TEST PLAN:

Create an InPlaceEdit inside a Modal, then focus the InPlaceEdit, hit Esc, the Modal should not close.


balzss (approved):
works great, thanks for the submission!

HerrTopi (approved):

PR #1426, opened 2024-02-02
fix(ui-overlays): add maskcounter

joyenjoyer:
**Test**: create a nested Modal setup and check if scrollbar is visible upon the closure of the nested modal.

Closes: INSTUI-3855

### DIFF
--- a/packages/ui-overlays/src/Mask/MaskCounter.ts
+++ b/packages/ui-overlays/src/Mask/MaskCounter.ts
@@ -1,0 +1,49 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015 - present Instructure, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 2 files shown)


packages/ui-overlays/src/Mask/MaskCounter.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const MaskCounter = (() => {
+
+  let counter = 0
+  const getCounter = () => counter
+  const setCounter = (value: number) => {
+    counter = value
+  }
+  const incrementCounter = () => {
+    setCounter(getCounter() + 1)
```
HerrTopi, line 33: Is this getter really necessary? I think you can access the counter directly for the same effect

packages/ui-overlays/src/Mask/index.tsx
```diff
@@ -72,7 +74,10 @@ class Mask extends Component<MaskProps> {
 
   componentWillUnmount() {
     if (this.props.fullscreen) {
-      noScroll.off()
+      MaskCounter.decrementCounter()
+      if (MaskCounter.getCounter() === 0) {
```
HerrTopi, line 78: Just to be extra careful, I'd check for counter being less than 1 (in case of rapid closure oof more than one modals)

HerrTopi (changes requested):

packages/ui-overlays/src/Mask/MaskCounter.ts
joyenjoyer: I prefer to leave this for encapsulation and immutability principles.

packages/ui-overlays/src/Mask/index.tsx
joyenjoyer: Good catch, thanks!

balzss (approved):

HerrTopi (approved):

PR #1441, opened 2024-02-13
test(ui-date-time-input,ui-modal): migrate old Modal tests

git-nandor:
Closes: INSTUI-3920


cypress/component/Modal.cy.tsx
```diff
@@ -0,0 +1,108 @@
...
+        onDismiss={onDismissSpy}
+        label="Modal Dialog"
+        shouldReturnFocus={false}
+      >
+        <p>Modal body text</p>
+      </Modal>
+    )
+    cy.root().should('contain', 'Modal body text')
+
+    cy.get('body').trigger('keydown', { keyCode: 27 })
```
balzss, line 59: this is fine like this but for these tests it might worth using the `cypress-real-events` to be closer how these work in the "real world"

cypress/component/Modal.cy.tsx
```diff
@@ -0,0 +1,108 @@
...
+    }
+    cy.mount(<Example handleDismiss={handleDismissSpy} />)
+
+    cy.root().should('contain', 'Modal body text')
+
+    cy.get('#increment-btn').click().wait(100)
+    cy.get('#value-indicator').should('contain', '1')
+    cy.wrap(handleDismissSpy).should('not.have.been.called')
+
+    cy.get('body').click(0, 0)
```
balzss, line 105: for clicks too

balzss (approved):
looks good, works fine locally. left some comments about using `cypress-real-events` in the future but I'm fine leaving these ones as they are since they are working fine. great job!

cypress/component/Modal.cy.tsx
git-nandor: I tried the following versions:
cy.realPress('Escape')
cy.realPress('Escape').wait(100)
cy.get('body').type('{esc}')

Unfortunately, in every case, they broke or became flaky.
Same for the clicks too.

joyenjoyer (approved):

PR #1460, opened 2024-02-28
Remove jsx pragma

HerrTopi:


HerrTopi:
Was needed in Nextjs 14.1, but not needed in Nextjs 14.2

PR #1462, opened 2024-03-05
fix(ui-options): add prop to move role to the outer tag

balzss:
INSTUI-3953


HerrTopi (approved):

balzss:
@HerrTopi I've updated the props name and the docs, please check again

HerrTopi (approved):

joyenjoyer (approved):

PR #1464, opened 2024-03-06
[BREAKING] remove desktopConfig prop from TopNavBar.Layout for v9

balzss:
INSTUI-4016

BREAKING CHANGE:
desktopConfig prop is removed from TopNavBar.Layout

should only be merged right before the v9 release

test plan:
- check docs for TopNavBar: every mention of `desktopConfig` or `hideActionsUserSeparator` should be removed
- add `desktopConfig` prop to `<TopNavBar.Layout />`: there should be a console error/warning about the deprecated prop with a url to the upgrade guide
- check the upgrade guide


packages/ui-react-utils/src/deprecated.ts
```diff
@@ -37,7 +37,7 @@ export interface DeprecatedDecorator {
 }
 
 const deprecated = (() => {
-  if (process.env.NODE_ENV === 'production') {
+  if (process.env.NODE_ENV === 'pproduction') {
```
HerrTopi, line 40: Probably not intentional :D Please fix this

HerrTopi (changes requested):

packages/ui-top-nav-bar/src/TopNavBar/TopNavBarLayout/DesktopLayout/index.tsx
```diff
@@ -51,7 +51,7 @@ private: true
 @deprecated(
   '9',
   { hideActionsUserSeparator: true },
-  'From v9, actionUserSeparator has to be visible at all times. Please do not use designs which hide it. Note: this prop is probably used inside the desktopConfig prop on [TopNavBar.Layout]'
+  'further info: https://google.com'
```
joyenjoyer, line 54: I love using Google when I have a concern about something but is this really the best thing to recommend here?

joyenjoyer (changes requested):
Please revisit the places I added a comment to.

joyenjoyer (approved):

HerrTopi (approved):

PR #1470, opened 2024-03-22
test(ui-date-time-input): migrate old DateTimeInput tests

git-nandor:


cypress/component/DateTimeInput.cy.tsx
```diff
@@ -0,0 +1,287 @@
...
+
+    cy.get('@timeInput').should('have.value', '')
+    cy.wrap(onChange).should('have.been.called')
+
+    cy.get('@dateInput').realClick().wait(100)
+
+    cy.contains('button', '22')
+      .realClick()
+      .wait(100)
+      .then(($btn) => {
```
balzss, line 75: why is the `$` notation? is this some best practice for cypress?

balzss (approved):
good job!

cypress/component/DateTimeInput.cy.tsx
git-nandor: It's standard practice based on [documentation](https://docs.cypress.io/guides/core-concepts/variables-and-aliases), so that's why I use it too.

HerrTopi (approved):

PR #1471, opened 2024-03-25
docs(ui-tabs): make id required and remove Tabs.Tab from the docs

HerrTopi:
Closes: INSTUI-4001


matyasf (changes requested):
looks good, but isnt this a breaking change? So it should be on the v9 branch with a breaking commit message?

PR #1474, opened 2024-03-26
fix(ui-drawer-layout): fix drawerlayout transition bug

joyenjoyer:
Closes: INSTUI-4022

When placement property is set to 'end', a weird scrollbar appears during transition. Transition type is set to 'fade' as a fix.

**Test:** check if ui-drawer-layout still has a weird animation if `placement='end'`


balzss (approved):
works well and looks great 👍

matyasf (approved):

PR #1476, opened 2024-04-01
fix(ui-tabs): fix automatic id generation in tabs

HerrTopi:
Closes: INSTUI-4001


matyasf (changes requested):
it will fail because `id`s are the same for different tabs

balzss (approved):

matyasf (approved):

PR #1481, opened 2024-04-04
WIP(ui-tooltip): add aria-label

HerrTopi:
Closes: INSTUI-3450


matyasf:
This version is likely not a good solution because its read twice by VoiceOver

matyasf:
We are closing this PR because after discussions with the a11y team it came out that we need a different solution to read tooltips in browse mode in NVDA

PR #1494, opened 2024-04-19
Inst UI 3384 bidirectional

joyenjoyer:


balzss (changes requested):
there is a lint error in `instructure-ui/packages/__docs__/components.js` you can find the details if you run `npm run test:patchset` locally

PR #1510, opened 2024-05-06
docs(instui-config): fix docs for v9

balzss:


matyasf (approved):
looks good, I think it should be merged to the release commit so it works properly

HerrTopi (approved):

PR #1515, opened 2024-05-14
chore: update ui-scripts version in package.json

joyenjoyer:


balzss:
seems like you branched off from a feature branch instead of master or something strange happened that messed up the git history. pls fix it

PR #1516, opened 2024-05-15
Inst UI 4066 investigation integration of vitest for unit testing

git-nandor:


balzss:
looks great, I think we should go forward with this

packages/ui-popover/src/Popover/__new-tests__/Popover.test.tsx
```diff
@@ -70,7 +71,7 @@ const PopoverContainer = ({
 
 describe('<Popover />', () => {
   afterEach(() => {
-    jest.resetAllMocks()
+    //jest.resetAllMocks()
```
joyenjoyer, line 74: Consider removing this instead of commenting.

balzss:
I guess `jest` as a dependency could also be removed in this pr

balzss (changes requested):
seems like vitest is running in watch mode by default (but not in CI), would be nice to disabled that

balzss:
also package-lock is not in synch with package.json, please update that

matyasf (changes requested):
dev-commands.md needs to be around the "npm run test" part, it still says that the tests are running in Chrome. otherwise good job!

balzss (approved):
great job!

matyasf (approved):

PR #1522, opened 2024-05-23
fix(ui-tabs): fix id generation when null is present as children

HerrTopi:
Closes: INSTUI-4069


matyasf (approved):
Now it works, but please try to add a test for this case
- there is a {false} tab
- user clicks a tab
- no error should be in the console. Also the selection is weird too, the last 2 tabs are selected, perhaps this is something we can test against instead

HerrTopi:
This fix can't be tested easily with the current unit tests, since it requires a complex example react component to be build (see in docs) and manipulated with clicks

PR #1535, opened 2024-06-06
chore(ui-menu,ui-navigation): improve typing

balzss:
INSTUI-4083


balzss:
@matyasf what do you think about loosening the type defs?

matyasf (approved):
This way is correct. This generic prop defines what the value of `event.currentTarget` will be for the `mouseEvent` (it defaults to `Element`), and even if we put there an InstUI component it still reads the underlying DOM element

HerrTopi (approved):

PR #1538, opened 2024-06-12
fix(ui-top-nav-bar): keep width unchanged when active status is set t…

joyenjoyer:
…o topnavbar.item

Closes: INSTUI-4085

Setting the letterSpacing as a workaround makes the width unchanged during state changes.

**Testing plan**
Set the `status` property to `active` on a TopNavBar.Item and observe whether the width of it changes. It should not.


balzss (approved):
interesting. where's this magic number coming from? maybe an explanatory comment would be nice

git-nandor (approved):

PR #1556, opened 2024-06-20
feat(ui-date-input): add year picker to DateInput and note to use the new api

balzss:
also fixed the broken a11y for DateInput

test plan:
- check the example for the DateInput with the year picker -> try if works as expected
- check both the example with and without the year picker for a11y
- check the notice for the old DateInput api

Closes: INSTUI-4097

### DIFF
--- a/packages/ui-date-input/src/DateInput/README.md
+++ b/packages/ui-date-input/src/DateInput/README.md
@@ -35,7 +35,42 @@ class Example extends React.Component {
 render(<Example />)
 ```
 
+### With year picker
+
+```javascript
+---
+type: example
+---
+class Example extends React.Component {
+  state = { value: '' }
+
+  render () {
+    return (
+      <DateInput
+        renderLabel="Choose a date"
+        assistiveText="Type a date or use arrow keys to navigate date picker."
+        width="20rem"
+        isInline
+        value={this.state.value}
+        onChange={(e, value)=> this.setState({value:value.value})}
+        invalidDateErrorMessage="Invalid date"
+        withYearPicker={{
+          screenReaderLabel: "Year picker",
+          startYear:1999,
+          endYear:2024,
+          maxHeight: "200px"
+        }}
+      />
+    )
+  }
+}
+
+render(<Example />)
+```
+
 ### Composing a DateInput in your Application
+
+> **Important note**: this is the legacy API for the DateInput. It is much easier to use the way described above and more importantly: less buggy and has better accesibility. Please don't use this version and if you are having an issue with the DateInput, check if you are using the new version already and consider updating if not.
 
 `DateInput` uses `Calendar` internally. See [Calendar](#Calendar) for more detailed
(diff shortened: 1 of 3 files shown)


matyasf (changes requested):
- when I enter an invalid date in the first example the invalid date message flashes for a second but then disappears
- as I mentioned before it should be nice if I could move with arrows in calendar dates, but as I understand this needs to happen in the Calendar component.

Mz biggest concern that its using `moment` which is an out of date library. if we switch to a more modern lib, it will be a breaking change for our users.
We should investigate how a polyfill for the new Temporal API can replace moment for us.

balzss:
> * when I enter an invalid date in the first example the invalid date message flashes for a second but then disappears
> * as I mentioned before it should be nice if I could move with arrows in calendar dates, but as I understand this needs to happen in the Calendar component.
> 
> Mz biggest concern that its using `moment` which is an out of date library. if we switch to a more modern lib, it will be a breaking change for our users. We should investigate how a polyfill for the new Temporal API can replace moment for us.

1. good catch, this is fixed now
2. yes, this should be done in Calendar
3. regarding moment: end users (devs) don't really interact or depend on moment because this component simply returns a date string (just like the original `DateInput`) but is uses moment internally to parse and validate the dates. both of these can be replaced somewhat easily with the native `Date` object but that causes a small mismatch in the validation of this and `Calendar` component. this is addressed in the code with a comment and it won't change anything for users when we remove it later

packages/ui-date-input/src/DateInput2/index.tsx
```diff
@@ -0,0 +1,262 @@
...
+    setInputMessages(messages || [])
+  }, [messages])
+
+  const handleInputChange = (e: SyntheticEvent, value: string) => {
+    onChange?.(e, value)
+  }
+
+  const handleDateSelected = (
+    dateString: string,
+    _momentDate: any, // real type is Moment but used any to avoid imporing the moment lib
```
joyenjoyer, line 122: typo: imporing

joyenjoyer:
When I set `size` to either `small` or `medium`, no height or width changes only when it is set to `large`. Is that an expected behavior?

balzss:
> When I set `size` to either `small` or `medium`, no height or width changes only when it is set to `large`. Is that an expected behavior?

nice catch! the icon wasn't changing size with the input so it prevented it to become small, now it's fixed

joyenjoyer (approved):
nice job!

matyasf (approved):
looks good, very nice work!
FYI we have a `@isWIP` tag that you can use in our markdown files that wont make them appear in the sidebar (but you can see it by typing the URL)

balzss:
> looks good, very nice work! FYI we have a `@isWIP` tag that you can use in our markdown files that wont make them appear in the sidebar (but you can see it by typing the URL)

good to know but I'm not sure if we want that since the component can more or less replace most of the current uses of the original DateInput and we explicitly recommend it in the docs

PR #1557, opened 2024-06-21
fix(ui-drilldown): index drilldown options to always preserve the correct order

balzss:
Closes: INSTUI-4098

test plan:

- open drilldown in the docs
- go to the `shouldHideOnSelect` example
- modify the example so the Drilldown.Options have numbers as ids in a random order (e.g.:
```
      <Drilldown.Option value="1" id="2">
        Option 1
      </Drilldown.Option>
      <Drilldown.Option value="2" id="1">
        Option 2
      </Drilldown.Option>
      <Drilldown.Option value="3" id="3">
        Option 3
      </Drilldown.Option>
```
- use arrow keys to navigate between options
- the highlighted option should change as expected (for the incorrect behaviour see our currently deployed docs page)

### DIFF
--- a/packages/ui-drilldown/src/Drilldown/index.tsx
+++ b/packages/ui-drilldown/src/Drilldown/index.tsx
@@ -85,7 +85,7 @@ type OptionData = {
 
 // Contains the Option ComponentElement and the extra data we need track on it,
 // but don't want to expose as props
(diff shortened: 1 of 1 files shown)


git-nandor (approved):

joyenjoyer (approved):
nice job

PR #1561, opened 2024-06-27
fix(ui-a11y-utils,ui-dom-utils): fix FocusRegion when its inside an iframe

matyasf:
WIP, needs unit tests

fix that will allow to create focus trap inside an iframe like
```
const main = document.getElementById('iframe').contentWindow.document.getElementById('main2')
FocusRegionManager.activateRegion(main)
```

### DIFF
--- a/packages/ui-dom-utils/src/getComputedStyle.ts
+++ b/packages/ui-dom-utils/src/getComputedStyle.ts
@@ -40,16 +40,21 @@ import type { UIElement } from '@instructure/shared-types'
  * @param { string | null | undefined } pseudoElt - A string specifying the pseudo-element to match. Omitted (or null ) for real elements.
  * @returns { Object } object containing css properties and values for the element
  */
-function getComputedStyle(el?: UIElement, pseudoElt?: string | null) {
-  let style = {}
+function getComputedStyle(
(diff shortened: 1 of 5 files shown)


packages/ui-dom-utils/src/getComputedStyle.ts
```diff
@@ -40,16 +40,21 @@ import type { UIElement } from '@instructure/shared-types'
  * @param { string | null | undefined } pseudoElt - A string specifying the pseudo-element to match. Omitted (or null ) for real elements.
  * @returns { Object } object containing css properties and values for the element
  */
-function getComputedStyle(el?: UIElement, pseudoElt?: string | null) {
-  let style = {}
+function getComputedStyle(
+  el?: UIElement,
+  pseudoElt?: string | null
+): CSSStyleDeclaration {
+  let style = {} // TODO(breaking) return undefined instead!
```
matyasf, line 47: this is a bug, but we can only fix it with a major release. sometimes just an empty object is returned, in this case functions that call this one can fail, e.g. in `findFocusable`: `getComputedStyle(element).getPropertyValue('position')`

matyasf:
we do not need this now, will reopen it if needed

PR #1563, opened 2024-06-28
fix(ui-text-input): fix height issues

joyenjoyer:
Closes: INSTUI-3950


balzss:
this should be closed, as discussed with design

joyenjoyer:
Won't do.

PR #1567, opened 2024-07-03
fix(ui-number-input): set numberinput type to number

joyenjoyer:
Closes: INSTUI-3857

**Test**
Observe if `NumberInput` component has type `number`. Default arrows of the html input component should be hidden in every browser. Please check this in multiple browsers.


balzss (changes requested):
the failing tests in CI are looking for a text input which is no longer present, I think those should be updated

balzss (approved):

matyasf (approved):
looks good in Chrome, FF, Safari and mobile FF

PR #1568, opened 2024-07-03
fix(ui-top-nav-bar): lowered topnavbar zindex so it goes below trays/popovers/modals/etc

balzss:
INSTUI-4116

explanation:

- topnavbar desktop layout z-index didn't have any effect because of the missing `position: relative`
- topnavbar was positioned above everything else (including tray/modal/popovers/etc) which is not the desired behavior according to the design team

test plan:

- add a tray component and a topnavbar to the same page and check if tray is above topnavbar is both small viewport and desktop
- add a select/popover/menu to tray and check if the stacking order is still correct (no menu behind navbar or tray)


joyenjoyer (approved):
nice job

matyasf (approved):

PR #1571, opened 2024-07-05
feat(ui-top-nav-bar): add workaround hack for new topnav design with the old api

balzss:
INSTUI-4121

test plan:

- open topnav example
- resize viewport to display smallviewportlayout
- add `renderNavbarStartDangerousHack` prop to the `smallViewportConfig`
- test if the "hack" shows up in small viewport (and not in desktop view)

### DIFF
--- a/packages/ui-top-nav-bar/src/TopNavBar/TopNavBarLayout/SmallViewportLayout/props.ts
+++ b/packages/ui-top-nav-bar/src/TopNavBar/TopNavBarLayout/SmallViewportLayout/props.ts
@@ -113,6 +113,11 @@ type SmallViewportLayoutOwnProps = {
    * Callback fired when an item is selected in the dropdown menu
    */
   onDropdownMenuSelect?: DrilldownProps['onSelect']
+
+  /**
+   * A way to add generic html/react content to the start (left side) of the nav bar. This is a temporary workaround if you are using a design that is not possible to achive in the TopNavBar normally.
+   */
+  renderNavbarStartDangerousHack?: React.ReactNode
 }
 
 type TopNavBarSmallViewportLayoutOwnProps = CommonTopNavBarLayoutProps &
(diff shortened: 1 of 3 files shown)


packages/ui-top-nav-bar/src/TopNavBar/TopNavBarLayout/SmallViewportLayout/props.ts
```diff
@@ -113,6 +113,11 @@ type SmallViewportLayoutOwnProps = {
    * Callback fired when an item is selected in the dropdown menu
    */
   onDropdownMenuSelect?: DrilldownProps['onSelect']
+
+  /**
+   * Stay away...
+   */
```
matyasf, line 119: Please add a proper description. We should also mention that this is a deprecated prop that people should not use

matyasf (approved):
other than the description it works well :)

packages/ui-top-nav-bar/src/TopNavBar/TopNavBarLayout/SmallViewportLayout/props.ts
```diff
@@ -113,6 +113,11 @@ type SmallViewportLayoutOwnProps = {
...
+  renderNavbarStartDangerousHack?: React.ReactNode
```
joyenjoyer, line 120: Maybe expand the type with HTMLElement if someone wants to add a simple html tag, otherwise nice job.

balzss: regular html tags should be compatible with `ReactNode`, at least in my testings it didn't throw any error for them

packages/ui-top-nav-bar/src/TopNavBar/TopNavBarLayout/SmallViewportLayout/props.ts
balzss: done

joyenjoyer (approved):
good job

PR #1573, opened 2024-07-08
INSTUI-4123 Upgrade command-utils to use EcmaScript modules

matyasf:
`eslint-plugin-import` was replaced with `eslint-plugin-import-x` because it does not support the exports field in package.json now, see https://github.com/import-js/eslint-plugin-import/issues/1810 

This necessitated upgrading `typescript-eslint` to v6 

TEST PLAN:
if build/test/chramotic scripts are running this change should be ok


package.json
```diff
@@ -87,14 +87,15 @@
     "danger": "^11.3.1",
     "esbuild": "^0.21.3",
     "eslint": "^8.57.0",
+    "@typescript-eslint/eslint-plugin": "^6.21.0",
     "husky": "^9.0.11",
```
matyasf, line 91: for this new version this needs to be in this top place for some reason

packages/command-utils/package.json
```diff
@@ -3,7 +3,11 @@
   "version": "9.1.0",
   "description": "Node CLI utilities made by Instructure Inc.",
   "author": "Instructure, Inc. Engineering and Product Design",
-  "main": "./lib/index.js",
+  "type": "module",
+  "exports": {
+    ".": "./lib/index.js",
+    "./package.json": "./package.json"
+  },
```
matyasf, line 10: This is how a standards compliant package using ESM should look like. In the future sadly we will have to add the legacy "module" field https://stackoverflow.com/questions/42708484/what-is-the-module-package-json-field-for

package.json
joyenjoyer: What do you mean by in this top place?

matyasf: I mean in the top package.json

balzss (approved):

package.json
```diff
@@ -71,7 +71,7 @@
     "@commitlint/config-conventional": "^17.8.1",
     "@emotion/cache": "^11.11.0",
     "@emotion/react": "^11.11.4",
-    "@instructure/ui-scripts": "8",
+    "@instructure/ui-scripts": "9",
```
balzss, line 74: is this dependency even needed?

balzss: probably, or before we "unified" lots of our scripts. other question: does `command-utils` need to be a public package?

matyasf: I've removed this dependency from here.

> probably, or before we "unified" lots of our scripts. other question: does command-utils need to be a public package?

nope, but I'll need to double check

matyasf: IMO we should set everything thats possible to private in a future pr

joyenjoyer (approved):
nice job

PR #1575, opened 2024-07-09
test(ui-portal): migrate old Portal tests

git-nandor:
Closes: INSTUI-4124


balzss (approved):
I guess this need to be updated to use vitest. otherwise looks good

PR #1577, opened 2024-07-10
Inst UI 4119 convert UI link to the new test format

ToMESSKa:


balzss (changes requested):
jest should be replaced with vitest

joyenjoyer (approved):

balzss (approved):

PR #1578, opened 2024-07-11
INSTUI-3794  update dependencies. Major updates are TS, Vitest

matyasf:
TEST PLAN:
run scripts that are not run automatically by CI, e.g. bump


packages/ui-a11y-utils/package.json
```diff
@@ -29,7 +29,7 @@
     "@instructure/ui-dom-utils": "9.2.0",
     "@instructure/ui-react-utils": "9.2.0",
     "@instructure/uid": "9.2.0",
-    "keycode": "^2.2.1"
```
matyasf, line 32: Its better to define supported versions for user-consimed packages in the broadest range possible to reduce the chance of including it twice in their builds. Luckily we just have a few such dependencies

balzss (approved):

PR #1581, opened 2024-07-12
test(ui-side-nav-bar): migrate old SideNavBar tests

git-nandor:
Closes: INSTUI-4131


cypress/component/SideNavBarItem.cy.tsx
```diff
@@ -21,20 +21,31 @@
...
-import { SideNavBar } from './index'
-import { SideNavBarItemLocator } from './SideNavBarItem/SideNavBarItemLocator'
+import { SideNavBarItem } from '../../packages/ui-side-nav-bar'
+import { IconAdminLine } from '../../packages/ui-icons'
+import '../support/component'
 
-export { SideNavBarItemLocator }
+describe('<SideNavBarItem/>', () => {
+  it('should show a tooltip when the nav is minimized ', async () => {
+    const onClick = cy.spy()
```
balzss, line 33: no need to change this but is there a reason this is a spy and not just something like `() => {}`?

balzss (approved):

joyenjoyer (approved):

cypress/component/SideNavBarItem.cy.tsx
git-nandor: This is due to the bulk rewriting of the old code stub(). It is entirely equivalent, with no particular reason. 
The only advantage is that this way the lint does not throw a warning: warning Unexpected empty method.

PR #1590, opened 2024-07-18
chore(ui-tag): fix typecheck error

balzss:
Fix typecheck and downgrade node from 22 to 20 because `netlify-cli` doesn't work with it nicely. It time out the `npm install` and failed all CI checks.


matyasf (approved):
good to go is CI passes. please add some description in the PR why we needed to downgrade node

git-nandor (approved):

PR #1592, opened 2024-07-19
feat(ui-date-input,ui-text-input): add custom validation function

balzss:
and fix small layout issue

### DIFF
--- a/packages/ui-date-input/src/DateInput2/README.md
+++ b/packages/ui-date-input/src/DateInput2/README.md
@@ -1,5 +1,5 @@
 ---
-describes: DateInput
+describes: DateInput2
 ---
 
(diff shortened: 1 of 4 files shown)


git-nandor (approved):

matyasf (changes requested):
something is off with the validation, it does not trigger if I click outside from the calendar just when I unfocus the text field. 

https://github.com/user-attachments/assets/1e0b65a5-3f30-4ae9-bdea-d6ea264636d5

balzss:
> something is off with the validation, it does not trigger if I click outside from the calendar just when I unfocus the text field.
> 
>  Screen.Recording.2024-07-23.at.11.23.38.mov

oh, you are right! it validates on input blur and I didn't consider this scenario. will fix it

PR #1601, opened 2024-07-25
chore: clean up npm scripts and pr validation workflow

balzss:


package.json
```diff
@@ -52,7 +44,7 @@
     "dev:examples": "lerna run start:watch --stream --scope docs-examples",
     "prestart:examples": "npm run bootstrap",
     "start:examples": "lerna run start --stream --scope docs-examples",
-    "husky:pre-commit": "npm-run-all update:package:list commit:package:list lint:staged ts:check:references",
+    "husky:pre-commit": "npm run update:package:list && npm run commit:package:list && npm run lint:staged && npm run ts:check:references",
```
matyasf, line 47: I think this will fail on Windows. It would be really nice if we could run `bootstrap`, `dev` and make + push a commit on Windows when we test there a11y issues

matyasf (approved):

git-nandor (approved):

PR #1606, opened 2024-07-26
chore: fix build on Windows

matyasf:
not fully fixed yet, cant commit due to https://github.com/conventional-changelog/commitlint/issues/4106

what works:
- `bootstrap`
- `dev`
- `test:vitest` almost  (1 test fails for some reason)

### DIFF
--- a/packages/__docs__/build-docs.mjs
+++ b/packages/__docs__/build-docs.mjs
@@ -1,4 +1,0 @@
-// scripts must be built from .ts files before this is run
-import {buildDocs} from './lib/build-docs.mjs'
-
-buildDocs()
--- a/packages/__docs__/build-docs.mjs
+++ b/packages/__docs__/build-docs.mjs
@@ -1,0 +1,4 @@
+// scripts must be built from .ts files before this is run
+import {buildDocs} from './lib/build-docs.mjs'
+
+buildDocs()
--- a/packages/__docs__/package.json
+++ b/packages/__docs__/package.json
@@ -11,7 +11,7 @@
   "homepage": "https://instructure.design",
   "bugs": "https://github.com/instructure/instructure-ui/issues",
   "scripts": {
-    "prestart": "npm run build:scripts:ts && node -e 'import(\"./lib/build-docs.mjs\").then(file => file.buildDocs());'\n",
+    "prestart": "npm run build:scripts:ts && node build-docs.mjs",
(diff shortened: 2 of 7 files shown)


packages/__docs__/package.json
```diff
@@ -11,7 +11,7 @@
   "homepage": "https://instructure.design",
   "bugs": "https://github.com/instructure/instructure-ui/issues",
   "scripts": {
-    "prestart": "npm run build:scripts:ts && node -e 'import(\"./lib/build-docs.mjs\").then(file => file.buildDocs());'\n",
```
matyasf, line 14: Windows does not like such a fancy command at all :/ had to move it to a file

packages/ui-scripts/lib/icons/build-icons.js
```diff
@@ -42,9 +43,9 @@ export default {
   },
   handler: async (argv) => {
     const configFile = path.join(process.cwd(), argv.config)
-    const config = await import(configFile)
+    const configFileURL = pathToFileURL(configFile)
+    const config = await import(configFileURL)
```
matyasf, line 47: `import()` for files must begin with `file://`, this is added by this built in helper

packages/__docs__/build-docs.mjs
```diff
@@ -0,0 +1,4 @@
+// scripts must be built from .ts files before this is run
+import {buildDocs} from './lib/build-docs.mjs'
+
+buildDocs()
```
balzss, line 4: was this file created in windows? I think this file ending is not compatible with unix EOF. I think editing this file in macos should solve this

balzss: also the same file name with `lib/build-docs.mjs` is a bit confusion. I'm wondering if there's that file already, can't that one be run instead?

balzss (changes requested):

packages/__docs__/build-docs.mjs
matyasf: There is a way not to create a new file and not use fancy syntax in `package.json` but its also ugly: https://stackoverflow.com/a/75760840 What do you think?

matyasf: fixed it, now it doesn't need an extra file

HerrTopi (approved):

balzss (approved):

PR #1617, opened 2024-08-01
test(ui-metric): migrate old Metric tests

ToMESSKa:
Closes: INSTUI-4150


balzss (approved):

packages/ui-metric/src/Metric/__new-tests__/Metric.test.tsx
```diff
@@ -23,65 +23,59 @@
...
-    const metric = await MetricLocator.find()
-    const label = await metric.findWithText('Grade')
-
-    expect(
-      await label.find('[role="rowheader"]', { expectEmpty: true })
-    ).to.not.exist()
+    expect(label).not.toHaveAttribute('role', 'rowheader')
   })
 
   it('should allow methods and ReactNodes as labels and values', async () => {
```
joyenjoyer, line 76: I see that it accepts a div but what about methods? The test mentions that too. Please revisit this.

joyenjoyer (changes requested):
Revisit this test please: `should allow methods and ReactNodes as labels and values`

joyenjoyer (approved):

PR #1618, opened 2024-08-02
WIP(instui-config,ui-modal): allow simple react nodes as modal children

joyenjoyer:


packages/ui-modal/src/Modal/README.md
```diff
@@ -509,6 +509,73 @@ class Example extends React.Component {
 render(<Example />)
 ```
 
+### Using custom children
+
+Occasionally, you might find it useful to incorporate custom components within a `Modal`, such as a higher-order component for `Modal.Header` or `Modal.Body` or not using built in child components at all. Although this approach is typically not advised, it can sometimes aid in code splitting or achieving more streamlined code, especially for more intricate and sizable `Modal`s. In these instances, it is crucial to ensure that the necessary props are consistently passed down manually.
```
balzss, line 514: I would describe/list these props

packages/ui-modal/src/Modal/index.tsx
```diff
@@ -160,23 +160,25 @@ class Modal extends Component<ModalProps, ModalState> {
...
-          })
-        } else {
-          return safeCloneElement(child, {
-            variant: variant
-          })
-        }
+    return Children.map(children, (child) => {
+      if (!child) return // ignore null, falsy children
+
+      if (matchComponentTypes<BodyChild>(child, [ModalBody])) {
```
balzss, line 166: will this part work with custom children? I mean the `variant` and `overflow` props? or do the users have to add these manually too?

joyenjoyer: No, they only work if they used our own children components.

packages/ui-modal/src/Modal/index.tsx
```diff
@@ -160,23 +149,14 @@ class Modal extends Component<ModalProps, ModalState> {
...
-            overflow: child.props.overflow || overflow
-          })
-        } else {
-          return safeCloneElement(child, {
-            variant: variant
-          })
-        }
-      }
-    )
+    return Children.map(children, (child: any) => {
```
matyasf, line 152: Please o not use `any` if possible.  This line should be `return Children.map(children as ReactElement, (child) => {` because `safeCloneElement` expects a `ReactElement`. This makes it also necessary to add extra checks for non-React component children like a simple string. I think the best check is to use `isvalidElement` like here https://github.com/instructure/instructure-ui/blob/master/packages/ui-table/src/Table/index.tsx#L136

packages/ui-modal/src/Modal/index.tsx
```diff
@@ -160,23 +149,14 @@ class Modal extends Component<ModalProps, ModalState> {
...
+      if (!child) return // ignore null, falsy children
+
+      return safeCloneElement(child, {
+        variant: variant,
+        overflow: child?.props?.overflow || overflow
+      })
```
matyasf, line 158: This line is causing hard to debug issues in some edge cases. If you add a `Tabs` or an `Alert` as a child you will get an exception because we use the `variant` prop often but not with the `inverse` value. AFAIK we cant do much here, I would just mention it in the docs that these 2 props are passed down to the children

packages/ui-modal/src/Modal/__new-tests__/Modal.test.tsx
```diff
@@ -350,29 +350,6 @@ describe('<Modal />', () => {
...
-      expect(dialog).toBeInTheDocument()
-      expect(consoleErrorMock).toHaveBeenCalledWith(
-        expect.any(String),
-        expect.any(String),
-        expect.stringContaining(expectedErrorMessage),
-        expect.any(String)
-      )
-    })
-
     it('should pass inverse variant to children when set', async () => {
```
matyasf, line 353: Please add a test that tests if custom/weird children work. Something like this https://github.com/instructure/instructure-ui/blob/master/packages/ui-table/src/Table/__new-tests__/Table.test.tsx#L170

packages/ui-modal/src/Modal/README.md
```diff
@@ -1051,6 +1051,99 @@ On smaller viewports (like mobile devices or scaled-up UI), we don't want to los
   render(<Example />)
   ```
 
+### Using custom children
+
+Occasionally, you might find it useful to incorporate custom components within a `Modal`, such as a higher-order component for `Modal.Header` or `Modal.Body` or not using built in child components at all. Although this approach is typically not advised, it can sometimes aid in code splitting or achieving more streamlined code, especially for more intricate and sizable `Modal`s.
+
+Below example demonstrates how to use a higher-order component for `Modal.Body`. `Modal` consists of a `Modal.Header`, a custom `WrappedModalBody` component, and a `View` component. `Modal.Header` and `WrappedModalBody` inherit the `variant` and `overflow` properties automatically, while the `View` component does not.
```
matyasf, line 1058: why does the view not inherit these props? as I see from the code it does

matyasf (changes requested):
see my comments

packages/ui-modal/src/Modal/README.md
```diff
@@ -1051,6 +1051,99 @@ On smaller viewports (like mobile devices or scaled-up UI), we don't want to los
...
+
+  handleButtonClick = () => {
+    this.setState(function (state) {
+      return { open: !state.open }
+    })
+  };
+
+  renderCloseButton () {
+    return (
+      <CloseButton
```
matyasf, line 1082: Please add `color="primary-inverse"` so this button is visible

packages/ui-modal/src/Modal/props.ts
```diff
@@ -156,7 +149,7 @@ type ModalOwnProps = {
   /**
    * The children to be rendered within the `<Modal />`. Children must be type of: `Modal.Header`, `Modal.Body`, `Modal.Footer`. The `Modal.Body` child is required, and they have to follow this order.
```
matyasf, line 150: This is no longer true. Id rewrite this that these are the recommended children, but one can use custom children and that the `variant` and `overflow` props are always passed down

matyasf (changes requested):
nice work, there are some small docs changes

matyasf (approved):

balzss (approved):

PR #1632, opened 2024-08-08
test(ui-codemods): migrate old Codemods tests

git-nandor:
Closes: INSTUI-4126


.github/workflows/release.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       - name: Set up project.
         run: npm run bootstrap
       - name: Run tests.
-        run: USE_REACT_STRICT_MODE=0 npm run test && npm run test:vitest && npm run test:jest && npm run cy:component
+        run: USE_REACT_STRICT_MODE=0 npm run test && npm run test:vitest && npm run cy:component
```
matyasf, line 23: we still have 1 `Jest` test suite in the `ui` package, but that will be converted to `vitest` next, so its OK that those tests are not running for a few builds.

packages/ui-codemods/ts-require-transform.ts
```diff
@@ -0,0 +1,52 @@
...
+    sourcefile: filename
+  })
+
+  return result.code
+}
+
+addHook(transformCode, {
+  exts: ['.ts', '.tsx'],
+  ignoreNodeModules: true
+})
```
matyasf, line 52: We are never unloading this hook. I guess its fine since tests are isolated https://vitest.dev/config/#isolate by default

matyasf (approved):
very nice work!

packages/ui-codemods/ts-require-transform.ts
joyenjoyer: Yes, this is not an issue.

HerrTopi (approved):

PR #1633, opened 2024-08-09
feat(ui-calendar,ui-date-input): improve DateInput2

balzss:
INSTUI-4167

this PR adds:

- return iso dateString from the `onChange` callback
- `formatDate` prop
- extend documentation

test plan:

- check docs for DateInput2
- check if the return values in the `onChange` callback are good (esp the dateString and whether you can parse it back to `Date`)
- try a custom `formatDate` function and see if it works as expected


HerrTopi (approved):

matyasf (approved):
looks good

PR #1635, opened 2024-08-13
update ESLint to v9 and other smaller dependencies

matyasf:
Closes: INSTUI-4099, INSTUI-4062, INSTUI-4183
to test: check docs app, check output of `npm run lint` for erroneous lint warnings. 

The good:
- `ui-eslint-config` package is gone :)
- Lots of small changes that were detected by the correctly configured linter, like `packages/template-` packages had some leftover files that could be deleted
- no weird ESLint errors when nothing was wrong
- no need for 100+ `.eslintignore` files
- it detected some duplicate tests, these were removed
- added a plugin for specifically linting `vitest` tests

The bad:
- `eslint-plugin-jsx-a11y` is not yet converted to ESLint v9, add it back when its ready https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/978
- `eslint-import-resolver-typescript` and `eslint-plugin-import-x` are not converted either, but these are less useful plugins
- Our own ESLint plugin is not converted yet to ESLint v9, now its turned off, should be converted in the near future (and wont need its own package).
- Docs app linting is not 100% yet because it has some messy interwoven browser+node code.

Other smaller changes:
- Fixed a warning in Table about the `isStacked` and `hover` prop that was always showing in the console
- Fixed a test that broke with the `moment` upgrade, turns out the test was bad and moment got better


eslint.config.mjs
```diff
@@ -0,0 +1,216 @@
+/*
```
matyasf, line 1: This file was the bulk of the work

packages/ui-time-select/src/TimeSelect/__tests__/TimeSelect.test.tsx
```diff
@@ -459,7 +459,7 @@ describe('<TimeSelect />', async () => {
     const select = await TimeSelectLocator.find()
     const input = await select.findInput()
     await input.change({ target: { value: '' } })
-    await input.typeIn('04:45:55 AM')
+    await input.typeIn('4:45:55 AM')
```
matyasf, line 462: new moment is more exact in strict parsing mode what we're using

packages/ui-utils/src/__new-tests__/hash.test.tsx
```diff
@@ -347,24 +347,6 @@ describe('hash', () => {
       expect(result1).toEqual(result2)
     })
 
-    it('hashes two different simple objects to different values', () => {
```
matyasf, line 350: duplicate test

packages/ui-buttons/src/Button/__new-tests__/Button.test.tsx
```diff
@@ -74,7 +74,7 @@ describe('<Button/>', () => {
     expect(button).toHaveTextContent(children)
   })
 
-  it('should render a button', () => {
+  it('should render a button with name prop', () => {
```
joyenjoyer, line 77: `Button` does not have a `name` property. The text content `Hello World` is its child.

packages/ui-source-code-editor/src/SourceCodeEditor/index.tsx
```diff
@@ -483,7 +483,7 @@ class SourceCodeEditor extends Component<SourceCodeEditorProps> {
     const { styles } = this.props
 
     if (!styles?.theme || !styles.highlightStyle) {
-      return
+      return undefined
```
joyenjoyer, line 486: Why is this needed here?

HerrTopi (approved):
Just a thought: shouldn't we turn "Unexpected any. Specify a different type" warnings off? Those are polluting the log and serve no real info.

Otherwise: great work!

packages/ui-source-code-editor/src/SourceCodeEditor/index.tsx
matyasf: ESLint expects that a getter always returns a value. IMO this is a good rule, and its OK if we make such edge cases more explicit

packages/ui-buttons/src/Button/__new-tests__/Button.test.tsx
matyasf: good catch, I'll delete it

matyasf:
> Just a thought: shouldn't we turn "Unexpected any. Specify a different type" warnings off? Those are polluting the log and serve no real info.

Yep, it gets a bit spammy because we have a lot of these. But I think they are useful, we should try to get rid of `any`s

PR #1640, opened 2024-08-15
feat(ui-date-input): improve DateInput2 api, extend docs

balzss:
/INSTUI-4219

this PR simplifies and improves the `DateInput2` api

the docs are also extended to address the experimental nature of the component and elaborate on usage

test plan:

try the following things in a couple different locales (e.g. `it-it`, `ar-eg`,  `pt-br`)
- try setting a date by typing it in manually and blurring the input
- try selecting a date from the picker
- check if the picker shows the correct date after manually typing or selecting from the picker and closing it
- try the year picker: see if the correct year is selected after closing or typing the date manually
- check the custom date validation, customize it and see if it works as expected
- test the default validation: try setting an invalid date
- try the date formatting: modify the example and see if it works as expected

### DIFF
--- a/packages/ui-date-input/src/DateInput2/README.md
+++ b/packages/ui-date-input/src/DateInput2/README.md
@@ -2,28 +2,37 @@
 describes: DateInput2
 ---
 
-This component is an updated version of [`DateInput`](/#DateInput) that's easier to configure for developers, has a better UX, better accessibility features and a year picker. We recommend using this instead of `DateInput` which will be deprecated in the future.
+> *Warning*: `DateInput2` is an **experimental** upgrade to the existing [`DateInput`](/#DateInput) component, offering easier configuration, better UX, improved accessibility, and a year picker. While it addresses key limitations of `DateInput`, it's still in the experimental phase, with some missing unit tests and potential API changes.
 
 ### Minimal config
 
 - ```js
   class Example extends React.Component {
-    state = { value: '' }
+    state = { inputValue: '', dateString: '' }
 
     render() {
       return (
-        <DateInput2
-          renderLabel="Choose a date"
-          screenReaderLabels={{
-            calendarIcon: 'Calendar',
-            nextMonthButton: 'Next month',
-            prevMonthButton: 'Previous month'
-          }}
-          value={this.state.value}
-          width="20rem"
-          onChange={(e, value) => this.setState({ value })}
-          invalidDateErrorMessage="Invalid date"
-        />
+        <div>
+          <DateInput2
+            renderLabel="Choose a date"
+            screenReaderLabels={{
+              calendarIcon: 'Calendar',
+              nextMonthButton: 'Next month',
+              prevMonthButton: 'Previous month'
+            }}
+            value={this.state.inputValue}
+            width="20rem"
+            onChange={(e, inputValue, dateString) => {
+              this.setState({ dateString, inputValue })
+            }}
+            invalidDateErrorMessage="Invalid date"
+          />
+          <p>
+            Input Value: <code>{this.state.inputValue}</code>
+            <br />
+            UTC Date String: <code>{this.state.dateString}</code>
+          </p>
+        </div>
       )
     }
   }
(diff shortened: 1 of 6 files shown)


packages/ui-date-input/src/DateInput2/README.md
```diff
@@ -2,28 +2,37 @@
 describes: DateInput2
 ---
 
-This component is an updated version of [`DateInput`](/#DateInput) that's easier to configure for developers, has a better UX, better accessibility features and a year picker. We recommend using this instead of `DateInput` which will be deprecated in the future.
+`DateInput2` is an experimental upgrade to the existing [`DateInput`](/#DateInput) component, offering easier configuration, better UX, improved accessibility, and a year picker. While it addresses key limitations of `DateInput`, it's still in the experimental phase, with some missing unit tests and potential (though unlikely) API changes. `DateInput` will be deprecated in the future, but for now, developers can start using `DateInput2` and provide feedback.
```
joyenjoyer, line 5: I would not expose this information to the public: whether we have unit tests or not.

joyenjoyer (approved):

packages/ui-date-input/src/DateInput2/README.md
balzss: I think we can disclose this for experimental components

packages/ui-date-input/src/DateInput2/props.ts
```diff
@@ -45,9 +45,9 @@ type DateInput2OwnProps = {
     nextMonthButton: string
   }
   /**
-   * Specifies the input value.
+   * Specifies the input value *before* formatting. The `formatDate` will be applied to it before displaying. Should be a valid, parsable date.
```
matyasf, line 48: I would mention here what a parseable date is, e.g. link this https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#date_time_string_format

matyasf (changes requested):
There is some weirdness going on with date parsing. I got these settings:
```
locale="hu-hu"
timezone="Australia/Sydney"
```
When I select say 10th August in the Calendar dropdown the first example will display:

```
UTC Date String: 2024-08-09T14:00:00.000Z
Input Value: 2024. augusztus 10.
```

Which is OK I guess, selecting 10th August (midnight) in Australia is 9th August in London.

But if I enter into this DateInput: `2024-08-10` and focus out, it treats the entered date as UTC:

```
UTC Date String: 2024-08-10T00:00:00.000Z
Input Value: 2024. augusztus 10.
```

matyasf:
Also watch out Date's parser is so broken that its scary. From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#date_time_string_format : "When the time zone offset is absent, date-only forms are interpreted as a UTC time and date-time forms are interpreted as local time." Either use Moment's or make 100% sure that we always add timezone offset

packages/ui-date-input/src/DateInput2/props.ts
balzss: that would be a good addition

matyasf:
Some funky locale bugs:

Set locale to `ar`, and I think that a slash is in the wrong place, it puts `7‏/1‏/2025` in the text input field (ISO date is 2025-01-06T23:00:00.000Z)
Set locale to `bg` or `fa`, select a date in the calendar, click on the textfield, then click out. It will display invalid date.

In the "with year picker" example I can move forward with the arrows to feb. 2025. Here the year picker will be empty. Same if I enter a date after 2024 manually. I dont have a good solution for this besides implementing some min/max prop.

balzss:
> Some funky locale bugs:
> 
> Set locale to `ar` (Arabic), and I think that a slash is in the wrong place, it puts `7‏/1‏/2025` in the text input field (ISO date is 2025-01-06T23:00:00.000Z) Set locale to `bg` (Bulgarian) or `fa` (Persian), select a date in the calendar, click on the textfield, then click out. It will display invalid date. You can check here what Canvas supports: https://community.canvaslms.com/t5/Canvas-Basics-Guide/Which-languages-does-Canvas-support/ta-p/19
> 
> In the "with year picker" example I can move forward with the arrows to feb. 2025. Here the year picker will be empty. Same if I enter a date after 2024 manually. I dont have a good solution for this besides implementing some min/max prop.

- arabic and persian locales should work but you have to switch the example to be displayed right to left
- bulgarian locale is tricky and I'm haven't found a good solution but fortunately it is not on the supported list that you've linked
- for the yearpicker limit: I'll try to think about a good solution

packages/ui-date-input/src/DateInput2/README.md
```diff
@@ -33,51 +42,149 @@ This component is an updated version of [`DateInput`](/#DateInput) that's easier
...
+
+### Parsing and formatting dates
+
+When typing in a date manually (instead of using the included picker), the component tries to parse the date as you type it in. By default parsing is based on the user's locale which determines the order of day, month and year (e.g.: a user with US locale will have MONTH/DAY/YEAR order, and someone with GB locale will have DAY/MONTH/YEAR order).
+
+Any of the following separators can be used when typing a date: `,`, `-`, `.`, `/` or a whitespace however on blur the date will be formatted according to the locale and separators will be changed and leading zeros also adjusted.
+
+If you want different parsing and formatting then the current locale you can use the `dateFormat` prop which accepts either a string with a name of a different locale (so you can use US date format even if the user is France) or a parser and formatter functions.
+
+The default parser also have a limitation of excluding years before `1000` and after `9999`. These values are invalid by default but not with custom parsers.
```
matyasf, line 84: The default parser also has a limitation of not working with years before...

packages/ui-date-input/src/DateInput2/README.md
```diff
@@ -33,51 +42,149 @@ This component is an updated version of [`DateInput`](/#DateInput) that's easier
...
+  render(<Example />)
+  ```
+
+### Parsing and formatting dates
+
+When typing in a date manually (instead of using the included picker), the component tries to parse the date as you type it in. By default parsing is based on the user's locale which determines the order of day, month and year (e.g.: a user with US locale will have MONTH/DAY/YEAR order, and someone with GB locale will have DAY/MONTH/YEAR order).
+
+Any of the following separators can be used when typing a date: `,`, `-`, `.`, `/` or a whitespace however on blur the date will be formatted according to the locale and separators will be changed and leading zeros also adjusted.
+
+If you want different parsing and formatting then the current locale you can use the `dateFormat` prop which accepts either a string with a name of a different locale (so you can use US date format even if the user is France) or a parser and formatter functions.
```
matyasf, line 82: Its not clear for me from this how the Dateformat prop as string differs from the `locale` prop. I would make a new paragraph for these where I would also write about how it gets calculated. You can reuse it from `DateTimeInput`:

"Localization
The component is localized via its locale and timezone parameters. Both are read from props, context and from the browser's locale in this priority order. locale determines the language and format dates and time are displayed in (e.g. month names, AM/PM or 24-hour format) and the beginning of the week (e.g. Monday in Germany, Sunday in the U.S.) in the dropdown calendar."

packages/ui-date-input/src/DateInput2/README.md
```diff
@@ -33,51 +42,149 @@ This component is an updated version of [`DateInput`](/#DateInput) that's easier
...
+        messages={[
+          {type: 'hint', text: 'YYYY-MM-DD'}
+        ]}
+        locale="en-us"
+        dateFormat={{
+          parser: (input) => {
+            // split input on '.', whitespace, '/', ',' or '-' using regex: /[.\s/.-]+/
+            // the '+' allows splitting on consecutive delimiters
+            const [year, month, day] = input.split(/[,.\s/.-]+/)
+            const newDate = new Date(year, month-1, day)
```
matyasf, line 129: This is interpreted in local time. I think it could lead to bugs if the component uses a different timezone -- it might return a date a day before/after?

packages/ui-date-input/src/DateInput2/index.tsx
```diff
@@ -44,150 +44,212 @@ import type { DateInput2Props } from './props'
...
+  //   Given a dateString, a locale and a timeZone. The dateString is assumed to be formatted according
+  //   to the locale. So if the locale is `en-us` the dateString is expected to be in the format of M/D/YYYY.
+  //   The dateString is also assumed to be in the given timeZone, so "1/1/2020" in "America/Los_Angeles" timezone is
+  //   expected to be "2020-01-01T08:00:00.000Z" in UTC time.
+  //   This function tries to parse the dateString taking these variables into account and return a javascript Date object
+  //   that is adjusted to be in UTC.
+
+  // Split string on '.', whitespace, '/', ',' or '-' using regex: /[.\s/.-]+/.
+  // The '+' allows splitting on consecutive delimiters.
+  // `.filter(Boolean)` is needed because some locales have a delimeter at the end (e.g.: "hu-hu")
```
matyasf, line 58: I think the filter description here is wrong. Or is there such a datestring as '2024/12/23hu-hu'?

packages/ui-date-input/src/DateInput2/index.tsx
```diff
@@ -44,150 +44,212 @@ import type { DateInput2Props } from './props'
...
+  // create utc date from year, month (zero indexed) and day
+  const date = new Date(Date.UTC(year, month - 1, day))
+
+  if (date.getMonth() !== month - 1 || date.getDate() !== day) {
+    // Check if the Date object adjusts the values. If it does, the input is invalid.
+    return null
+  }
+
+  // Format date string in the provided timezone
+  const parts = new Intl.DateTimeFormat('en-US', {
```
matyasf, line 91: why is here `en-US` hardcoded?

packages/ui-date-input/src/DateInput2/index.tsx
balzss: what I meant is hungarian have an extra `.` at the end (e.g.: `2024. 09. 20.`), I'll make this comment more clear

packages/ui-date-input/src/DateInput2/index.tsx
balzss: the goal here is the convert the date to the given `timeZone` which can be done via the `Intl.DateTimeFormat` which needs a locale but in this case, the locale is completely irrelevant. it could be any other locale and it would still work fine

packages/ui-date-input/src/DateInput2/README.md
balzss: might be true, I will check that

matyasf (approved):
nice work!

PR #1643, opened 2024-08-22
docs(ui-overlays,ui-pages): function based examples are ad…

ToMESSKa:
…ded to Overlay, Pages

Closes: INSTUI-4182

- for Pages, the conditional rendering of email, spouse, parent properties is reworked with a workaround because of of a bug in the Table component. Table cannot accept custom components and throws error when condtional rendering results in a null


matyasf (approved):

joyenjoyer (approved):
nice work

PR #1646, opened 2024-08-23
Convert instructure ESLint plugin to the v9 syntax, delete old package

matyasf:
TEST PLAN:
Try a local relative import, e.g. `import {getTextDirection} from '../../packages/i18n'`, it should fail linting with a recommendation for the good import


balzss:
works well for me, but I was wondering on thing. it says: `Relative import from another package is not allowed. Use "@instructure/ui-i18n/es" instead of "../../../ui-i18n" `. shouldn't it be without the `.../es` part? not a big deal just asking

HerrTopi (approved):

matyasf:
> works well for me, but I was wondering on thing. it says: `Relative import from another package is not allowed. Use "@instructure/ui-i18n/es" instead of "../../../ui-i18n" `. shouldn't it be without the `.../es` part? not a big deal just asking

I've fixed this

PR #1650, opened 2024-08-26
feat(ui-menu): add renderLabelInfo to Menu

joyenjoyer:
Closes: INSTUI-4192

Add `renderLabelInfo` - either a string or a React.Node - prop to a `Menu`. It should be shown near the label. Override the `minWidth` theme variable if the width is not enough for the renderLabelInfo content.


HerrTopi (approved):

balzss:
why have the pr actions not run for this?

joyenjoyer:
> why have the pr actions not run for this?

Good q. I can't remember clearly but first this was a draft pr, that might be the cause.

balzss:
I was able to add the prop to a menuitem and it works as expected but when I add it to the Menu itself, nothing (seems to) happen. is it supposed to work on Menu too? it is listed in its props

PR #1652, opened 2024-08-26
  INSTUI-4188 New Table API to allow custom children

matyasf:
- [x] use Context instead of passing down props
- [x] Allow custom children
- [x] write test for custom children
- [x] make sure that one can use falsey/null children
- [x] write example and docs for the new API

To test:

- Check out the new examples and play with them.   
- Try to add custom components as children that just wrap existing components.               
- Try to make fully new custom components.  
- Add stupid things as children to components (see one of the new tests), it should not result in an exception

### DIFF
--- a/docs/contributor-docs/v11-upgrade-guide.md
+++ b/docs/contributor-docs/v11-upgrade-guide.md
@@ -1,0 +1,25 @@
+---
+title: Upgrade Guide for Version 11
+category: Guides
+order: 99
+isWIP: true
+---
+
+# Upgrade Guide for Version 11
+
+## Removal of deprecated props/components
+
+### Table
+
+[Table](/#Table) is now using `TableContext` to pass down data to its child components, the following props are no longer passed down to their children:
+
+| Component | prop        | Substitute / Notes              |
+| --------- | ----------- | ------------------------------- |
+| `Row`     | `isStacked` | is now stored in `TableContext` |
+| `Body`    | `isStacked` | is now stored in `TableContext` |
+| `Body`    | `hover`     | is now stored in `TableContext` |
+| `Body`    | `headers`   | is now stored in `TableContext` |
+
+## Changes
+
+- `ui-dom-utils`/`getComputedStyle` can now return `undefined`: In previous versions sometimes returned an empty object which could lead to runtime exceptions when one tried to call methods of `CSSStyleDeclaration` on it.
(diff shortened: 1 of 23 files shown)


packages/ui-table/src/Table/TableContext.ts
```diff
@@ -0,0 +1,54 @@
...
+  /**
+   * Contents of the first row of cells. Has value if `isStacked` is `true`.
+   */
+  headers?: Renderable[]
+}
+
+/**
+ * React context created by the `Table` component to hold its data
+ */
+const TableContext = createContext<TableContextType>({
```
matyasf, line 47: The major changes is that instead of sending down props trough multiple components it uses now a Context. Props are still passed down for compatibility reasons

git-nandor (approved):

packages/ui-table/src/Table/Body/index.tsx
```diff
@@ -70,11 +71,17 @@ class Body extends Component<TableBodyProps> {
         css={styles?.body}
         role={isStacked ? 'rowgroup' : undefined}
       >
-        {Children.map(children as RowChild[], (child) =>
+        {Children.map(children as any[], (child) =>
           safeCloneElement(child, {
             key: child.props.name,
+            // Sent down for compatibility with custom components
+            // TODO DEPRECATED, remove in next version
```
joyenjoyer: I may add a specific version number (v11 or v12) here otherwise we could forget this and it may stay here forever.

packages/ui-table/src/Table/Head/index.tsx
```diff
@@ -177,12 +178,15 @@ class Head extends Component<TableHeadProps> {
...
   render() {
-    const { children, isStacked, styles } = this.props
-
-    return isStacked ? (
+    const { children, styles } = this.props
+    return this.context.isStacked ? (
       this.renderSelect()
     ) : (
-      <thead {...omitProps(this.props, Head.allowedProps)} css={styles?.head}>
+      // TODO remove 'hover' in the next version, its passed down for compatibility with custom components
```
joyenjoyer: Again please add a specific version here.

packages/ui-table/src/Table/Body/index.tsx
matyasf: fixed

packages/ui-table/src/Table/Head/index.tsx
matyasf: fixed

docs/contributor-docs/v11-upgrade-guide.md
```diff
@@ -0,0 +1,25 @@
+---
+title: Upgrade Guide for Version 11
+category: Guides
+order: 99
+isWIP: true
```
matyasf, line 5: this hides it in the docs app, but one can see it when typing in the URL

joyenjoyer (approved):
Cool, nice job!

PR #1658, opened 2024-08-29
docs(many): function based examples are added to SimpleSelect, ToggleDetails, Tooltip

Kadirsaglm:


packages/ui-tooltip/src/Tooltip/README.md
```diff
@@ -59,74 +59,121 @@ type: example
...
-  <MyComponent />
-</Tooltip>
-```
+  ;<Tooltip renderTip="Tooltip text to display">
+    <MyComponent />
+  </Tooltip>
+  ```
+
+- ```js
+  const MyComponent = React.forwardRef((props, ref) => {
```
Kadirsaglm, line 165: I'm not sure where the imports for the docs examples are defined, but if you could point out where they are defined, I can replace this with just `forwardRef` instead.

matyasf: They are defined in https://github.com/instructure/instructure-ui/blob/master/packages/__docs__/globals.js , feel free to add new imports from React here

Kadirsaglm: Done, thanks!

ToMESSKa (approved):

matyasf (approved):

PR #1659, opened 2024-08-29
docs(ui-text-input,ui-view): functional examples are added to TextInput, View

Kadirsaglm:


packages/ui-text-input/src/TextInput/README.md
```diff
@@ -108,79 +185,131 @@ Focusable content will be focused separately from the input itself.
...
+            onChange={this.handleChange}
+            renderBeforeInput={
+              <View display="block" padding="xxx-small 0">
+                {this.state.value !== '' && (
+                  <Tag
+                    text={this.state.value}
+                    margin="xxx-small xxx-small xxx-small none"
+                    onClick={function () {
+                      console.log(this.state.value)
+                    }}
```
ToMESSKa, line 215: This throws an error in the console: Uncaught TypeError: Cannot read properties of undefined (reading 'state').
I think if you convert this to an arrow function (like you did in the function based example) it will work fine. Could you please fix?
Feel free to rewrite the original class-based component if you notice an error. :)

packages/ui-view/src/View/README.md
```diff
@@ -458,47 +458,212 @@ The focus outline adjusts to account for the shape of the View. For example, the
...
+    updateFocused = (event) => {
+      this.setState({ isFocused: event.target.checked })
+    }
+
+    updateInset = (event) => {
+      this.setState({ inset: event.target.checked })
+    }
+
+    render() {
+      const { isFocused, focusPosition } = this.state
```
ToMESSKa, line 482: Here these values are deconstructed but never used later. So i don't think this line is necessary.

packages/ui-view/src/View/README.md
```diff
@@ -607,61 +774,126 @@ class FocusedExample extends React.Component {
...
+  class OverflowExample extends React.Component {
+    constructor(props) {
+      super(props)
 
-    this.state = {
-      overflowY: 'visible'
+      this.state = {
+        overflowY: 'visible'
+      }
     }
```
ToMESSKa, line 803: Here only overFlowY is declared, but overFlowX is missing. Could you please add overFlowX too?

ToMESSKa:
Nice job. All of the recommended changes are in the original class-based examples.

packages/ui-text-input/src/TextInput/README.md
Kadirsaglm: Fixed this and i changed other anonymous functions to arrow functions to be more coherent

packages/ui-view/src/View/README.md
Kadirsaglm: Removed

packages/ui-view/src/View/README.md
Kadirsaglm: Done

matyasf (approved):
nice work!

ToMESSKa (approved):

PR #1660, opened 2024-08-30
Fix mismatched vitest and internal dev dependencies versions

matyasf:
TEST PLAN:
tests should pass. package-lock.json should be a bit smaller


packages/ui-a11y-utils/package.json
```diff
@@ -36,7 +36,7 @@
     "@instructure/ui-color-utils": "10.2.0",
     "@testing-library/jest-dom": "^6.4.5",
     "@testing-library/react": "^15.0.7",
-    "vitest": "^1.6.0"
```
matyasf, line 39: we were using vitest version 2.0.2 version except in these few places

packages/ui-tag/src/Tag/__new-tests__/Tag.test.tsx
```diff
@@ -35,8 +35,6 @@ import { View } from '@instructure/ui-view'
 
 const originalOmitViewProps = View.omitViewProps
 
-// https://github.com/vitest-dev/eslint-plugin-vitest/issues/511
-// eslint-disable-next-line vitest/valid-describe-callback
 describe('<Tag />', async () => {
```
matyasf, line 38: the issue was fixed, so the disable is no longer needed

joyenjoyer (approved):

PR #1663, opened 2024-09-01
docs(ui-table): function based examples are added to Table

Kadirsaglm:


packages/ui-table/src/Table/README.md
```diff
@@ -310,126 +765,64 @@ class SortableTable extends React.Component {
...
-          </Table.Head>
+        <Table caption="Set text-align for columns">
+          <Table.Head>{renderHeaderRow()}</Table.Head>
           <Table.Body>
             <Table.Row>
-              {Object.entries(colTextAligns).map(([headerId, textAlign]) => {
-                return (
-                  <Table.Cell
-                    key={headerId}
-                    width={headers.find(header => header.id === headerId).width}
```
ToMESSKa, line 336: This width props was deleted from the original class based example and missing from the functional too. Is there a reason for that?

packages/ui-table/src/Table/README.md
```diff
@@ -167,141 +351,412 @@ class Example extends React.Component {
...
+            </Table.Body>
+          </Table>
+        </ToggleGroup>
+      )
+
+      return (
+        <FormField id="columnTextAlign" label="Set column text-align">
+          <Flex margin="0 0 medium">
+            {Object.entries(colTextAligns).map(([headerId, textAlign]) => {
+              return (
```
ToMESSKa, line 545: This whole Formfield is missing from the function-based component.

packages/ui-table/src/Table/README.md
```diff
@@ -167,141 +351,412 @@ class Example extends React.Component {
...
+    />
+  )
+  ```
 
+- ```javascript
+  const SortableTable = ({ caption, headers, rows }) => {
     const initialColWidth = {}
     headers.forEach((header) => {
       initialColWidth[header.id] = 'start'
     })
```
ToMESSKa, line 717: This whole functional example doesn't work, it throws an error: TypeError: Cannot read properties of undefined (reading 'onRequestSort'). Can you please look into this again?

packages/ui-table/src/Table/README.md
```diff
@@ -167,141 +351,412 @@ class Example extends React.Component {
...
+        ...prevState,
+        [id]: value
+      }))
+    }
 
-    return (
+    const renderHeaderRow = (direction) => (
       <Table.Row>
-        {(headers || []).map(({ id, text, width }) => (
+        {headers.map(({ id, text, width }) => (
```
ToMESSKa, line 751: Here there is truthy check i in the class-based example "((headers || []).map(...", which is missing from here.

packages/ui-table/src/Table/README.md
```diff
@@ -1136,93 +1997,218 @@ Also you need the following props on the components:
...
+      const Tag = isStacked ? 'div' : 'tr'
+      const rowStyle =
+        hover && this.state.isHovered
+          ? { backgroundColor: 'SeaGreen' }
+          : { backgroundColor: 'white' }
+
+      return (
+        <Tag
+          style={rowStyle}
+          role={isStacked ? 'row' : undefined}
```
ToMESSKa, line 2047: This role prop is missing from the functional example

packages/ui-table/src/Table/README.md
```diff
@@ -1136,93 +1997,218 @@ Also you need the following props on the components:
...
+          onMouseOver={this.toggleHoverOn}
+          onMouseOut={this.toggleHoverOff}
+        >
+          {React.Children.toArray(this.props.children)
+            .filter(React.isValidElement)
+            .map((child, index) => {
+              return React.cloneElement(child, {
+                key: child.props.name,
+                // used by `CustomTableCell` to render its column title in `stacked` layout
```
ToMESSKa, line 2056: I think this comment should be in the functional example too.

packages/ui-table/src/Table/README.md
Kadirsaglm: According to the docs `Table.Cell` doesn't have `width` prop and also in console below warning is shown

> console.js:53 Warning: [Cell] prop 'width' is not allowed. 

That's why I removed the width prop. If it is needed for some kind of a workaround solution I can add it back.

packages/ui-table/src/Table/README.md
Kadirsaglm: I just realized that the `renderOptions` function in the class-based example has two return statements, with the second one including the FormField. I think we should remove the unreachable return statement completely from the class-based component.

packages/ui-table/src/Table/README.md
Kadirsaglm: I couldn't reproduce this issue in my local. I can sort the table as expected. Could you provide more context like if there is any like special settings set etc.?

packages/ui-table/src/Table/README.md
Kadirsaglm: Done

packages/ui-table/src/Table/README.md
Kadirsaglm: Added

packages/ui-table/src/Table/README.md
Kadirsaglm: Added

packages/ui-table/src/Table/README.md
ToMESSKa: okay, thanks for the clarification

packages/ui-table/src/Table/README.md
ToMESSKa: I agree, it should be removed

packages/ui-table/src/Table/README.md
ToMESSKa: I checked it again, it works now 👍

packages/ui-table/src/Table/README.md
Kadirsaglm: Done

ToMESSKa (approved):

PR #1668, opened 2024-09-03
INSTUI-4080 fix(ui-selectable): fix Select options not being selectable on iOS Safari with VoiceOver on

matyasf:
`aria-owns` seems to be misused here. It rearranges the DOM seen by the screenreader, so the [subtree owned by the element will be its child](https://tink.uk/using-the-aria-owns-attribute/). In our case it meant that a popup will the the child of an `input` component.
Also its very rarely used, see https://www.makethingsaccessible.com/guides/aria-controls-vs-aria-owns/ , there is just one sample use case for it at W3C ([a navigation treeview](https://www.w3.org/WAI/ARIA/apg/example-index/#examples_by_role_label))

To test:
- Test all examples of Select, SimpleSelect, Selectable with keyboard navigation and VoiceOver. Test on mobile devices with TalkBack/VoiceOver and Windows too if possible

Note: This will need to be ported to InstUI v8 and v9

### DIFF
--- a/packages/ui-selectable/src/Selectable/__tests__/Selectable.test.tsx
+++ b/packages/ui-selectable/src/Selectable/__tests__/Selectable.test.tsx
@@ -800,14 +800,12 @@ describe('<Selectable />', async () => {
 
       expect(input.getAttribute('aria-expanded')).to.equal('false')
       expect(input.getAttribute('aria-controls')).to.not.exist()
-      expect(input.getAttribute('aria-owns')).to.not.exist()
 
       await subject.setProps({ isShowingOptions: true })
       expect(input.getAttribute('aria-expanded')).to.equal('true')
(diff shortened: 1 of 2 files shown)


ToMESSKa:
- on Mac with keyboard navigation and Voiceover: **tested, it worked**
- on Android with Talkback: **it doesn't work for me,** I can select the input field, but it doesn't react to double tap or anything else

ToMESSKa (approved):

matyasf:
> * on Mac with keyboard navigation and Voiceover: **tested, it worked**
> * on Android with Talkback: **it doesn't work for me,** I can select the input field, but it doesn't react to double tap or anything else

I've verified that this fix does not fix the issue on Android (but it does on iOS)

HerrTopi (approved):

PR #1669, opened 2024-09-04
chore: codeSandbox error fixed in Select, SimpleSelect, Selectable

ToMESSKa:
Closes: INSTUI-4201

test plan:

- test if examples in Select, SimpleSelect, Selectable are opened without error in CodeSandbox

### DIFF
--- a/packages/__docs__/src/CodeSandboxButton/index.tsx
+++ b/packages/__docs__/src/CodeSandboxButton/index.tsx
@@ -73,14 +73,16 @@ import 'moment/min/locales'
       ', '
     )}} from "@instructure/ui"\n\n`
 
-    const codeLines = this.props.code.split('\n')
-    const shouldRender = !codeLines[codeLines.length - 1]
-      .split(' ')
-      .join('')
-      .includes('render(<')
-    const codeBlock = shouldRender
-      ? `render(${this.props.code})`
-      : this.props.code
+    // removes all whitespace (spaces, tabs, newlines, etc.) from the code string
+    const cleanedCode = this.props.code.replace(/\s+/g, '')
+    const lastRenderIndex = cleanedCode.lastIndexOf('render(<')
+    const endsWithRenderStatement =
+      lastRenderIndex !== -1 && cleanedCode.endsWith('>)')
+
+    const codeBlock = endsWithRenderStatement
+      ? this.props.code
+      : `render(${this.props.code})`
+
     const renderStatement = `const render = (el) => { ReactDOM.render(el, document.getElementById('app')) }\n`
     const codeSandboxData = {
       title: this.props.title,


packages/__docs__/src/CodeSandboxButton/index.tsx
```diff
@@ -73,14 +73,13 @@ import 'moment/min/locales'
...
-    const codeBlock = shouldRender
-      ? `render(${this.props.code})`
-      : this.props.code
+    const cleanedCode = this.props.code.replace(/\s+/g, '')
+    const renderEndMatcher = /render\(<.*>\)$/
+    const endsWithRenderStatement = renderEndMatcher.test(cleanedCode)
+    const codeBlock = endsWithRenderStatement
+      ? this.props.code
+      : `render(${this.props.code})`
+
```
ToMESSKa, line 82: The old code split the code into lines and checked if the last line included 'render(<'. However, sometimes the last line was just a ')' despite the code ending in a render function. This resulted in multiple render functions in CodeSandbox. Now, all newline characters and extra spaces are removed from the code, and a regex is used to check whether the code actually ends with 'render(<...>)', for example, 'render(<Example />)' or 'render(<div><Component /></div>)'.

packages/__docs__/src/CodeSandboxButton/index.tsx
```diff
@@ -73,14 +73,15 @@ import 'moment/min/locales'
...
-      : this.props.code
+    const cleanedCode = this.props.code.replace(/\s+/g, '')
+    const lastRenderIndex = cleanedCode.lastIndexOf('render(<')
+    const endsWithRenderStatement =
+      lastRenderIndex !== -1 && cleanedCode.endsWith('>)')
+
+    const codeBlock = endsWithRenderStatement
+      ? this.props.code
+      : `render(${this.props.code})`
+
```
ToMESSKa, line 84: The old code split the code into lines and checked if the last line included 'render(<'. However, sometimes the last line was just a ')' despite the code ending in a render function. This resulted in multiple render functions in CodeSandbox. Now, all newline characters and extra spaces are removed from the code, and it is checked whether the code actually ends with render(<...>) like 'render(<Example />)' or 'render(<View><Component /></View>)

packages/__docs__/src/CodeSandboxButton/index.tsx
```diff
@@ -73,14 +73,15 @@ import 'moment/min/locales'
...
 
-    const codeLines = this.props.code.split('\n')
-    const shouldRender = !codeLines[codeLines.length - 1]
-      .split(' ')
-      .join('')
-      .includes('render(<')
-    const codeBlock = shouldRender
-      ? `render(${this.props.code})`
-      : this.props.code
+    const cleanedCode = this.props.code.replace(/\s+/g, '')
```
matyasf, line 76: regexp are so hard to read, please add here a comment on what this one does

matyasf (changes requested):
see my comment

packages/__docs__/src/CodeSandboxButton/index.tsx
ToMESSKa: comment added

matyasf (approved):

balzss (approved):

PR #1670, opened 2024-09-05
docs(ui-text-area,ui-tree-browser): functional examples are added to TextArea, TreeBrowser

Kadirsaglm:


packages/ui-tree-browser/src/TreeBrowser/README.md
```diff
@@ -7,141 +7,428 @@ to receive a normalized data structure, examples can be seen at https://github.c
...
+  render(<Example />)
+  ```
 
-```
+- ```js
+  const Example = () => {
+    const [size, setSize] = useState('medium')
+    const sizes = ['small', 'medium', 'large']
+
+    const handleSizeSelect = (e, size) => {
```
ToMESSKa, line 183: I think this example is duplicated, there should be only two there, the class and the function based.

Kadirsaglm: Removed the duplicate

HerrTopi (approved):

ToMESSKa (approved):

PR #1671, opened 2024-09-05
fix(ui-radio-input): only radio button and label are clickable

joyenjoyer:
Closes: INSTUI-4197

**Test plan**

Go through all of the RadioInput/RadioInputGroup examples in the docs app:

- Check whether the design is not broken
- click events should be registered if you click on the radio group icon or label. If you click on the remaining area of the radio input, click should not be registered.


joyenjoyer:
This is wip, I am working on checkboxes and toggles.

ToMESSKa (approved):

git-nandor (approved):

PR #1674, opened 2024-09-12
fix(ui-simple-select): ensure input value updates correctly when options change

Kadirsaglm:
- Fixed an issue where the input value was not updated appropriately when the list of options was changed.
- Implemented behavior to clear the input if the previously selected option is no longer available in the updated options.
- Ensured the input value persists if the previously selected option remains in the updated options list.

### DIFF
--- a/packages/ui-simple-select/src/SimpleSelect/__new-tests__/SimpleSelect.test.tsx
+++ b/packages/ui-simple-select/src/SimpleSelect/__new-tests__/SimpleSelect.test.tsx
@@ -22,7 +22,7 @@
  * SOFTWARE.
  */
 import React from 'react'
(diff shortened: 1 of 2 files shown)


balzss:
thank you for the PR, nice work! two things:

- the ts build is failing, please take a look at that (you can locally run `npm run bootstrap` which runs the ts build too)
- we are in the process of migrating our test suit from karma to vitest. the old tests are in the `__tests__` folder and new ones are in the `__new_tests__` dir, please write your test in the new format

matyasf:
I like the solution and the code, thanks for the contribution!

matyasf (approved):

balzss (approved):
great work!

PR #1675, opened 2024-09-12
refactor(console): remove babel macro

matyasf:
It was only used to insert the `#__PURE__` annotation, but it was buggy and it got inserted twice, both in the wrong place (it must be before a function call, we put it before an export statement).
https://rollupjs.org/configuration-options/#pure
https://webpack.js.org/guides/tree-shaking/#mark-a-function-call-as-side-effect-free

TEST PLAN:
Compare the built binaries with the previous version. There should be no differences in other packages. You can do this online here: https://npmdiff.dev/%40instructure%2Fui-tabs/10.2.1/10.2.2-pr-snapshot-1726213252671 (the SNAPSHOT version is from a manual release in this branch  https://github.com/instructure/instructure-ui/actions/runs/10845048343 )


packages/console/src/index.ts
```diff
@@ -49,11 +39,11 @@ export {
...
-  logDebug,
-  logAssert,
-  consoleLog
+  /*#__PURE__*/ error as logError,
+  /*#__PURE__*/ warn as logWarn,
+  /*#__PURE__*/ warnDeprecated as logWarnDeprecated,
+  /*#__PURE__*/ info as logInfo,
+  /*#__PURE__*/ debug as logDebug,
+  /*#__PURE__*/ assert as logAssert,
+  /*#__PURE__*/ log as consoleLog
```
matyasf, line 48: as I could tell the only thing the plugin did is to insert these PURE comments (in a wrong way because it did it twice)

packages/ui-babel-preset/lib/index.js
```diff
@@ -174,7 +174,7 @@ function getWebEnvConfig(opts) {
     },
     useBuiltIns: 'entry',
     // this version has to match the version in package.json
-    corejs: '3.26.1',
```
matyasf, line 177: This is a separate small fix, doesnt seem to change anything

ToMESSKa (approved):

balzss (approved):
nice job getting to the bottom of this

PR #1695, opened 2024-09-23
fix(ui-progress,ui-range-input): use just the native elements instead of ARIA tags

matyasf:
Adding these ARIA tags is not recommended for these since the screen readers can voice them properly, no need for ARIA overrides

TEST PLAN:
Test `ProgressBar`, `ProgressCircle`, `RangeInput` with screenreaders. They should work as good or better than in the latest release


packages/ui-file-drop/src/FileDrop/index.tsx
```diff
@@ -349,6 +349,7 @@ class FileDrop extends Component<FileDropProps, FileDropState> {
         height={height}
         elementRef={this.handleElementRef}
       >
+        {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
         <label
```
matyasf, line 353: this component works fine in VO

packages/ui-progress/src/ProgressBar/index.tsx
```diff
@@ -126,10 +126,6 @@ class ProgressBar extends Component<ProgressBarProps> {
             css={styles?.htmlProgress}
             max={valueMax}
             value={valueNow}
-            role="progressbar"
-            aria-valuetext={valueText}
-            aria-valuenow={valueNow}
-            aria-valuemax={valueMax}
```
matyasf, line 132: These are not needed. I guess they were there because screenreaders years ago did not support `<progress>` natively? Now it works fine without these, e.g. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/progressbar_role#prefer_html

packages/ui-range-input/src/RangeInput/index.tsx
```diff
@@ -174,18 +175,14 @@ class RangeInput extends Component<RangeInputProps, RangeInputState> {
...
       }
-
+      const props = { [RangeInput.outputLocatorAttribute]: this.id }
       return (
         <ContextView background="inverse" placement="end center">
-          <output
-            htmlFor={this.id}
-            css={this.props.styles?.rangeInputInputValue}
-          >
+          <div {...props} css={this.props.styles?.rangeInputInputValue}>
```
matyasf, line 183: `output` was misused here, its meant to be used as a [result of a calculation not to display a value](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/output). It caused the value to be read twice in VO and NVDA

packages/ui-range-input/src/RangeInput/index.tsx
```diff
@@ -212,16 +209,12 @@ class RangeInput extends Component<RangeInputProps, RangeInputState> {
...
-            role="slider"
             id={this.id}
             min={this.props.min}
             max={this.props.max}
             step={this.props.step}
             value={this.value}
             onChange={this.noopChange}
-            aria-valuenow={this.value}
-            aria-valuemin={this.props.min}
-            aria-valuemax={this.props.max}
```
matyasf, line 224: Again, native HTML is recommended:  https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/slider_role : "It is recommended to use a native [input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) of type range, [input type="range"](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range), rather than the slider role."

ToMESSKa (commented):
- on Android with TalkBack, in ProgressBar and ProgressCircle the value is read 3 times : "40.0, 40, loading completion 40/60, progress indicator" vs previous version: "40/60, loading completion 40/60, progress indicator" (here only 2 times)
- the stepper buttons in NumberInput do not respond on Mac, iOS or Android, but this is true for the previous version
- I could not test on Windows

matyasf:
> * on Android with TalkBack, in ProgressBar and ProgressCircle the value is read 3 times : "40.0, 40, loading completion 40/60, progress indicator" vs previous version: "40/60, loading completion 40/60, progress indicator" (here only 2 times)

I've done a small change, now it reads "Loading completion 40/ 60 progress indicator" on Talkback/Chrome

HerrTopi (approved):

ToMESSKa:
> > * on Android with TalkBack, in ProgressBar and ProgressCircle the value is read 3 times : "40.0, 40, loading completion 40/60, progress indicator" vs previous version: "40/60, loading completion 40/60, progress indicator" (here only 2 times)
> 
> I've done a small change, now it reads "Loading completion 40/ 60 progress indicator" on Talkback/Chrome

With TalkBack, ProgressBar works fine, but ProgressCircle still reads the value 3 times: "40.0, 40 loading completion 40 out of 60, progress indicator"

ToMESSKa (approved):

PR #1698, opened 2024-09-25
fix(ui-radio-input): fixing the issue of TalkBack reading the radio g…

ToMESSKa:
…rop options incorrectly

Closes: INSTUI-4231

The issue: if RadioInputButton is wrapped in a FormFieldGroup, Android TalkBack cannot read the numbering of the elements correctly, e.g. it keeps reading every option as "1 out of 3". FormFieldGroup puts on extra HTML elements, which may be introducing ambiguity for TalkBack reading the correct sequence of items in a radio group. 

TEST PLAN:

- check the 4th example of the RadioInputGroup in the documentation. TalkBack should read the numbering of the elements correctly now eg: "1 out of 3", "2 out of 3", "3 out of 3"
- check if the examples on the RadioInputGroup page are read correctly by other screenreaders.


matyasf (approved):
nice!

HerrTopi (approved):

PR #1701, opened 2024-09-27
(WIP) Move from npm to pnpm

matyasf:
build, docs, tests seem to work.
Storybook does not.


matyasf:
I will close this as I could not get Storybook working. Webpack compiles endlessly because something is off with `require.context`, mayble related to https://github.com/storybookjs/storybook/issues/6408 
We will first need to move off from the `storiesOf` API in Storybook

PR #1702, opened 2024-09-27
chore: rewrite deploy action to be able to deploy multiple versions t…

joyenjoyer:
…o gh pages

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,19 +1,41 @@
-name: Deploy docs from master commit
+name: Deploy docs to gh-pages
 on:
   push:
     branches:
       - master
+      - v8_maintenance
+      - v9_maintenance
+  workflow_dispatch:
 jobs:
   deploy-docs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - run: npm ci && npm run bootstrap
+      - name: Set build directory and deployment path based on branch
+        id: set-build-dir
+        run: |
+          echo "::set-output name=build_dir::./packages/__docs__/__build__"
+          case "${{ github.ref }}" in
+            "refs/heads/master")
+              echo "::set-output name=deploy_dir::./"
+              ;;
+            "refs/heads/v8_maintenance")
+              echo "::set-output name=deploy_dir::v8"
+              ;;
+            "refs/heads/v9_maintenance")
+              echo "::set-output name=deploy_dir::v9"
+              ;;
+          esac
       - name: Build docs-app
         run: npm run build:docs
-      - uses: JamesIves/github-pages-deploy-action@v4
+      - name: Copy redirects config file to the __build__ directory.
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy.yml
```diff
@@ -1,19 +1,40 @@
...
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - run: npm ci && npm run bootstrap
+      # Conditional build depending on branch
+      - name: Set build directory based on branch
+        id: set-build-dir
+        run: |
+          if [ "${{ github.ref }}" == "refs/heads/master" ]; then
+            echo "::set-output name=build_dir::./packages/__docs__/__build__"
```
balzss, line 20: could this be set before the if/elif statement? it's the same for every condition

packages/__docs__/_redirects_gh_pages
balzss, line 1: is this file needed? isn't it for netflity redirects?

packages/__docs__/src/Header/index.tsx
```diff
@@ -69,9 +69,12 @@ class Header extends Component<HeaderProps> {
     // If we select the latest version from the dropdown,
     // then navigate to the index (instructure.design/#currentHash).
     // In every other case eg.: v6,v7 navigate to --> instructure.design/v6/#currentHash
+    const rootToAdd = window.location.origin.includes('github.io')
```
balzss, line 72: please add and explanatory comment for this. could be a TODO with something like "remove when we permanently migrate to github.io"

packages/__docs__/src/versionData.js
```diff
@@ -33,17 +33,14 @@
...
 
   if (!isLocalHost) {
-    // eslint-disable-next-line compat/compat
-    const result = await fetch(`${window.location.origin}/versions.json`, {
-      signal
-    })
-    const versionsData = await result.json()
-
-    return versionsData
+    let input = window.location.hostname.includes('github.io')
```
balzss, line 39: please add an explanation here too, maybe with a TODO remove later

packages/__docs__/src/versionData.js
```diff
@@ -53,7 +50,9 @@ const fetchVersionData = async (signal) => {
  * if we are on the docs page of a legacy version,
  * the path includes the version number, e.g. `/v7` or `/v8`
  */
-const [versionInPath] = window.location.pathname.split('/').filter(Boolean)
+const lastSegment =
+  window.location.pathname.split('/').filter(Boolean).pop() || ''
+const versionInPath = /v\d$/.test(lastSegment) ? lastSegment : undefined
```
balzss, line 55: these 2 lines are not really straight forward. what do they do?

balzss (changes requested):
please add some explanations to the code, otherwise nice work!

HerrTopi (approved):

matyasf (approved):

balzss (approved):

PR #1705, opened 2024-09-30
fix(shared-types,ui-table): fix color contrast for table sorting icons

HerrTopi:
Closes: INSTUI-4238


balzss (approved):
looks better, works better

balzss:
note: it'd be nice if chromatic could catch these type of changes...

matyasf (approved):

HerrTopi:
Chromatic would probably catch this if we had examples for this exact config, which we don't

HerrTopi:
Currently blocked by design. Final color decision is coming soon

PR #1706, opened 2024-09-30
fix(ui-pagination): fix pagination with numberinput when onBlur it calls  onPage change every time, even where there are no changes

HerrTopi:
TEST_PLAN: check if on an input type Pagination with a simple config if the onPageChange callback fires when the input field is blurred, only if the value has actually changed
Closes: INSTUI-4239


matyasf (approved):
looks good!

balzss (approved):
looks good but it's funny that state.number can be a float

PR #1707, opened 2024-09-30
Remove ui-stylelint-config, update dependecies

matyasf:
Remove `ui-stylelint-config` because:
- It was used to lint `.css` files, which is kinda useless by itself because we only have a few `.css` files and most of them are autogenerated.
- It was not invoked by our `lint` command because we did not call our `lint` command on the few packages that have `.css` files.
- It was not configured properly, just throwing an error when invoked on an actual `.css` file.

Also updated dependencies, mostly minor version updates


packages/ui-list/src/InlineList/__new-tests__/InlineList.test.tsx
```diff
@@ -56,10 +56,8 @@ describe('<InlineList />', () => {
         <InlineList.Item>List item 2</InlineList.Item>
         <InlineList.Item>List item 3</InlineList.Item>
         <InlineList.Item>List item 4</InlineList.Item>
-        {/* eslint-disable-next-line no-constant-binary-expression */}
-        {null && <InlineList.Item>ignore me 1</InlineList.Item>}
-        {/* eslint-disable-next-line no-constant-binary-expression */}
-        {false && <InlineList.Item>ignore me 2</InlineList.Item>}
+        {null}
+        {false}
```
matyasf, line 60: new TS version is clever and figures out that the part after `&&` is never executed

packages/ui-scripts/lib/icons/build-icons.js
```diff
@@ -52,10 +52,11 @@ export default {
 
     // optimize svgs in place
     runCommandSync('svgo', [
-      '-r',
-      '-f',
+      '--quiet',
```
matyasf, line 55: this will make `bootstrap` a bit less spammy

balzss (approved):
great work!

HerrTopi (approved):
Nice!

PR #1708, opened 2024-10-01
docs(ui-select,ui-selectable): function based examples are added to S…

ToMESSKa:
…elect, Selectable and Accessing the DOM

Closes: INSTUI-4221

### DIFF
--- a/docs/guides/accessing-the-dom.md
+++ b/docs/guides/accessing-the-dom.md
@@ -25,70 +25,120 @@ class MyComponent extends React.Component {
     return <div ref={this.ref}>Content</div>
   }
 }
+
+```
+
+```javascript
+---
+type: code
+---
+const MyComponent = React.forwardRef((props, ref) => {
+  return <div ref={ref}>Content</div>
+})
 ```
 
 Good Usage Example with `ref`:
 
-```js
----
-type: example
----
-
-class GoodComponent extends React.Component {
-  constructor(props) {
-    super(props)
-    this.ref = React.createRef()
+- ```js
+  class GoodComponent extends React.Component {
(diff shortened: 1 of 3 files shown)


docs/guides/accessing-the-dom.md
```diff
@@ -25,70 +25,120 @@ class MyComponent extends React.Component {
...
-        placement='end center'
-        offsetX='20px'
-      >
-        <span style={{ padding: '8px', background: 'white' }}>Positioned content</span>
-      </Position>
-    </View>
-  )
-}
+- ```js
+  const BadComponent = React.forwardRef((props, ref) => {
```
balzss, line 124: this example doesn't display the positioned component on the docs page

balzss (changes requested):

docs/guides/accessing-the-dom.md
ToMESSKa: The example is purposefully bad usage example because it's without ref. Unfortunately I couldn't find a way to write a "bad example" without ref that displays the positioned component. Any suggestion?

balzss: I see, in that case I'd might remove that example all together and present it in a different way. Maybe just the code itself or something like that.

ToMESSKa: I converted the bad examples to just a code.

ToMESSKa: I have converted the bad examples back to the Class/Function example format

balzss (approved):
thanks for trying my suggestion and sorry I made you revert it!

HerrTopi (approved):

PR #1709, opened 2024-10-01
chore(ui-calendar): add keyboard navigation and focus controll for cale…

git-nandor:
…ndar

### DIFF
--- a/packages/ui-calendar/src/Calendar/index.tsx
+++ b/packages/ui-calendar/src/Calendar/index.tsx
@@ -23,7 +23,14 @@
  */
 
 /** @jsx jsx */
-import React, { Children, Component, ReactElement, MouseEvent } from 'react'
+import React, {
+  Children,
+  Component,
+  ReactElement,
+  MouseEvent,
+  KeyboardEvent,
+  FocusEvent
+} from 'react'
 
 import { View } from '@instructure/ui-view'
 import {
@@ -82,6 +89,9 @@ class Calendar extends Component<CalendarProps, CalendarState> {
     role: 'table'
   }
 
+  // Create ref for each calendar day for programmatic focus management
+  dayRefs: (HTMLElement | null )[]
+
   ref: Element | null = null
   private _weekdayHeaderIds = (
     this.props.renderWeekdayLabels || this.defaultWeekdays
@@ -90,6 +100,7 @@ class Calendar extends Component<CalendarProps, CalendarState> {
   }, {})
   constructor(props: CalendarProps) {
     super(props)
+    this.dayRefs = []
 
     this.state = this.calculateState(
       this.locale(),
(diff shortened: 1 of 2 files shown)


packages/ui-calendar/src/Calendar/index.tsx
```diff
@@ -90,18 +98,91 @@ class Calendar extends Component<CalendarProps, CalendarState> {
   }, {})
   constructor(props: CalendarProps) {
     super(props)
-
-    this.state = this.calculateState(
-      this.locale(),
-      this.timezone(),
-      props.currentDate
+    const dayRefs = Array.from({ length: Calendar.DAY_COUNT }, () =>
```
balzss, line 101: some comment about why is this needed would be nice

packages/ui-calendar/src/Calendar/index.tsx
```diff
@@ -90,18 +98,91 @@ class Calendar extends Component<CalendarProps, CalendarState> {
...
+      createRef<HTMLElement>()
     )
+
+    this.state = {
+      ...this.calculateState(this.locale(), this.timezone(), props.currentDate),
+      dayRefs
+    }
+  }
+
+  setFirstDayTabIndex = (tabIndex: number) => {
```
balzss, line 111: typing could be more restrictive, maybe `tabIndex: 0 | -1`

packages/ui-calendar/src/Calendar/index.tsx
```diff
@@ -90,18 +98,91 @@ class Calendar extends Component<CalendarProps, CalendarState> {
...
+    const firstDay = this.state.dayRefs[0]?.current?.firstChild as HTMLElement
```
balzss, line 112: I would also add why the `firstChild` bit is needed (because you want the button children and not the table cell)

balzss: maybe also call the variable `firstDayButton`

packages/ui-calendar/src/Calendar/index.tsx
```diff
@@ -90,18 +98,91 @@ class Calendar extends Component<CalendarProps, CalendarState> {
...
+    firstDay.tabIndex = tabIndex
   }
 
   handleRef = (el: Element | null) => {
     this.ref = el
   }
 
+  handleOnBlur = (e: FocusEvent<Element>) => {
```
balzss, line 120: I would name this function something like `handleDaysTableBlur`

packages/ui-calendar/src/Calendar/index.tsx
```diff
@@ -90,18 +98,91 @@ class Calendar extends Component<CalendarProps, CalendarState> {
...
+    const dataCid = e.relatedTarget?.getAttribute('data-cid')
+
+    if (dataCid !== 'Day') {
```
balzss, line 123: why is this needed?

packages/ui-calendar/src/Calendar/index.tsx
```diff
@@ -322,7 +404,13 @@ class Calendar extends Component<CalendarProps, CalendarState> {
     return (
       <table role={this.role}>
         <thead>{this.renderWeekdayHeaders()}</thead>
-        <tbody>{this.renderDays()}</tbody>
+        {/* eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions */}
+        <tbody
+          onBlur={(e) => this.handleOnBlur(e)}
+          onKeyDown={(e) => this.handleKeyDown(e)}
```
balzss, line 410: this is not needed here

packages/ui-calendar/src/Calendar/index.tsx
```diff
@@ -90,18 +98,91 @@ class Calendar extends Component<CalendarProps, CalendarState> {
...
+
+    const moveFocus = (targetIndex: number) => {
+      const targetDay = this.state.dayRefs[targetIndex]?.current
+        ?.firstChild as HTMLElement
+
+      targetDay?.focus()
+      this.setFirstDayTabIndex(-1)
+    }
+
+    if (typeof dayIndex === 'number') {
```
balzss, line 140: this check is not needed, make the `dayIndex` param non-optional

packages/ui-calendar/src/Calendar/index.tsx
```diff
@@ -408,17 +496,33 @@ class Calendar extends Component<CalendarProps, CalendarState> {
...
+              <td
+                key={day.props.date}
+                role={role}
+                ref={dayRef}
+                onKeyDown={(e) => this.handleKeyDown(e, dayIndex)}
+              >
+                {role === 'presentation'
+                  ? safeCloneElement(day, {
+                      'aria-describedby': this._weekdayHeaderIds[i],
+                      tabIndex: dayIndex === 0 ? 0 : -1
```
balzss, line 518: check if the onkeydown can be added to the child prop instead of the `td` (if yes, you might need to refactor the code where you call `.firstChild`)

balzss: also maybe the dayref?

PR #1714, opened 2024-10-04
fix(ui-options): subgroup titles in Options are now announnced by Tal…

ToMESSKa:
…kBack and iOS VoiceOver

Closes: INSTUI-4235

ISSUE:
Subgroup titles are not read in Options with Android Talkback and iOS Voiceover. It works perfectly on Mac Voiceover and on Windows.

TEST PLAN:
- open the Options page and go to the second example (teh one with the subgroups)
- try reading the sub-group titles using Android TalkBack and iOS VoiceOver
- If possible, also check on Mac and Windows to see if it still works

### DIFF
--- a/packages/ui-options/src/Options/index.tsx
+++ b/packages/ui-options/src/Options/index.tsx
@@ -48,6 +48,7 @@ import type { OptionsSeparatorProps } from './Separator/props'
 
 import type { OptionsProps } from './props'
 import { allowedProps, propTypes } from './props'
+import { isAndroidOrIOS } from '@instructure/ui-utils'
 
 type ItemChild = React.ComponentElement<OptionsItemProps, Item>
 type SeparatorChild = React.ComponentElement<OptionsSeparatorProps, Separator>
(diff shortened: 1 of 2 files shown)


balzss (approved):

matyasf (approved):
Looks good, just a hint, be careful when using DOM globals like `navigator`, a lot of them are `undefined` in Node, resulting in errors when using SSR. I've tested this one with our regression testing app and it works OK because Node returns `Node.js/22`, but this code might break in some SSR edge cases like using Bun (havent tested it) or just this value is used in a mobile browser

packages/ui-options/src/Options/index.tsx
```diff
@@ -104,11 +104,12 @@ class Options extends Component<OptionsProps> {
 
   renderLabel() {
     const { renderLabel, styles } = this.props
+    const isAndroidOrIOS = /Android|iPhone|iPad/i.test(navigator.userAgent)
     return (
       <span
         id={this._labelId}
         role="presentation"
-        aria-hidden="true"
+        aria-hidden={isAndroidOrIOS ? 'false' : 'true'}
```
matyasf, line 112: Can you please leave a comment detailing why this is needed?

ToMESSKa: Rewrote the code with ua-parser-js and left a comment too

matyasf (approved):

PR #1729, opened 2024-10-11
required asterisk and new form error messages

balzss:
INSTUI-4321

test plan:

- please go through the code thoroughly to check the implementation details
- try to manually reproduce the new error state in multiple components based on the design docs
- check the TextInput readme for a showcase of all the different components and states
- check the [new form error guide](https://instructure.github.io/instructure-ui/pr-preview/pr-1729/#form-errors)

### DIFF
--- a/packages/ui-checkbox/src/Checkbox/CheckboxFacade/props.ts
+++ b/packages/ui-checkbox/src/Checkbox/CheckboxFacade/props.ts
@@ -40,6 +40,10 @@ type CheckboxFacadeOwnProps = {
    * Visual state showing that child checkboxes are a combination of checked and unchecked
    */
   indeterminate?: boolean
+  /**
+   * Indicate if the parent component (`Checkbox`) is invalid to set the style accordingly.
+   */
+  invalid?: boolean
 }
 
 type PropKeys = keyof CheckboxFacadeOwnProps
@@ -57,7 +61,8 @@ const propTypes: PropValidators<PropKeys> = {
   focused: PropTypes.bool,
   hovered: PropTypes.bool,
   size: PropTypes.oneOf(['small', 'medium', 'large']),
-  indeterminate: PropTypes.bool
+  indeterminate: PropTypes.bool,
+  invalid: PropTypes.bool
 }
 
 const allowedProps: AllowedPropKeys = [
--- a/packages/ui-checkbox/src/Checkbox/ToggleFacade/props.ts
+++ b/packages/ui-checkbox/src/Checkbox/ToggleFacade/props.ts
@@ -39,6 +39,10 @@ type ToggleFacadeOwnProps = {
   focused?: boolean
   size?: 'small' | 'medium' | 'large'
   labelPlacement?: 'top' | 'start' | 'end'
+  /**
+   * Indicate if the parent component (`Checkbox`) is invalid to set the style accordingly.
+   */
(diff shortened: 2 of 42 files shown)


packages/ui-checkbox/src/Checkbox/CheckboxFacade/props.ts
```diff
@@ -40,6 +40,7 @@ type CheckboxFacadeOwnProps = {
    * Visual state showing that child checkboxes are a combination of checked and unchecked
    */
   indeterminate?: boolean
+  error?: boolean
```
matyasf, line 43: Please add some description about this new prop

packages/ui-checkbox/src/Checkbox/ToggleFacade/props.ts
```diff
@@ -59,7 +60,8 @@ const propTypes: PropValidators<PropKeys> = {
   readOnly: PropTypes.bool,
   focused: PropTypes.bool,
   size: PropTypes.oneOf(['small', 'medium', 'large']),
-  labelPlacement: PropTypes.oneOf(['top', 'start', 'end'])
+  labelPlacement: PropTypes.oneOf(['top', 'start', 'end']),
+  error: PropTypes.bool
```
matyasf, line 64: here too

matyasf (changes requested):
Looks good as a WIP, just some things missing:
- description for some props
- I would make a page about forms in InstUI under Guides, where one can see all the components with some toggles that can set error, new error, required states.
- Please write unit tests too for these new props

ToMESSKa:
in FileDrop, the error icon isn't aligned with the text the same way as in the rest of the components (it seems to be aligned to the bottom end vs to the top in the rest)

packages/ui-file-drop/src/FileDrop/index.tsx
```diff
@@ -136,7 +136,7 @@ class FileDrop extends Component<FileDropProps, FileDropState> {
     if (this.hasMessages) {
       return (
         this.props.messages!.findIndex((message: FormMessage) => {
-          return message.type === 'error'
+          return message.type === 'error' || message.type === 'newError'
```
ToMESSKa, line 139: in FileDrop, the error icon isn't aligned with the text the same way as in the rest of the components (it seems to be aligned to the bottom end vs to the top in the rest)

ToMESSKa (changes requested):
In the second example on the CheckBoxGroup page, if the type is newError, the error message seems to be misplaced when layout='inline' (it appears in a separate column)

matyasf (approved):
Looks good, I really like the new 'form errors' page. Perhaps in the future we could expand it to a full documentation about forms in InstUI.

Also can you please write a VRT test for this (check out `TooltipPositioning.tsx`)

packages/ui-radio-input/src/RadioInputGroup/props.ts
```diff
@@ -87,20 +92,27 @@ type RadioInputGroupOwnProps = {
    * any children (ones that aren't `RadioInput` are passed through)
    */
   children?: React.ReactNode
+  /*
```
ToMESSKa, line 95: I think one more asterisk might be needed here e.g: /** because I cannot see this description on the preview page?

ToMESSKa (approved):

PR #1735, opened 2024-10-12
chore: convert __examples__ to esm

balzss:


balzss:
this is probably won't work until we remove the `storiesOf` API
https://github.com/storybookjs/storybook/issues/14877#issuecomment-1360943030

PR #1737, opened 2024-10-15
chore: add ESM build to icon fonts and refactor the icon docs page

balzss:
- add esm option to build icon fonts and keep the current for compatibility
- refactor docs to not depend on the `ui-icons` commonjs build and make it overall simpler

test plan:
- check the icon docs (for completeness and performance too)


packages/__docs__/src/index.js
```diff
@@ -30,9 +30,7 @@ import { InstUISettingsProvider } from '@instructure/emotion'
 import '../globals'
 
 createRoot(document.getElementById('app')).render(
-  <React.StrictMode>
```
matyasf, line 33: Why did you take StrictMode out?

packages/ui-scripts/lib/icons/generate-icon-fonts.js
```diff
@@ -0,0 +1,51 @@
...
+  normalize: true // Normalize SVG sizes
+}
+
+export default function generateIconFonts({
+  inputDir,
+  outputDir,
+  prefix,
+  name
+}) {
+  generateFonts({ ...config, inputDir, outputDir, prefix, name })
```
matyasf, line 43: This is pretty cool, we can deprecate lots of code :)
Please mark the code parts in `ui-icons-build` as deprecated (it would be the best if we could delete this package in the future IMO)

matyasf (changes requested):
Looks great, much much simpler code than before. Please add a comment to our icons page about the deprecations and the `ui-icons-build` code

packages/__docs__/src/index.js
balzss: it was only for mistake and forgot to put it back

matyasf (approved):

packages/__docs__/src/Icons/index.tsx
```diff
@@ -23,304 +23,342 @@
...
+        <Alert variant="warning" margin="small 0">
+          Icon Font is a deprecated format and only here for compatibility
+          reasons. It doesn&apos;t have right-to-left support and some icons
+          have visual artifacts due to svg-to-ttf conversion. We recommend using
+          the React format.
+        </Alert>
+      )}
+      {selectedFormat === 'svg' && (
+        <Alert variant="info" margin="small 0">
+          The SVG format doesn&apos;t have right-to-left support. I you need
```
joyenjoyer, line 291: Please fix this type: **If** you need that

packages/ui-icons/gulpfile.js
```diff
@@ -22,6 +22,7 @@
  * SOFTWARE.
  */
 
+// TODO deprecated: remove in v11
```
joyenjoyer, line 25: Make a task in JIRA about removal so we won't forget it.

joyenjoyer (changes requested):
Nice work.

joyenjoyer (approved):

PR #1746, opened 2024-10-21
fix(ui-table): fix table header

HerrTopi:


matyasf (approved):
lets not forget to add this to v10 too. Also can you write a test/make a ticket for this? Thanks!

balzss (approved):

PR #1747, opened 2024-10-24
INSTUI-4241: fix(ui-toggle-details): do not put aria-expanded and aria-controls on  the toggle if there is nothing to toggle

matyasf:
Also do not render an empty div for the details in this case. Note that this might break tests since
they might find these DOM parts via these ARIA tags. Also the empty div took up a small space when expanded (0.375rem), this might cause layout changes

TEST PLAN:
Make a ToggleDetails without any children, it should have no aria tags. Should still listen to mouse events

### DIFF
--- a/packages/ui-expandable/src/Expandable/index.tsx
+++ b/packages/ui-expandable/src/Expandable/index.tsx
@@ -43,9 +43,7 @@ class Expandable extends Component<ExpandableProps, ExpandableState> {
   static propTypes = propTypes
   static allowedProps = allowedProps
   static defaultProps = {
-    defaultExpanded: false,
-    onToggle: function () {},
-    children: null
+    defaultExpanded: false
   }
 
   _contentId: string
@@ -92,7 +90,7 @@ class Expandable extends Component<ExpandableProps, ExpandableState> {
     if (!this.isControlled()) {
       this.setState(toggleExpanded)
     }
-    this.props.onToggle(event, !this.expanded)
+    this.props.onToggle?.(event, !this.expanded)
   }
 
   render() {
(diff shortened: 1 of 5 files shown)


packages/ui-expandable/src/Expandable/index.tsx
```diff
@@ -43,9 +43,7 @@ class Expandable extends Component<ExpandableProps, ExpandableState> {
   static propTypes = propTypes
   static allowedProps = allowedProps
   static defaultProps = {
-    defaultExpanded: false,
-    onToggle: function () {},
-    children: null
```
matyasf, line 48: this is just removing some old old code, ancient InstUI loved default values even when it was not needed.

packages/ui-expandable/src/Expandable/index.tsx
```diff
@@ -101,15 +99,12 @@ class Expandable extends Component<ExpandableProps, ExpandableState> {
     if (typeof render === 'function') {
       return render({
         expanded: this.expanded,
-        getToggleProps: (props = {} as any) => {
+        getToggleProps: (props = {}) => {
           return {
+            ...props,
```
matyasf, line 104: this is better, users cannot override our props this way

packages/ui-toggle-details/src/ToggleDetails/index.tsx
```diff
@@ -101,13 +101,16 @@ class ToggleDetails extends Component<ToggleDetailsProps> {
...
+      ? toggleProps
+      : { onClick: toggleProps.onClick }
     const props = {
       ...omitProps(this.props, ToggleDetails.allowedProps),
-      ...toggleProps,
+      ...tProps,
       children: this.renderSummary(expanded)
-      // spread operator makes toggleProps loose Record<string, any>>
-    } as Record<string, any>
+    } as Record<string, unknown>
```
matyasf, line 113: The actual fix

packages/ui-toggle-details/src/ToggleDetails/index.tsx
```diff
@@ -152,11 +155,11 @@ class ToggleDetails extends Component<ToggleDetailsProps> {
   }
 
   renderDetails(expanded: boolean, detailsProps: { id: string }) {
-    const { children } = this.props
+    if (!this.props.children) return null
     const expandedStyles = expanded ? { display: 'block' } : { display: 'none' }
     return (
       <div {...detailsProps} css={[this.props.styles?.details, expandedStyles]}>
```
matyasf, line 161: This `div` is rendering a line with 0.375rem when open even when it has no children. I've removed this, hopefully nothing will break

git-nandor (approved):

PR #1748, opened 2024-10-24
fix(ui-buttons): add back ic-brand theme tokens to basebutton

HerrTopi:
Test plan: check if button's ic-brand variables can be overridden and the changes apply or not

Closes: INSTUI-4338


matyasf (approved):

matyasf:
it would be good to make an example about these brand variables in our "Using theme overrides" guide, since this is a recommended way to do theme overrides

joyenjoyer (approved):

PR #1749, opened 2024-10-25
fix: docs screenreader alerts are no longer screenreader focusable

matyasf:
The issue is that our `ScreenReaderContent` class does not hide its text from Screenreader focus, and this is used by `Alert` to display it in an ARIA live region.
I've tried to fix the `ScreenReaderContent` class, but could not do it, so the solution way to add a timeout to the screenreader alerts, so users will not have time to go back with the focus

Also some better types in the docs app and the debounce utility

TEST PLAN:
Enter some text in the docs in the top search field with screenreader on. After this focus out, and find the beginning of the page. It should not read the search results again

### DIFF
--- a/packages/__docs__/buildScripts/DataTypes.mts
+++ b/packages/__docs__/buildScripts/DataTypes.mts
@@ -213,17 +213,16 @@ export type ParsedDoc = {
   descriptions: Record<string, string>
   // Hold minimal information about each document that is needed for search
   // functionality
-  docs: Record<
-    string,
-    {
-      title: string
-      order?: string
-      category?: string
-      isWIP?: boolean
-      tags?: string
-    }
-  >
+  docs: ParsedDocSummary
 }
+
+export type ParsedDocSummary = Record<string,{
+  title: string
+  order?: string
+  category?: string
+  isWIP?: boolean
+  tags?: string
+}>
 
 type IconGlyph = {
   name: string
--- a/packages/__docs__/src/Search/index.tsx
+++ b/packages/__docs__/src/Search/index.tsx
@@ -33,6 +33,7 @@ import { Select } from '@instructure/ui-select'
 import { SearchStatus } from '../SearchStatus'
 
 import type { SearchProps, SearchState, OptionType } from './props'
+import { debounce } from '@instructure/debounce'
 
 class Search extends Component<SearchProps, SearchState> {
   static defaultProps = {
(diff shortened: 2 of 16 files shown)


packages/__docs__/buildScripts/DataTypes.mts
```diff
@@ -213,18 +213,17 @@ export type ParsedDoc = {
...
-    {
-      title: string
-      order?: string
-      category?: string
-      isWIP?: boolean
-      tags?: string
-    }
-  >
+  docs: ParsedDocSummary
 }
```
matyasf, line 217: the types in the docs app were pretty bad, I've corrected them

packages/__docs__/src/Search/index.tsx
```diff
@@ -70,18 +73,22 @@ class Search extends Component<SearchProps, SearchState> {
       })
     })
   }
+  componentDidUpdate(_prevProps: SearchProps, prevState: SearchState) {
+    if (this.state.announcement != prevState.announcement) {
+      this._debounced({ announcement: null })
+    }
```
matyasf, line 79: This is the actual fix. Now the SR alert disappears after 1 second. this is enough to read it, but when tabbing out from the search field it will no longer be in the beginning of the DOM

packages/debounce/src/debounce.ts
```diff
@@ -28,10 +28,9 @@ interface DebounceOptions {
   trailing?: boolean
 }
 
-export type Debounced = {
-  (...args: unknown[]): unknown
+export type Debounced<F extends (...args: any) => any> = F & {
   cancel: () => void
-  flush: () => void
```
matyasf, line 34: Also I've fixed the type here. Now when you use this function it will correctly have the same parameter types and return types as the  function debounced with the help of some TS generics trickery.

packages/ui-a11y-content/src/ScreenReaderContent/styles.ts
```diff
@@ -29,25 +29,23 @@ import type { ScreenReaderContentStyle } from './props'
...
+      width: '0.0625rem !important',
+      height: '0.0625rem !important',
+      margin: '-0.0625rem !important',
+      padding: '0 !important',
       position: 'absolute',
       top: 0,
       insetInlineStart: 0,
-      overflow: 'hidden',
-      clip: 'rect(0 0 0 0)',
-      border: 0
```
matyasf, line 50: Some minor fixes to make even more sure that the SR only alerts dont affect the visible DOM

HerrTopi (approved):

joyenjoyer (approved):
nice work

PR #1759, opened 2024-11-06
INSTUI-4266 : The open/close sidebar button does not cause focus loss

matyasf:
Now the focus moved to the 'close' button of the opened sidebar

TEST PLAN:
open and close the menu on the docs app with keyboard. Focus should not get lost. Also test with
screenreaders

Also better typing for some docs TS types.

### DIFF
--- a/packages/__docs__/buildScripts/DataTypes.mts
+++ b/packages/__docs__/buildScripts/DataTypes.mts
@@ -224,23 +224,16 @@ export type ParsedDocSummary = Record<string,{
   tags?: string
 }>
 
-type IconGlyph = {
+type Glyph = {
+  bidirectional: boolean
+  lineSrc: string
+  solidSrc: string
   name: string
-  variant: any
   glyphName: string
-  deprecated: boolean
-}
-
-type IconFormat = {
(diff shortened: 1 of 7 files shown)


packages/__docs__/buildScripts/DataTypes.mts
```diff
@@ -224,23 +224,16 @@ export type ParsedDocSummary = Record<string,{
   tags?: string
 }>
 
-type IconGlyph = {
+type Glyph = {
+  bidirectional: boolean
+  lineSrc: string
+  solidSrc: string
```
matyasf, line 230: this was not updated with the icons page rewrite

packages/__docs__/src/App/index.tsx
```diff
@@ -641,6 +613,11 @@ class App extends Component<AppProps, AppState> {
               shape="circle"
               color="secondary"
               size="medium"
+              ref={(button) => {
+                if (button) {
+                  button.focus()
+                }
+              }}
```
matyasf, line 620: This is the fix

balzss (approved):

HerrTopi (approved):

PR #1762, opened 2024-11-06
INSTUI-4322 Make it possible to not use global theme registry. Deprecate global theme registry.

matyasf:
The main goal of this PR is to make it possible to run InstUI v8/v9 in the same page as InstUI v10 when the host app uses the older version and a later initialized one v10.
Currently this causes the v10 app to look buggy because `ThemeRegistry` will return an incompatible v8 theme object:

in `canvas.ts`:
```
const theme = ThemeRegistry.registerTheme(__theme)
export default theme
```
in `ThemeRegistry.registerTheme`:
```
if (theme.key && registry.themes[theme.key]) {
    return registry.themes[theme.key] as RegisteredTheme<T>
  }
```

This PR makes it possible to get the theme directly circumventing `ThemeRegistry` and its theme cache. Also deprecates `ThemeRegistry` (which is used for global themes)

### DIFF
--- a/packages/ui-themes/src/themes/canvas/index.ts
+++ b/packages/ui-themes/src/themes/canvas/index.ts
@@ -59,6 +59,11 @@ export type CanvasTheme = BaseTheme & {
   key: 'canvas'
 } & typeof sharedThemeTokens & { colors: Colors } & CanvasBrandVariables
 
+/**
+ * Canvas theme without the `use` function and `variables` prop.
+ * Not affected by global theme overrides (`.use()` function).
+ * Will be default in the next major version of InstUI
+ */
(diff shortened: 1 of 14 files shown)


packages/ui-themes/src/themes/canvas/index.ts
```diff
@@ -69,3 +74,4 @@ const __theme: CanvasTheme = {
 const theme = ThemeRegistry.registerTheme(__theme)
 
 export default theme
```
matyasf, line 76: This is the main change and what we will recommend people to use

balzss (approved):

HerrTopi (approved):

PR #1763, opened 2024-11-07
feat(ui-view): borderColor prop added to ContextView, borderColor now…

ToMESSKa:
… can be a HEX code in View

Closes: INSTUI-4346

TEST PLAN:

- check if ContextView accepts the borderColor prop e.g.: borderColor="alert",  borderColor="#FF0000" (including the arrow)
- check if View's borderColor prop accepts HEX code e.g.: borderColor="#FF0000" (including the arrow)
- check the description of the borderColor in ContextView and View documentation

### DIFF
--- a/packages/ui-view/src/ContextView/index.tsx
+++ b/packages/ui-view/src/ContextView/index.tsx
@@ -97,7 +97,8 @@ class ContextView extends Component<ContextViewProps> {
       stacking,
       style,
       textAlign,
-      styles
+      styles,
+      borderColor
     } = this.props
 
     return (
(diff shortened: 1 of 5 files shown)


HerrTopi (approved):
Nice work

balzss (changes requested):
when using the old tokens for borderColor (`info`, `danger`, etc.) the small arrow always have black borders

balzss:
note: I've added a commit to remove the option from ContextView to support old color tokens (`info`, `danger`, etc.). You can still set them which would cause a mismatched border color but it is considered an invalid usecase.

balzss (approved):

HerrTopi (approved):

PR #1767, opened 2024-11-11
v9: backport new error messages from v10

balzss:
INSTUI-4355

also backported a fix for the gh pages deploy action

test plan:

- make sure everything has been properly backported from v10


git-nandor (approved):

matyasf (changes requested):
image

CheckboxGroup and RadioInputGroup has messages in the wrong place, on master they are above the controls, in the PR below. Which one is right?

balzss:
> image
> CheckboxGroup and RadioInputGroup has messages in the wrong place, on master they are above the controls, in the PR below. Which one is right?

good catch! the one in master is correct. fixed it

matyasf (approved):

PR #1771, opened 2024-11-13
fix(ui-color-picker): make ColorPicker tooltip VoiceOver focusable

balzss:
Closes: INSTUI-4259

test plan:

- check colorpicker example with tooltip if voiceover reads the tooltip context correctly

### DIFF
--- a/packages/ui-color-picker/src/ColorPicker/index.tsx
+++ b/packages/ui-color-picker/src/ColorPicker/index.tsx
@@ -26,7 +26,7 @@
 /** @jsxFrag React.Fragment */
 import React, { Component } from 'react'
 
-import { withStyle, jsx } from '@instructure/emotion'
+import { withStyle, jsx, InstUISettingsProvider } from '@instructure/emotion'
 import { warn, error } from '@instructure/console'
 import { omitProps } from '@instructure/ui-react-utils'
 import { testable } from '@instructure/ui-testable'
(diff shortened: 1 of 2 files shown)


ToMESSKa (approved):

matyasf (changes requested):
I dont like that the SR is focusing on nothing and the tooltip does not show with this solution.

I would solve the underlying issue, that icons set `aria-hidden` to `true` (in `InlineSVG`) by default. You just need a small change there to allow this attribute:

image

balzss:
@matyasf I've went with a different approach. According to [w3c](https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/) and [mdn](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role) tooltips should be focusable with keyboard.

matyasf (changes requested):
I dont think this is a good solution. NVDA does not read the tooltip. Also its wonky on VO too, when I move the SR focus it does not read it, but after I've tabbed trough it, it will read it
I think its because the `svg` icon is set to `aria-hidden` which is a no for focusable elements https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden#description , so NVDA just removes is from the SR DOM

balzss:
@matyasf can you please check it again? now it also reads the tooltip when the input is focused but I don't think we can prevent that. I think it always reads the label when an input gets focused...

also fyi: it has a ts error that's probably introduced from [this pr](https://github.com/instructure/instructure-ui/pull/1771/files#diff-aacd5a00cfdc942924434e62ae19741c1a8d335477045d8cb9cdc143745454dfR397) so a [temporary ts-ignore](https://github.com/instructure/instructure-ui/pull/1771/files#diff-aacd5a00cfdc942924434e62ae19741c1a8d335477045d8cb9cdc143745454dfR397) was added

packages/ui-svg-images/src/InlineSVG/index.tsx
```diff
@@ -150,7 +150,6 @@ class InlineSVG extends Component<InlineSVGProps> {
       title,
       description,
       focusable,
-      children,
```
matyasf, line 153: why did you remove this?

matyasf (approved):
nice work!

packages/ui-color-picker/src/ColorPicker/index.tsx
```diff
@@ -389,9 +389,22 @@ class ColorPicker extends Component<ColorPickerProps, ColorPickerState> {
       <span>
         <span css={styles?.label}>{label}</span>
         <span>
-          <Tooltip renderTip={tooltip}>
-            <IconInfoLine tabIndex={-1} />
-          </Tooltip>
+          <InstUISettingsProvider
+            theme={{
+              themeOverrides: {
+                componentOverrides: {
```
matyasf, line 395: I did not knew that this is a valid syntax... easier one would be

```
  theme={{
    componentOverrides: {
      BaseButton: {
        smallHeight: 'auto'
      }
    }
  }}
```

PR #1780, opened 2024-11-19
fix(emotion,shared-types): better TS types for theme objects and their overrides

matyasf:
Also convert some of the final files in the docs app to TS and improve their typing

TEST PLAN:
Try to make various theme and override objects, they should have nice autocomplete and no errors


packages/emotion/src/EmotionTypes.ts
```diff
@@ -87,3 +141,62 @@ export type {
   GenerateStyle,
   ComponentStyle
 }
+/*
```
matyasf, line 144: I've left this here intentionally to test this complex type

packages/emotion/src/EmotionTypes.ts
```diff
@@ -29,21 +29,75 @@ import type {
   DeepPartial
```
matyasf, line 29: The type changes in this file and the main change

packages/__docs__/components.ts
```diff
@@ -130,4 +130,4 @@ export { Drilldown } from '@instructure/ui-drilldown'
 export { SourceCodeEditor } from '@instructure/ui-source-code-editor'
 export { TopNavBar } from '@instructure/ui-top-nav-bar'
 export { TruncateList } from '@instructure/ui-truncate-list'
-export { canvas, canvasHighContrast, instructure } from '@instructure/ui-themes'
```
matyasf, line 133: This was left here from v9... we really need to type check the docs app

packages/__docs__/src/compileMarkdown.tsx
```diff
@@ -39,7 +39,32 @@ import { Playground } from './Playground'
...
+function trimIndent(str: string) {
+  const lines = `${str.replace(/\r\n/g, '\n').replace(/\r/g, '\n')}\n`.split(
+    '\n'
+  )
+  let indentFound = false
+  let trimmed = ''
+  let indent = ''
+  lines.forEach((line, _i) => {
+    // matches whitespace at the end of a string
+    line.replace(/\s*$/, '')
```
ToMESSKa, line 52: I think .replace() does not change the original string, so the whitespace isn't actually removed. The result should be reassigned e.g.: line = line.replace(/\s*$/, '')

ToMESSKa (changes requested):

packages/__docs__/src/compileMarkdown.tsx
matyasf: good catch, fixed

HerrTopi (approved):

ToMESSKa (approved):

PR #1785, opened 2024-11-21
docs(ui-dialog): add a11y clarifications to Dialog docs

ToMESSKa:
Closes: INSTUI-4293, INSTUI-4291

ISSUES:
- the form labels in the dialog example are not descriptive enough
- the Dialog example does not have a role or appropriate description

TEST PLAN:
- when opening the Dialog example, the screen reader should announce that it is a dialog and read the dialog's label (e.g., "Full name form").
- input labels should be descriptive in the example, such as "First name" instead of just "First", when using a screenreader
- review the added recommendations for dialog accessibility


matyasf (approved):
nice work :)

balzss (approved):

PR #1788, opened 2024-11-22
fix(ui-breadcrumb): add and update aria tags in Breadcrumb

ToMESSKa:
Closes: INSTUI-4268

### DIFF
--- a/packages/ui-breadcrumb/src/Breadcrumb/index.tsx
+++ b/packages/ui-breadcrumb/src/Breadcrumb/index.tsx
@@ -62,6 +62,16 @@ class Breadcrumb extends Component<BreadcrumbProps> {
     this.ref = el
   }
 
+  addAriaCurrent = (child: React.ReactNode) => {
+    const updatedChild = React.cloneElement(
+      child as React.ReactElement<{ 'aria-current'?: string }>,
+      {
+        'aria-current': 'true'
+      }
+    )
+    return updatedChild
+  }
+
   componentDidMount() {
     this.props.makeStyles?.()
   }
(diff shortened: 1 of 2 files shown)


packages/ui-breadcrumb/src/Breadcrumb/index.tsx
```diff
@@ -78,9 +88,13 @@ class Breadcrumb extends Component<BreadcrumbProps> {
...
     }
     return React.Children.map(children, (child, index) => {
+      const isLastElement = index === numChildren - 1
+      const hasHref =
+        React.isValidElement(child) && (child.props as { href?: string }).href
+
       return (
         <li css={styles?.crumb} style={inlineStyle}>
-          {child}
+          {isLastElement && hasHref ? this.addAriaCurrent(child) : child}
```
matyasf, line 97: As I understand we can always put this ARIA tag here.
From https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current "If the element representing the current page in the breadcrumb was not a link, the aria-current is optional.", lets put it always there. same in mdn's example: https://developer.mozilla.org/en-US/docs/Web/CSS/Layout_cookbook/Breadcrumb_Navigation

matyasf (changes requested):
See my comment. Also please write in the Readme in the DO section that the current location should always be the last element in the breadcrumb, this is not stated in the page explicitly

packages/ui-breadcrumb/src/Breadcrumb/index.tsx
ToMESSKa: done

matyasf (changes requested):

packages/ui-breadcrumb/src/Breadcrumb/BreadcrumbLink/props.ts
```diff
@@ -63,6 +63,11 @@ type BreadcrumbLinkOwnProps = {
    * Place the icon before or after the text in the Breadcrumb.Link
    */
   iconPlacement?: 'start' | 'end'
+  /**
+   * Whether the page this breadcrumb points to is the current one. If true, it sets aria-current="page".
+   * If this prop is not set to true on any breadcrumb, the element recieving the aria-current="page" will always be the last element on the right by default, unless the last last element's isCurrentPage prop is explicity set to false too.
+   */
+  isCurrentPage?: boolean
```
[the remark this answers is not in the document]
ToMESSKa: Added the a11y box and removed the "on the right by default" part

joyenjoyer (approved):

PR #1790, opened 2024-11-24
fix(ui-form-field): Remove hardcoded colors and use theme provided ones

szalonna:
The form field message component used hardcoded colors for the different message types, independently from the theme currently used. Now it will use the theme-provided colors.


matyasf:
Thanks for the PR, we will look at it

matyasf (approved):
looks good, thanks for the contribution!

szalonna:
After our meeting, I consider this an obsolete

PR #1794, opened 2024-11-26
fix(ui-text-area): make focus line follow resized textarea

joyenjoyer:
Closes: INSTUI-4380

Test: add the `resize` property to `TextArea` component, resize it and observe whether the focusline follows the resizing properly.

### DIFF
--- a/packages/ui-text-area/src/TextArea/index.tsx
+++ b/packages/ui-text-area/src/TextArea/index.tsx
@@ -117,8 +117,8 @@ class TextArea extends Component<TextAreaProps> {
         if (this._highlightRef) {
           const entryStyle = window.getComputedStyle(entry.target)
           this._highlightRef.style.transition = 'none'
-          this._highlightRef.style.width = `calc(${entryStyle.width}px + 2.125rem)`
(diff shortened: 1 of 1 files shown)


packages/ui-text-area/src/TextArea/index.tsx
```diff
@@ -117,8 +117,8 @@ class TextArea extends Component<TextAreaProps> {
         if (this._highlightRef) {
           const entryStyle = window.getComputedStyle(entry.target)
           this._highlightRef.style.transition = 'none'
-          this._highlightRef.style.width = `calc(${entryStyle.width}px + 2.125rem)`
-          this._highlightRef.style.height = `calc(${entryStyle.height}px + 2.125rem)`
+          this._highlightRef.style.width = `calc(${entryStyle.width} + 0.5rem)`
```
joyenjoyer, line 120: entryStyle.width is a string, which already has 'px' in it. I removed the unneccessary px string from the `calc()` function and added the proper rem amount.

matyasf (approved):
nice work!
An extra unit test would be nice, but it might be too tricky

balzss:
nice work!

balzss (approved):

PR #1800, opened 2024-11-28
docs: add more documentation on how to do theme overrides

matyasf:
Content mostly recovered from the deleted parts in the v9 version, but I also tried to make the examples more clear


balzss:
looks good, I'd add an extra section about overriding "nested" components. e.g. if you want to override the `borderRadius` for `DateInput2` you cannot do it directly and you have to do it like this:

```
<InstUISettingsProvider
  theme={{
    componentOverrides: {
      TextInput: {
        borderRadius: '50%'
      }
    }
  }}
    >
    <DateInput2 />
</InstUISettingsProvider>
```

balzss (approved):

HerrTopi (approved):

PR #1805, opened 2024-12-02
feat(ui-progress): add isRounded to ProgressBar

adamiku:


balzss:
Hi @adamiku , thanks for the PR. Please do not add `isRounded` as a prop but add a new theme variable for `borderRadius` in progressbar instead so you can override it with your value (`20px`) in your codebase. Also we do not use hardcoded variables in `styles.ts` files.

adamiku:
> Hi @adamiku , thanks for the PR. Please do not add `isRounded` as a prop but add a new theme variable for `borderRadius` in progressbar instead so you can override it with your value (`20px`) in your codebase. Also we do not use hardcoded variables in `styles.ts` files.

I close this PR as the InstUi team will implement it.

PR #1808, opened 2024-12-03
feat(shared-types,ui-progress): add customization options for progressbar

balzss:


packages/ui-progress/src/ProgressBar/index.tsx
```diff
@@ -130,11 +131,11 @@ class ProgressBar extends Component<ProgressBarProps> {
           />
 
           <span css={styles?.track} role="presentation" aria-hidden="true">
-            <span css={styles?.trackValue}></span>
+            <span css={styles?.trackValue}>{renderValueInside && value}</span>
           </span>
         </span>
 
-        {value && (
+        {value && !renderValueInside && (
```
matyasf, line 138: Can you try to write a test for this? e.g if renderValueInside is `true` then the value text should be inside the track's DOM

matyasf (changes requested):
I like the API, tested it with SR too, it works, an extra unit test would be nice

packages/ui-progress/src/ProgressBar/index.tsx
balzss: done

matyasf (approved):

adamiku:
Looks good to me! Thanks!

adamiku (approved):

HerrTopi (approved):

PR #1809, opened 2024-12-03
fix(ui-text-input): fix long before elements overflowing in TextInput, Select,SimpleSelect

matyasf:
Closes: INSTUI-4344

When adding lots of elements like Tags it was overflowing the input field, this commit fixes it

This bug was introduced with https://github.com/instructure/instructure-ui/pull/1696 when we added a Flex layout without wrap.

TEST PLAN:
Add lots of elements to renderBeforeInput to Select,SimpleSelect,TextInput. They should wrap not overflow

### DIFF
--- a/packages/ui-text-input/src/TextInput/index.tsx
+++ b/packages/ui-text-input/src/TextInput/index.tsx
@@ -77,7 +77,6 @@ class TextInput extends Component<TextInputProps, TextInputState> {
     super(props)
     this.state = {
       hasFocus: false,
-      beforeElementHasWidth: undefined,
       afterElementHasWidth: undefined
     }
     this._defaultId = props.deterministicId!()
@@ -87,7 +86,6 @@ class TextInput extends Component<TextInputProps, TextInputState> {
   ref: Element | null = null
 
   private _input: HTMLInputElement | null = null
-  private _beforeElement: HTMLSpanElement | null = null
   private _afterElement: HTMLSpanElement | null = null
 
   private readonly _defaultId: string
@@ -112,13 +110,10 @@ class TextInput extends Component<TextInputProps, TextInputState> {
         'focus',
         this.handleFocus
       )
-
       this.setState({
-        beforeElementHasWidth: this.getElementHasWidth(this._beforeElement),
         afterElementHasWidth: this.getElementHasWidth(this._afterElement)
       })
     }
-
     this.props.makeStyles?.(this.makeStyleProps())
   }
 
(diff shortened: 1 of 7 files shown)


ToMESSKa (approved):

balzss:
I've checked it and seems like it was also buggy in v9, so might not be a regression in the commit you've linked. Also the spacing is a bit odd:
image

I would suggest using the `gap` css prop instead of adding margins to every pill/tag.

Another thing: shouldn't the input tag wrap with the pill too, instead of being on a new line:
image

balzss (changes requested):
see my comment above

packages/ui-text-input/src/TextInput/styles.ts
```diff
@@ -198,30 +186,21 @@ const generateStyle = (
...
-      ...flexItemBase,
-      minWidth: '0.0625rem',
-      flexShrink: 1,
-      flexGrow: 1
+      justifyContent: 'flex-start',
+      flexDirection: 'row',
+      // Apply the style if the first child has NOT the inputLayout class
+      '&:has(> :first-child:not([class$="textInput__inputLayout"]))': {
+        paddingInlineStart: componentTheme.padding
+      },
```
matyasf, line 198: This method is slightly worse than the previous one, but does not need JS (now it applies padding if someone puts in a 0 width `beforeElement`)

packages/ui-text-input/src/TextInput/index.tsx
```diff
@@ -360,39 +349,25 @@ class TextInput extends Component<TextInputProps, TextInputState> {
       >
         <span css={styles?.facade}>
           {renderBeforeOrAfter ? (
-            <div>
-              <span css={styles?.layout}>
-                {beforeElement && (
+            <span css={styles?.layout}>
+              {beforeElement && <div css={styles?.beforePadding}></div>}
```
matyasf, line 353: I dont like this solution, that we have an extra DOM element just to add a bit of padding. But I could not find a better solution, what I've tried:
1. Use a tricky CSS, that adds padding if the first child of the `layout` `span` is NOT the `inputLayout`: `'&:has(> :first-child:not([class$="textInput__inputLayout"]))'`. While this works nice in the browser it crashes in Vite tests because of a bug in a dependency: https://github.com/dperini/nwsapi/issues/123
So using this solution means that users will need to use either the latest Jest or a patched version of Jest where `nwsapi` is at the latest version, which is not cool IMO.

2. Do some JS magic to figure out the width of the `renderBeforeInput` elements: This is what the old code was doing, now it would need to be even more complicated because we're not wrapping `renderBeforeInput` in a `div` anymore, so we need to take care of the case that its a React fragment/text/whatever weirdness users put there. 

While the current solution is a bit ugly because of the extra DOM element and the padding is added if the `renderBeforeInput`'s width is 0, its still the best I could find.

joyenjoyer: As we discussed adding padding to the container would be an easier solution, please fix.

joyenjoyer (approved):

balzss (approved):
looks good. later we could simplify this by removing the margins and paddings from the Tags and using flex gap but that would require rewriting our examples and breaking existing code that users need to update

PR #1812, opened 2024-12-06
WIP: new spacing

balzss:


balzss:
@matyasf @HerrTopi this is the first part of the spacing update. this pr adds the tokens, the guide and the margin prop for some components. in order to keep the pr normal size and make the review easier, other components will get the margin prop in other PRs

packages/ui-color-picker/src/ColorPicker/props.ts
```diff
@@ -225,6 +225,11 @@ type ColorPickerOwnProps = {
    * If true, alpha slider will be rendered. Defaults to false
    */
   withAlpha?: boolean
+
+  /**
+   * Margin around the component. Accepts a `Spacing` token. See token values and example usage in [this guide](/#layout-spacing).
```
matyasf, line 230: we should put the full link here, so it works in IDEs

docs/guides/layout-spacing.md
```diff
@@ -0,0 +1,92 @@
+---
+title: Layout Spacing
+category: Guides
+order: 8
+---
+
+# Layout Spacing
+
+Our design system provides a set of spacing tokens for consistent layouts and components. Some tokens share values but should be used semantically. For instance, while both `space12` and `buttons` are 12px, `buttons` should be used for spacing between buttons.
+
```
matyasf, line 10: we should mention the legacy tokens here too

packages/ui-themes/src/sharedThemeTokens/spacing.ts
```diff
@@ -33,7 +34,37 @@ const spacing: Spacing = Object.freeze({
   medium: '1.5rem', // 24px
   large: '2.25rem', // 36px
   xLarge: '3rem', // 48px
-  xxLarge: '3.75rem' // 60px
+  xxLarge: '3.75rem', // 60px
+
+  // new spacing tokens:
+  space0: '0px',
```
matyasf, line 40: why are these not in rem units?

matyasf:
Nice work, I like it so far :)

It would be good to have a visual test for all these margins. I suggest to make a large single test (like with`TooltipPositioning.tsx`) that has all the components with a set margin besides each other

docs/guides/layout-spacing.md
balzss: good point, I'll do that

packages/ui-themes/src/sharedThemeTokens/spacing.ts
balzss: I agree it should be updated but then `space60` will be a bit misleading when the user is on a zoomed in screen with larger than 16px rems. what do you think the correct solution is here?

packages/__examples__/.storybook/stories/SpacingTokens.tsx
```diff
@@ -0,0 +1,69 @@
...
+  return (
+    <div>
+      {spaceTokens.map((token) => (
+        <Button margin={token as any} key={token}>{token}</Button>
+      ))}
+    </div>
+  )
+}
+
+export default SpacingTokens
```
matyasf, line 69: I'd also put here all components with some fixed spacing value to test that it works for all of them. something like:
```
<Button margin="space4">Button</Button>
<NumberInput margin="space4">NumberInput</NumberInput>
<TextInput margin="space4">TextInput</TextInput>
...
```

HerrTopi (approved):
Lgtm when the already mentioned issues are fixed.

Nice work!

matyasf (changes requested):
Looks good except that the `Margin prop` stories are not in Storybook, I think you need to add them to `stories.ts`
Also can you check the Storybook stories and accept them if they look good?

packages/shared-types/src/ComponentThemeVariables.ts
```diff
@@ -519,6 +519,7 @@ export type ColorPickerTheme = {
   popoverFooterPadding: Spacing['xSmall']
   popoverFooterTopBorderWidth: Border['widthSmall']
   colorMixerButtonContainerLeftMargin: Spacing['xSmall']
+  spacing: any
```
matyasf, line 522: Can you type this?

matyasf (changes requested):
I think there is something off with the composite margin values, e.g.

`<TextInput margin="0 0 buttons 0" renderLabel="Name"/>` adds a margin around all sides. Same is happening in `ColorPicker`.
If this is a bug can you add a test for it? (I guess it needs a cypress component test?)

packages/shared-types/src/ComponentThemeVariables.ts
HerrTopi: This is very tricky to type, almost impossible if I'm correct.

matyasf (approved):

HerrTopi (approved):

PR #1817, opened 2024-12-09
feat(link): add textUnderlineOffset to Link component theme override

igera93:
Added text-underline-offset option to theme override variables.
Default value is set to auto.

Screenshot 2024-12-09 at 15 14 24

Screenshot 2024-12-09 at 15 07 46


balzss (approved):

HerrTopi (approved):
Nice work

PR #1818, opened 2024-12-09
refactor: upgrade dependencies

matyasf:
Just smaller dependency updates and related minor fixes


packages/pkg-utils/lib/get-package.js
```diff
@@ -32,6 +32,11 @@ exports.getPackage = function getPackage(options) {
   return new Package(result.packageJson, path.dirname(result.path))
 }
 
+/**
+ * Reads a package.json
+ * @param options {readPkgUp.NormalizeOptions}
+ * @returns {readPkgUp.NormalizedPackageJson}
+ */
```
matyasf, line 39: new version of TS complains about not able to infer types if this comment is not here

packages/ui-scripts/lib/icons/generate-icon-fonts.js
```diff
@@ -41,9 +41,8 @@ export default function generateIconFonts({
   name
 }) {
   generateFonts({ ...config, inputDir, outputDir, prefix, name })
-    .then((result) => {
-      // eslint-disable-next-line no-console
-      console.log('Icon font generated:', result)
+    .then((_result) => {
+      //console.log('Icon font generated:', result)
```
matyasf, line 45: This is generating the 'icon created' spam on the beginning of `npm run boorstrap` I've turned this off since its generating like 5000 line of logs

packages/ui-scripts/lib/index.js
```diff
@@ -27,9 +27,10 @@
 import yargs from 'yargs'
 import { yargCommands } from './commands/index.js'
 import { hideBin } from 'yargs/helpers'
-import { readEnv } from './utils/readEnv.js'
+import { findUpSync } from 'find-up'
+import { config } from 'dotenv'
 
-readEnv()
+config({ path: findUpSync('.env') || process.cwd() })
```
matyasf, line 33: Do we need this?

joyenjoyer (approved):

packages/uid/src/__new-tests__/uid.test.tsx
```diff
@@ -34,11 +34,11 @@ describe('uid', () => {
...
     const results = new Set()
-    for (let x = 0; x < 5000; x++) {
-      const id = uid('', 7)
+    for (let x = 0; x < 500; x++) {
+      const id = uid('', 17)
       results.add(id)
     }
-    expect(results.size).toBe(5000)
+    expect(results.size).toBe(500)
   })
```
matyasf, line 42: The original test was failing quite often simply because of probability, now generating 2 similar IDs and the test failing should happen much much less often

git-nandor (approved):

PR #1823, opened 2024-12-11
chore(many): make meta package export everything which is exported in…

ToMESSKa:
… components

Closes: INSTUI-4381

- added everything to the meta file which is exported in the ui-component-name/src/index.ts files
- also added sub component exports to some of the ui-component-name/src/index.ts files where the sub components props were exported but not the sub components themselves (eg, AppNavItem was missing from AppNav)


matyasf (approved):
nice work!

balzss (approved):

matyasf:
FYI i've changed commit type to `feat`

PR #1825, opened 2024-12-12
feat(ui-tabs): extends the customization of Tab component

szalonna:
Closes: [CLX-276](https://instructure.atlassian.net/browse/CLX-276)

Adds the possibility to disable built-in styling of a `Tab` to replace with a custom component through `renderTitle`
- adds `customTab` flag to `Panel` to remove the default styling from the `Tab`
- adds `alignTab` prop to `Panel` to allow to customzie the `Tabs`'s relative alignment in the container flexbox

[CLX-276]: https://instructure.atlassian.net/browse/CLX-276?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/ui-tabs/src/Tabs/Panel/index.tsx
+++ b/packages/ui-tabs/src/Tabs/Panel/index.tsx
@@ -56,7 +56,8 @@ class Panel extends Component<TabsPanelProps> {
     isSelected: false,
     padding: 'small',
     active: false,
-    unmountOnExit: true
+    unmountOnExit: true,
+    customTab: false
   }
 
   componentDidMount() {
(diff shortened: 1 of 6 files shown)


balzss:
thanks for the PR @szalonna! this is an OK solution but I think it would be more in line with our ways of doing custom styling if instead of adding these new props, you would introduce new theme variables (e.g. for alignment) so it could be done with themeoverrides. what do you think @matyasf @HerrTopi ?

HerrTopi:
> thanks for the PR @szalonna! this is an OK solution but I think it would be more in line with our ways of doing custom styling if instead of adding these new props, you would introduce new theme variables (e.g. for alignment) so it could be done with themeoverrides. what do you think @matyasf @HerrTopi ?

I agree as long as it can b

PR #1826, opened 2024-12-12
fix(ui-date-time-input,ui-form-field): make DateTimeInput compatible …

balzss:
…with the new error format

INSTUI-4405

test plan:

- check the datetimeinput examples in the error forms guide
- setting the `isRequired` prop should display two separate asterisks for each field
- when required and displaying an error, the asterisks should turn red
- the errors should be displayed under the input field like it was a single field and not a group

### DIFF
--- a/docs/guides/form-errors.md
+++ b/docs/guides/form-errors.md
@@ -135,6 +135,34 @@ const Example = () => {
           renderMessages={() => messages}
         />
 
+        <DateTimeInput
+          description={`DateTimeInput (layout="column")`}
+          datePlaceholder="Choose a date"
+          dateRenderLabel="Date"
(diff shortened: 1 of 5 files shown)


matyasf (approved):
looks good. I wonder why there is a test failure, when I run the test jsx code in thr docs app the time-input label is there.. If you figure out whats wrong (maybe just rerun the test?) then its good from me

balzss:
> looks good. I wonder why there is a test failure, when I run the test jsx code in thr docs app the time-input label is there.. If you figure out whats wrong (maybe just rerun the test?) then its good from me

fixed. it was a missing `*`

HerrTopi (approved):

PR #1828, opened 2024-12-13
feat(many): add new Typography tokens and update text and heading

HerrTopi:
Closes: INSTUI-4403

### DIFF
--- a/docs/guides/typography-system.md
+++ b/docs/guides/typography-system.md
@@ -1,0 +1,73 @@
+---
+title: Typography
+category: Guides
+order: 8
+---
+
+# Typography
+
+InstUI provides various typography tokens on theme and component level as well.
+
+## Typography components
+
+Your every text-based need should be covered by one of our two typography components, namely `<Text>` and `<Heading>`.
+These components have a `variant` prop with a handful of semantic values to choose from. These values describe multiple aspects of the text, such as `font-weight`, `font-size`, `line-height` and for `<Heading>` the DOM element that should be rendered, (i.e.:`<h1>`, `<h2>`, `<h3>`, `<h4>`, `<h5>`).
+
+While you can set all these properties separately as well, our strong recommendation is to use the variant with semantic values and only deviate from it in cases absolutely necessary.
+
+## <Text>
+
+Here is the list of semantic variant values and what they do in `<Text>`
+
+This is always rendered as `<span>`
+
+The values (e.g.:`weightRegular`) are from the currently used theme's typography sections, for example:
+[Canvas Theme](https://instructure.design/#canvas)
+
+| variant            | fontStyle | weight          | size               | lineHeight    |
+| ------------------ | --------- | --------------- | ------------------ | ------------- |
+| descriptionPage    | normal    | weightRegular   | descriptionPage    | lineHeight150 |
+| descriptionSection | normal    | weightRegular   | descriptionSection | lineHeight150 |
+| content            | normal    | weightRegular   | content            | lineHeight150 |
+| contentImportant   | normal    | weightImportant | content            | lineHeight150 |
+| contentQuote       | normal    | weightRegular   | content            | lineHeight150 |
+| contentSmall       | normal    | weightRegular   | contentSmall       | lineHeight150 |
+| legend             | normal    | weightRegular   | legend             | lineHeight150 |
+
+## <Heading>
+
+Here is the list of semantic variant values and how are they rendered in `<Heading>`
+
+| variant          | dom element |
+| ---------------- | ----------- |
+| titlePageDesktop | h1          |
+| titlePageMobile  | h1          |
+| titleSection     | h2          |
+| titleCardSection | h2          |
+| titleModule      | h3          |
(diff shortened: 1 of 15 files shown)


packages/ui-heading/src/Heading/README.md
```diff
@@ -4,6 +4,31 @@ describes: Heading
 
 Heading is a component for creating typographic headings.
 
+## Variant
+
+Variant takes care of - almost - all use-cases when it comes hedings on pages. Their name reflects the places they meant to be used. It sets the `level` prop and takes care of the style of the heading
+We recommend using `variants` instead of the level (and `as`) props.
```
matyasf, line 10: Are these deprecated? If yes move their examples to the bottom of this page please

packages/ui-heading/src/Heading/index.tsx
```diff
@@ -67,12 +83,26 @@ class Heading extends Component<HeadingProps> {
...
   }
 
+  checkProps() {
+    const { variant, level, as } = this.props
+    if (variant) {
+      if (level) {
+        console.warn("Don't use 'level' with 'variant' ")
+      }
+      if (as) {
+        console.warn("Don't use 'as' with 'variant' ")
```
matyasf, line 93: I would add to the warning that this is coming from the `Heading` component

packages/ui-text/src/Text/README.md
```diff
@@ -4,6 +4,28 @@ describes: Text
 
 A component for styling textual content
 
+## Variant
+
+Variant takes care of - almost - all use-cases when it comes to texts on pages. Their name reflects the places they meant to be used. It sets `size`, `weight`, `fontStyle` and `lineHeight`
+We recommend using `variants` instead of the aforementioned props.
+
+NOTE: when `variant` is set, `size`, `weight`, `fontStyle` and `lineHeight` props are ignored
```
matyasf, line 12: Why is this behaving differently from Heading? We are there not ignoring the values set by `variant`, just logging a warning. IMO we should be consistent, I vote for variant overwriting these props

packages/ui-text/src/Text/props.ts
```diff
@@ -57,10 +57,41 @@ type TextOwnProps = {
...
+   */
+  variant?:
+    | 'descriptionPage'
+    | 'descriptionSection'
+    | 'content'
+    | 'contentImportant'
+    | 'contentQuote'
+    | 'contentSmall'
+    | 'legend'
+  weight?: 'normal' | 'light' | 'bold' | 'weightRegular' | 'weightImportant'
```
matyasf, line 94: Can you please edit the storybook files? These are generating way too many stories.

matyasf (changes requested):
see my comments

packages/ui-heading/src/Heading/README.md
HerrTopi: I don't think so. I'd consider deprecating them, but this needs more investigation imho.

packages/ui-heading/src/Heading/index.tsx
HerrTopi: Done

packages/ui-text/src/Text/README.md
HerrTopi: They work the same way. When variant is set, every other prop that effect the same things as the variant is being ignored or overwritten. Same thing in this case. The warning is for cleaning up the users code

packages/ui-text/src/Text/README.md
```diff
@@ -158,3 +176,7 @@ type: example
   This<sup>is</sup> some<sub>text</sub>.
 </Text>
 ```
+
+### DEPRECATED legacy values
+
+Multiple values from `size`, `weight` and `lineHeight` are deprecated, but still supported for backward compatibility reasons. Please only use the above listed, semantic values.
```
matyasf, line 182: I think it would be good to list here exactly what is deprecated. (same for Heading). We do not list anywhere the deprecated values

docs/guides/typography-system.md
```diff
@@ -0,0 +1,56 @@
...
+| titleModule      | h3          |
+| titleCardLarge   | h4          |
+| titleCardRegular | h4          |
+| titleCardMini    | h4          |
+| label            | h5          |
+| labelInline      | h5          |
+
+## Legacy values
+
+For compatibility reasons we still provide the legacy typography tokens (xxLarge, medium, etc.) so updating InstUI is easier but these tokens shouldn't be used when creating new screens.
```
matyasf, line 56: Please list here what is exactly deprecated. Also can we make some conversation table like 

`lineHeight 'lineHeightCondensed'` --> `'lineHeight125'`

matyasf (changes requested):
looks good, except some doc changes (see my comment)

packages/ui-text/src/Text/README.md
HerrTopi: Nothing is deprecated for heading. I changed for Text

balzss (approved):

PR #1829, opened 2024-12-16
fix(ui-date-time-input): fix DateTimeInput displaying wrong value of its value is changed in a onChange callback

matyasf:
Closes: INSTUI-4406

I could only reproduce it in CodeSandbox, but according to React documentation setState is not synchronous so the results are not deterministic.

TEST PLAN:
Enter '2' to the date input here: https://codesandbox.io/p/sandbox/stupefied-almeida-d8qxtr observe that the DateTimeInput's message field displays the correct date. 
Change the dependency to InstUI 10.9.0, reload the preview and do the same. DateTimeInput's value should be wrong


HerrTopi (approved):
lgtm

balzss (approved):

PR #1831, opened 2024-12-18
fix(ui-menu): screenreaders should read the correct number of menu items

matyasf:
Closes: INSTUI-4285

Tested on VoiceOver, NVDA, JAWS. Also simplified DOM structure quite a bit, there are multiple examples of Menu online that do not use a List to wrap the whole thing, it was not needed

TEST PLAN:
Open the Menu example and go trough it with NVDA, JAWS, VoiceOver. You should be able to navigate and they should read the correct menu item number (e.g. 3 of 7).
Check the DOM structure with the inspector to see if it looks OK

### DIFF
--- a/packages/ui-menu/src/Menu/MenuItem/index.tsx
+++ b/packages/ui-menu/src/Menu/MenuItem/index.tsx
@@ -187,8 +187,6 @@ class MenuItem extends Component<MenuItemProps, MenuItemState> {
         return 'menuitemcheckbox'
       case 'radio':
         return 'menuitemradio'
-      case 'flyout':
-        return 'button'
       default:
         return 'menuitem'
     }
--- a/packages/ui-menu/src/Menu/MenuItemGroup/index.tsx
+++ b/packages/ui-menu/src/Menu/MenuItemGroup/index.tsx
@@ -82,11 +82,8 @@ class MenuItemGroup extends Component<MenuGroupProps, MenuGroupState> {
         selected: this.selectedFromChildren(props) || props.defaultSelected!
       }
     }
-
-    this._labelId = props.deterministicId!('MenuItemGroup')
   }
 
-  private _labelId: string
   ref: Element | null = null
 
   handleRef = (el: Element | null) => {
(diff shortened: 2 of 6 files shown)


packages/ui-menu/src/Menu/MenuItem/index.tsx
```diff
@@ -187,8 +187,6 @@ class MenuItem extends Component<MenuItemProps, MenuItemState> {
         return 'menuitemcheckbox'
       case 'radio':
         return 'menuitemradio'
-      case 'flyout':
-        return 'button'
```
matyasf, line 191: A Menu having a `button` not a `menuitem` was driving NVDA crazy. NVDA really seems to be the most strict of them all

packages/ui-menu/src/Menu/MenuItemGroup/index.tsx
```diff
@@ -239,18 +231,15 @@ class MenuItemGroup extends Component<MenuGroupProps, MenuGroupState> {
...
         ref={this.handleRef}
       >
-        <span id={this._labelId}>{this.renderLabel()}</span>
-        <ul
-          role="menu"
+        {this.renderLabel()}
+        <div
           css={this.props.styles?.items}
           aria-disabled={this.props.disabled ? 'true' : undefined}
-          aria-labelledby={this._labelId}
```
matyasf, line 250: there is no need for `aria-labelledby` . Also its not usable for generic elements

packages/ui-menu/src/Menu/MenuItemGroup/index.tsx
```diff
@@ -239,18 +231,15 @@ class MenuItemGroup extends Component<MenuGroupProps, MenuGroupState> {
       <span
         {...props}
         css={this.props.styles?.menuItemGroup}
-        role="presentation"
         ref={this.handleRef}
       >
-        <span id={this._labelId}>{this.renderLabel()}</span>
-        <ul
-          role="menu"
```
matyasf, line 247: This should be used for submenus and menus, not for groups per specification https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role

packages/ui-menu/src/Menu/index.tsx
```diff
@@ -350,71 +350,59 @@ class Menu extends Component<MenuProps> {
           this.props.controls
```
balzss, line 350: I can't add a comment to line 338 but I think the `<li>` wrapper around the menuseparator should be removed since you removed those elsewhere too and the `<ul>` parent so now that line doesn't make sense anymore

balzss (changes requested):
left one comment but otherwise looks and works well, great job

packages/ui-menu/src/Menu/index.tsx
matyasf: yea, you're right, I missed that one

HerrTopi (approved):

balzss (approved):

PR #1835, opened 2024-12-19
chore: move commitizen config to commitlint;

balzss:
also removed cz-lerna-changelog pkg and skip commit hook on rebase and amend

`update:package:list` and `commit:package:list` was removed from package.json as I wasn't able to figure out why they were needed (probably legacy usecase)

the package `instui-config` was removed too as it isn't serving any purpose anymore

test plan:
- check out the branch and make any change and commit it -> check if the prompt works as intended
- make a change 2 different packages -> check if scopes are correctly calculated
- in the scope select menu, select more then 3 packages -> the final messages should display `many` as the scope
- char limit should be 150 -> after the prompt, extend the message over the limit -> commitlint should give an error
- amend a commit -> prompt should not appear
- rebase the branch -> prompt should not appear


matyasf (changes requested):
very nice work! there is something off with the package.lock.json, when i run `npm i` it generates some changes. I think you just need to do this and commit it.

HerrTopi:
> very nice work! there is something off with the `package-lock.json`, when i run `npm i` it generates some changes. I think you just need to do this and commit it.
> 
> Also for whatever reason `npm` does not delete our old packages from `package-lock.json`, but leaves them there with an `extraneous=true` prop. If I delete these entries they are gone for good, can you please do this?

This should have a better solution than manually managing the package lock, shouldn't it?

HerrTopi (approved):

PR #1839, opened 2025-01-07
fix(ui-pagination): fix Pagination sometimes displaying non `li` elements in lists, fix key errors too

matyasf:
Closes: INSTUI-4422

TEST PLAN:
check all examples in inspector, the pagination buttons should all be in lists. There should be no duplicate key warning in the console

### DIFF
--- a/packages/ui-pagination/src/Pagination/index.tsx
+++ b/packages/ui-pagination/src/Pagination/index.tsx
@@ -22,7 +22,7 @@
  * SOFTWARE.
  */
 /** @jsx jsx */
-import React, { Component } from 'react'
+import React, { Component, ReactElement, ReactNode } from 'react'
 
 import { View } from '@instructure/ui-view'
 import { testable } from '@instructure/ui-testable'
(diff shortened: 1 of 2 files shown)


packages/ui-pagination/src/Pagination/index.tsx
```diff
@@ -104,7 +104,7 @@ class Pagination extends Component<PaginationProps> {
     currentPage: 1,
     siblingCount: 1,
     boundaryCount: 1,
-    ellipsis: <li style={{all: 'unset'}}>...</li>,
```
matyasf, line 107: I think that this was a bug too, that its a `li` should be an internal part of the component

packages/ui-pagination/src/Pagination/index.tsx
```diff
@@ -434,22 +446,22 @@ class Pagination extends Component<PaginationProps> {
 
       if (sliceStart - firstIndex > 1)
         visiblePages.unshift(
-          <span key="first" aria-hidden="true">
-            &hellip;
-          </span>
+          <li style={{ all: 'unset' }} key="first" aria-hidden="true">
+            {this.props.ellipsis}
```
matyasf, line 450: now the `ellipsis` prop works in the legacy code too

balzss (approved):

HerrTopi (approved):

PR #1842, opened 2025-01-09
docs(ui-radio-input,ui-source-code-editor): add functional examples t…

ToMESSKa:
…o SourceCodeEditor and class-based examples to RadioInputGroup

Closes: INSTUI-4419

TEST PLAN:

**- test all the class-based examples in RadioInputGroup if they work the same as the functional examples and without error
- test all the functional examples in SourceCodeEditor if they work the same as the class-based and without error**

### DIFF
--- a/packages/ui-radio-input/src/RadioInputGroup/README.md
+++ b/packages/ui-radio-input/src/RadioInputGroup/README.md
@@ -11,87 +11,210 @@ The `name` prop sets the [same low level HTML attribute](https://developer.mozil
 Adjust the size of the radio button and label text via the `size` prop. The default size is
 `medium`.
 
-```js
----
-type: example
----
-function Example () {
-  const inputs = [
-    {value: 'foo', label: 'Radio option one' },
-    {value: 'bar', label: 'Radio option two' },
-    {value: 'baz', label: 'Radio option three' }
-  ]
-  const handleChange = function (event, value) {
-    console.log(value)
+- ```js
+  class Example extends React.Component {
+    constructor(props) {
+      super(props)
+      this.state = {
+        inputs: [
+          { value: 'foo', label: 'Radio option one' },
+          { value: 'bar', label: 'Radio option two' },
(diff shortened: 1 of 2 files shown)


packages/ui-source-code-editor/src/SourceCodeEditor/README.md
```diff
@@ -54,190 +54,369 @@ Setting the correct language adds **syntax highlighting** and other helpful feat
...
+        </Flex>
+      )
+    }
+  }
+
+  render(<LanguageExamples />)
+  ```
+
+- ```js
+  const languages = {
```
balzss, line 251: the source code has extra padding in the functional examples

packages/ui-radio-input/src/RadioInputGroup/README.md
```diff
@@ -11,87 +11,210 @@ The `name` prop sets the [same low level HTML attribute](https://developer.mozil
...
-function Example () {
-  const inputs = [
-    {value: 'foo', label: 'Radio option one' },
-    {value: 'bar', label: 'Radio option two' },
-    {value: 'baz', label: 'Radio option three' }
-  ]
-  const handleChange = function (event, value) {
-    console.log(value)
+- ```js
+  class Example extends React.Component {
```
balzss, line 15: in the (far away) future we will probably only need functional examples for components so I'd suggest not to add class based ones if we don't have them already

balzss (changes requested):

packages/ui-source-code-editor/src/SourceCodeEditor/README.md
ToMESSKa: fixed it

packages/ui-radio-input/src/RadioInputGroup/README.md
ToMESSKa: I talked to Matyi too and removed the class based examples

balzss (approved):

PR #1844, opened 2025-01-13
test(ui-tray): migrate old Tray tests

git-nandor:


cypress/component/Tray.cy.tsx
```diff
@@ -21,10 +21,49 @@
...
+      >
+        Hello Tray
+        <input type="text" />
+        <input type="text" id="my-input" />
+      </Tray>
+    )
+
+    cy.get('[aria-label="Tray Example"]').as('tray')
+
+    cy.get('@tray').realPress('Escape')
```
ToMESSKa, line 66: The original test have a focus check on the tray before pressing the Esc key, which I don't see here.

ToMESSKa (changes requested):

cypress/component/Tray.cy.tsx
balzss: good catch but the `Tray` should close on escape whether it contains the focus or not so I think it's better to leave it out. maybe a separate test could be added to check with the `shouldContainFocus` prop set to true and some tabbing if it keeps the focus or not but that's out of scope for this PR

balzss (approved):

ToMESSKa (approved):

PR #1846, opened 2025-01-14
fix(ui-menu): make Menu.Item apply target prop

ToMESSKa:
Closes: INSTUI-4425

ISUUE:
- Menu.Item does not apply the target prop to the link, so the link always opens in the same tab

TEST PLAN:
- go the Menu component's example
- try the second menu item's (Default (Grid view)) with different target props like "_blank", "_self" or insert an iframe inside the View (above the first Menu.Item) and set the target to "myFrame":
```
        <iframe name="myFrame" src="about:blank" width="600" height="400"></iframe>
```
- **the link should open according to the given target prop: "_blank" - new tab, "_self" - same tab, "myFrame" - in the iframe**


HerrTopi (approved):
Good work!

PR #1854, opened 2025-01-31
fix(ui-modal): make Modal's header non-sticky with small window height

ToMESSKa:
Closes: INSTUI-4391


packages/ui-modal/src/Modal/ModalHeader/styles.ts
```diff
@@ -83,6 +83,7 @@ const generateStyle = (
       borderBottomWidth: '0.0625rem',
       borderBottomStyle: 'solid',
       borderBottomColor: componentTheme.borderColor,
+      ...(smallPortView ? { position: 'relative' } : {}),
```
ToMESSKa, line 86: I needed to add this because otherwise the close button starts scrolling with the text instead of staying in the Modal.Header

packages/ui-modal/src/Modal/index.tsx
```diff
@@ -146,21 +156,85 @@ class Modal extends Component<ModalProps, ModalState> {
...
+
+    // Separate header and body elements
+    const filteredChildren = childrenArray.filter((child) => {
       if (isValidElement(child)) {
-        return safeCloneElement(child, {
-          variant: variant,
-          overflow: (child?.props as { overflow: string })?.overflow || overflow
-        })
-      } else {
-        return child
```
ToMESSKa, line 161: I moved this to the cloneChildWithProps function

packages/ui-modal/src/Modal/index.tsx
```diff
@@ -86,7 +86,8 @@ class Modal extends Component<ModalProps, ModalState> {
 
     this.state = {
       transitioning: false,
-      open: props.open ?? false
+      open: props.open ?? false,
+      windowHeight: window.innerHeight
```
matyasf, line 90: We should always put null checks to DOM APIs because they dont exist when the user uses SSR. You can use our `canUseDOM` function to check whether its rendered on the server. Its safe to use them in `componentDidMount` AFAIK (please double check!), but definitely not in a static method like here

matyasf (changes requested):
I like the concept! There will be some issues with SSR, but thats an easy fix. After the a11y team agrees  its good from me

packages/ui-modal/src/Modal/index.tsx
ToMESSKa: I set windowHeight to 99999 in the state.(TypeScript didn't allow me not to define it). Also replaced a check with the canUseDOM.

matyasf (approved):
nice work!

packages/ui-modal/src/Modal/ModalHeader/props.ts
```diff
@@ -55,7 +56,8 @@ type ModalHeaderStyle = ComponentStyle<'modalHeader'>
 const propTypes: PropValidators<PropKeys> = {
   children: PropTypes.node,
   variant: PropTypes.oneOf(['default', 'inverse']),
-  spacing: PropTypes.oneOf(['default', 'compact'])
+  spacing: PropTypes.oneOf(['default', 'compact']),
+  smallPortView: PropTypes.bool
```
balzss, line 60: minor thing but why is it `portview` and not `viewport`?

balzss (approved):
I tested it and works as it should. Left a minor comment but other than that it's ok. Good job!

packages/ui-modal/src/Modal/ModalHeader/props.ts
ToMESSKa: Thanks, I've changed it to viewport!

PR #1857, opened 2025-02-04
fix(ui-color-picker): add hex to aria-label

joyenjoyer:
INSTUI-4273

**Test:**

1. you have to test `ColorPreset` and `ColorContrast` separately, go to the docs page examples
2. try it with JAWS, NVDA and VoiceOver
3. go through the colors, the hex content of the color should be read only once
4. if you select one color, the hex content and a 'selected' text should be read
5. check if you notice other accessibility/screenreader issues
6. check the components with Accessibility Insights chrome extension

### DIFF
--- a/packages/ui-color-picker/src/ColorPreset/index.tsx
+++ b/packages/ui-color-picker/src/ColorPreset/index.tsx
@@ -266,7 +266,11 @@ class ColorPreset extends Component<ColorPresetProps, ColorPresetState> {
       cursor={this.props.disabled ? 'not-allowed' : 'auto'}
       as="button"
       {...(selectOnClick ? { onClick: () => this.props.onSelect(color) } : {})}
-      {...(this.isSelectedColor(color) ? { 'aria-label': 'selected' } : {})}
+      {...{
+        'aria-label': `${color}${
+          this.isSelectedColor(color) ? ' selected' : ''
+        }`
+      }}
     >
       <div>
         <ColorIndicator color={color} shape="rectangle" role="presentation" />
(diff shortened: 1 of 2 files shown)


packages/ui-color-picker/src/ColorPreset/index.tsx
```diff
@@ -282,9 +286,25 @@ class ColorPreset extends Component<ColorPresetProps, ColorPresetState> {
...
 
-  renderIndicatorTooltip = (child: React.ReactElement, color: string) => (
-    <Tooltip renderTip={<div>{color}</div>}>{child}</Tooltip>
-  )
+  renderIndicatorTooltip = (child: React.ReactElement, color: string) => {
+    return (
+      <Tooltip
+        renderTip={<div>{color}</div>}
+        elementRef={(element) => {
+          if (
```
joyenjoyer, line 294: `aria-describedby` on the Tooltip`s button has to be removed otherwise the `Tooltip` text is read aloud by screenreaders. I am using the Tooltip's elementref property and the Element's API call to access the button.

packages/ui-color-picker/src/ColorPreset/index.tsx
```diff
@@ -266,7 +266,11 @@ class ColorPreset extends Component<ColorPresetProps, ColorPresetState> {
       cursor={this.props.disabled ? 'not-allowed' : 'auto'}
       as="button"
       {...(selectOnClick ? { onClick: () => this.props.onSelect(color) } : {})}
-      {...(this.isSelectedColor(color) ? { 'aria-label': 'selected' } : {})}
+      {...{
+        'aria-label': `${color}${
+          this.isSelectedColor(color) ? ' selected' : ''
```
joyenjoyer, line 271: If the color is selected, I append the 'selected' text to the hex code so screenreaders read that too.

ToMESSKa (approved):

ToMESSKa:
Don't forget to merge the commits into a single one. :)

matyasf (approved):
nice work :)

PR #1858, opened 2025-02-04
docs(ui-date-time-input): fix Calendar and DateTimeInput page examples clipped off in docs

matyasf:
This fix makes margins much smaller in small viewports, so pages look much better. This fixes Calendar examples being clipped off.

To test check out the DateTimeInput and Calendar pages in docs with width set to 320px via devtools.

closes: INSTUI-4256

### DIFF
--- a/packages/__docs__/src/App/index.tsx
+++ b/packages/__docs__/src/App/index.tsx
@@ -81,6 +81,7 @@ import type {
   ParsedDocSummary
 } from '../../buildScripts/DataTypes.mjs'
 import { logError } from '@instructure/console'
+import type { Spacing } from '@instructure/emotion'
 
 type AppContextType = {
   themeKey: keyof MainDocsData['themes']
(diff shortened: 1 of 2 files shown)


packages/__docs__/src/App/index.tsx
```diff
@@ -824,6 +821,9 @@ class App extends Component<AppProps, AppState> {
               <div css={this.props.styles?.hamburger}>
                 <InstUISettingsProvider>
                   <IconButton
+                    themeOverride={{
+                      secondaryBorderColor: '#343434'
+                    }}
```
matyasf, line 826: Now the hamburger menu is overlapping the content, so I made its border have a bit more contrast

git-nandor (approved):

packages/__docs__/src/App/index.tsx
matyasf: They dont have to agree on smaller changes in our docs app :)

joyenjoyer (approved):

PR #1862, opened 2025-02-06
feat(ui-popover, ui-select): allow overriding Select's dropdown border

ToMESSKa:
Closes: INSTUI-4395

ISUUE: the border of Select's dropdown cannot be overridden now. CLX team need a version without borders. The border radius need to be changed too, but that can be done with a InstUISettingsProvider.

TEST PLAN:

- open an example in Select
- apply following themeoverride on the Select component:
```
themeOverride={{
    popoverBorderWidth: 'none'
  }}
```
- the dropdown should not have a border now
- it should work with other values too eg. 'large'
- test the borderRadius too with InstUISettingsProvider

### DIFF
--- a/packages/ui-popover/src/Popover/props.ts
+++ b/packages/ui-popover/src/Popover/props.ts
@@ -276,7 +276,10 @@ type PopoverOwnProps = {
    */
   elementRef?: (element: Element | null) => void
   /**
-   * Accepts the familiar CSS shorthand to designate border widths corresponding to edges with the following values: 'none', 'small', 'medium','large'(e.g. 'none large none large).
+   * Allowed values: 0, 'none', 'small', 'medium', 'large'.
+   *
+   * Accepts the familiar CSS shorthand to designate border widths corresponding to edges. (e.g. 'none large none large).
+   *
    * Only applies to a Popover without an arrow.
    */
   borderWidth?: BorderWidth


packages/ui-popover/src/Popover/props.ts
```diff
@@ -275,6 +275,11 @@ type PopoverOwnProps = {
    * Provides a reference to the underlying HTML root element
    */
   elementRef?: (element: Element | null) => void
+  /**
+   * Accepts the familiar CSS shorthand to designate border widths corresponding
+   * to edges. Only applies to a Popover without an arrow.
```
ToMESSKa, line 280: I inserted this info about the no arrow version, because the arrow version uses ComponentView instead of View which in turn would need to be rewritten too.

matyasf:
So as I see one can now use this prop and after merging #1860 the border radius will work nice too.
After #1860 is merged can you please rebase this and ask the CLX team for a review?

This is the sample code that they need:

```
<InstUISettingsProvider
  theme={{
        componentOverrides: {
          Select: { popoverBorderWidth: 'large' },
          View: {borderRadiusMedium: '1rem'}
        }
  }}
>
  <SimpleSelect renderLabel="Uncontrolled Select" >
    <SimpleSelect.Option id="foo" value="foo"
                         renderBeforeLabel={(props) => {
                           return <IconCheckSolid />
                         }}>
      Foo
    </SimpleSelect.Option>
    <SimpleSelect.Option id="bar" value="bar">
      Bar
    </SimpleSelect.Option>
    <SimpleSelect.Option id="baz" value="baz">
      Baz
    </SimpleSelect.Option>
  </SimpleSelect>
</InstUISettingsProvider>
```

ToMESSKa:
> So as I see one can now use this prop and after merging #1860 the border radius will work nice too. After #1860 is merged can you please rebase this and ask the CLX team for a review?
> 
> This is the sample code that they need:
> 
> ```
> <InstUISettingsProvider
>   theme={{
>         componentOverrides: {
>           Select: { popoverBorderWidth: 'large' },
>           View: {borderRadiusMedium: '1rem'}
>         }
>   }}
> >
>   <SimpleSelect renderLabel="Uncontrolled Select" >
>     <SimpleSelect.Option id="foo" value="foo"
>                          renderBeforeLabel={(props) => {
>                            return <IconCheckSolid />
>                          }}>
>       Foo
>     </SimpleSelect.Option>
>     <SimpleSelect.Option id="bar" value="bar">
>       Bar
>     </SimpleSelect.Option>
>     <SimpleSelect.Option id="baz" value="baz">
>       Baz
>     </SimpleSelect.Option>
>   </SimpleSelect>
> </InstUISettingsProvider>
> ```

@matyasf I rebased it and added @igera93 for review

igera93 (approved):
Thanks for this change, LGTM 👍

balzss (approved):

PR #1863, opened 2025-02-06
(WIP) Fix description text not read on hover by NVDA

matyasf:


packages/ui-form-field/src/FormFieldLayout/index.tsx
```diff
@@ -184,9 +169,9 @@ class FormFieldLayout extends Component<FormFieldLayoutProps> {
         css={styles?.formFieldLayout}
         style={{ width }}
         aria-describedby={this.hasMessages ? this._messagesId : undefined}
+        aria-labelledby={this.props.label ? this._labelId : undefined}
         ref={this.handleRef}
       >
-        {this.elementType === 'fieldset' && this.renderLegend()}
```
matyasf, line 189: The old solution used here a `fieldset` element. SR only reads its `<legend>` if its the first element. But we use a `Grid`, so the solution was to add a `<legend>` as `ScreenReaderContent` and use `aria-hidden` on the visible label.

The new solution uses `aria-labelledby` instead, so there is just one, visible DOM element for the label

packages/ui-checkbox/src/CheckboxGroup/index.tsx
```diff
@@ -148,6 +148,7 @@ class CheckboxGroup extends Component<CheckboxGroupProps, CheckboxGroupState> {
         vAlign="top"
         messagesId={this._messagesId}
         elementRef={this.handleRef}
+        role="group"
```
matyasf, line 151: This makes SRs read these a bit nicer

packages/ui-form-field/src/FormFieldLayout/index.tsx
```diff
@@ -166,15 +147,25 @@ class FormFieldLayout extends Component<FormFieldLayoutProps> {
...
+    let labelConnector = {}
+    if (ElementType == 'label') {
+      // this is a `<label>`, it needs to be connected to a control
+      // This is the case for e.g. `FormField`, `TextInput`, `NumberInput`
+      labelConnector = { htmlFor: this.props.id }
+    } else if (this.props.label) {
+      // This is a FormFieldGroup (span), it needs to be connected to its label
+      // This is the case for `FormFieldGroup`
+      labelConnector = { 'aria-labelledby': this._labelId }
+    }
```
matyasf, line 168: This is a mess. I just moved this logic here, no change

cypress/component/FormField.cy.tsx
```diff
@@ -36,7 +37,7 @@ describe('<FormField/>', () => {
       </FormFieldLayout>
     )
 
-    cy.get('span[class$="-formFieldLabel"]')
```
matyasf, line 39: For tests likely this will be most of the change

packages/ui-form-field/src/FormFieldLayout/index.tsx
```diff
@@ -84,133 +79,122 @@ class FormFieldLayout extends Component<FormFieldLayoutProps> {
...
     return (
       <ElementType
-        {...omitProps(props, [
-          ...FormFieldLayout.allowedProps,
-          ...Grid.allowedProps
-        ])}
+        {...omitProps(props, [...FormFieldLayout.allowedProps])}
         css={styles?.formFieldLayout}
         style={{ width }}
-        aria-describedby={this.hasMessages ? this._messagesId : undefined}
```
matyasf, line 186: `aria-describedby` is no longer needed because SRs work fine if the `label`'s child is the control

packages/ui-form-field/src/FormField/index.tsx
```diff
@@ -68,7 +68,6 @@ class FormField extends Component<FormFieldProps> {
         label={this.props.label}
         vAlign={this.props.vAlign}
         as="label"
-        htmlFor={this.props.id}
```
matyasf, line 71: This is not needed because the `label`s child is the control, https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#defining_an_implicit_label

ToMESSKa (changes requested):
DateTimeInput error messages are not read by VoiceOver for me

HerrTopi (approved):

packages/ui-form-field/src/FormField/index.tsx
matyasf: EDIT 2 weeks later: we should not do this, it will bring the last leaf on the focus causing issues with e.g. multiple select

PR #1864, opened 2025-02-07
test(ui-select,ui-options): migrate old Options tests

git-nandor:


packages/ui-options/src/Options/Item/__tests__/Item.test.tsx
```diff
@@ -1,246 +0,0 @@
...
-    const link = await item.find('[href="/helloWorld"]')
-
-    expect(link).to.exist()
-    expect(link.getDOMNode().tagName.toLowerCase()).to.equal('a')
-  })
-
-  describe('with generated examples', async () => {
-    generateA11yTests(Item, ItemExamples)
-  })
-})
```
ToMESSKa, line 246: How about these? I cannot see these in the new tests.

ToMESSKa (changes requested):

packages/ui-options/src/Options/Item/__tests__/Item.test.tsx
git-nandor: It returns with 0 generated test cases.
However, among the Options tests, there are generated a11y tests that include different variations of the Items, so I deleted this from here.

ToMESSKa (approved):

balzss (approved):

PR #1865, opened 2025-02-07
fix(ui-source-code-editor): link label to input field programmatically

ToMESSKa:
Closes: INSTUI-4289

ISSUE: SourceCodeEditor label is not linked to the input field pragmatically. Therefore screenreader does not read the label when tabbed into the input field

TEST PLAN:
- open the first example in SourceCodeEditor
- with a screenreader go into the input field
- the screenreader should announce the label 'javascript SourceCodeEditor with syntax highlight" when entering the input field
- the input (textbox) element should have an "aria-labelledby" attribute pointing to the label element by id.

### DIFF
--- a/packages/ui-source-code-editor/src/SourceCodeEditor/index.tsx
+++ b/packages/ui-source-code-editor/src/SourceCodeEditor/index.tsx
@@ -302,6 +302,7 @@ class SourceCodeEditor extends Component<SourceCodeEditorProps> {
     if (indentOnLoad) {
       this.indentAll()
     }
+    this.assignAriaLabel()
   }
 
   componentWillUnmount() {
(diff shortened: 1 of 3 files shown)


packages/ui-source-code-editor/src/SourceCodeEditor/index.tsx
```diff
@@ -655,7 +665,7 @@ class SourceCodeEditor extends Component<SourceCodeEditorProps> {
           omitProps(restProps, SourceCodeEditor.allowedProps)
         )}
       >
-        <label css={styles?.label} htmlFor={this._id}>
+        <label css={styles?.label} id={this._id}>
```
ToMESSKa, line 668: html for does not work in div elements and the textbox is a div element in CodeMirror

packages/ui-source-code-editor/src/SourceCodeEditor/index.tsx
```diff
@@ -644,6 +645,15 @@ class SourceCodeEditor extends Component<SourceCodeEditorProps> {
     }
   }
 
+  assignAriaLabel = () => {
+    if (this._containerRef) {
+      const editorDiv = this._containerRef.querySelector('[role="textbox"]')
```
ToMESSKa, line 650: because the CodeMirror does not provide any way to reach this element, I had to find it like this

joyenjoyer: An alternative way to find textbox is to go through child elements of the `containerRef` if it is an element but this is good as well.

joyenjoyer (approved):

PR #1868, opened 2025-02-12
fix(ui-view,ui-file-drop,ui-buttons): make focus ring radius fit the …

ToMESSKa:
…enclosed element's radius

INSTUI-4375

ISSUE: When a border radius in a Button or FileDrop component is modified, the focus ring radius does not follow.

TEST PLAN:
- test the examples in Button, BaseButton, IconButton, CondensedButton and FileDrop if the focus ring radius will follow the border radius override:

```
themeOverride={{
    borderRadius: '100px'
  }}
```
- the examples's focus ring should work as expected without the override


balzss (approved):
this is a good short-term solution but later we should think about making it automatically follow the border radius (without an extra prop)

PR #1870, opened 2025-02-13
docs(ui-focusable): make screenreader annouce Tooltip in Focusable ex…

ToMESSKa:
…ample by  providing aria props

INSTUI-4313

ISSUE:
In Focusable, there is a tooltip called "I'm Focused!", but no hint is provided for screen reader users about its presence.

TEST PLAN:
- navigate to the first example in [Focusable](https://instructure.design/pr-preview/pr-1870/#Focusable) using different screen readers 
- screen readers should announce both the button text and the tooltip text.
- inspect the HTML code to verify that the elements are correctly associated using aria-labelledby and id attributes

### DIFF
--- a/packages/ui-focusable/src/Focusable/README.md
+++ b/packages/ui-focusable/src/Focusable/README.md
@@ -15,7 +15,7 @@ type: example
     console.log(options)
     return <span>
       <div>
-        <Button>Focus me!</Button>
+        <Button aria-labelledby="focusable-example1-button focusable-example1-tooltip" id="focusable-example1-button">Focus me!</Button>
       </div>
       {options.focused && (
         <ContextView
@@ -26,6 +26,8 @@ type: example
           padding="small"
           borderWidth="small"
           display="block"
+          id="focusable-example1-tooltip"
+          role="tooltip"
         >
           I&#39;m focused!
         </ContextView>


packages/ui-focusable/src/Focusable/README.md
```diff
@@ -15,7 +15,7 @@ type: example
     console.log(options)
     return <span>
       <div>
-        <Button>Focus me!</Button>
```
matyasf, line 18: there is no need to associate the button with itself. Just the tooltip is enough:
`<Button aria-labelledby="tooltip">Focus me!</Button>` Also I suggest to give a more unique ID to items in examples (e.g. `focusable-example1-tooltip`, who knows that else could use this ID in the docs app

matyasf (changes requested):

packages/ui-focusable/src/Focusable/README.md
ToMESSKa: @matyasf I associated the button with itself because otherwise, the button's title would not be announced—only the "tooltip" text would be.
(The first example in Tooltip works similarly; the tooltip text is read after the button's title.)

matyasf (approved):

PR #1872, opened 2025-02-19
test(ui-tabs): migrate old Tabs tests

git-nandor:


packages/ui-tabs/src/Tabs/Panel/__new-tests__/Panel.test.tsx
```diff
@@ -23,30 +23,31 @@
...
+    render(
       <Panel isSelected renderTitle="Panel Title">
         Panel contents
       </Panel>
     )
+    const tabPanel = screen.getByRole('tabpanel')
 
-    const tabPanel = await find('[role="tabpanel"]')
-    expect(tabPanel.getAttribute('role')).to.equal('tabpanel')
+    expect(tabPanel).toBeInTheDocument()
```
ToMESSKa, line 51: It might be better to explicitly check for the aria-role in the new test as well since screen.getByRole('tabpanel') does not necessarily mean the element has the tabpanel attribute, the role might have just been inferred."

packages/ui-tabs/src/Tabs/__tests__/Tabs.test.tsx
```diff
@@ -1,467 +0,0 @@
...
-
-    const tabs = await TabsLocator.find()
-
-    expect(await tabs.findSelectedTab(':contains(Second Tab)')).to.exist()
-    const secondPanel = await tabs.findTabPanel(':contains(Tab 2 content)')
-    expect(secondPanel.getAttribute('aria-hidden')).to.not.exist()
-  })
-
-  it('should not allow selecting a disabled tab', async () => {
-    await mount(
```
ToMESSKa, line 175: This title implies for me that the tab cannot be clicked, but the old and the new tests both just check if it has an aria-hidden role (which has little to do with being selectable).

packages/ui-tabs/src/Tabs/__tests__/Tabs.test.tsx
```diff
@@ -1,467 +0,0 @@
...
-
-    await wait(() => {
-      expect(onChange).to.have.been.calledOnce()
-      expect(onChange.getCall(0).args[1].index).to.equal(1)
-    })
-  })
-
-  it('should update the selected tab when clicked', async () => {
-    const onChange = stub()
-    await mount(
```
ToMESSKa, line 302: What about this test? I cannot see the new version of it.

packages/ui-tabs/src/Tabs/__tests__/Tabs.test.tsx
```diff
@@ -1,467 +0,0 @@
...
-      )
-
-      const subject = await mount(<Example width="150px" />)
-      const tabs = await TabsLocator.find()
-
-      expect(await tabs.find('[class$="-tabs__endScrollOverlay"]')).to.exist()
-
-      const tabPanels = await tabs.findAllTabPanels()
-      tabPanels[2].click()
-
```
ToMESSKa, line 455: I cannot see this click event in the new test, is that for a reason?

packages/ui-tabs/src/Tabs/Panel/__new-tests__/Panel.test.tsx
git-nandor: Thanks! Modified.

packages/ui-tabs/src/Tabs/__tests__/Tabs.test.tsx
git-nandor: In this case, the selection is controlled via the isSelected prop, and the correct aria-hidden attribute values serve as proof of the final rendered state.

packages/ui-tabs/src/Tabs/__tests__/Tabs.test.tsx
git-nandor: This was a duplication.
The test on line 199, "should call onRequestTabChange when selection changes via click", tests the same behavior and more:
```
expect(onChange).to.have.been.called()
expect(onChange.getCall(0).args[1].index).to.equal(1)
expect(onChange.getCall(0).args[1].id).to.equal('two')
```

The deleted assertions were:
```
expect(onChange).to.have.been.called()
expect(onChange.getCall(0).args[1].index).to.equal(1)
```

ToMESSKa (approved):

balzss (approved):

joyenjoyer (approved):

PR #1875, opened 2025-02-20
feat(ui-date-input): add feature to disable dates and access the input's ref

balzss:
INSTUI-4430

test plan:

- add `disabledDates` prop to the component as a list of strings (`['2025-02-20', '2025-02-21', '2025-02-22']`) and open the calendar to check if dates are disabled
- add `disabledDate` prop with an array and see if the dates are disabled (`(date) => date.startsWith('2025-02')`)
- add inputRef prop as a function and see if it returns the inputref
- call that inputref with the `.focus()` fn and see if it focuses the input

### DIFF
--- a/packages/ui-date-input/src/DateInput/README.md
+++ b/packages/ui-date-input/src/DateInput/README.md
@@ -2,7 +2,7 @@
 describes: DateInput
 ---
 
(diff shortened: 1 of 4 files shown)


joyenjoyer (approved):
nice work

PR #1877, opened 2025-02-24
fix(ui-rating): add text value to Rating examples

matyasf:
Also remove a not needed div and make the icon aria-hidden because VO reads sometimes a lots of empty groups

closes: INSTUI-4307


packages/ui-rating/src/RatingIcon/index.tsx
```diff
@@ -110,20 +110,19 @@ class RatingIcon extends Component<RatingIconProps, RatingIconState> {
 
     return (
       <span
+        aria-hidden="true"
```
matyasf, line 113: Sometimes VO was reading "group group group group", this prevents it.

joyenjoyer (approved):

PR #1881, opened 2025-02-28
fix(ui-pagination): remove padding from legacy Pagination

tolads:
<ul> padding was unset and so inherited from the browser's stylesheet

Context: https://instructure.slack.com/archives/C0JCJ63TR/p1740483851746669?thread_ts=1737474461.141169&cid=C0JCJ63TR
Easy change, tried to do it myself:)

You can easily test what has changed by removing the padding in the browser's DevTools.


balzss (approved):
good catch, thank you for the fix!

PR #1885, opened 2025-03-04
fix(ui-popover): add scrollbar to popover if it out of view

joyenjoyer:
INSTUI-4453

**Test 1**

1. Create a Popover with content enough to go offscreen if you Zoom in 400%
2. Zoom in 400%
3. Check if a scrollbar gets displayed

**Test 2**
1. Add a Popover inside a Modal (trigger button should be on a modal) with content enough to go offscreen if you Zoom in 400%
2. Zoom in 400%
3. Check if a scrollbar gets displayed

The Popover being slightly off screen during zoom is not an issue according to @Reka.

### DIFF
--- a/packages/ui-popover/src/Popover/index.tsx
+++ b/packages/ui-popover/src/Popover/index.tsx
@@ -25,35 +25,34 @@
 import React, { Component } from 'react'
 import keycode from 'keycode'
 
+import type { PositionProps } from '@instructure/ui-position'
 import {
-  Position,
+  mirrorHorizontalPlacement,
   parsePlacement,
-  mirrorHorizontalPlacement
+  Position
 } from '@instructure/ui-position'
+import type { ContextViewProps, ViewProps } from '@instructure/ui-view'
 import { ContextView, View } from '@instructure/ui-view'
+import type { DialogProps } from '@instructure/ui-dialog'
 import { Dialog } from '@instructure/ui-dialog'
 import { textDirectionContextConsumer } from '@instructure/ui-i18n'
+import type { RequestAnimationFrameType } from '@instructure/ui-dom-utils'
 import {
-  findDOMNode,
   containsActiveElement,
-  requestAnimationFrame,
-  handleMouseOverOut
+  findDOMNode,
+  handleMouseOverOut,
+  requestAnimationFrame
 } from '@instructure/ui-dom-utils'
 
 import {
-  safeCloneElement,
   callRenderProp,
+  safeCloneElement,
   withDeterministicId
 } from '@instructure/ui-react-utils'
-import { createChainedFunction, shallowEqual, px } from '@instructure/ui-utils'
+import { createChainedFunction, px, shallowEqual } from '@instructure/ui-utils'
 import { logError as error } from '@instructure/console'
 import { testable } from '@instructure/ui-testable'
 import { FocusRegion } from '@instructure/ui-a11y-utils'
-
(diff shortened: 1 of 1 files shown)


packages/ui-popover/src/Popover/index.tsx
```diff
@@ -192,6 +192,15 @@ class Popover extends Component<PopoverProps, PopoverState> {
     }
   }
 
+  isContentOutOfView() {
+    if (this._contentElement) {
+      const rect = this._contentElement.getBoundingClientRect()
+      const windowHeight = window.innerHeight
+      return rect.bottom > windowHeight || rect.top < 0
```
joyenjoyer, line 199: I am using the ref for the displayed content and check whether it crosses the top and bottom sides of the browser window.

packages/ui-popover/src/Popover/index.tsx
```diff
@@ -578,7 +591,15 @@ class Popover extends Component<PopoverProps, PopoverState> {
 
         return (
           <ContextView {...viewProps} borderColor={styles?.borderColor}>
-            {content}
+            <div
+              style={{
```
joyenjoyer, line 595: I left this here instead of putting into the styles file because 1/ it contains logic, 2/ this is not purely styling.

packages/ui-popover/src/Popover/index.tsx
```diff
@@ -192,6 +192,15 @@ class Popover extends Component<PopoverProps, PopoverState> {
     }
   }
 
+  isContentOutOfView() {
+    if (this._contentElement) {
+      const rect = this._contentElement.getBoundingClientRect()
+      const windowHeight = window.innerHeight
```
matyasf, line 198: I think trying to access `window` with SSR will lead to a crash? Can you please use `canUseDOM` and if its false this function could just return `false`.

matyasf (commented):
If I use the test code below it first adds a scrollbar, but if I close and open the popover again it does not

```
  state = {
    isShowingContent: false
  }

  renderCloseButton() {
    return (
      <CloseButton
        placement="end"
        offset="small"
        onClick={() => this.setState({ isShowingContent: false })}
        screenReaderLabel="Close"
      />
    )
  }

  render() {
    return (
      <View>
        <Popover
          renderTrigger={<Button>Sign In</Button>}
          isShowingContent={this.state.isShowingContent}
          onShowContent={(e) => {
            this.setState({ isShowingContent: true })
          }}
          onHideContent={(e, { documentClick }) => {
            this.setState({ isShowingContent: false })
          }}
          on="click"
          screenReaderLabel="Popover Dialog Example"
          shouldContainFocus
          shouldReturnFocus
          shouldCloseOnDocumentClick
          offsetY="16px"
        >
          <View padding="medium" display="block" as="form">
            {this.renderCloseButton()}
            <FormFieldGroup description="Log In">
              <TextInput
                renderLabel="Username"
                inputRef={(el) => {
                  if (el) {
                    this._username = el
                  }
                }}
              />
              <TextInput renderLabel="Password" type="password" />
              <TextInput renderLabel="Password" type="password" />
              <TextInput renderLabel="Password" type="password" />
              <TextInput renderLabel="Password" type="password" />
              <TextInput renderLabel="Password" type="password" />
              <TextInput renderLabel="Password" type="password" />
              <TextInput renderLabel="Password" type="password" />
              <TextInput renderLabel="Password" type="password" />
              <TextInput renderLabel="Password" type="password" />
            </FormFieldGroup>
          </View>
        </Popover>
      </View>
    )
  }
}

render(<Example />)
```

PR #1887, opened 2025-03-04
refactor(ui-form-field): fix wrong bottom margin when layout is inline

matyasf:
Also correct the margin between controls and messages in DateTimeInput

### DIFF
--- a/packages/ui-form-field/src/FormFieldLayout/index.tsx
+++ b/packages/ui-form-field/src/FormFieldLayout/index.tsx
@@ -154,7 +154,11 @@ class FormFieldLayout extends Component<FormFieldLayoutProps> {
         )
       }
       // needs to be wrapped because it needs an `id`
-      return <div id={this._labelId}>{this.props.label}</div>
+      return (
+        <div id={this._labelId} style={{ display: 'contents' }}>
+          {this.props.label}
+        </div>
+      )
     } else return null
   }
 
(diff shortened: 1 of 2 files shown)


packages/ui-form-field/src/FormFieldLayout/styles.ts
```diff
@@ -150,6 +150,7 @@ const generateStyle = (
           )
         },
       columnGap: '0.375rem',
+      rowGap: '0.75rem',
```
matyasf, line 153: I'm using rowGap instead of margin, this makes it look correct for small inline layouts for e.g. RangeInput

packages/ui-form-field/src/FormFieldLayout/index.tsx
```diff
@@ -154,7 +154,11 @@ class FormFieldLayout extends Component<FormFieldLayoutProps> {
         )
       }
       // needs to be wrapped because it needs an `id`
-      return <div id={this._labelId}>{this.props.label}</div>
+      return (
+        <div id={this._labelId} style={{ display: 'contents' }}>
```
matyasf, line 158: display:contents fixes the gap if the label is a ScreenreaderContent, this way it does not add a double margin

packages/ui-form-field/src/FormFieldLayout/styles.ts
```diff
@@ -158,20 +159,16 @@ const generateStyle = (
     },
     formFieldLabel: {
       label: 'formFieldLayout__label',
-      ...(hasVisibleLabel && {
-        ...labelStyles,
-        // NOTE: needs separate groups for `:is()` and `:-webkit-any()` because of css selector group validation (see https://www.w3.org/TR/selectors-3/#grouping)
-        '&:is(label)': labelStyles,
-        '&:-webkit-any(label)': labelStyles
-      })
+      ...labelStyles,
```
matyasf, line 162: hasVisibleLabel is not needed here because this check already happens in `index.tsx`

git-nandor (approved):

joyenjoyer (approved):

PR #1894, opened 2025-03-07
chore(ui-icons): remove ui-icons-build package as it is no longer used

balzss:
INSTUI-4336

test plan:

go to the preview docs and check all icon formats


matyasf (changes requested):
When I run `npm i` it makes changes to the `package.lock.json`.
Also we can see that its a huge gain, it went from 42K LOC to 35K LOC

balzss:
> When I run `npm i` it makes changes to the `package.lock.json`. Also we can see that its a huge gain, it went from 42K LOC to 35K LOC

good catch, I will update it

HerrTopi (approved):
Best feeling of the job: removing unnecessary files.
Nice work!

PR #1908, opened 2025-03-17
introduce new visual regression tests with chromatic instead of storybook

balzss:
INSTUI-4486

test plan:

- check the code changes
- check the images generated in chromatic with the button tests
- check out the branch locally and run the regression app:
  - run install and bootstrap in the root folder
  - go to the `regression-test` folder
  - run install there too
  - run `npm run dev`
  - go to `localhost:3000` and check out the button example


regression-test/package.json
```diff
@@ -8,28 +8,29 @@
     "build": "next build",
     "start": "next start",
     "lint": "next lint",
-    "test": "cypress run"
+    "test": "ELECTRON_EXTRA_LAUNCH_ARGS=--remote-debugging-port=9222 npx cypress run"
```
matyasf, line 11: Can you please add a comment why this is needed?

regression-test/tsconfig.json
```diff
@@ -19,7 +19,8 @@
     ],
     "paths": {
       "@/*": ["./src/*"]
-    }
+    },
+    "target": "ES2017"
```
matyasf, line 23: Can you add a comment here why we need to use a 7 year old version?

.github/workflows/visual-regression.yml
```diff
@@ -0,0 +1,64 @@
+name: "Chromatic"
+
+on: push
+
+jobs:
+  cypress:
+    name: Run Cypress
```
matyasf, line 7: Can you please rename this to something like "Run Cypress for visual regression testing"?

.github/workflows/visual-regression.yml
```diff
@@ -0,0 +1,64 @@
...
+        run: npm ci
+        working-directory: regression-test
+      - name: Run Cypress tests
+        uses: cypress-io/github-action@v6
+        env:
+          ELECTRON_EXTRA_LAUNCH_ARGS: "--remote-debugging-port=9222"
+        with:
+          start: npm run dev
+          working-directory: regression-test
+      - uses: actions/upload-artifact@v4
```
matyasf, line 31: Can you add a 'name' here?

.github/workflows/visual-regression.yml
```diff
@@ -0,0 +1,64 @@
+name: "Chromatic"
+
+on: push
+
+jobs:
+  cypress:
+    name: Run Cypress
+    runs-on: ubuntu-latest
+    container:
+      image: cypress/browsers:node-22.13.1-chrome-132.0.6834.83-1-ff-134.0.2-edge-132.0.2957.115-1
```
matyasf, line 10: Can you use here the latest version? https://hub.docker.com/layers/cypress/browsers/latest/images/sha256-497f1681f759e5f707fc57395bb31aa93df2907afdee1b5b571508d7591b1048

.github/workflows/visual-regression.yml
```diff
@@ -0,0 +1,64 @@
...
+      options: --user 1001
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
```
matyasf, line 18: This is likely not needed, its included in the base Docker image

matyasf (changes requested):
see my comments, nice work!

.github/workflows/visual-regression.yml
```diff
@@ -0,0 +1,64 @@
+name: "Chromatic"
+
+on: push
```
joyenjoyer, line 3: As we discussed, please change this to:
`on: [pull_request]`

joyenjoyer (changes requested):
great job, check the comments

.github/workflows/visual-regression.yml
balzss: done

.github/workflows/visual-regression.yml
balzss: indeed. it has been removed

.github/workflows/visual-regression.yml
balzss: done

regression-test/tsconfig.json
balzss: looks like it wasn't needed so I've removed it

regression-test/package.json
balzss: looks like this isn't needed so I've updated the script

matyasf (approved):

joyenjoyer (approved):

PR #1909, opened 2025-03-18
fix(ui-side-nav-bar): make screenreaders announce Badge text in SideN…

ToMESSKa:
…avBar

INSTUI-4292

ISSUE:
- in Safari with VoiceOver, and with NVDA and JAWS, the Badge count in SideNavBar example was not announced, it worked only in Chrome with VoiceOver

TEST PLAN:
- go over the SideNavBar example with Voiceover in Safari and Chrome, it should announce the text "You have 99 unread messages."
- go over the SideNavBar example with NVDA and JAWS, they should announce the text "You have 99 unread messages."


matyasf (approved):
nice work :)

joyenjoyer (approved):

PR #1912, opened 2025-03-20
fix(ui-pagination): make Pagination wrap on smaller screen sizes and prevent scrollbars

ToMESSKa:
INSTUI-4461

**ISSUE:** 
- on smaller screen sizes the Pagination has a horizontal scrollbar
- in some cases it has a vertical scrollbar and page buttons and their focus ring aren't fully visible when the parent has no padding

**TEST PLAN:**

**test no horizontal scrollbars:**
- open the examples in Pagination
- click Inspect then switch to mobile device view
- when exceeding the width of the container, the pagination numbers should wrap instead of a horizontal scrollbars appearing, and the new lines should be centered (similarly to the the legacy examples)

**test no vertical scrollbars:**
- go to the first example, click Inspect and on the Styles tab remove the preview parent element's (class="css-kinttp-preview") padding
- while the parent's padding removed, try when no margin prop is provided
- while the parent's padding removed when no margin prop is provided, the component should have NO vertical scrollbars (make sure you hover over the component and try scrolling with the mouse too)
- while the parent's padding removed and when no margin prop is provided, the click on the page buttons and make sure the buttons including their focus rings should remain fully visible (e.g. their top or bottom edges should not covered by the parent checkerboard container)

### DIFF
--- a/packages/ui-pagination/src/Pagination/styles.ts
+++ b/packages/ui-pagination/src/Pagination/styles.ts
@@ -41,7 +41,9 @@ const generateStyle = (componentTheme: PaginationTheme): PaginationStyle => {
       all: 'unset',
       display: 'flex',
       alignItems: 'center',
-      gap: componentTheme.pageIndicatorGap
+      gap: componentTheme.pageIndicatorGap,
+      flexWrap: 'wrap',
+      justifyContent: 'center'
     },
     pagination: {
       label: 'pagination',
--- a/cypress/component/Pagination.cy.tsx
+++ b/cypress/component/Pagination.cy.tsx
@@ -22,7 +22,7 @@
  * SOFTWARE.
  */
 import React from 'react'
-import { Pagination, ScreenReaderContent } from '../../packages/ui'
+import { Pagination, ScreenReaderContent } from '@instructure/ui'
 
 import '../support/component'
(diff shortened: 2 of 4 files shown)


packages/ui-pagination/src/Pagination/styles.ts
```diff
@@ -50,7 +52,8 @@ const generateStyle = (componentTheme: PaginationTheme): PaginationStyle => {
     pages: {
       label: 'pagination__pages',
       display: 'inline-flex',
-      alignItems: 'center'
+      alignItems: 'center',
+      margin: '0.3rem'
```
balzss, line 56: why is this needed? and why this value?

balzss: instead of adding a margin here (so focusring isn't cut of), you could set the default margin to something like `space8` (just make sure to use something from the new set of spacing tokens and not something like `xSmall`). this makes sure that when devs are implementing a design they don't have to deal with random margin around the component

ToMESSKa: @balzss removed the 0.3rem margin and set a default 'space8' margin

balzss (approved):

PR #1915, opened 2025-03-24
refactor(many): remove checkerboard background in examples, fix TS type errors, use types from react-docgen

matyasf:
The checkedboard needed to go because dark squared did not meet WCAG 2.1 AA contrast requirements against red error text.
Also type checking of docs during bootstrap and use types from `react-docgen`

To test:
- check pages in the docs app that they look OK and there are no console errors
- check code examples that the checkerboard is gone and things look OK

Fixes INSTUI-4282


packages/__docs__/buildScripts/DataTypes.mts
```diff
@@ -92,99 +93,6 @@ type JSDocFunctionReturns = {
     names: string[]
   }
 }
-// TODO these are from React-docgen Documentation.d.ts,
-// remove when react-docgen exports them
```
matyasf, line 96: `react-docgen` exports these types now, so it can be deleted from here

packages/__docs__/tsconfig.build.json
```diff
@@ -273,6 +273,9 @@
     {
       "path": "../ui-babel-preset/tsconfig.build.json"
     },
+    {
+      "path": "../ui-color-picker/tsconfig.build.json"
+    },
```
matyasf, line 278: This will type check the docs app during build, it just increases build time by ~1 sec

git-nandor (approved):

PR #1917, opened 2025-03-26
Fix generating same ids when using different apps, e.g. in module federation

matyasf:
The cause of this issue was that we generate IDs on the ESM module level, and if 2 or more InstUI applications are executed as a downloaded .js file (like in Module federation) the instance counters are not aware of each other and are generating the same IDs causing all kinds of issues, e.g. with Radio buttons.

To test: Check out https://github.com/matyasf/module-federation-instui and use this build as a dependency -- version `10.14.1-pr-snapshot-1743089682418`
- Radiobuttons should work if both the host and the guest app use this code.
- Radiobuttons should work if one of the apps uses this code, the other one an earlier build (e.g. InstUI v9
- Read the updated docs


packages/ui-top-nav-bar/src/TopNavBar/utils/mapItemsForDrilldown.tsx
```diff
@@ -62,7 +59,7 @@ const mapItemsForDrilldown = (
   const submenus: ItemMappedForDrilldownOption[] = []
   const { currentPageId, renderOptionContent } = options
 
-  const customPopoverIdMap = generateInstanceCounterMap()
+  const customPopoverIdMap = new Map<string, number>()
```
matyasf, line 62: using the global counter caused here a weird bug, I think because it was re-rendered multiple times the IDs got out of sync causing the menu buttons in SmallViewport mode to break.

regression-test/package.json
```diff
@@ -18,17 +18,14 @@
...
-    "tailwindcss": "^3"
+    "eslint-config-next": "14.2.26",
+    "typescript": "^5"
   },
   "dependencies": {
     "@instructure/ui": "snapshot",
-    "next": "^14.2.3",
+    "next": "^14.2.26",
     "react": "^18",
     "react-dom": "^18"
```
matyasf, line 30: while testing this with SSR, I've updated and simplified our regression test a bit

balzss (commented):
I'm rewriting a lot of the regression test in [this commit](https://github.com/instructure/instructure-ui/pull/1908) so you changes would be deleted (or updated) anyway, can you leave it out from this pr?

HerrTopi (approved):

packages/ui-react-utils/src/DeterministicIdContext/DeterministicIdContext.ts
```diff
@@ -22,9 +22,31 @@
...
 import React from 'react'
-import { generateInstanceCounterMap } from './generateInstanceCounterMap'
+import type { DeterministicIdProviderValue } from './DeterministicIdContextProvider'
+
+declare global {
+  // eslint-disable-next-line no-var
+  var __INSTUI_GLOBAL_INSTANCE_COUNTER__: Map<string, number>
+}
+const __INSTUI_GLOBAL_INSTANCE_COUNTER__ = '__INSTUI_GLOBAL_INSTANCE_COUNTER__'
+
```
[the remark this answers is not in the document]
matyasf: Good catch, I've renamed it

balzss (approved):

PR #1924, opened 2025-04-01
fix(ui-form-field): update form error message color to have better contrast

balzss:
INSTUI-4491

test plan:

- check a newError message in the docs app -> the red should be darker than the current/previous color


ToMESSKa:
@balzss Now the input field border and the exclamation mark icon have different red than the text. Is that okay?

ToMESSKa (commented):
@balzss Now the input field border and the exclamation mark icon have different red than the text. Is that okay?

balzss:
> @balzss Now the input field border and the exclamation mark icon have different red than the text. Is that okay?

you are right, probably need to change them too!

balzss:
@ToMESSKa from the discussion with designers: 
> "Only the text should be updated. Strokes, icons, surfaces can still use the 4570 versions, because they have different A11y requirements."

ToMESSKa (approved):

joyenjoyer (approved):

PR #1926, opened 2025-04-02
fix(ui-buttons): fix tabindex=0 added unnecessarly to Buttons

matyasf:
This is added by the browser for interactive element. Also fix builds failing due to missing import in Breadcrumb

Fixes INSTUI-4499

### DIFF
--- a/packages/ui-breadcrumb/package.json
+++ b/packages/ui-breadcrumb/package.json
@@ -41,6 +41,7 @@
     "@instructure/ui-react-utils": "10.15.1",
     "@instructure/ui-testable": "10.15.1",
     "@instructure/ui-truncate-text": "10.15.1",
+    "@instructure/ui-tooltip": "10.15.1",
     "@instructure/ui-utils": "10.15.1",
     "@instructure/ui-view": "10.15.1",
     "prop-types": "^15.8.1"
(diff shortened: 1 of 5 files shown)


packages/ui-breadcrumb/package.json
```diff
@@ -41,6 +41,7 @@
     "@instructure/ui-react-utils": "10.15.0",
     "@instructure/ui-testable": "10.15.0",
     "@instructure/ui-truncate-text": "10.15.0",
+    "@instructure/ui-tooltip": "10.15.0",
```
matyasf, line 44: This is used by Breadcrumb, but was not added leading to build errors

balzss:
I've read through the raised issue and I don't see any reasoning why this should be fixed. I mean I get that this is redundant but what actual issue is caused by the tabindex being there? would be nice to get better reasons than it being redundant since if that's the only issue with it than this is very low priority (and not moderate as it was indicated in the ticket) but I might have missed some crucial information

balzss (approved):
to me it seems this adds unnecessary complexity to our code, maybe we should limit the possible values for the `as` prop since I can't really imagine too many usecases other `a`

matyasf:
> I've read through the raised issue and I don't see any reasoning why this should be fixed. I mean I get that this is redundant but what actual issue is caused by the tabindex being there?

Good questions, I'll ask them, since it has a high priority

git-nandor (approved):

PR #1928, opened 2025-04-03
fix(ui-text-input): make TextInput maintain focus when renderAfterInp…

ToMESSKa:
…ut is conditionally rendered

INSTUI-4466

### DIFF
--- a/packages/ui-text-input/src/TextInput/index.tsx
+++ b/packages/ui-text-input/src/TextInput/index.tsx
@@ -309,7 +309,11 @@ class TextInput extends Component<TextInputProps, TextInputState> {
       ? callRenderProp(renderAfterInput)
       : null
 
-    const renderBeforeOrAfter = !!beforeElement || !!afterElement
+    const renderBeforeOrAfter =
+      !!beforeElement ||
+      !!afterElement ||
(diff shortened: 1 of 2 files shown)


packages/ui-text-input/src/TextInput/index.tsx
```diff
@@ -145,12 +145,15 @@ class TextInput extends Component<TextInputProps, TextInputState> {
...
+    const beforeElement = this.props.renderBeforeInput
+      ? callRenderProp(this.props.renderBeforeInput)
+      : null
     return {
       disabled: interaction === 'disabled',
       invalid: this.invalid,
       focused: hasFocus,
       afterElementHasWidth: afterElementHasWidth,
-      beforeElementExists: this.props.renderBeforeInput != undefined
+      beforeElementExists: !!beforeElement
```
ToMESSKa, line 156: A paddingInlineStart in styles.js is being applied based on beforeElementExists (which checks if the prop is defined), rather than whether beforeElement actually renders content. This would result in unwanted padding after I modified `const renderBeforeOrAfter` (see below)

HerrTopi (approved):

matyasf (approved):

PR #1930, opened 2025-04-04
fix: update PropTypes to align with the new spacing tokens

ToMESSKa:
INSTUI-4501

https://github.com/instructure/instructure-ui/pull/1912

In the above fix, a new spacing token was set as default and only the TypeScript types were updated and PropTypes weren't.


matyasf (approved):
Are we using now `ThemeablePropTypes.spacing` anywhere? If not I think we should delete it, its misleading because it does not contain all our spacing values

balzss (approved):
thanks for the quick fix

balzss (approved):

matyasf (approved):
Looks good! Can you please make a ticket to clear up the remains of typing in `ThemeablePropValues`? Thanks

PR #1935, opened 2025-04-08
fix(ui-select,ui-form-field): fix iOS VoiceOver with Select and SimpleSelect announcing 'readonly' and 'textinput'

matyasf:
Also fix formFieldLayout spacing issue if there are multiple messages and the first one is an empty string

To test:
Check Select read-only, SimpleSelect, TimeSelect examples with VoiceOver on iOS and Desktop

Fixes INSTUI-4500

### DIFF
--- a/packages/ui-select/src/Select/index.tsx
+++ b/packages/ui-select/src/Select/index.tsx
@@ -697,10 +697,15 @@ class Select extends Component<SelectProps> {
     // popup buttons rather than comboboxes.
     const overrideProps: Partial<TextInputProps> = !isEditable
       ? {
-          // Given that Safari with Voiceover does not support proper combobox
-          // handling, a button role is set as a workaround.
+          // We need role="combobox" for the 'open list' button shortcut to work
+          // with desktop screenreaders.
+          // But desktop Safari with Voiceover does not support proper combobox
+          // handling, a 'button' role is set as a workaround.
           // See https://bugs.webkit.org/show_bug.cgi?id=236881
-          role: utils.isSafari() ? 'button' : 'combobox',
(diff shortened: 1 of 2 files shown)


packages/ui-select/src/Select/index.tsx
```diff
@@ -697,10 +697,15 @@ class Select extends Component<SelectProps> {
...
+          // We need role="combobox" for the 'open list' button shortcut to work
+          // with desktop screenreaders.
+          // But desktop Safari with Voiceover does not support proper combobox
+          // handling, a 'button' role is set as a workaround.
           // See https://bugs.webkit.org/show_bug.cgi?id=236881
-          role: utils.isSafari() ? 'button' : 'combobox',
+          // Also on iOS Chrome with role='combobox' it announces unnecessarily
+          // that its 'read-only' and that this is a 'textfield', see INSTUI-4500
+          role:
+            utils.isSafari() || utils.isAndroidOrIOS() ? 'button' : 'combobox',
```
matyasf, line 708: This is the fix. When we use `combobox` role, it doesnt seem to change the role of the trigger element itself, just speaks extra stuff like that it can be opened. If we use the `button` role, it does not announce that its actually an `input`, and neither that its `read-only`. 

On Android sadly its still broken (so no change), I did not use something like `isIOS` because I want to minimize the number of different browser detectors.

joyenjoyer (approved):
It works on desktop. I could not test it on iOS but the code looks good to me. @HerrTopi @ToMESSKa please test this on iOS devices.

ToMESSKa (approved):

PR #1937, opened 2025-04-10
feat(ui-link): add variants to <Link>

HerrTopi:
TEST_PLAN:
Check the [ticket](https://instructure.atlassian.net/browse/INSTUI-4506) and if it's criteria is met. Check the README and if the explanation is it's understandable

INSTUI-4506

### DIFF
--- a/packages/ui-link/src/Link/README.md
+++ b/packages/ui-link/src/Link/README.md
@@ -22,6 +22,37 @@ type: example
 </View>
 ```
 
+### Variant
+
+In order to make it easy to get the most commonly used links, we have the variant prop. It will set all the necessary styles (fontSize, lineHeight, and textDecoration).
+
+```js
+---
+type: example
+---
+<div>
+<div>
+In a line of text you should use the <Link variant="inline" renderIcon={<IconUserLine />} href="https://instructure.github.io/instructure-ui/">inline</Link> link variant.
+</div>
+
+<br></br>
+<div>
+<Text variant="contentSmall">In a line of text, where the text is smaller, use the <Link variant="inline-small"  renderIcon={<IconUserLine />} href="https://instructure.github.io/instructure-ui/">inline-small</Link> link variant
+</Text>
+</div>
+
+<br></br>
+<div>
+If the link is standalone (not in a text), use the <code>standalone</code> <Link display="block" variant="standalone" renderIcon={<IconUserLine />} href="https://instructure.github.io/instructure-ui/">standalone</Link>
+</div>
+
+<br></br>
+<div>
+If the link is standalone (not in a text), but smaller, use the <code>standalone-small</code> <Link display="block" variant="standalone-small"  renderIcon={<IconUserLine />} href="https://instructure.github.io/instructure-ui/">standalone-small</Link>
+</div>
+</div>
+```
+
 ### Adding margin
 
 Use the `margin` prop to add space to the left or right of the Link. Because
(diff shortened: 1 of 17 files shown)


git-nandor (approved):

packages/ui-link/src/Link/styles.ts
```diff
@@ -40,10 +40,33 @@ const generateStyle = (
   props: LinkProps,
   state: LinkStyleProps
 ): LinkStyle => {
-  const { isWithinText, renderIcon, iconPlacement, color } = props
+  const { isWithinText, renderIcon, iconPlacement, color, variant } = props
   const { containsTruncateText, hasVisibleChildren } = state
   const inverseStyle = color === 'link-inverse'
 
+  const variantStyles = {
```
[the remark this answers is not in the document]
HerrTopi: These are indeed in the styles file. I don't know what you mean

packages/ui-link/src/Link/index.tsx
```diff
@@ -79,12 +79,23 @@ class Link extends Component<LinkProps, LinkState> {
   }
   ref: Element | null = null
 
+  checkProps = () => {
+    const { variant, isWithinText } = this.props
+    if (variant && isWithinText) {
+      console.warn(
```
matyasf, line 85: Do not use bare `console`, use our `console` package

HerrTopi: Changed

packages/ui-link/src/Link/README.md
```diff
@@ -22,6 +22,37 @@ type: example
...
+
+<br></br>
+<div>
+<Text variant="contentSmall">In a line of text, where the text is smaller, use the <Link variant="inline-small"  renderIcon={<IconUserLine />} href="https://instructure.github.io/instructure-ui/">inline-small</Link> link variant
+</Text>
+</div>
+
+<br></br>
+<div>
+If the link is standalone (not in a text), use the `standalone` <Link display="block" variant="standalone" renderIcon={<IconUserLine />} href="https://instructure.github.io/instructure-ui/">standalone</Link>
```
matyasf, line 46: use here `<code>standalone</code>` so it renders properly (same with standalone-small)

packages/ui-link/src/Link/index.tsx
```diff
@@ -79,12 +79,23 @@ class Link extends Component<LinkProps, LinkState> {
   }
   ref: Element | null = null
 
+  checkProps = () => {
+    const { variant, isWithinText } = this.props
+
+    warn(
+      !!variant && !!isWithinText,
```
matyasf, line 86: This does not work. First this needs to be `false` for a warning to display and second it will not work since we default `isWithinText` to `true`, so it displays the warning by default. I think it can be removed

matyasf (changes requested):

packages/ui-link/src/Link/index.tsx
```diff
@@ -79,12 +79,23 @@ class Link extends Component<LinkProps, LinkState> {
...
+      '<Link> when `variant` is used, `isWithinText` is ignored. Please remove it!'
```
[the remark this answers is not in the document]
HerrTopi: Removed anyways

matyasf (approved):
2 small things: 
- Link font size does not come from the theme, this is a bit strange to me
- The gap between the icon and the text is the same for all sizes, is this OK?

PR #1938, opened 2025-04-10
fix(ui-date-input,ui-avatar): add ref support to functional components

matyasf:
This is needed for things that try to access the `ref` prop, like our [Transition](https://github.com/instructure/instructure-ui/blob/7226e21d53f5e36820440205bfc2e269281e47bd/packages/ui-motion/src/Transition/BaseTransition/index.ts#L342) or [Position](https://github.com/instructure/instructure-ui/blob/7226e21d53f5e36820440205bfc2e269281e47bd/packages/ui-position/src/Position/index.tsx#L295) package.
Also we have a page asking users to add the `ref` prop, see https://instructure.design/#accessing-the-dom
Finally in React 19 this code will be much simpler, see https://react.dev/blog/2024/12/05/react-19#ref-as-a-prop 

To test: check is `Transition`, `Position` works with `Avatar` and `DateInput2`

Fixes INSTUI-4520

### DIFF
--- a/packages/ui-date-input/src/DateInput2/index.tsx
+++ b/packages/ui-date-input/src/DateInput2/index.tsx
@@ -22,7 +22,14 @@
  * SOFTWARE.
  */
 
-import { useState, useEffect, useContext } from 'react'
+import {
+  useState,
+  useEffect,
+  useContext,
+  forwardRef,
+  ForwardedRef,
+  ValidationMap
+} from 'react'
 import type { SyntheticEvent } from 'react'
 import { Calendar } from '@instructure/ui-calendar'
 import { IconButton } from '@instructure/ui-buttons'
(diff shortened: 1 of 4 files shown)


packages/ui-date-input/src/DateInput2/index.tsx
```diff
@@ -132,224 +139,234 @@ function parseLocaleDate(
...
-  const getLocale = () => {
-    if (locale) {
-      return locale
-    } else if (localeContext.locale) {
-      return localeContext.locale
-    }
-    // default to the system's locale
-    return Locale.browserLocale()
-  }
+const DateInput2 = forwardRef(
```
matyasf, line 142: The actual change is just wrapping this in a `forwardRef`

packages/ui-date-input/src/DateInput2/index.tsx
```diff
@@ -132,224 +139,234 @@ function parseLocaleDate(
...
-            visibleMonth={selectedDate}
-            locale={getLocale()}
-            timezone={getTimezone()}
-            renderNextMonthButton={
+    const selectedDate = parseDate(value)[1]
+
+    return (
+      <TextInput
+        {...passthroughProps(rest)}
+        ref={ref}
```
matyasf, line 297: this is the only change in the component

HerrTopi (approved):

PR #1940, opened 2025-04-11
Rewrite focus ring logic in `View`

matyasf:
The main change is to use CSS `outline` instead of an absolute positioned `:before` element. This has the following effects:

`+` Now focus ring is not part of the size calculation, so there will be no scrollbars e.g. if the `overflow` is set to `auto` in some parent element.
`+` `View` now can display focus rings even if its `position` is not `absolute`
`+` Page CSS and code became simpler since we are not using `:before` CSS pseudo tags
`+` Focus ring will follow fancy `border` settings (see `View`'s focus ring examples)
`~` The focus animation is very slightly different. (the old one was using `scale`, this uses `outline-offset`)
`-` We were using the `outline` prop to display the debug border with the `withVisualDebug` prop. To fix this conflict the debug border now uses box-shadow; this overwrite the normal shadow of the `View` and is not a dashed but a solid line.

For future components we should use `View`'s focus ring for all our components, this could be done with the https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within CSS pseudo-class, I've added a new prop for it (`focusWithin`). But sadly we define how the focus ring should look like in several component's theme and use there theme variables, so removing these and switching to this new prop would be a breaking change.

To test:
- Check the focus examples in View https://instructure.design/pr-preview/pr-1940/#View/#Indicating%20that%20a%20View%20is%20focused , play with View props (e.g. `border`) and theme variables that change how the focus ring and border behave
- Test the new `focusWithin` prop with the current focus/border props

### DIFF
--- a/packages/ui-flex/src/Flex/README.md
+++ b/packages/ui-flex/src/Flex/README.md
@@ -279,7 +279,7 @@ type: example
 ### Handling overflow
 
 When `direction` is set to `column`, Flex.Items' `overflowY` property is automagically set
-to `auto` to account for content overflow with a vertical scrollbar.
+to `auto` to account for content overflow with a vertical scrollbar. Add padding, so focus rings are not cut off.
 
 > To override this default, simply set `overflowY` on the Flex.Item to either `visible` or `hidden`.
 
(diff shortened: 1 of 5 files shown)


packages/ui-flex/src/Flex/README.md
```diff
@@ -279,25 +279,28 @@ type: example
 ### Handling overflow
 
 When `direction` is set to `column`, Flex.Items' `overflowY` property is automagically set
-to `auto` to account for content overflow with a vertical scrollbar.
+to `auto` to account for content overflow with a vertical scrollbar. Add padding, so focus rings are not cut off.
```
matyasf, line 282: I don't like that we are doing this automagically, but I think it would be a breaking change to remove

balzss (approved):

PR #1944, opened 2025-04-16
fix(many): make Select accessible for iOS VoiceOver

ToMESSKa:
INSTUI-4428

**ISSUE:**
- Select does not work properly on iOS with Voiceover. After opening the listbox by double tapping on the input field, the listbox closes when the user try to swipe right to jump from the input to the listbox items

**TEST PLAN:**
- make sure you have the latest iOS update
- test the examples in Select, SimpleSelect, TimeSelect, DateTimeInput, Table (sortable responsive), Calendar (with yearpicker) examples **with VoiceOver on iPad and/or iPhone**
- open the listbox by double tapping on the input field and swipe right
- you should be able to jump to a listbox item after swiping right
- the listbox should remain open after swiping right
- you should be able to navigate through the listbox items by swiping left and right
- after double tapping on a list item, the focus should return to the input field and the tapped item should be selected
- you should be able to navigate through the listbox item even when the options are filtered by typing something into the input field when autocomplete is avaliable
- try the examples in as many other configuration as possible, they should remain accessible (on macOS with Chrome/Safari, on Windows with NVDA/JAWS)

### DIFF
--- a/packages/__docs__/src/Preview/styles.ts
+++ b/packages/__docs__/src/Preview/styles.ts
@@ -80,7 +80,6 @@ const generateStyle = (
   const previewStyle = {
     boxSizing: 'border-box',
     margin: 0,
-    position: 'relative',
     overflow: 'auto',
 
     ...backgroundVariants[background!],
--- a/packages/ui-select/src/Select/README.md
+++ b/packages/ui-select/src/Select/README.md
@@ -26,6 +26,14 @@ describes: Select
       selectedOptionId: this.props.options[0].id,
       announcement: null
     }
+    inputElement = null
+
+    focusInput = () => {
+      this.inputElement.blur()
+      if (this.inputElement) {
+        this.inputElement.focus()
+      }
+    }
 
     getOptionById(queryId) {
       return this.props.options.find(({ id }) => id === queryId)
@@ -69,6 +77,7 @@ describes: Select
     }
 
     handleSelectOption = (event, { id }) => {
+      this.focusInput()
       const option = this.getOptionById(id).label
       this.setState({
         selectedOptionId: id,
@@ -99,6 +108,9 @@ describes: Select
             onRequestHideOptions={this.handleHideOptions}
             onRequestHighlightOption={this.handleHighlightOption}
             onRequestSelectOption={this.handleSelectOption}
+            inputRef={(el) => {
+              this.inputElement = el
+            }}
           >
             {this.props.options.map((option) => {
               return (
@@ -158,6 +170,14 @@ describes: Select
     const [highlightedOptionId, setHighlightedOptionId] = useState(null)
     const [selectedOptionId, setSelectedOptionId] = useState(options[0].id)
     const [announcement, setAnnouncement] = useState(null)
+    const [inputElement, setInputElement] = useState(null)
+
+    const focusInput = () => {
+      inputElement.blur()
+      if (inputElement) {
+        inputElement.focus()
+      }
+    }
 
     const getOptionById = (queryId) => {
       return options.find(({ id }) => id === queryId)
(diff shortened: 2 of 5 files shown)


packages/__docs__/src/Preview/styles.ts
```diff
@@ -80,7 +80,6 @@ const generateStyle = (
   const previewStyle = {
     boxSizing: 'border-box',
     margin: 0,
-    position: 'relative',
```
ToMESSKa, line 83: if this is not removed, the listbox Popover is covered by the preview box when the mountNode is not window.body. Mounting the listbox Popover on the end of the document make sure to be always on the top, but the listbox need to be mounted elsewhere to use it with VoiceOver iOS.

packages/ui-select/src/Select/README.md
```diff
@@ -69,6 +77,7 @@ describes: Select
     }
 
     handleSelectOption = (event, { id }) => {
+      this.focusInput()
```
ToMESSKa, line 80: All the examples need to be updated in Select with removing then reapplying focus on the input field because VoiceOver cursor on iOS follows the browser focus to return to the input field after selecting and item. Without this, the cursor jumps to an element other that the input because there was not focus change registered on the input (the input remained focused all the way while the listbox is open and a element was selected)

matyasf (changes requested):
Very nice work, congratulations, this was a very tough issue :)
I've tested it on iOS Chrome+VO, Windows Chrome+JAWS, Windows Chrome+NVDA and it worked nice everywhere.

There is just one smaller documentation change needed, see my comment.

packages/ui-select/src/Select/README.md
```diff
@@ -2532,3 +2699,15 @@ To display icons (or other elements) before or after an option, pass it via the
...
+---
+type: embed
+---
+<Guidelines>
+  <Figure recommendation="a11y" title="Accessibility">
+    <Figure.Item>To ensure Select is accessible for iOS VoiceOver users, the input field’s focus must be blurred and then reapplied after selecting an option and closing the listbox. The examples above demonstrate this behavior.
+    </Figure.Item>
+  </Figure>
+</Guidelines>
+```
```
ToMESSKa, line 2713: @matyasf I added this a11y guideline

packages/ui-select/src/Select/README.md
```diff
@@ -26,6 +26,14 @@ describes: Select
       selectedOptionId: this.props.options[0].id,
       announcement: null
     }
+    inputElement = null
+
+    focusInput = () => {
```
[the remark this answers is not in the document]
ToMESSKa: added

HerrTopi (approved):

packages/ui-select/src/Select/index.tsx
```diff
@@ -735,7 +740,17 @@ class Select extends Component<SelectProps> {
       shouldNotWrap,
       display: isInline ? 'inline-block' : 'block',
       renderBeforeInput: this.handleRenderBeforeInput(),
-      renderAfterInput: this.handleRenderAfterInput(),
+      // On iOS VoiceOver, if there is a custom element instead of the changing up and down arrow button
```
ToMESSKa, line 743: This is another potential VoiceOver bug and its workaround: when the button that opens the listbox does not update (e.g., it remains a search icon instead of an down arrow changing to an up arrow), the listbox closes unexpectedly when swiping.

packages/ui-select/src/Select/index.tsx
```diff
@@ -798,7 +813,16 @@ class Select extends Component<SelectProps> {
             <Popover
               constrain={constrain}
               placement={placement}
-              mountNode={mountNode}
+              // On iOS VoiceOver, the Popover is mounted right after the input
+              // in order to be able to navigate through the list items with a swipe.
+              // The swipe would result in closing the listbox if mounted elsewhere.
```
ToMESSKa, line 818: The bug here is that VoiceOver touchscreen focus does not move to the listbox popup when it is mounted elsewhere in the DOM (e.g., at the end of <body>), despite correct ARIA attributes. Despite the workaround, I also created a WebKit bug report for this: https://bugs.webkit.org/show_bug.cgi?id=289338

PR #1945, opened 2025-04-16
fix(ui-date-input): correctly format the DateInput2 placeholder in every timezone

balzss:
INSTUI-4498

test plan

- go to the dateinput2 docs page and set the first example to every possible timezone offset (e.g. `Etc/GMT+4`, `Etc/GMT+5`, `Etc/GMT-10`, etc.) -> the placeholder should never break

### DIFF
--- a/packages/ui-date-input/src/DateInput2/index.tsx
+++ b/packages/ui-date-input/src/DateInput2/index.tsx
@@ -214,7 +214,7 @@ const DateInput2 = ({
     return date ? [formatDate(date), date.toISOString()] : ['', '']
   }
 
-  const formatDate = (date: Date): string => {
+  const formatDate = (date: Date, timeZone: string = getTimezone()): string => {
     // use formatter function if provided
     if (typeof dateFormat !== 'string' && dateFormat?.formatter) {
       return dateFormat.formatter(date)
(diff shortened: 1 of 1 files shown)


ToMESSKa (changes requested):
If I provide an invalid timezone like "Invalid/Timezone", it does not give an error anymore. It might be worth noting this in the documentation.

packages/ui-date-input/src/DateInput2/index.tsx
```diff
@@ -214,21 +214,21 @@ const DateInput2 = ({
...
       typeof dateFormat === 'string' ? dateFormat : getLocale(),
-      { timeZone: getTimezone(), calendar: 'gregory', numberingSystem: 'latn' }
+      { timeZone, calendar: 'gregory', numberingSystem: 'latn' }
     )
   }
 
   const getDateFromatHint = () => {
     const exampleDate = new Date('2024-09-01')
-    const formattedDate = formatDate(exampleDate)
+    const formattedDate = formatDate(exampleDate, 'UTC') // exampleDate is in UTC so format it as such
```
matyasf, line 231: In the date constructor above it would be nicer to explicitly use the UTC timezone, not to rely on a JS quirk, e.g. `Date.UTC('2024-09-01')`

matyasf (changes requested):

matyasf (approved):

balzss:
> If I provide an invalid timezone like "Invalid/Timezone", it does not give an error anymore. It might be worth noting this in the documentation.

@ToMESSKa after some considerations, I think the new behaviour is actually more logical since it falls back to the default. But if it causes confusion in the future, I'll extend the docs

ToMESSKa (approved):

PR #1946, opened 2025-04-22
fix(ui-text-input,ui-simple-select,ui-select): diabled is replaced wi…

ToMESSKa:
…th aria-disabled

INSTUI-4485

### DIFF
--- a/packages/ui-select/src/Select/index.tsx
+++ b/packages/ui-select/src/Select/index.tsx
@@ -746,7 +746,8 @@ class Select extends Component<SelectProps> {
 
       onFocus,
       onBlur: utils.createChainedFunction(onBlur, onRequestHideOptions),
-      ...overrideProps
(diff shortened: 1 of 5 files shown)


packages/ui-utils/src/getBrowser.ts
```diff
@@ -55,7 +55,7 @@ const isIE = () => {
 const isChromium = () => {
   const chromiumBasedBrowsers = ['Chrome', 'Chromium', 'Opera', 'Edge']
   return chromiumBasedBrowsers.some((browser) => {
-    getBrowser().name?.includes(browser)
+    return getBrowser().name?.includes(browser)
```
ToMESSKa, line 58: this did not work properly

matyasf: nice catch :)

PR #1950, opened 2025-04-22
feat(emotion): [InstUISettingsProvider] should be able to access the current theme

HerrTopi:
INSTUI-4525

### DIFF
--- a/packages/emotion/src/getTheme.ts
+++ b/packages/emotion/src/getTheme.ts
@@ -52,25 +52,6 @@ import type {
 const getTheme =
   (themeOrOverride: ThemeOrOverride) =>
   (ancestorTheme = {} as BaseTheme) => {
-    try {
-      // If a valid InstUI theme is given, it just returns the theme.
-      if (isBaseTheme(themeOrOverride)) {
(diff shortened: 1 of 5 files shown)


packages/emotion/src/getTheme.ts
```diff
@@ -86,16 +67,40 @@ const getTheme =
       currentTheme = ancestorTheme
     }
 
+    let calcedThemeOrOverride = themeOrOverride
```
joyenjoyer, line 70: I would use a different name, `calced` is used in mathematical or computational contexts. Maybe `resolvedThemeOrOverride` is less confusing?

packages/emotion/src/getTheme.ts
```diff
@@ -86,16 +67,40 @@ const getTheme =
...
+
+    if (typeof calcedThemeOrOverride === 'function') {
```
matyasf, line 72: Can you write a unit test for this new feature?

matyasf (changes requested):
see my comment above

balzss (approved):

PR #1952, opened 2025-04-24
fix(ui-options,ui-menu): improve hover animations

HerrTopi:
TEST PLAN:
check options and menu hover animations. They should look better now. Whatever better means :D 
(on hover the color changes should be instant, on mouse leave it should have a 200ms transition time)

INSTUI-4502


matyasf (approved):
nice, I like it :)

balzss (approved):

PR #1956, opened 2025-04-25
fix(ui-table): fix table crashing in stacked layout when using falsy children

matyasf:
- fix table crashing in stacked layout when using falsy children
- a `Table` example that was using `Children` instead of `React.Children`
- CodeSandboxes crashing because missing imports.
- Make CodeSandbox examples nicer by only importing samples when needed

Fixes INSTUI-4534

To test:
1. try this code in this PR, it should not crash:
```
<Table caption="Test table" layout="stacked">
        <Table.Head>
          {/* @ts-expect-error error is normal here */}
          <Table.Row>
            <Table.Cell>Foo</Table.Cell>
            {}
            {false}
          </Table.Row>
        </Table.Head>
        <Table.Body>
          <Table.Row>
            <Table.RowHeader>1</Table.RowHeader>
            <Table.Cell>The Shawshank Redemption</Table.Cell>
            <Table.Cell>1994</Table.Cell>
            <Table.Cell>9.3</Table.Cell>
          </Table.Row>
          <Table.Row>
            <Table.RowHeader>2</Table.RowHeader>
            <Table.Cell>The Godfather</Table.Cell>
            <Table.Cell>1972</Table.Cell>
            <Table.Cell>9.2</Table.Cell>
          </Table.Row>
        </Table.Body>
      </Table>
```
2. Try to insert invalid children into other places (e.g. text, `{false}`), it should never crash.
3. Open some of our example codes in CodeSandbox, it should only import sample media (e.g. `lorem`) when its used, they should not crash

### DIFF
--- a/packages/__docs__/src/CodeSandboxButton/index.tsx
+++ b/packages/__docs__/src/CodeSandboxButton/index.tsx
@@ -42,25 +42,51 @@ class CodeSandboxButton extends Component<CodeSandboxButtonProps> {
   render() {
     //TODO: we should not import everything everytime
     //somehow conditionally import needed modules
-    const importStatements = `import {useState, useEffect} from "react"
+    const importStatements = `import React, {useState, useEffect, useContext, Children} from "react"
 import ReactDOM from "react-dom"
 import moment from 'moment'
-import {avatarSquare, avatarPortrait, lorem} from "./samplemedia"
-import {placeholderImage} from "./samplemedia"
-import iconExample from "!svg-inline-loader!./heart_lg.svg"
 import 'moment/min/locales'
 `
-    const reactComponentPattern = /<[A-Z]\w+|Icon\w+/gm
-    const neededClasses = this.props.code
-      .match(reactComponentPattern)
-      ?.map((className) => className.replace(/</gm, '').trim())
-
-    if (this.props.code.includes('Calendar')) {
-      neededClasses?.push('Calendar')
+    let importExampleIcon = ''
+    if (this.props.code.includes('iconExample(')) {
+      importExampleIcon =
+        'import iconExample from "!svg-inline-loader!./heart_lg.svg"'
     }
 
-    const uniqueClasses = [...new Set(neededClasses)]
+    // import sample media
+    const neededSampleMedia: string[] = []
+    if (this.props.code.includes('avatarSquare')) {
+      neededSampleMedia.push('avatarSquare')
+    }
+    if (this.props.code.includes('avatarPortrait')) {
+      neededSampleMedia.push('avatarPortrait')
+    }
+    if (this.props.code.includes('lorem.')) {
+      neededSampleMedia.push('lorem')
+    }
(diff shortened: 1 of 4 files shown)


packages/__docs__/src/CodeSandboxButton/index.tsx
```diff
@@ -42,25 +42,51 @@ class CodeSandboxButton extends Component<CodeSandboxButtonProps> {
   render() {
     //TODO: we should not import everything everytime
     //somehow conditionally import needed modules
-    const importStatements = `import {useState, useEffect} from "react"
+    const importStatements = `import React, {useState, useEffect, useContext, Children} from "react"
```
matyasf, line 45: not importing these caused some CodeSandbox examples to crash

packages/__docs__/src/CodeSandboxButton/index.tsx
```diff
@@ -42,25 +42,51 @@ class CodeSandboxButton extends Component<CodeSandboxButtonProps> {
...
+      .match(reactComponentPattern)
+      ?.forEach((className) =>
+        neededClasses.push(className.replace(/</gm, '').trim())
+      )
+    if (this.props.code.includes('Calendar')) {
+      neededClasses.push('Calendar')
+    }
+    this.props.code
+      .match(/[A-Z]\w+Context/gm)
+      ?.forEach((className) => neededClasses.push(className))
```
matyasf, line 88: now CodeSandbox examples are not littered with unused sample imports

packages/ui-table/src/Table/Head/index.tsx
```diff
@@ -117,7 +117,7 @@ class Head extends Component<TableHeadProps> {
     let count = 0
     Children.forEach(firstRow.props.children, (grandchild) => {
       count += 1
-      if (!grandchild.props) return
+      if (!grandchild?.props) return // grandchild can be false
```
matyasf, line 120: This is the actual fix in the `Table` component

ToMESSKa (approved):

packages/__docs__/src/CodeSandboxButton/index.tsx
```diff
@@ -42,25 +42,51 @@ class CodeSandboxButton extends Component<CodeSandboxButtonProps> {
   render() {
     //TODO: we should not import everything everytime
```
balzss, line 43: wouldn't this pr make this comment obsolete?

balzss (approved):

balzss (approved):

packages/__docs__/src/CodeSandboxButton/index.tsx
matyasf: not fully, we are still importing `moment`, some React functions that might not be needed every time. Some areas to improve:
- Remove moment import if its not used, do not add to the dependencies it if its not used.
- The imports from Instructure are pretty dumb, e.g. if we define a custom class it imports it from `@instructure/ui` (this does not cause an issue though)

PR #1959, opened 2025-05-05
Ai avatar

HerrTopi:
TEST_PLAN:
Check out the ticket and its design. Also check out the new avatar docs.

INSTUI-4538


packages/ui-avatar/src/Avatar/styles.ts
```diff
@@ -48,47 +49,108 @@ const generateStyle = (
...
+    size,
+    color,
+    hasInverseColor,
+    shape,
+    src,
+    showBorder,
+    variant
+  } = params
+
+  // NOTE: this is needed due to design changes. The size of the component is calculated from "em" which means it is
```
balzss, line 63: can you please add that this is temporary and only here for compatibility with old versions and it will be cleaned up when the component is redesigned?

packages/ui-avatar/src/Avatar/README.md
```diff
@@ -4,6 +4,34 @@ describes: Avatar
 
 The avatar component can be used to display a user's avatar. When an image src is not supplied the user's initials will display.
 
+## Variant
+
+Avatar has a variant prop, which currently only toggles it between `ai` and `default` behavior. `ai` is a preset where you can only change the `size` and `margin` visual props, all others are preset. In the following example, there are the `ai` variants.
+
+### ai
```
balzss, line 11: I'd capitalize this title to `AI`

balzss: and also change the order. so default is first, but this is not that important

packages/ui-avatar/src/Avatar/README.md
```diff
@@ -4,6 +4,34 @@ describes: Avatar
...
+
+```js
+---
+type: example
+readonly: true
+---
+<div>
+   <View display="block" padding="small medium">
+    <Avatar variant="ai" size="xx-small" margin="0 small 0 0" />
```
balzss, line 20: please use the new spacing tokens, `small` is deprecated and will be removed

balzss (changes requested):

matyasf (changes requested):
This code does not render the AI icon:

```
<Avatar variant="ai" size="xx-large" src={avatarSquare}/>
```

This code renders it in a weird shape:

```
<Avatar variant="ai" size="xx-large" shape/>
```

it would be nice to fix these if its not hard

balzss (approved):

joyenjoyer (approved):

PR #1961, opened 2025-05-06
docs(ui-select): fix_focus_input_function_in_select_examples

ToMESSKa:
INSTUI-4539

IISSUE:
- the line handling the input blur isn't in the right place
- handling the inputRef with states is not a good practice

TEST PLAN:
- check the class and functional examples on iPad/iPhone with VoiceOver, the focus should be returned to the input field after selecting an option
- functional examples should handle inputRef with useRef

### DIFF
--- a/packages/ui-select/src/Select/README.md
+++ b/packages/ui-select/src/Select/README.md
@@ -29,8 +29,8 @@ describes: Select
     inputElement = null
 
     focusInput = () => {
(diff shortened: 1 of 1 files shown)


matyasf (approved):
looks good!

HerrTopi (approved):
Good job! Works as expected, at least on iPhone

PR #1962, opened 2025-05-07
fix(ui-alerts): add variantScreenReaderLabel prop to Alert to improve screenreader usability

balzss:
INSTUI-4533

test plan:

- open the alert component's page in the docs and go through with screenreader -> it should announce the type of the alert

### DIFF
--- a/packages/ui-alerts/src/Alert/props.ts
+++ b/packages/ui-alerts/src/Alert/props.ts
@@ -47,6 +47,10 @@ type AlertOwnProps = {
    */
   variant?: 'info' | 'success' | 'warning' | 'error'
   /**
+   * How the screen reader should announce the alert variant. While the `variant` prop sets the color and icon for the alert component, this label should be a textual representation of that information. So e.g. if the variant is `info`, this label could be "Information," or "Information alert,". Note the `,` at the end of the label which helps the screenreader to be more natural sounding.
+   */
+  variantScreenReaderLabel?: string
+  /**
    * Function that returns the DIV where screenreader alerts will be placed.
    */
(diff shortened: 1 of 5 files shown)


packages/ui-alerts/src/Alert/props.ts
```diff
@@ -46,6 +46,10 @@ type AlertOwnProps = {
    * Determines color and icon
    */
   variant?: 'info' | 'success' | 'warning' | 'error'
+  /**
+   * How the screen reader should announce the alert variant. While the `variant` prop sets the color and icon for the alert component, this label should be a textual representation of that information. So e.g. if the variant is `info`, this label could be "Information:" or "Information alert:". Note the `:` at the end of the label which helps the screenreader to be more natural sounding.
```
ToMESSKa, line 50: @balzss JAWS unfortunately reads the ':' as 'colon', but I checked that ',' works as a pause with VoiceOver and JAWS too, so I would recommend that instead.

ToMESSKa (changes requested):
@balzss only the first example has the new prop, maybe it would be a good idea to add it to the rest of the examples too

matyasf (changes requested):
`variantScreenReaderLabel` does nothing when added to the `liveregion` example in VoiceOver, as I see the text is not added to the `liveRegion`.

Also please add some text about this new prop, e.g. to the info boxes

packages/ui-alerts/src/Alert/props.ts
balzss: done

ToMESSKa (approved):

PR #1965, opened 2025-05-08
fix(ui-tooltip,ui-popover): put aria-expanded on trigger element

HerrTopi:
INSTUI-4544


packages/ui-drilldown/src/Drilldown/index.tsx
```diff
@@ -121,7 +121,8 @@ class Drilldown extends Component<DrilldownProps, DrilldownState> {
     shouldReturnFocus: true,
     withArrow: true,
     offsetX: 0,
-    offsetY: 0
+    offsetY: 0,
+    shouldSetARIAExpanded: true
```
matyasf: I would use camel case for this variable name (e.g. `shouldSetAriaExpanded`, we are never writing `ARIA` in all caps

packages/ui-popover/src/Popover/props.ts
```diff
@@ -283,6 +283,10 @@ type PopoverOwnProps = {
    * Only applies to a Popover without an arrow.
    */
   borderWidth?: BorderWidth
+  /**
+   * If `true` (default), aria-expanded is calculated automatically on the trigger
```
matyasf: I'd rewrite this to:

If `true` (default), then the `aria-expanded` prop is added to the trigger. If its supplied via the `aria-expanded` prop then it takes the given value, otherwise its calculated automatically based on whether the content is shown.

(same for Drilldown)

packages/ui-popover/src/Popover/index.tsx
```diff
@@ -468,7 +469,7 @@ class Popover extends Component<PopoverProps, PopoverState> {
         }
       }
 
-      if (shouldContainFocus) {
+      if (this.props.shouldSetARIAExpanded) {
         // only set aria-expanded if popover can contain focus
         expanded = this.shown ? 'true' : 'false'
```
matyasf, line 475: the comment on line 478 should say "// wants to remove/override aria-expanded even when shouldSetAriaExpanded"

packages/ui-drilldown/src/Drilldown/index.tsx
```diff
@@ -1553,6 +1556,7 @@ class Drilldown extends Component<DrilldownProps, DrilldownState> {
             this._trigger = el
           },
           'aria-haspopup': this.props.role,
+          'aria-expanded': 'true',
```
matyasf: As I see this does here nothing. If `shouldSetARIAExpanded` is `false` , `aria-expanded` is removed, if its true then the value on the trigger is overwritten

PR #1968, opened 2025-05-12
fix(ui-top-nav-bar,ui-buttons): display a focus ring in TopNavBar if a button has a Popover open

matyasf:
When a menu is active, it should be indicated as the “active” element for a11y reasons

To test: The desktop layout TopNavBar should show an outline around a button if it has some popover, like a submenu or a popup open. There should be no other change

Fixes part of INSTUI-4323


ToMESSKa (changes requested):
in the smallViewport version, the hambuger/X menu icon does not have a focus ring when the drilldown is open

joyenjoyer (approved):

PR #1976, opened 2025-05-15
fix(ui-byline): remove figure and figcaption

joyenjoyer:
INSTUI-4549

Test plan:

1. Go through a Byline example in the docs app with NVDA and VoiceOver, make sure no figure of figcaption is read by SR apps.

### DIFF
--- a/packages/ui-byline/src/Byline/index.tsx
+++ b/packages/ui-byline/src/Byline/index.tsx
@@ -82,12 +82,11 @@ class Byline extends Component<BylineProps> {
         {...passthroughProps}
         elementRef={this.handleRef}
         css={this.props.styles?.byline}
-        as="figure"
         margin={this.props.margin}
         maxWidth={this.props.styles?.maxWidth}
       >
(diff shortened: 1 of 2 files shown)


packages/ui-byline/src/Byline/index.tsx
```diff
@@ -82,12 +82,11 @@ class Byline extends Component<BylineProps> {
         {...passthroughProps}
         elementRef={this.handleRef}
         css={this.props.styles?.byline}
-        as="figure"
         margin={this.props.margin}
         maxWidth={this.props.styles?.maxWidth}
       >
         <div css={this.props.styles?.figure}>{this.props.children}</div>
-        <figcaption css={this.props.styles?.caption}>
```
joyenjoyer, line 90: This was introduced in 2016. The use of `<figure>` and `<figcaption>` (which NVDA announces unnecessarily) is no longer needed for proper screen reader support.

HerrTopi (approved):

PR #1991, opened 2025-05-27
fix(ui-time-select,ui-simple-select,ui-select): add missing keyboard …

ToMESSKa:
…interactions and fix duplicate SR announcements

INSTUI-4526

**ISSUE:**
- in Select, SimpleSelect and TimeSelect, while navigation through the options, an option was announced twice: 
  - for VoiceOver, the cause was that while navigating the options with keyboard, the input value updates too, causing the screen reader to announce the changed input value as well
  - for NVDA the cause was that that some of the examples included an Alert, which caused reading option twice
- in Select, SimpleSelect and TimeSelect, when no option is selected and user presses down arrow, the focus does not jump to the first option as recommended by the official APG patterns
- in Select, SimpleSelect and TimeSelect, when no option is selected and user presses up arrow, the focus does not jump to the last option as recommended by the official APG patterns
- for Select, these keyboard interactions were implemented in the examples, while for SimpleSelect and TimeSelect, the source code itself was modified

**TEST PLAN:**

screenreader annoucements:
- go through every example in Select, SimpleSelect and TimeSelect, VoiceOver and NVDA should announce an option only once when navigating with down and up arrow buttons: eg 'Alaska not selected' instead of 'Alaska, Alaska not selected'

keyboard interactions:
- in Select, SimpleSelect and TimeSelect, set the initially selected option to none in the example if needed in order to make the input empty e.g.:
  - In Select set the intial inputValue and selectedOptionId state values to null
  - in SimpleSelect/TimeSelect setting  value or defaultValue to value=' ' or defaultValue=' '
- go through the example with keyboard interactions only
- when no option is selected initially and user presses down arrow, the listbox should open and the first option should be highlighted and focused (instead of not focusing/highlighting anything)
- when no option is selected initially and user presses up arrow, the listbox should open and the last option should be highlighted and focused (instead of not focusing/highlighting anything)
- in SimpleSelect it should work with grouped options and mixed grouped and non-grouped options e.g.: only the last option is a non-grouped option

### DIFF
--- a/packages/ui-select/src/Select/README.md
+++ b/packages/ui-select/src/Select/README.md
@@ -40,14 +40,27 @@ describes: Select
     }
 
     handleShowOptions = (event) => {
+      const { options } = this.props
+      const { inputValue, selectedOptionId } = this.state
+
       this.setState({
         isShowingOptions: true
       })
+      if (inputValue || selectedOptionId || options.length === 0) return
+
+      switch (event.key) {
+        case 'ArrowDown':
+          return this.handleHighlightOption(event, { id: options[0].id })
+        case 'ArrowUp':
+          return this.handleHighlightOption(event, {
+            id: options[options.length - 1].id
+          })
+      }
     }
 
     handleHideOptions = (event) => {
       const { selectedOptionId } = this.state
-      const option = this.getOptionById(selectedOptionId).label
+      const option = this.getOptionById(selectedOptionId)?.label
       this.setState({
         isShowingOptions: false,
         highlightedOptionId: null,
@@ -68,17 +81,17 @@ describes: Select
       const nowOpen = !this.state.isShowingOptions
         ? `List expanded. ${optionsAvailable}`
         : ''
-      const option = this.getOptionById(id).label
+      const option = this.getOptionById(id)?.label
       this.setState((state) => ({
         highlightedOptionId: id,
-        inputValue: event.type === 'keydown' ? option : state.inputValue,
+        inputValue: state.inputValue,
         announcement: `${option} ${nowOpen}`
       }))
     }
 
     handleSelectOption = (event, { id }) => {
       this.focusInput()
-      const option = this.getOptionById(id).label
+      const option = this.getOptionById(id)?.label
       this.setState({
         selectedOptionId: id,
         inputValue: option,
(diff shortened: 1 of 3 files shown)


packages/ui-select/src/Select/README.md
```diff
@@ -205,15 +221,15 @@ describes: Select
       const nowOpen = !isShowingOptions
         ? `List expanded. ${optionsAvailable}`
         : ''
-      const option = getOptionById(id).label
+      const option = getOptionById(id)?.label
       setHighlightedOptionId(id)
-      setInputValue(event.type === 'keydown' ? option : inputValue)
```
ToMESSKa, line 210: with VoiceOver, while navigating the options with keyboard, the input value updates too, causing the screen reader to announce the changed input value and the option as well. This behavior (the input changing while just navigation through options) is impemented in any official APG examples

packages/ui-simple-select/src/SimpleSelect/index.tsx
```diff
@@ -264,6 +289,27 @@ class SimpleSelect extends Component<SimpleSelectProps, SimpleSelectState> {
     if (typeof this.props.onShowOptions === 'function') {
       this.props.onShowOptions(event)
     }
+    const keyboardEvent = event as React.KeyboardEvent
```
matyasf: This code is a bit misleading, since one might think that this is always a `keyboardEvent`. This is not the case, this can be either a `MouseEvent` or a `KeyboardEvent`. I think the best way would be to check the `event.type`, e.g. 

```
if (event.type.startsWith('key')) {
```

BTW we should type these `onRequest...` functions properly, not just to a generic `SyntheticEvent`

packages/ui-time-select/src/TimeSelect/index.tsx
```diff
@@ -383,6 +383,26 @@ class TimeSelect extends Component<TimeSelectProps, TimeSelectState> {
       highlightedOptionId: this.state.selectedOptionId
     })
     this.props.onShowOptions?.(event)
+
+    const keyboardEvent = event as React.KeyboardEvent
```
matyasf: Same issue as the code in SimpleSelect

packages/ui-simple-select/src/SimpleSelect/index.tsx
ToMESSKa: I fixed this, check please.

packages/ui-time-select/src/TimeSelect/index.tsx
ToMESSKa: I fixed this, check please.

balzss (approved):

PR #1992, opened 2025-05-28
fix(ui-view): fix focus outline showing when setting a global :hover CSS rule for outline

matyasf:
Currently we are animating the focus ring on focus, on the normal state its a CSS outline with 0 alpha. If the outline is overridden irt will be animated because its different from the global defaulrt.


packages/ui-view/src/View/styles.ts
```diff
@@ -284,40 +284,35 @@ const getFocusStyles = (props: ViewProps, componentTheme: ViewTheme) => {
...
+    ...(shouldAnimateFocus && {
+      transition: 'outline-color 0.2s, outline-offset 0.25s'
+    }),
+    outlineStyle: componentTheme.focusOutlineStyle,
+    outlineWidth: componentTheme.focusOutlineWidth,
+    ...outlineOffsetAndColor,
+    '&:hover': {
+      // apply the same style so it's not overridden by some global style
+      ...outlineOffsetAndColor
+    },
```
matyasf, line 308: this is the actual fix, others are just refactor

packages/ui-view/src/View/styles.ts
```diff
@@ -284,40 +284,35 @@ const getFocusStyles = (props: ViewProps, componentTheme: ViewTheme) => {
...
-        : {})
-    }
+    ...(shouldAnimateFocus && {
+      transition: 'outline-color 0.2s, outline-offset 0.25s'
+    }),
+    ...outlineStyle,
+    '&:hover, &:active': {
+      // apply the same style so it's not overridden by some global style
+      ...outlineStyle
+    },
```
matyasf, line 308: This is the actual fix

balzss (approved):

joyenjoyer (approved):

PR #1997, opened 2025-05-30
refactor(many): rename tests folders and clean up old theme tests, ch…

git-nandor:
I removed the old Karma-based component theme tests and renamed the test folders from __new-tests__ to __tests__. The test folders that are still waiting to be migrated have been temporarily renamed to __old-tests__. I also updated the config files and the workflow accordingly. Additionally, I removed the codemods tests.

Test:
Please verify that the Vitest-based tests run correctly, and check that the workflow and the package.json scripts are set up properly.


matyasf (approved):
looks good, nice work!

balzss (approved):
everything looks good and runs as expected. good job, nice to see this after so much effort :)

PR #1999, opened 2025-06-02
fix(ui-date-input): make DateInput2 date parsing work in every locale and timezone

balzss:
INSTUI-4571

test plan:

- check all examples for dateinput2, they should work as expected
- input a date manually (with keyboard) for each example and blur the input -> it should parse the date successfully
- placeholder hints should not have any number in them, just `D`, `M` and `Y` letters
- use chrome devtools and change your location (3 dots -> more tools -> sensors -> location)
- check the component examples again with different locales
- check if manually changing locale and timezone works as expected

### DIFF
--- a/packages/ui-date-input/src/DateInput2/index.tsx
+++ b/packages/ui-date-input/src/DateInput2/index.tsx
@@ -25,7 +25,6 @@
 import {
   useState,
   useEffect,
-  useContext,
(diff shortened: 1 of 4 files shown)


joyenjoyer (approved):
tested all examples, worked fine, gj

PR #2000, opened 2025-06-02
fix(many): fix "not a valid selector" exception when an option ID contains quotes

matyasf:
The issue is that we are letting users choose any string as an ID, and the querySelector method only accepts CSS safe IDs. Luckily there is a built in method to escape stuff that it does not like

Fixes INSTUI-4570

To test:

- This code should not throw errors:

```
const options = ["Option 1", 'Option 2 with " inside']

render(
  <SimpleSelect renderLabel="Uncontrolled Select">
    {options.map((option) => (
      <SimpleSelect.Option id={option} value={option}>
        {option}
      </SimpleSelect.Option>
    ))}
  </SimpleSelect>
)
```

- In the 'Video Settings' Drilldown example make an option's ID contain a double quote ("). when hovering over it the example should not crash

### DIFF
--- a/packages/ui-drilldown/src/Drilldown/__new-tests__/Drilldown.test.tsx
+++ b/packages/ui-drilldown/src/Drilldown/__new-tests__/Drilldown.test.tsx
@@ -104,6 +104,33 @@ describe('<Drilldown />', () => {
       expect(options.length).toBe(0)
       expect(notAllowedChild).not.toBeInTheDocument()
     })
+
+    it('should not crash for weird option ids', async () => {
+      const onSelect = vi.fn()
+      const weirdID = 'some"_weird!@#$%^&*()\\|`id'
+      render(
+        <Drilldown rootPageId="page0" onSelect={onSelect}>
+          <Drilldown.Page id="page0">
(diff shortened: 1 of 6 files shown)


packages/ui-drilldown/src/Drilldown/__new-tests__/Drilldown.test.tsx
```diff
@@ -104,6 +104,34 @@ describe('<Drilldown />', () => {
...
+
+    it('should not crash for weird option ids', async () => {
+      const onSelect = vi.fn()
+      const weirdID = 'some"_weird!@#$%^&*()\\|`id'
+      render(
+        <Drilldown rootPageId="page0" onSelect={onSelect}>
+          <Drilldown.Page id="page0">
+            {data.map((option) => (
+              <Drilldown.Option
+                //id={option.id}
```
ToMESSKa: what's the point of this comment?

matyasf: this was a mistake, I removed it

git-nandor (approved):

PR #2005, opened 2025-06-03
feat(ui-icons): wip: compare icon

balzss:


packages/ui-scripts/lib/icons/generate-react-components.js
```diff
@@ -22,7 +22,7 @@
  * SOFTWARE.
  */
 import fs from 'fs'
-import svg2jsx from 'svg-to-jsx'
+import { svg2jsx } from './svg2jsx.js'
```
joyenjoyer, line 25: svg-to-jsx package should be removed from package.json if we don't use it anymore.

packages/ui-scripts/lib/icons/svg2jsx.js
```diff
@@ -0,0 +1,109 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+// this is a custom svg to jsx converter function which should be able to handle all of our svgs with some assumptions:
+// - the root tag is always <svg>
+// - there are no inline styles
+// - there are no classnames
+// - there is no `for` of `htmlFor` attr
+export function svg2jsx(svgString) {
```
joyenjoyer, line 30: Please write a test for this function.

balzss: I agree we should test this, I've opened a ticket to add tests for the ui-scripts package

docs/contributor-docs/adding-icons.md
```diff
@@ -94,3 +87,9 @@ type: code
   Draw inside the lines.
 
 - Fill the space edge-to-edge as much as possible. The build process will add margins as needed.
+
+- Don't use inline styles
+
+- Don't use `class` or `for` attributes
+
+- Always have `<svg>` as the root tag
```
matyasf, line 95: This was a limitation with the old solution too, right?

balzss: probably yes, but it is definitely a "limitation" after this change

joyenjoyer (approved):

PR #2009, opened 2025-06-03
chore: upgrade jsdoc to latest to support node 23+

HerrTopi:
TEST_PLAN: check if everything works in docs with latest(24.x) and older(20.x) node

INSTUI-4574

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,5 +127,8 @@
   },
   "browserslist": [
     "extends @instructure/browserslist-config-instui"
-  ]
+  ],
+  "dependencies": {
+    "clipboardy": "^4.0.0"
+  }
 }
--- a/packages/__docs__/webpack.config.mjs
+++ b/packages/__docs__/webpack.config.mjs
@@ -59,15 +59,15 @@ const config = merge(baseConfig, {
       directory: outputPath,
     },
     host: '0.0.0.0',
-    onListening: function () {
+    onListening: async function () {
       // devServer is watching source files by default and hot reloading the docs page if they are changed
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -127,5 +127,8 @@
   },
   "browserslist": [
     "extends @instructure/browserslist-config-instui"
-  ]
+  ],
+  "dependencies": {
+    "clipboardy": "^4.0.0"
```
balzss, line 132: cant this be a dev dep?

balzss: or is this even needed?

packages/__docs__/webpack.config.mjs
```diff
@@ -59,15 +59,15 @@ const config = merge(baseConfig, {
...
       // devServer is watching source files by default and hot reloading the docs page if they are changed
       // however markdown files (i.e. README.md) need to be recompiled hence the need for chokidar
       const paths = globbySync(['packages/**/*.md', 'docs/**/*.md'], { cwd: '../../' }).map(p => '../../' + p)
       chokidar
         .watch(paths)
-        .on('change', (evt) => {
+        .on('change', async (evt) => {
           const fullPath = resolvePath(import.meta.dirname, evt)
-          processSingleFile(fullPath)
+          await processSingleFile(fullPath)
```
matyasf, line 70: Both the Webpack onListening https://webpack.js.org/configuration/dev-server/#devserveronlistening
and Chokidar's on https://github.com/paulmillr/chokidar are not async.

I think this should work fine for 99% of use cases (unless you somehow modify the same file in very quick succession?) but please put a comment to both functions, that these do not wait for the promise to return.

PR #2011, opened 2025-06-04
fix(ui-form-field): make JAWS read input field labels and form error …

ToMESSKa:
…separately

INSTUI-4568

**ISSUE:**
- JAWS announces input field labels and form error messages together if form error is followed by a label

**TEST PLAN:**
- go to the first example in FormFieldGroup
- with JAWS, go through the example using down and up arrow keys to read 'Invalid name' and 'Favorite Side Dish' texts
- JAWS should announce these texts separately: first 'Invalid name', then by pressing the down arrow key 'Favorite Side Dish' (previously JAWS announced them together)
- check the example with other screenreaders too, they should have no issues announcing them 
- compare the error messages to the current version in the 'Form Errors', 'FormField' and 'FormFieldGroup' pages, there should be no visual changes


matyasf (approved):
works nicely :)

PR #2012, opened 2025-06-04
fix(ui-buttons): make Button have a focus ring in Safari

ToMESSKa:
INSTUI-4527

**ISSUE:**
- In Safari, the same BaseButton or Button component doesn't show a focus ring, whereas the ring does appear in Chrome and other browsers - this is due to a missing 0 tabindex, which Safari needs
- Modal trigger button does not have a focus ring after closing the Modal due to this

**TEST PLAN:**
- make sure you have the latest Safari version
- open BaseButton and Button pages **in Safari**
- the first few examples should get a focus ring when they are clicked on - unlike in the latest InstUI release in Safari
- they should have a tabIndex=0 in Safari but not in other browsers
- open the first example of Modal **in Safari,** after closing it, the Modal should have a focus ring - unlike in the latest InstUI release in Safari

### DIFF
--- a/packages/ui-buttons/src/BaseButton/index.tsx
+++ b/packages/ui-buttons/src/BaseButton/index.tsx
@@ -36,6 +36,7 @@ import { isActiveElement } from '@instructure/ui-dom-utils'
 import { hasVisibleChildren } from '@instructure/ui-a11y-utils'
 import { View } from '@instructure/ui-view'
 import type { ViewProps } from '@instructure/ui-view'
+import { isSafari } from '@instructure/ui-utils'
 
 import { withStyle } from '@instructure/emotion'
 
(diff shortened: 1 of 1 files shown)


packages/ui-buttons/src/BaseButton/index.tsx
```diff
@@ -36,6 +36,7 @@ import { isActiveElement } from '@instructure/ui-dom-utils'
 import { hasVisibleChildren } from '@instructure/ui-a11y-utils'
 import { View } from '@instructure/ui-view'
 import type { ViewProps } from '@instructure/ui-view'
+import * as utils from '@instructure/ui-utils'
```
matyasf: please dont use `*` imports, they might break in Vite and also increase bundle size (because it always imports everything)

ToMESSKa: Thanks, i fixed this

PR #2013, opened 2025-06-05
fix(ui-buttons): remove underline from disabled Button with href

ToMESSKa:
INSTUI-4540

### DIFF
--- a/packages/ui-buttons/src/BaseButton/styles.ts
+++ b/packages/ui-buttons/src/BaseButton/styles.ts
@@ -56,7 +56,7 @@ const generateStyle = (
     isCondensed
   } = props
 
-  const { isDisabled, hasOnlyIconVisible } = state
+  const { isDisabled, hasOnlyIconVisible, isEnabled } = state
 
   const shapeVariants = {
     circle: { borderRadius: '50%' },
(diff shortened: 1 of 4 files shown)


packages/ui-buttons/src/BaseButton/styles.ts
```diff
@@ -401,26 +401,30 @@ const generateStyle = (
...
+          },
+          '&:focus': {
+            textDecoration: 'none'
+          },
+          '&:active > [class$=-baseButton__content]':
+            colorVariants[color!].active,
+          '&:hover > [class$=-baseButton__content]': colorVariants[color!].hover
+        }
+      : {
+          textDecoration: 'none'
```
matyasf: Please add here these:
```
          label: 'baseButton',
          appearance: 'none',
```

Label is needed, so it's easier to identity what this is in the DOM (also unit tests might use it), appearance: none might prevent some custom CSS to override things here (although I could not override it)

matyasf (changes requested):
see my comment

packages/ui-buttons/src/BaseButton/styles.ts
ToMESSKa: @matyasf I added them

matyasf (approved):

balzss (approved):

PR #2019, opened 2025-06-12
chore: announce search results in side navigation

ToMESSKa:
INSTUI-4243

**ISSUE:**
- search results in documentation's side navigation are not announced

**TEST PLAN:**
- navigate to the "Find…" search bar in the side navigation using different screen readers
- type a search term (e.g., "color" or "select")
- the screen reader should:
  - announce the number of results available
  - clearly indicate when no results are available


matyasf (approved):
looks good, I've tested with VoiceOver on OSX, JAWS and NVDA on Windows 11

joyenjoyer (approved):
gj

PR #2022, opened 2025-06-13
feat(ui-simple-select,ui-select): add layout prop to Select and Simpl…

ToMESSKa:
…eSelect

INSTUI-4482

**ISUUE:**
- Select and SimpleSelect do not have a layout prop, so the label in SimpleSelect cannot be placed to the right/left of the input field

**TEST PLAN:**
- check the examples in Select and SimpleSelect by adding the layout='inline' prop
- the label should be placed to the right/left of the input field (depending on text direction) the same way as for example in TextInput

### DIFF
--- a/packages/ui-select/src/Select/props.ts
+++ b/packages/ui-select/src/Select/props.ts
@@ -256,6 +256,14 @@ type PropsFromTextInput = {
    * when available space is exceeded.
    */
   shouldNotWrap?: boolean
+
+  /**
+   * In `stacked` mode the input is below the label.
+   *
+   * In `inline` mode the input is to the right/left (depending on text direction) of the label,
+   * and the layout will look like `stacked` for small screens.
+   */
+  layout?: 'stacked' | 'inline'
 }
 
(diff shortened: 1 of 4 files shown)


packages/ui-select/src/Select/props.ts
```diff
@@ -256,6 +256,12 @@ type PropsFromTextInput = {
    * when available space is exceeded.
    */
   shouldNotWrap?: boolean
+
+  /**
+   * In `stacked` mode the input is below the label, in `inline` mode the
+   * container is to the right/left (depending on text direction)
+   */
```
matyasf, line 265: better text suggestion (use the same for SimpleSelect too):

In `stacked` mode the input is below the label.
In `inline` mode the input is to the right/left (depending on text direction) of the label, and the layout will look like `stacked` for small screens

ToMESSKa: @matyasf thanks, I added this to both.

balzss (approved):

PR #2023, opened 2025-06-13
fix(ui-alerts): fix Alert border radius override and add new prop for…

ToMESSKa:
… custom icon

INSTUI-4572

**ISSUE:**
- overriding the border radius of Alert results the icon's border not changing
- the icon cannot be customized

**TEST PLAN:**

- go over the examples in Alert by overriding borderRadius e.g.:
```
themeOverride={{
    borderRadius: '1rem'
  }}
```
- the icons border radius should also change (previously is was unaffected)
- it should work in RTL too
- try adding a custom icon by using the renderCustomIcon prop e.g.:
```
renderCustomIcon={()=> <IconUserLine />}
```
- the icon should change

### DIFF
--- a/packages/ui-alerts/src/Alert/styles.ts
+++ b/packages/ui-alerts/src/Alert/styles.ts
@@ -95,6 +95,9 @@ const generateStyle = (
       justifyContent: 'center',
       fontSize: '1.125rem',
       borderRight: `${componentTheme.borderWidth} ${componentTheme.borderStyle}`,
+      margin: -1,
+      borderStartStartRadius: componentTheme.borderRadius,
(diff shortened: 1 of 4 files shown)


packages/ui-alerts/src/Alert/styles.ts
```diff
@@ -95,6 +95,9 @@ const generateStyle = (
       justifyContent: 'center',
       fontSize: '1.125rem',
       borderRight: `${componentTheme.borderWidth} ${componentTheme.borderStyle}`,
+      margin: -1,
```
ToMESSKa, line 98: this is needed to make a white line between the Alert's and the icon's border to disappear

PR #2027, opened 2025-06-16
feat(ui-pagination): add prop to customize screenreader label on buttons

matyasf:
The new prop, called `screenReaderLabelPageButton` allows to customize screenreader messages on the page buttons (e.g. 1,2,3,4...)

To test: Add this new prop to some of the new examples, and test with various screenreaders, that the given text is announced not the one on the button

INSTUI-4602

### DIFF
--- a/packages/ui-pagination/src/Pagination/PaginationButton/index.tsx
+++ b/packages/ui-pagination/src/Pagination/PaginationButton/index.tsx
@@ -69,6 +69,9 @@ class PaginationButton extends Component<PaginationPageProps> {
           {...props}
           aria-current={this.props.current ? 'page' : undefined}
           elementRef={this.handleRef}
+          {...(this.props.screenReaderLabel
+            ? { 'aria-label': this.props.screenReaderLabel }
+            : {})}
         >
           {this.props.children}
         </BaseButton>
--- a/packages/ui-pagination/src/Pagination/README.md
+++ b/packages/ui-pagination/src/Pagination/README.md
@@ -63,9 +63,10 @@ The pagination component provides props to handle most of the pagination use-cas
   render(<Example />)
   ```
 
-You can set any `totalPageNumber`, the component can handle it easily. Furthermore, you can set
-`siblingCount`, which indicates how many pages are visible on either side of the `currentPage` and the
-`boundaryCount`, which indicates how many pages are visible in the beginning and end.
+You can set any `totalPageNumber`, the component can handle it easily.\
+Furthermore, you can set `siblingCount`, which indicates how many pages are visible on either side of the `currentPage` and the
+`boundaryCount`, which indicates how many pages are visible in the beginning and end.\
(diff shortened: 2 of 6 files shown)


packages/ui-pagination/src/Pagination/README.md
```diff
@@ -87,6 +88,9 @@ You can set any `totalPageNumber`, the component can handle it easily. Furthermo
           onPageChange={(nextPage) => this.setState({ currentPage: nextPage })}
           siblingCount={3}
           boundaryCount={2}
+          screenReaderLabelPageButton={(currentPage, totalPageNumber) =>
+            `Page ${currentPage} of ${totalPageNumber}`
```
ToMESSKa, line 92: add this please to the functional example too

ToMESSKa (changes requested):

packages/ui-pagination/src/Pagination/index.tsx
```diff
@@ -332,6 +332,12 @@ class Pagination extends Component<PaginationProps> {
           key={i}
           onClick={() => this.handleNavigation(i, currentPage)}
           current={i === currentPage}
+          {...(this.props.screenReaderLabelPageButton && {
+            screenReaderLabel: this.props.screenReaderLabelPageButton(
+              i,
+              this.props.totalPageNumber!
+            )
+          })}
```
HerrTopi: This would be cleaner if you'd use the `...(predicate?{}:{})` pattern.  The current pattern with the && could result in undesirable behaviour. 
When the predicate is false, the result will be the boolean `false`. Which in this case would not matter much, but `...false` is not an intended expression

packages/ui-pagination/src/Pagination/README.md
matyasf: done

packages/ui-pagination/src/Pagination/index.tsx
matyasf: done

packages/ui-pagination/src/Pagination/PaginationButton/index.tsx
```diff
@@ -69,6 +69,9 @@ class PaginationButton extends Component<PaginationPageProps> {
           {...props}
           aria-current={this.props.current ? 'page' : undefined}
           elementRef={this.handleRef}
+          {...(this.props.screenReaderLabel && {
+            'aria-label': this.props.screenReaderLabel
+          })}
```
HerrTopi: Sorry, I missed one. Could you do the same pred?{}:{} pattern here as well?

HerrTopi (approved):

packages/ui-pagination/src/Pagination/PaginationButton/index.tsx
matyasf: done

ToMESSKa (approved):

PR #2040, opened 2025-06-17
test(ui-codemods): add back tests for v10 codemods with a nice Vitest config

matyasf:
Update Vitest config so it can handle both Node and web tests, make codemod tests with less magic


packages/ui-codemods/lib/__node_tests__/__testfixtures__/colors.input.ts
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable */
```
matyasf, line 1: the test fixtures in this PR are revived from our old code, they don't need to be reviewed

packages/ui-codemods/lib/helpers/replaceDeprecatedProps.ts
```diff
@@ -37,6 +37,7 @@ import type {
 import { printWarning } from './codemodHelpers'
 
 /**
+ * DEPRECATED, this should be deleted in the future
```
matyasf, line 40: This file is only used for very old codemods (I think v7->v8), they can be deleted in a later PR

ToMESSKa (changes requested):
if I run `npm run test:vitest-watch` and then modify the codemod test files, the test do not fail and do not seem to run again despite being in watch mode.
They fail if I modify them the run 'npm run test:vitest'.

packages/ui-codemods/lib/__node_tests__/__testfixtures__/updateV10Breaking/colors.input.ts
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable */
```
matyasf, line 1: The test fixtures are revived from deleted files, no need to review them

packages/ui-codemods/lib/__node_tests__/testHelpers.ts
```diff
@@ -0,0 +1,77 @@
...
+import { Transform } from 'jscodeshift'
+
+/**
+ * A helper function to run codemods. For it work properly some conventions
+ * must be kept:
+ * - Fixtures must be in the `./__testfixtures__/[codemod name]/` subfolder
+ * - Fixtures must be named `XY.input.[ts|tsx|js]` and `XY.output.[ts|tsx|js]`
+ * @param codemod The codemod to run
+ */
+export function runTest(codemod: Transform) {
```
matyasf, line 38: This is the main way of running codemod tests

vitest.config.mts
```diff
@@ -25,12 +25,45 @@
...
+        // matches any file inside the __testfixtures__ directory
+        pattern: /^.*\/ui-codemods\/.*\/__testfixtures__\/.*$/,
+        // file is the full path. e.g.  /Users/MyUser/code/instructure-ui/packages/ui-codemods/lib/__node_tests__/__testfixtures__/colors.input.ts
+        testsToRun: (file, _match) => {
+          const dirName = path.basename(path.dirname(file))
+          // reruns all tests that match the directory name of the test fixture
+          return `packages/ui-codemods/lib/__node_tests__/${dirName}.test.ts`
+        }
+      }
+    ],
```
matyasf, line 44: This tricky code allows tests to rerun in watch mode when one makes changes to a fixture. For it to work the filename/directory conventions mentioned above must be kept.

packages/ui-codemods/lib/__node_tests__/__testfixtures__/updateV10Breaking/colors.input.ts
```diff
@@ -0,0 +1,40 @@
...
+  return {
+    logoSubtitle: {
+      label: 'logoSubtitle',
+      marginTop: '-3px'
+    },
+    optionWrapper: {
+      label: 'optionWrapper',
+      margin: '-0.5rem -0.75rem',
+      padding: '0.5rem 0.75rem',
+      border: '0 solid ' + canvas.colors.tiara,
```
balzss, line 15: typescript throws an error for me here because `tiara` doesn't exist on `canvas.colors`, maybe canvas.colors should have an `any` type?

packages/ui-codemods/lib/__node_tests__/__testfixtures__/updateV10Breaking/colors.input.ts
```diff
@@ -0,0 +1,40 @@
...
+      display: 'flex',
+      maxWidth: '100%',
+      overflow: 'hidden',
+      alignItems: 'center'
+    },
+    optionIcon: {
+      label: 'optionIcon',
+      width: '36px',
+      height: '36px',
+      backgroundColor: canvasHighContrast.colors.oxford,
```
balzss, line 26: same issue here

packages/ui-codemods/lib/__node_tests__/__testfixtures__/updateV10Breaking/colors.input.ts
balzss: ok, I just saw that you put eslint-disable on the top. maybe ts-ignore could go there too

matyasf: yep, Ive added `ts-nocheck` (dont know why its needed, its already taken out from compilation in the `tsconfig`)

ToMESSKa:
@matyasf If I modify both input files the same time, the result of the test only shows one of them failing, it also says '1 failed'. Shouldn't it say '2 failed'? Both fail if I modify them independently tho.

matyasf:
> input files the same time, the result of the test only shows one of them failing, it also says '1 failed'. Shouldn't it say '2 failed'? Both fail if I modify them independently

I think this is OK, the 2 test fixtures are for one test. Usually we will make these one by one

ToMESSKa (approved):

PR #2041, opened 2025-06-18
feat(ui-color-picker): add colorScreenReaderLabel prop to provide mor…

ToMESSKa:
…e color information

INSTUI-4273

### DIFF
--- a/packages/ui-color-picker/src/ColorPreset/index.tsx
+++ b/packages/ui-color-picker/src/ColorPreset/index.tsx
@@ -264,12 +264,12 @@ class ColorPreset extends Component<ColorPresetProps, ColorPresetState> {
       padding="0"
       cursor={this.props.disabled ? 'not-allowed' : 'auto'}
       as="button"
+      aria-label={`${color}${
+        this.props.colorScreenReaderLabel
+          ? `, ${this.props.colorScreenReaderLabel}`
+          : ''
+      }`}
       {...(selectOnClick ? { onClick: () => this.props.onSelect(color) } : {})}
-      {...{
-        'aria-label': `${color}${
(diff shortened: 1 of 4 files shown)


packages/ui-color-picker/src/ColorPreset/index.tsx
```diff
@@ -264,12 +264,12 @@ class ColorPreset extends Component<ColorPresetProps, ColorPresetState> {
...
       as="button"
+      aria-label={`${color}${
+        this.props.colorScreenReaderLabel
+          ? `, ${this.props.colorScreenReaderLabel}`
+          : ''
+      }`}
       {...(selectOnClick ? { onClick: () => this.props.onSelect(color) } : {})}
-      {...{
-        'aria-label': `${color}${
-          this.isSelectedColor(color) ? ' selected' : ''
```
ToMESSKa, line 270: hard coding 'selected' isn't a good practice from a i18n perspective, so I introduced colorScreenReaderLabel prop instead

ToMESSKa:
> I think it would be nicer if the `colorScreenReaderLabel` would accept a function, like for example here: https://github.com/instructure/instructure-ui/blob/master/packages/ui-pagination/src/Pagination/PaginationPageInput/props.ts#L56
> 
> it could have 2 parameters, the color and whether its selected. This way users don't need to fiddle with the selected logic (and it will be easier to add uncontrolled mode if needed) and can come up with their own wording for i18n, e.g. "color with hex code EFEFEF" or "hex code EFEFEF selected"
@matyasf 
Nice idea, I converted colorScreenReaderLabel to accepting a function, it should be able to handle the special wordings now.

PR #2042, opened 2025-06-19
feat(ui-progress): add screenReaderValue prop and add aria-label to p…

ToMESSKa:
…rogressbar via the screenReaderLabel prop

INSTUI-4599


packages/ui-progress/src/ProgressBar/index.tsx
```diff
@@ -103,10 +103,6 @@ class ProgressBar extends Component<ProgressBarProps> {
       typeof formatScreenReaderValue === 'function'
         ? formatScreenReaderValue({ valueNow: valueNow!, valueMax: valueMax! })
         : undefined
-    // consolidating the label and aria-valuetext to put in aria-label because
-    // NVDA does not read aria-valuetext: https://github.com/nvaccess/nvda/issues/913
-    // But leaving aria-valuetext because JAWS ignores aria-label
-    const labelAndValueText = `${screenReaderLabel} ${valueText}`
```
ToMESSKa, line 109: being more that 6 years old, these are not valid anymore:
- NVDA does read `aria-valuetext` as of now
- JAWS does not read `aria-valuetext` in the latest InstUI version. But based on my testing, JAWS also fails to announce any accessible label or text (aria-label, aria-labelledby, aria-descriibedby, aria-value, aria-valuetext...) for a progress bar (or a `<div>`with role='progressbar') - except reading the current value. This seems to be a known bug:
[JAWS does not narrate progressbar name with virtual cursor · Issue #650 · FreedomScientific/standards-support](https://github.com/FreedomScientific/standards-support/issues/650).

packages/ui-progress/src/ProgressBar/index.tsx
```diff
@@ -126,7 +122,8 @@ class ProgressBar extends Component<ProgressBarProps> {
             css={styles?.htmlProgress}
             max={valueMax}
             value={valueNow}
-            aria-valuetext={labelAndValueText}
+            aria-valuetext={valueText}
+            aria-label={screenReaderLabel}
```
ToMESSKa, line 126: label now is added to 'aria-label', and 'aria-valuetext' does not include the label anymore. I did not feel the need to introduce the `screenReaderValue` prop suggested by @balzss in the slack channel, because I think `formatScreenReaderValue` does the same thing

matyasf (approved):

balzss (approved):
great work

PR #2043, opened 2025-06-19
feat(ui-heading): enable setting level for headings with variant

HerrTopi:
TEST_PLAN:  Check if all the variants still set the `h` tags as before, but if level is set, it overrides the defaults.
Also check the wording and completeness of the docs

INSTUI-4612

### DIFF
--- a/packages/ui-heading/src/Heading/README.md
+++ b/packages/ui-heading/src/Heading/README.md
@@ -6,10 +6,15 @@ Heading is a component for creating typographic headings.
 
 ## Variant
 
-Variant covers almost all use cases for headings on pages. Their name reflects the places they meant to be used. It sets the `level` prop and takes care of the style of the heading
-We recommend using `variants` instead of the `level` (and `as`) props.
+Variant covers almost all use cases for headings on pages. Their name reflects the places they meant to be used. It takes care of the style of the heading
 
-NOTE: when `variant` is set, `level` and `as` props are ignored
+NOTE 1: for legacy reasons, each `variant` has a default `level` set. This is not the recommended way and will be removed in a later major release. Please specify the `level` directly!
+
+NOTE 2: when `variant` is set the `as` prop is ignored
+
+IMPORTANT A11Y NOTE 1: there can be only one `h1` tag in a page
+
+IMPORTANT A11Y NOTE 2: `h` tags can not skip a level, so for example an `h1` followed by an `h3` not allowed
 
 ```js
 ---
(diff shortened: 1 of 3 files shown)


packages/ui-heading/src/Heading/README.md
```diff
@@ -6,10 +6,15 @@ Heading is a component for creating typographic headings.
...
+Variant covers almost all use cases for headings on pages. Their name reflects the places they meant to be used. It takes care of the style of the heading
 
-NOTE: when `variant` is set, `level` and `as` props are ignored
+NOTE 1: for legacy reasons, each `variant` has a default `level` set. This is not the recommended way and will be removed in a later major release. Please specify the `level` directly!
+
+NOTE 2: when `variant` is set the `as` prop is ignored
+
+IMPORTANT A11Y NOTE 1: there can be only one `h1` tag in a page
+
+IMPORTANT A11Y NOTE 2: `h` tags can not skip a level, so for example an `h1` followed by an `h3` not allowed
```
matyasf, line 17: followed by an `h3` **is** not allowed

packages/ui-heading/src/Heading/index.tsx
```diff
@@ -138,13 +135,16 @@ class Heading extends Component<HeadingProps> {
...
         return 'span'
+      } else if (level) {
+        return level
+      }
+
+      if (variant) {
+        return variantLevels[variant]
       } else {
-        return level!
+        return 'span'
```
matyasf, line 147: This code is not reachable in a meaningful way, it only ends up here if `level` is `null` (and not for `undefined` for some reason) and `variant` is not set. In this case its just a simple text, no need for `Heading`. Or is this just meant as a fallback for this edge case? Either was, `null` and `undefined` should behave the same.

packages/ui-heading/src/Heading/README.md
balzss, line 17: this sounds like there is a limit in our lib. instead of `not allowed`, I would say something like `incompatible with wcag` or not passing it or something like that

balzss (approved):

packages/ui-heading/src/Heading/index.tsx
HerrTopi: Yes, a fallback case. Mainly for ts to not complain

packages/ui-heading/src/Heading/README.md
HerrTopi: Fixed

HerrTopi: Fixed

matyasf (approved):

PR #2045, opened 2025-06-23
feat(ui-heading,shared-types): add new aiHeading prop and design. Also add h6 level to heading

HerrTopi:
TEST_PLAN:
check the ticket's Figma and compare the docs to it. Also check the wording and structure of the docs

INSTUI-4616


packages/ui-heading/src/Heading/index.tsx
```diff
@@ -107,15 +108,43 @@ class Heading extends Component<HeadingProps> {
   }
 
   renderContent() {
-    const { children, renderIcon } = this.props
-
-    if (renderIcon) {
+    const { children, renderIcon, aiVariant } = this.props
+    // const resolvedRenderIcon = aiVariant ? IconAiColoredSolid : renderIcon
```
balzss: i think this comment line can be removed

packages/ui-heading/src/Heading/README.md
```diff
@@ -34,11 +34,19 @@ type: example
...
+### AI Heading
+
+Pre-configured and with unique styles, the `ai-headings` are used for standardized, ai-related components.
+
 ```js
 ---
 type: example
 ---
-<Heading>Default Heading</Heading>
+<div>
```
balzss: i'd add a gap between the examples to better separate them visually, something like <div style={{display: 'flex', flexDirection: 'col', gap: '24px'>

packages/ui-heading/src/Heading/styles.ts
```diff
@@ -202,7 +207,29 @@ const generateStyle = (
...
+          }
+        : {})
+    },
+    igniteAI: {
+      label: 'igniteAI',
+      background: `
+        linear-gradient(to bottom, ${componentTheme.aiTextTopGradientColor} 0%, ${componentTheme.aiTextBottomGradientColor} 100%) text`,
+      border: 'solid transparent',
+      WebkitTextFillColor: 'transparent',
+      paddingRight: '.25rem'
```
balzss, line 211: shouldn't this use a spacing token from the theme? idk if need, just asking

packages/ui-heading/src/Heading/styles.ts
```diff
@@ -202,7 +207,29 @@ const generateStyle = (
...
+    },
+    igniteAIStacked: {
+      label: 'igniteAIStacked',
+      fontSize: '1rem',
+      lineHeight: '1.25rem',
```
balzss, line 216: same as above: shouldn't a token be used here?

balzss (changes requested):

packages/ui-heading/src/Heading/styles.ts
HerrTopi: Nope, this has to be hardcoded according the design. Which is ok, it wouldn't be a breaking change later if we need it

packages/ui-heading/src/Heading/styles.ts
HerrTopi: Same as my previous answer: nope

packages/ui-heading/src/Heading/README.md
HerrTopi: Good idea. Added.

balzss (approved):

matyasf (commented):
- the SR support could be better, JAWS and NVDA read the text without a break e.g. `IgniteAINutritition facts` (I think a comma would solve it), but VO is weird its saying things like "Heading level 2 Ignite AI level 1 titleardSection level 1" It would be very good if it would not say those "level 1 " stuff
- Heading levels h2,h5,h6 have the IgniteAI text not set to bold, according to design this should look the same independent from the level (when I dont set `variant`)

PR #2046, opened 2025-06-23
feat(ui-text-input,ui-date-input): properly support IconButtons inside TextInputs

balzss:
INSTUI-4614

test plan:

- check new docs and example for TextInput with the iconbutton rendered at the end
- check out dateinput2
- check the code, please try to think about edge cases. I might have forgot about something that could break

### DIFF
--- a/packages/ui-text-input/src/TextInput/README.md
+++ b/packages/ui-text-input/src/TextInput/README.md
@@ -182,9 +182,6 @@ TextInput accepts focusable and non-focusable content before and/or after
 the input text. A common use case is adding an icon or avatar to the input.
 Focusable content will be focused separately from the input itself.
 
-> Note: For any content larger than an icon or small avatar (multiple [Tags](#Tag), for example),
-> use the `renderBeforeInput` property.
-
 - ```javascript
   class ExtraContentExample extends React.Component {
     constructor(props) {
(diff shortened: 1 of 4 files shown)


packages/ui-text-input/src/TextInput/README.md
```diff
@@ -301,6 +298,47 @@ Focusable content will be focused separately from the input itself.
...
+---
+type: example
+---
+<TextInput
+  renderLabel="Search"
+  type="search"
+  placeholder="Find something..."
+  renderBeforeInput={<IconSearchSolid/>}
+/>
+```
```
matyasf: isnt it an a11y problem that the close button here does not have a focus ring and is not in the keyboard tab order?

packages/ui-text-input/src/TextInput/README.md
```diff
@@ -301,6 +298,47 @@ Focusable content will be focused separately from the input itself.
...
+    <TextInput
+      renderLabel="Password"
+      type={showPassword ? 'text' : 'password'}
+      placeholder="Find something..."
+      value={passwordValue}
+      onChange={(e, newValue) => setPasswordValue(newValue)}
+      renderAfterInput={
+        <IconButton withBorder={false} withBackground={false} onClick={() => setShowPassword(prevState => !prevState)}>
+          {showPassword ? <IconOffLine/> : <IconEyeLine/>}
+        </IconButton>
```
matyasf: SRs say here just Button, please add some text for them, something like "show/hide text"

packages/ui-text-input/src/TextInput/README.md
```diff
@@ -301,6 +298,47 @@ Focusable content will be focused separately from the input itself.
...
+      }
+    />
+  )
+}
+render(<InputsWithButtonsExample />)
+```
+
+Having a clear button for a search field is also common but you don't need to implement anything to make that work, only set the `type` of the input to `search`:
```
matyasf: I dont see a close button in FF for this example.

matyasf (changes requested):
see my comments. Also

packages/ui-text-input/src/TextInput/README.md
balzss: tbh I wasn't really sure where that feature was coming from but seems like it's a chrome thing. I've added this example to show people that they don't need the iconbutton when it's `type="search"`. I guess this feature is not really a complete solution so I'll remove this example. later on we could discuss what to do/suggest with `type=search` input fields

packages/ui-text-input/src/TextInput/README.md
balzss: see my comment about the search input below

packages/ui-text-input/src/TextInput/README.md
balzss: done

packages/ui-text-input/src/TextInput/README.md
balzss: I've removed this example

PR #2048, opened 2025-06-24
fix(ui-tabs): fix bottom border of Tab with scrolling overflow in sec…

ToMESSKa:
…ondary variant

INSTUI-4604


packages/ui-tabs/src/Tabs/styles.ts
```diff
@@ -62,6 +62,10 @@ const generateStyle = (
       flexFlow: 'row wrap'
     },
     scroll: {
+      marginBottom:
+        variant === 'secondary'
+          ? `calc(${componentTheme.tabVerticalOffset} * -1)`
```
ToMESSKa, line 67: tabVerticalOffset gives effect of selected tab "bar" overlaying bottom border, see used in like 50 above

PR #2051, opened 2025-06-30
Refactor: remove jsdoc-api from the docs app

matyasf:
Changes:   
- It uses the [TypeScript compiler API](https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API) instead of `JSDoc-api` JSDoc is outdated and its meant for JavaScript projects.  We use this to parse everything thats not a React component or a Markdown file. (the `utilities` menu in the sidebar).  
-  It gets type information for **functions** from the TypeScript types, so it can be deleted in the jsdoc (the `{Object}` part in `@param {Object} bla`).
- For non-function or complex utilities it gets the info from the JSDoc so we need to keep it (e.g. `withStyle`, `contains`).
- Removed `DateTime` and `conversions` and some custom PropTypes from the doc since they were parsed in a special way and contained no useful info.
- note that it cannot parse types if the jsdoc is at the export e.g.
   ```
   function myFunc() ...
   
   export {
   /**
   * here is the doc. The parser doesnt know the types, do not use this pattern
   */
   myFunc
   }
   ```

To test: 
- check all the functions in the `utilities` submenu, they should display equal or more information

### DIFF
--- a/docs/contributor-docs/writing-docs.md
+++ b/docs/contributor-docs/writing-docs.md
@@ -12,15 +12,40 @@ This page documents how to add and maintain documentation.
 
 The documentation site is generated from the **source code** and the `.md` files. There are two types of `.md` files: `README.md` and `named-md-files.md`, under the `docs folder`.
 
-The **source code** is parsed with [JSDoc](https://jsdoc.app/) with the exeption of react components, which are parsed by [react-docgen](https://react-docgen.dev/) and it provides `type` information for the docs.
+The **source code** is parsed with the [TypeScript compiler API](https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API) with the exception of React components, which are parsed by [react-docgen](https://react-docgen.dev/). These tools parse out JSDoc comments, type information of React component props and types for functions.
+
+## Utility functions
+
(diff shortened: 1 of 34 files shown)


packages/__docs__/src/Params/index.tsx
```diff
@@ -54,33 +55,78 @@ class Params extends Component<ParamsProps> {
...
-    return type ? type.names.join(' | ') : null
-  }
-
   renderDescription(description: string) {
     return <div>{description && compileMarkdown(description)}</div>
   }
 
-  render() {
-    const { layout } = this.props
+  renderGenericRows() {
```
matyasf, line 62: now it can display generic parameters

docs/contributor-docs/writing-docs.md
```diff
@@ -12,15 +12,40 @@ This page documents how to add and maintain documentation.
 
 The documentation site is generated from the **source code** and the `.md` files. There are two types of `.md` files: `README.md` and `named-md-files.md`, under the `docs folder`.
 
-The **source code** is parsed with [JSDoc](https://jsdoc.app/) with the exeption of react components, which are parsed by [react-docgen](https://react-docgen.dev/) and it provides `type` information for the docs.
+The **source code** is parsed with the [TypeScript compiler API](https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API) with the exception of React components, which are parsed by [react-docgen](https://react-docgen.dev/). These tools parse out JSDoc comments, type information of React component props and types for functions.
+
+## Utility functions
```
matyasf, line 17: Heres the high level information how it works

PR #2067, opened 2025-07-14
chore(ui-codemods): make local themes default

git-nandor:


packages/ui-codemods/lib/__node_tests__/runTest.ts
```diff
@@ -42,9 +43,14 @@ export function runTest(codemod: Transform) {
   )
   let fixturesRun = 0
   entries.forEach((entry) => {
-    if (entry.isFile() && entry.name.indexOf('input') > -1) {
+    if (entry.isFile() && entry.name.includes('input')) {
+      const isWarningTest = entry.name.includes('.warning.input')
```
matyasf, line 47: This is pretty good :) Can you please mention this in https://instructure.design/#codemods ?

packages/ui-codemods/lib/utils/codemodHelpers.ts
```diff
@@ -503,6 +503,62 @@ function addImportIfNeeded(
...
+ * @param oldName - Original imported name to replace
+ * @param newName - New name to use
+ * @returns boolean indicating if any changes were made
+ */
+function renameImportAndUsages(
+  j: JSCodeshift,
+  root: Collection,
+  oldName: string,
+  newName: string
+): boolean {
```
matyasf: For this to be a good utility function one should also specify the import path. While your use case here (import names like `canvasThemeLocal`) are pretty much guaranteed to be unique, if we want to rename e.g. `Button` with this then I bet that it would make lots of false renames. Also this new param should work like in `replaceImport`, meaning one should be able to specify either a string or an array of strings

matyasf (changes requested):
see my comments

packages/ui-codemods/lib/__node_tests__/runTest.ts
```diff
@@ -53,17 +59,27 @@ export function runTest(codemod: Transform) {
...
-      // Also, this is closer to how its used, there is always a correct
-      // filename
-      runInlineTest(codemod, {}, { path: inputPath, source: input }, expected)
+
+      if (isWarningTest) {
+        const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {})
+
+        try {
+          // Just run codemod, no file output comparison needed
+          runInlineTest(codemod, {}, { path: inputPath, source: input }, input)
```
ToMESSKa: @git-nandor Based on this, I think the warning check will only work if the codemod does not alter the input. So this will not work with a textfixture file where the code is modified by the codemod and a warning needs to be printed the same time.  Because if you compare the input to input in line 68 and codemod modifies the input, the runInlineTest will encounter a change and the rest of the code won't run. Maybe it needs no be rethinked?

git-nandor: Based on what we discussed, I have updated the warning testing so that multiple warnings can now be tested, and there is no longer any input file comparison. This way, it works correctly even when the file changes.

packages/ui-codemods/lib/__node_tests__/runTest.ts
git-nandor: I added a new section about warning testing.

packages/ui-codemods/lib/utils/codemodHelpers.ts
git-nandor: Thank you for the feedback. I’ve made the necessary changes.

packages/ui-codemods/lib/__node_tests__/runTest.ts
ToMESSKa: @git-nandor Nice job, now the warnings can be tested even if the input changes.
As for multiple warnings, the warning test passes if I expect only one warning but two are emitted. I think it only checks if a warning containing that string was called at least once, but it doesn't verify that _only those_ expected warnings were called.
I checked, something like this would work if I only expect one warning but two are emitted.
```
         // Get all actual warning calls
          const actualWarnings = warnSpy.mock.calls.map(call => call[0] as string);
          // 1. First verify the exact number of warnings
          expect(warnSpy).toHaveBeenCalledTimes(warnings.length);
          // 2. Then verify each expected warning exists exactly once
          warnings.forEach((expected) => {
            // Check that the warning was called at least once with this content
            expect(warnSpy).toHaveBeenCalledWith(expect.stringContaining(expected))
            // Verify it appears exactly once
            const matchingCount = actualWarnings.filter(w => 
              w.includes(expected)
            ).length;
            expect(matchingCount).toBe(1)
          })
```

packages/ui-codemods/lib/renameCanvasThemesCodemod.ts
```diff
@@ -0,0 +1,103 @@
...
+const renameCanvasThemesCodemod: Transform = (
+  file,
+  api,
+  options?: { fileName?: string; usePrettier?: boolean }
+) => {
+  return instUICodemodExecutor(renameCanvasThemes, file, api, options)
+}
+
+const renameCanvasThemes: InstUICodemod = (j, root, filePath) => {
+  const importPath = '@instructure/ui-themes'
```
matyasf: This should be `['@instructure/ui-themes', '@instructure/ui'] because its also exported from our `ui` meta-package https://github.com/instructure/instructure-ui/blob/master/packages/ui/src/index.ts#L406

packages/ui-codemods/lib/renameCanvasThemesCodemod.ts
```diff
@@ -0,0 +1,103 @@
...
+          filePath,
+          lineNumber,
+          `Found ${objectName}.use(). This will be deleted. Users should wrap each React root in its own InstUISettingsProvider instead.`
+        )
+        didWarn = true
+      }
+    }
+  })
+
+  return didRename || didWarn
```
matyasf: You only need to return `true` if there are modifications, this is used in `instUICodemodExecutor` to run `prettier` after the rewrite

packages/ui-codemods/lib/__node_tests__/runTest.ts
git-nandor: Thanks for the feedback! I updated the warning tests to be stricter.

packages/ui-codemods/lib/renameCanvasThemesCodemod.ts
git-nandor: Thanks for the feedback! I added the new path.

packages/ui-codemods/lib/renameCanvasThemesCodemod.ts
git-nandor: Thanks! I modified

packages/ui-codemods/lib/__node_tests__/runTest.ts
ToMESSKa: @git-nandor  Okay I sorry, I realized the code I posted does not handle the case when there are three warnings and the same warning appears twice, only cases where each warning is different...

I checked, this could handle that
```
          const actualWarnings = warnSpy.mock.calls.map(
            (call) => call[0] as string
          )

          expect(warnSpy).toHaveBeenCalledTimes(expectedWarnings.length)

          expectedWarnings.forEach((expected) => {
            // Count how many times this warning was expected
            const expectedCount = expectedWarnings.filter(
              (e) => e === expected
            ).length

            // Count how many times it actually appeared
            const actualCount = actualWarnings.filter((w) =>
              w.includes(expected)
            ).length

            expect(actualCount).toBe(expectedCount) // Now handles duplicates!
          })
```

PR #2070, opened 2025-07-16
chore(ui-codemods): add codemod to remove maxWidth style variable fr…

ToMESSKa:
…om Tag

INSTUI-4632

**TEST PLAN**

- run the codemod tests, they should pass
- think of cases the current test cases do not cover

### DIFF
--- a/packages/ui-codemods/lib/__node_tests__/__testfixtures__/removeMaxWidthFromTag/test1.output.tsx
+++ b/packages/ui-codemods/lib/__node_tests__/__testfixtures__/removeMaxWidthFromTag/test1.output.tsx
@@ -1,0 +1,83 @@
+// @ts-nocheck
+
+import { Tag } from '@instructure/ui-tag'
+import { InstUISettingsProvider } from '@instructure/emotion'
+
+function test() {
+  return (
+    <div>
+      <Tag
+        text="hello"
+        themeOverride={{
+          defaultBackground: 'custom value'
+        }}
+      />
+      <Tag
+        text="hello"
+        themeOverride={() => ({
+          defaultBackground: 'custom value'
+        })}
+      />
+      <Tag
+        text="hello"
+        themeOverride={{
+          defaultBackground: 'custom value'
+        }}
+      />
+      <Button
+        text="hello"
+        themeOverride={{
+          maxWidth: '10px'
+        }}
+      />
+      <Tag text="hello" />
+      <Tag
+        text="hello"
+        themeOverride={{
+          ...someOverrides
+        }}
+      />
+      <Tag
+        text="hello"
+        themeOverride={{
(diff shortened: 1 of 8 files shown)


packages/ui-codemods/lib/removeMaxWidthFromTag.ts
```diff
@@ -0,0 +1,352 @@
...
+              if (
+                attr.value?.type === 'JSXExpressionContainer' &&
+                (attr.value.expression?.type === 'CallExpression' ||
+                  attr.value.expression?.type === 'Identifier')
+              ) {
+                const line = attr.loc?.start.line
+                printWarning(
+                  filePath,
+                  line,
+                  'maxWidth style variable has been removed but cannot analyze dynamic themeOverride (function call or variable). Please check manually for maxWidth usage.'
```
matyasf: As I understand the code, it does not modifications in this case? (which is fine, users should do it manually for these cases, just the text is wrong)

packages/ui-codemods/lib/__node_tests__/__testfixtures__/removeMaxWidthFromTag/test1.output.tsx
```diff
@@ -0,0 +1,83 @@
...
+                defaultBackground: 'custom value'
+              }
+            }
+          }
+        }}
+      >
+        <Tag text="hello" />
+      </InstUISettingsProvider>
+
+      <Tag text="hello" />
```
matyasf, line 76: some strange things are happening here. This tag does not exists in the input

packages/ui-codemods/lib/removeMaxWidthFromTag.ts
```diff
@@ -0,0 +1,352 @@
...
+              if (
+                attr.value?.type === 'JSXExpressionContainer' &&
+                attr.value.expression?.type === 'ArrowFunctionExpression'
+              ) {
+                const func = attr.value.expression
+                let themeObj: namedTypes.ObjectExpression | null = null
+                const line = attr.loc?.start.line
+
+                // Subcase 2a: Handle concise arrow function (implicit return)
+                // Example: themeOverride={() => ({ maxWidth: '10px' })}
```
matyasf, line 104: (for the future) While its pretty cool, that it can handle functions, if this gets too complicated I'd just eit a warning in such cases. Its enough that we handle direct assignments e.g. `themeOverride={{asd:bla}}`

packages/ui-codemods/lib/removeMaxWidthFromTag.ts
ToMESSKa: @matyasf Thank you, I corrected the print messages. I also added a warning test based on Nándi's PR.

PR #2071, opened 2025-07-16
fix(ui-modal): remove tabbability from modalBody if there is no scrollbar

HerrTopi:
TEST_PLAN:

-  check if tabbing works ok in general through a modal
- make sure you can't tab to the modal body if it has no scrollbar
- make sure you can tab to the modal body if it has scrollbar

INSTUI-4648

### DIFF
--- a/packages/ui-modal/src/Modal/ModalBody/index.tsx
+++ b/packages/ui-modal/src/Modal/ModalBody/index.tsx
@@ -113,14 +113,10 @@ class ModalBody extends Component<ModalBodyProps> {
         as={as}
         css={this.props.styles?.modalBody}
         padding={padding}
-        // Setting this to 0 is necessary for findFocusable to find this DOM
-        // element.
-        // Note that VoiceOver does not read the contents properly
-        // if there is a scrollbar, and it has no kb just SR focus.
-        // To prevent this we could set tabIndex to -1
-        // but this would make the scrollbar non-focusable, so the modal cannot
-        // be scrolled with keyboard.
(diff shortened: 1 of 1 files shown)


packages/ui-modal/src/Modal/ModalBody/index.tsx
```diff
@@ -113,14 +113,10 @@ class ModalBody extends Component<ModalBodyProps> {
...
-        // Setting this to 0 is necessary for findFocusable to find this DOM
-        // element.
-        // Note that VoiceOver does not read the contents properly
-        // if there is a scrollbar, and it has no kb just SR focus.
-        // To prevent this we could set tabIndex to -1
-        // but this would make the scrollbar non-focusable, so the modal cannot
-        // be scrolled with keyboard.
-        tabIndex={0}
+        //check if there is a scrollbar, if so, the element has to be tabbable to be able to scroll with keyboard only
+        {...(this.ref?.scrollHeight !== this.ref?.getBoundingClientRect().height
```
balzss, line 117: works fine as is, but maybe move this calculation to a variable outside the render function so it's a bit cleaner. good job nonetheless

balzss (approved):

packages/ui-modal/src/Modal/ModalBody/index.tsx
HerrTopi: Could be. To be honest we use this pattern in lots of places and is farely easy to understand imho and I like that we don't have to go elsewhere to see the implementation

PR #2072, opened 2025-07-17
fix(many): fix focus ring not respecting theme overrides in Button and FileDrop

matyasf:
This was already fixed in INSTUI-4375, but it broke with the focus rewrite. This PR introduces a more stable method to accomplish the same thing. Instead of a new prop, now the borderRadius prop on View accepts CSS units like '1rem'

To test:
- Set different `borderRadius` values in the `View` examples when it has a focus ring.
- Check if `focusOutlineWidth`, `focusOutlineOffset`, `focusOutlineInset` theme variables are working the same for `View` by overriding them and comparing the result on `master`
- Override the `borderRadius` theme variable in the `Button` and `FileDrop` examples. it should now respect the theme variable and not use the default value of the View.

Fixes INSTUI-4641

### DIFF
--- a/packages/__docs__/src/Returns/index.tsx
+++ b/packages/__docs__/src/Returns/index.tsx
@@ -25,6 +25,7 @@
 import { Component } from 'react'
 import { Table } from '@instructure/ui-table'
 import type { ReturnsProps } from './props'
+import Markdown from 'marked-react'
 
 class Returns extends Component<ReturnsProps> {
   render() {
@@ -41,7 +42,9 @@ class Returns extends Component<ReturnsProps> {
             <Table.Cell>
               <code>{this.props.types.type}</code>
             </Table.Cell>
-            <Table.Cell>{this.props.types.description}</Table.Cell>
+            <Table.Cell>
+              <Markdown>{this.props.types.description}</Markdown>
+            </Table.Cell>
           </Table.Row>
         </Table.Body>
       </Table>
--- a/packages/emotion/src/styleUtils/getShorthandPropValue.ts
+++ b/packages/emotion/src/styleUtils/getShorthandPropValue.ts
@@ -29,55 +29,58 @@ import { logError as error } from '@instructure/console'
  * ---
  * category: utilities/themes
  * ---
- * Given a theme object, a string of space delimited prop values,
- * and a propName prefix, combines each prop value with the
- * propName prefix and replaces it with a corresponding value
- * from the theme object.
+ * Gets prop values from a theme object for the given values.
  * @module getShorthandPropValue
  *
- * @param {String} componentName - the name of the component (for error messages)
- * @param {Object} componentTheme - a theme object of keys and values
- * @param {String} propValue - a space delimited string of values
- * @param {String} propName - a prefix to combine with each propValue
- * @returns {String} a string with each value replaced with a value from the theme object
+ * @param componentName The name of the component (just for error messages)
+ * @param componentTheme A theme object of keys and values
+ * @param propValue A space delimited string of values e.g. `"small large none"`
+ * @param propName A prefix to combine with each propValue
+ * @param matchCSSUnits Match valid CSS strings too like `1rem`. It's a simple
+ * matcher, just checks whether the given string starts with a number and ends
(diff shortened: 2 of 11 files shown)


packages/emotion/src/styleUtils/getShorthandPropValue.ts
```diff
@@ -29,55 +29,58 @@ import { logError as error } from '@instructure/console'
...
-      }
-
-      if (shortHandValue === 'pill') {
+      } else if (shortHandValue === 'pill') {
         return '999em'
+      } else if (matchCSSUnits) {
+        // try to match it as a CSS unit
+        // matches [number][string] like '4rem', '.5em', '2.6px'
+        if (shortHandValue.match(/^([+-]?(\d+\.?\d*|\.\d+))([a-zA-Z]+)$/)) {
+          return shortHandValue
```
matyasf, line 75: This is not the best method to detect that something is a valid number+CSS variable, but I think we can safely assume that our theme tokens (e.g. `large`, `x-small`, ... check `ThemeablePropValues.ts`) will not start with a number or dot.

Making a perfect CSS unit+number matcher would much more complex, something like

```
function isValidCSSLength(s) {
    // Regular expression to match the numeric part and the unit part
    const regex = /^([+-]?(\d+\.?\d*|\.\d+))([a-zA-Z]*)$/;
    const match = s.match(regex);
    if (!match) return false;

    const numStr = match[1];
    const unitStr = match[3];

    // Check if the numeric part is zero
    const isZero = numStr.replace(/[0.]/g, '') === '';

    // List of valid CSS length units (case-insensitive)
    const validUnits = new Set([
        'cm', 'mm', 'q', 'in', 'pc', 'pt', 'px',
        'em', 'ex', 'cap', 'ch', 'ic', 'rem', 'lh', 'rlh',
        'vw', 'vh', 'vi', 'vb', 'vmin', 'vmax',
        'svw', 'svh', 'lvw', 'lvh', 'dvw', 'dvh',
        'svi', 'svb', 'lvi', 'lvb', 'dvi', 'dvb',
        'smin', 'lmin', 'dmin', 'smax', 'lmax', 'dmax'
    ]);

    if (isZero) {
        // For zero, the unit is optional but must be valid if present
        return unitStr === '' || validUnits.has(unitStr.toLowerCase());
    } else {
        // For non-zero, the unit must be present and valid
        return unitStr !== '' && validUnits.has(unitStr.toLowerCase());
    }
}
```

packages/ui-view/src/View/styles.ts
```diff
@@ -243,47 +192,6 @@ const getFocusStyles = (props: ViewProps, componentTheme: ViewTheme) => {
...
-        borderRadius: `calc(${componentTheme.borderRadiusMedium} - (${componentTheme.focusOutlineInset} + ${componentTheme.focusOutlineWidth}))`
-      },
-      'focusRing--radiusLarge': {
-        borderRadius: `calc(${componentTheme.borderRadiusLarge} - (${componentTheme.focusOutlineInset} + ${componentTheme.focusOutlineWidth}))`
-      },
-      'focusRing--radiusCustom': {
-        borderRadius: `calc(${focusRingBorderRadius} - (${componentTheme.focusOutlineOffset} + ${componentTheme.focusOutlineWidth}))`
-      }
-    }
-  }
```
matyasf, line 286: the cool part is that we're getting rid of this spaghetti code. I think it was needed because it needed to 'copy' CSS props to the `:before` tag previously

packages/__docs__/src/Returns/index.tsx
```diff
@@ -41,7 +42,9 @@ class Returns extends Component<ReturnsProps> {
             <Table.Cell>
               <code>{this.props.types.type}</code>
             </Table.Cell>
-            <Table.Cell>{this.props.types.description}</Table.Cell>
+            <Table.Cell>
+              <Markdown>{this.props.types.description}</Markdown>
```
matyasf, line 46: an unrelated small fix that allows to render Markdown properly in the "Returns" part of the docs (see `getShorthandPropValue`)

ToMESSKa (approved):

PR #2075, opened 2025-07-22
fix(ui-text-input): set IconButton height in TextInput to not break l…

joyenjoyer:
…ayout

INSTUI-4651

### DIFF
--- a/packages/ui-text-input/src/TextInput/index.tsx
+++ b/packages/ui-text-input/src/TextInput/index.tsx
@@ -97,6 +97,18 @@ class TextInput extends Component<TextInputProps, TextInputState> {
 
     this.ref = el
 
+    // Find the IconButton's span with class ending in 'content' and set its height
+    // This is a workaround until this component is redesigned (see INSTUI-4651).
+    // TODO: Remove this if INSTUI-4642 is completed.
+    if (el) {
+      const contentSpan = el.querySelector(
+        'span[class$="baseButton__content"]'
+      ) as HTMLElement | null
+      if (contentSpan) {
+        contentSpan.style.height = '2.25rem'
+      }
+    }
+
     if (typeof elementRef === 'function') {
       elementRef(el)
     }
--- a/packages/ui-text-input/src/TextInput/styles.ts
+++ b/packages/ui-text-input/src/TextInput/styles.ts
@@ -211,8 +211,6 @@ const generateStyle = (
       '& > :not(button):not([data-position^="Popover"])': {
         marginRight: componentTheme.padding
       },
-      marginTop: '1px',
-      marginBottom: '2px',
       display: 'flex',
       alignItems: 'center',
       ...sizeVariants[size!],


packages/ui-text-input/src/TextInput/index.tsx
```diff
@@ -97,6 +97,18 @@ class TextInput extends Component<TextInputProps, TextInputState> {
 
     this.ref = el
 
+    // Find the IconButton's span with class ending in 'content' and set its height
+    // This is a workaround until this component is redesigned (see INSTUI-4651).
+    // TODO: Remove this if INSTUI-4642 is completed.
```
matyasf: We should put here the actual explanation (its ok if it takes up a few paragraphs), not just link tickets.

packages/ui-text-input/src/TextInput/index.tsx
```diff
@@ -97,6 +97,18 @@ class TextInput extends Component<TextInputProps, TextInputState> {
...
+    if (el) {
+      const contentSpan = el.querySelector(
+        'span[class$="baseButton__content"]'
```
matyasf: hmm I really dont like this part, we should only depend on CSS selectors that there are really no other possibilities. Also this will fail if the user puts a button into `renderBeforeInput`. Searching in the children of `_afterElement` is better if I understand this correctly

packages/ui-text-input/src/TextInput/index.tsx
```diff
@@ -97,6 +97,18 @@ class TextInput extends Component<TextInputProps, TextInputState> {
...
+      ) as HTMLElement | null
+      if (contentSpan) {
+        contentSpan.style.height = '2.25rem'
```
matyasf: where is this value coming from?

packages/ui-text-input/src/TextInput/styles.ts
```diff
@@ -208,14 +208,22 @@ const generateStyle = (
...
-        marginRight: componentTheme.padding
+        marginRight: componentTheme.padding,
+        ...(sizeVariants[size!] && {
+          lineHeight: sizeVariants[size!].lineHeight
+        })
       },
-      marginTop: '1px',
-      marginBottom: '2px',
       display: 'flex',
+      marginTop: '1.5px',
```
joyenjoyer: The sum of the top (1.5px) and bottom (1.5px) margin is 3px. Previously, the top margin was 1px and bottom margin was 2px so the `IconButton` was closer to the top border of the `TextInput`.

packages/ui-text-input/src/TextInput/index.tsx
```diff
@@ -141,6 +142,35 @@ class TextInput extends Component<TextInputProps, TextInputState> {
     this.props.makeStyles?.(this.makeStyleProps())
   }
 
+  adjustAfterElementHeight() {
+    const afterElementChild = this._afterElement
```
joyenjoyer, line 146: In my first attempt, I chose css query selectors, it was ugly. Here, I set the style of nested elements using the `afterElement` ref.

packages/ui-text-input/src/TextInput/index.tsx
```diff
@@ -141,6 +142,35 @@ class TextInput extends Component<TextInputProps, TextInputState> {
...
+            .firstElementChild as HTMLElement | null)
+        : null
+
+    const targetSpan = buttonContentSpan ?? popoverContentSpan
+
+    if (targetSpan) {
+      // Set the height to 2.438rem (the height of a medium TextInput) to avoid layout shift when the afterElement content changes
+      // this temporary workaround is necessary because otherwise the layout breaks, later on IconButton's default height will be adjusted to the TextInput size
+      // so this workaround will not be needed anymore
+      targetSpan.style.height = '2.438rem'
```
balzss: i guess this value is 39px coverted to rem. since it's directly coming from a pixel value, i'd use pixels instead of rems here with some added explanation on why this is the exact value. also hardcoding this value here makes the code more fragile since if the values change (e.g. themeoverride the border stroke width) this may break

packages/ui-text-input/src/TextInput/styles.ts
balzss: please add a comment why these margin values need to be 1.5px (an not, for example, 1)

balzss (commented):
is it intended that the textinput without an iconbutton is 38px tall and the one with an iconbutton is 39px tall?

packages/ui-text-input/src/TextInput/index.tsx
joyenjoyer: Regardless of having an `IconButton` inside, height is the same now (set to `36px`, or if you check the element with the `facade` postfix, it has `38px` height).

packages/ui-text-input/src/TextInput/styles.ts
joyenjoyer: I was able to remove margins completely.

balzss (approved):
looks good

PR #2077, opened 2025-07-22
build(ui-scripts): allow to specify custom port for dev server

matyasf:
One can run now commands like "npm run dev -- -p 1234"

To test:
Run `npm run dev -- -p 1234`
Check if it runs on port 1234


docs/contributor-docs/dev-commands.md
```diff
@@ -11,7 +11,7 @@ This is a quick developer reference for common Instructure UI repository command
...
 | --------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `npm run bootstrap`                                                         | Clean all node modules and build output, reinstall dependencies, create fresh build including regenerating design tokens and type declaration files.                                                            |
-| `npm run dev`                                                               | Run a local instance of the documentation app in dev mode. Docs will be served at `http://0.0.0.0:9090`                                                                                                         |
+| `npm run dev`                                                               | Run a local instance of the documentation app in dev mode. Docs will be served at `http://0.0.0.0:9090`. Port can be specified too e.g. `npm run dev -- -p 1234`                                                |
```
balzss, line 14: if you put `--` in the package.json after the script, doesn't that mean the user doesn't need to specify it? so it would be `npm run dev -p 1234`

balzss (approved):
looks good, left one question

docs/contributor-docs/dev-commands.md
matyasf: nope, it just means that the parameters are forwarded. Sadly you need to use `npm run dev -- -p 1234`

PR #2084, opened 2025-07-24
chore(ui-react-utils)Remove hack decorator

git-nandor:


balzss:
this was done separately in the v11 branch and it is already in there

PR #2086, opened 2025-07-28
fix(ui-testable): handle ref forwarding in testable decorator

aaronshaf:
The testable decorator was passing ref as a regular prop instead of using
React's ref forwarding mechanism, causing React warnings when components
decorated with @testable() received a ref prop.
    
This fix updates the decorator to use React.forwardRef and filters out
the ref prop before passing props to the wrapped component, eliminating
the warning: "ref is not a prop. Trying to access it will result in
undefined being returned."

### DIFF
--- a/packages/ui-testable/src/testable.ts
+++ b/packages/ui-testable/src/testable.ts
@@ -22,7 +22,7 @@
  * SOFTWARE.
  */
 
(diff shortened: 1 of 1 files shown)


matyasf:
We will fix this in the upcoming v11 release by removing `@testable` completely and hardcoding the `data-cid` values
We will close this.

PR #2093, opened 2025-08-08
feat(ui-flex): add order prop to Flex.Item

ToMESSKa:
INSTUI-4663

**ISSUE:**
- Flex.Items cannot be ordered

**TEST PLAN:**
- try the example below, the item 'Three' should appear in the fourth place as the default value of the other items is '0' so they precede it
- try to assign different order numbers (e.g. from 1 to 4) to each Flex.Item in the example below, the items should be appear in that order
- when direction='row-reverse' is set to Flex in the example below, 'Three' should appear in the first place
- try setting the order prop in other examples in the documentation too, they should work as expected
```
<Flex withVisualDebug margin="none none large">
    <Flex.Item>One</Flex.Item>
    <Flex.Item>Two</Flex.Item>
    <Flex.Item order='1'>Three</Flex.Item>
    <Flex.Item>Four</Flex.Item>
  </Flex>
```

### DIFF
--- a/packages/ui-flex/src/Flex/Item/props.ts
+++ b/packages/ui-flex/src/Flex/Item/props.ts
@@ -113,11 +113,9 @@ type FlexItemOwnProps = {
   /**
    * Specifies the order of the `Flex.Item` inside the `Flex` component.
    *
-   * Items with the same order revert to source order.
-   *
-   * Utilizes the order flex CSS property. Default order CSS value is 0.
+   * Utilizes the order flex CSS property.
    */
-  order?: string | number
+  order?: number
 }
 
 type PropKeys = keyof FlexItemOwnProps
@@ -145,7 +143,7 @@ const propTypes: PropValidators<PropKeys> = {
   shouldShrink: PropTypes.bool,
   size: PropTypes.string,
   withVisualDebug: PropTypes.bool,
-  order: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
+  order: PropTypes.number
 }
 
 const allowedProps: AllowedPropKeys = [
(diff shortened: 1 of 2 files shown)


packages/ui-flex/src/Flex/Item/styles.ts
```diff
@@ -53,6 +53,7 @@ const generateStyle = (
       boxSizing: 'border-box',
       minWidth: '0.0625rem',
       flexBasis: size,
+      ...(order !== undefined && { order }),
```
HerrTopi: I would write it like this:
...(order!==undefined?{order}:{})

The current solution is ok as well, however it makes little sense to spread "false" inside the object, even if js ignores it

packages/ui-flex/src/Flex/Item/props.ts
```diff
@@ -110,6 +110,14 @@ type FlexItemOwnProps = {
...
    */
   withVisualDebug?: boolean
+  /**
+   * Specifies the order of the `Flex.Item` inside the `Flex` component.
+   *
+   * Items with the same order revert to source order.
+   *
+   * Utilizes the order flex CSS property. Default order CSS value is 0.
+   */
+  order?: number
```
HerrTopi, line 118: I don't know if I'd over explain the behavior. It works like the native css stuff, maybe we could just say that? I'm not sure
Also, I don't think it defaults to 0. It defaults to undefined

HerrTopi (changes requested):

packages/ui-flex/src/Flex/Item/styles.ts
ToMESSKa: thanks, I changed this

packages/ui-flex/src/Flex/Item/props.ts
ToMESSKa: I simplified the description

PR #2094, opened 2025-08-08
[v11] Fix Transition tests, add back ReactDOM.findDOMNode for old React versions

matyasf:


docs/guides/accessing-the-dom.md
```diff
@@ -8,9 +8,12 @@ order: 5
 
 Accessing the DOM directly in React is discouraged, because it pierces the DOM abstraction. Still there are some cases when InstUI is required to do so (for example for focus management or positioning). This happens with the [findDOMNode](#findDOMNode) utility function.
```
matyasf, line 10: We will need to compile a list for our users when they need to add this prop. The good news is that we have been spamming their console with warnings about this since InstUI 9

packages/ui-dom-utils/src/findDOMNode.ts
```diff
@@ -81,8 +82,14 @@ function findDOMNode(el?: UIElement): Element | Node | Window | undefined {
...
     console.error(
-      `Error: ${elName} doesn't have "ref" property.\nReactDOM.findDOMNode is removed in React 19, consider using refs instead. From InstUI v9, components must have the "ref" property for findDOMNode to work.\nSee more here: https://instructure.design/#accessing-the-dom`
+      `Error: ${elName} doesn't have "ref" property.\n
+      your code will likely not work with React 19+, because ReactDOM.findDOMNode() was removed.\n
+      See more here: https://instructure.design/#accessing-the-dom`
     )
+    const ReactDomInstance = ReactDOM as any
+    if (typeof ReactDomInstance.findDOMNode === 'function') {
+      return (ReactDOM as any).findDOMNode(node)
+    }
```
matyasf: note: I have not tested this with React 18, we will need to test the whole RC with it (and with React 17)

balzss (approved):
vitest test are not running in CI for some reason but it works fine locally

PR #2096, opened 2025-08-11
WIP fix(ui-avatar): avatar now renders images correctly when using SSR; more regression tests

matyasf:
Also add regression tests for Avatar, Alert. Change regression testing suite to use NextJS 14 because NextJS 15 uses React 19 which InstUI does not support yet


regression-test/package.json
```diff
@@ -8,7 +8,8 @@
     "build": "next build",
     "start": "next start",
     "lint": "next lint",
-    "cypress": "ELECTRON_EXTRA_LAUNCH_ARGS=--remote-debugging-port=9222 cypress run"
+    "cypress": "ELECTRON_EXTRA_LAUNCH_ARGS=--remote-debugging-port=9222 cypress run",
+    "cypress-chrome": "cypress open --browser chrome"
```
matyasf, line 12: This command can be used to launch the `cypress` GUI to debug tests

regression-test/cypress/e2e/spec.cy.ts
```diff
@@ -65,7 +65,7 @@ function terminalLog(violations: Result[]) {
 const axeOptions: { runOnly: RunOnly } = {
   runOnly: {
     type: 'tag',
-    values: ['wcag2a', 'wcag2aa', 'section508', 'best-practice']
```
matyasf, line 68: `best-practice` is not needed for any WCAG rating, its just nitpicks that should not fail a build

regression-test/src/app/avatar/page.tsx
```diff
@@ -0,0 +1,283 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+'use client'
+import React from 'react'
+import { Avatar as avv } from 'instructure-ui/ui-avatar/es/index'
+import {
+  IconGroupLine as igl,
+  IconAiSolid
+} from 'instructure-ui/ui-icons/es/index'
```
matyasf, line 30: This is ugly. But its needed, since full type information is not in the compiled code in the `es` folder, and TS React types are being strict and marks any prop as error.
As a workaround I cast it to `any`, so the type checker does not complain

matyasf:
>  question: in the test plan you mentioned to check if avatar can load images from url but that was working before too, am i correct?

Before it could not load images when using SSR (because in SSR `img` sometimes does not fire the `onLoad` event https://www.linkedin.com/posts/imagekit-io_when-using-the-onload-event-on-img-tags-activity-7318580136155459584-efcv/ ). Now it should work, you can check it by running the regression test

@balzss

PR #2098, opened 2025-08-11
docs: change logo

HerrTopi:
TEST_PLAN:

Check the logo in three places: 

1. favicon
2. top-right corner
3. top of the left drawer tray

 all three should be replaced by the new icon.

logo

INSTUI-4666


adamlobler:
In the drawer an the top-right corner I think those are great, but I think we can improve on the favicon.

In dark mode a favicon like this doesn't have much contrast to the dark background that's why it looks a little bit weird:
Screenshot 2025-08-12 at 17 55 42

+ I'm not sure if we need to support Safari as well, but there the favicon doesn't work at for any reason:
Screenshot 2025-08-12 at 17 57 06

packages/__docs__/src/App/index.tsx
```diff
@@ -727,7 +727,7 @@ class App extends Component<AppProps, AppState> {
         minHeight="100vh"
         width="18.75rem"
       >
-        <View display="block" textAlign="end" margin="xx-small x-small none">
+        <View display="block" textAlign="end" margin="xx-small x-small small">
```
balzss: let's use the newer spacing tokens here, these will be deprecated

HerrTopi: Done

matyasf (approved):
looks good! The whole docs app seems to be much more modern with this design

PR #2101, opened 2025-08-12
fix last 3 failing tests after react 19 upgrade

balzss:
notes:
react 19 only supposed to get rid of `defaultProps` for functional components, however after the update it broke in multiple different class components. my best guess is the `withStyle` decorator which should "forward" the default props but maybe it isn't anymore? I don't know and I wasn't able to find out or verify...

test plan:
- all vitest tests should be green
- check the code. it's not pretty, feels a bit hacky, what do you think? could it be improved?

### DIFF
--- a/packages/ui-drilldown/src/Drilldown/index.tsx
+++ b/packages/ui-drilldown/src/Drilldown/index.tsx
@@ -826,11 +826,13 @@ class Drilldown extends Component<DrilldownProps, DrilldownState> {
     if (groupProps?.selectableType) {
       this.handleGroupOptionSelected(event, selectedOption)
     } else {
+      // TODO workaround for react 19 default props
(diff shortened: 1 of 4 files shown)


packages/ui-drilldown/src/Drilldown/index.tsx
```diff
@@ -826,11 +826,13 @@ class Drilldown extends Component<DrilldownProps, DrilldownState> {
     if (groupProps?.selectableType) {
       this.handleGroupOptionSelected(event, selectedOption)
     } else {
+      // TODO workaround for react 19 default props
+      const optionWithDefaultProps = {...selectedOptionChild, props: {...selectedOptionChild.props, role: 'menuitem'}}
```
matyasf, line 830: Why did you need to add `role: 'menuitem'`?

balzss: it is the default value for the `role` prop

PR #2105, opened 2025-08-14
INSTUI-4672 Add the (almost complete) regression test suite, fix small a11y issues, SSR errors

matyasf:
These tests are mostly AI generated based on our README examples.

- Also fix an SSR error in `Calendar`: inThe weekday header IDs were getting out of sync because they were not deterministically generated
- Also fix an a11y issue in `Drilldown` (see comments)
- Also fix an a11y issue in `Options` (see comments)

Note: The large rendering delays are needed because there is a large layout shift in some components that use `FormFieldLayout` when the hydration happens because we calculate the grid placement in `styles.ts`, so the barebones DOM is a half-complete grid.

To test: 
- Launch the regression test app, check the examples. For now we're not looking for 100% coverage on all visual features, just the happy paths.
- Check the changes in Chromatic

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
     "@types/react-dom": "^18.3.0",
     "@vitejs/plugin-react": "^4.5.1",
     "@vitest/eslint-plugin": "^1.2.1",
+    "@types/node": "^22",
     "chai": "^4.4.1",
     "chalk": "^4.1.2",
     "commitizen": "^4.3.1",
@@ -101,7 +102,7 @@
     "@types/jest": "needed because https://github.com/testing-library/jest-dom/issues/544 recheck if fixed"
   },
   "engines": {
-    "node": ">=18",
+    "node": ">=22",
     "yarn": "YARN NO LONGER USED - use npm instead."
   },
   "config": {
--- a/packages/ui-calendar/src/Calendar/index.tsx
+++ b/packages/ui-calendar/src/Calendar/index.tsx
@@ -28,11 +28,11 @@ import { View } from '@instructure/ui-view'
 import {
   safeCloneElement,
   callRenderProp,
-  omitProps
+  omitProps,
+  withDeterministicId
 } from '@instructure/ui-react-utils'
 import { createChainedFunction } from '@instructure/ui-utils'
 import { logError as error } from '@instructure/console'
-import { uid } from '@instructure/uid'
 import { AccessibleContent } from '@instructure/ui-a11y-content'
 
 import { testable } from '@instructure/ui-testable'
@@ -64,6 +64,7 @@ import { SimpleSelect } from '@instructure/ui-simple-select'
 category: components
 ---
 **/
+@withDeterministicId()
 @withStyle(generateStyle, generateComponentTheme)
 @testable()
 class Calendar extends Component<CalendarProps, CalendarState> {
@@ -82,14 +83,15 @@ class Calendar extends Component<CalendarProps, CalendarState> {
   }
 
   ref: Element | null = null
-  private _weekdayHeaderIds = (
-    this.props.renderWeekdayLabels || this.defaultWeekdays
-  ).reduce((ids: Record<number, string>, _label, i) => {
-    return { ...ids, [i]: uid(`weekday-header-${i}`) }
-  }, {})
+  private _weekdayHeaderIds
+
   constructor(props: CalendarProps) {
     super(props)
-
+    this._weekdayHeaderIds = (
+      this.props.renderWeekdayLabels || this.defaultWeekdays
+    ).reduce((ids: Record<number, string>, _label, i) => {
+      return { ...ids, [i]: this.props.deterministicId!('weekday-header') }
+    }, {})
     this.state = this.calculateState(
       this.locale(),
       this.timezone(),
--- a/packages/ui-codemods/lib/__node_tests__/runTest.ts
+++ b/packages/ui-codemods/lib/__node_tests__/runTest.ts
@@ -45,10 +45,10 @@ export function runTest(codemod: Transform) {
   entries.forEach((entry) => {
     if (entry.isFile() && entry.name.includes('input')) {
       const isWarningTest = entry.name.includes('.warning.input')
-      const inputPath = path.join(entry.path, entry.name)
+      const inputPath = path.join(entry.parentPath, entry.name)
       const input = fs.readFileSync(inputPath, 'utf8')
       const expectedName = entry.name.replace('input', 'output')
-      const expectedPath = path.join(entry.path, expectedName)
+      const expectedPath = path.join(entry.parentPath, expectedName)
       const expected = fs.readFileSync(expectedPath, 'utf8')
 
       // eslint-disable-next-line no-console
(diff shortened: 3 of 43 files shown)


packages/ui-drilldown/src/Drilldown/index.tsx
```diff
@@ -1126,7 +1126,7 @@ class Drilldown extends Component<DrilldownProps, DrilldownState> {
     if (subPageId) {
       optionProps.renderAfterLabel = <IconArrowOpenEndSolid />
       optionProps['aria-haspopup'] = true
-      optionProps.role = customRole || 'button'
+      optionProps.role = customRole || 'menuitem'
```
matyasf, line 1129: Drilldown has `role=menu`, which should not have `role=button` children. We had a similar issue in the past: https://github.com/instructure/instructure-ui/pull/1831 Also see the W3 spec: https://w3c.github.io/aria/#menu . Also this triggers a critical axe check failure (on our docs page too)

regression-test/cypress/e2e/spec.cy.ts
```diff
@@ -37,7 +37,7 @@ afterEach(() => {
   // After each test, assert that console.error was not called
   // Add a small wait if your application might log errors asynchronously
   cy.wait(100).then(() => {
-    expect(windowErrorSpy).to.not.be.called
+    expect(windowErrorSpy).to.have.callCount(0)
```
matyasf, line 40: This makes it to output the actual error message to the console

regression-test/src/app/alert/page.tsx
```diff
@@ -0,0 +1,62 @@
+/*
```
matyasf: Note: All the component test pages were generated by AI using our READMEs as a basis. I dont think the code of these needs more than a very superficial checking, whats important how they look in Chromatic

packages/ui-options/src/Options/Item/styles.ts
```diff
@@ -63,12 +63,11 @@ const generateStyle = (
...
       color: componentTheme.highlightedLabelColor
     },
-    disabled: { cursor: 'not-allowed', opacity: 0.5 },
+    disabled: { cursor: 'not-allowed', opacity: 0.68 },
     'highlighted-disabled': {
       background: componentTheme.highlightedBackground,
       color: componentTheme.highlightedLabelColor,
-      cursor: 'not-allowed',
-      opacity: 0.5
+      cursor: 'not-allowed'
```
matyasf, line 70: These needed the opacity to be increased, otherwise they would fail color contract a11y tests. AFAIK we are not using this directly anywhere, so it should be OK

packages/ui-calendar/src/Calendar/index.tsx
```diff
@@ -82,14 +83,15 @@ class Calendar extends Component<CalendarProps, CalendarState> {
...
+  private _weekdayHeaderIds
+
   constructor(props: CalendarProps) {
     super(props)
-
+    this._weekdayHeaderIds = (
+      this.props.renderWeekdayLabels || this.defaultWeekdays
+    ).reduce((ids: Record<number, string>, _label, i) => {
+      return { ...ids, [i]: this.props.deterministicId!('weekday-header') }
+    }, {})
```
matyasf, line 94: SSR bugfix due to ID not being deterministic

packages/ui-heading/src/Heading/README.md
```diff
@@ -49,9 +49,9 @@ Pre-configured and with unique styles, the `ai-headings` are used for standardiz
 type: example
 ---
 <div style={{display: 'flex', flexDirection: 'column', gap: '24px'}}>
- <Heading aiVariant="stacked">Nutrition Facts</Heading>
- <Heading aiVariant="horizontal">Nutrition Facts</Heading>
- <Heading aiVariant="iconOnly">Nutrition Facts</Heading>
+  <Heading aiVariant="stacked" level="h2">Nutrition Facts</Heading>
+  <Heading aiVariant="horizontal" level="h3">Nutrition Facts</Heading>
+  <Heading aiVariant="iconOnly" level="h4">Nutrition Facts</Heading>
```
matyasf, line 54: `variant` should be specified according to our own docs

package.json
```diff
@@ -101,7 +102,7 @@
     "@types/jest": "needed because https://github.com/testing-library/jest-dom/issues/544 recheck if fixed"
   },
   "engines": {
-    "node": ">=18",
+    "node": ">=22",
```
matyasf, line 105: I've updated this to be in sync with what we use in Github actions

packages/ui-codemods/lib/__node_tests__/runTest.ts
```diff
@@ -45,10 +45,10 @@ export function runTest(codemod: Transform) {
   entries.forEach((entry) => {
     if (entry.isFile() && entry.name.includes('input')) {
       const isWarningTest = entry.name.includes('.warning.input')
-      const inputPath = path.join(entry.path, entry.name)
+      const inputPath = path.join(entry.parentPath, entry.name)
```
matyasf, line 48: `entry.path` is removed in Node 24, this is the recommended way since Node 20

package.json
```diff
@@ -64,6 +64,7 @@
     "@types/react-dom": "^18.3.0",
     "@vitejs/plugin-react": "^4.5.1",
     "@vitest/eslint-plugin": "^1.2.1",
+    "@types/node": "^22",
```
matyasf, line 67: This must be added otherwise it will just pick up a `@types/node` version from a dependency

regression-test/cypress/e2e/spec.cy.ts
```diff
@@ -89,10 +89,181 @@ describe('visual regression test', () => {
     cy.checkA11y('.axe-test', axeOptions, terminalLog)
   })
 
+  it('check badge', () => {
```
balzss: please remove "check" from test names so it is cleaner (esp. in chromatic ui)

regression-test/src/app/button/page.tsx
```diff
@@ -41,20 +54,117 @@ export default function ButtonPage() {
...
+        >
+          <Flex.Item shouldShrink shouldGrow>
+            <h2>A heading</h2>
+          </Flex.Item>
+          <Flex.Item padding="none none none medium">
+            <CloseButton size="medium" screenReaderLabel="Close" />
+          </Flex.Item>
+        </Flex>
+      </View>
+      <div style={{ display: 'flex', gap: '0.5rem' }}>
```
balzss, line 108: these iconbuttons look "squashed" in chromatic:

Image

balzss (changes requested):
there are some strange things with text inputs:
1. they don't display focusrings in storybook mode (other components, e.g. buttonns or checkboxes do)
2. textarea has the right and left side cut off:

Image

3. left padding seems to be missing?

Image

regression-test/src/app/button/page.tsx
matyasf: I could fix this with a small delay

regression-test/cypress/e2e/spec.cy.ts
matyasf: done

packages/ui-heading/src/Heading/README.md
joyenjoyer: You mean `level` instead of `variant` I guess.

HerrTopi (approved):

PR #2106, opened 2025-08-14
fix(ui-form-field): should not change layout when only screenreader-only message is present

HerrTopi:
TEST_PLAN:
try setting a `FormField` `messages` prop  with a `screenreader-only` message and check if the visuals change from a `FormField` without any `messages`. 
(add this prop for example: `messages={[{ text: "announcement", type: 'screenreader-only' }] }`)
Check if any other view of `FormField` has changed

INSTUI-4673

### DIFF
--- a/packages/ui-form-field/src/FormFieldLayout/styles.ts
+++ b/packages/ui-form-field/src/FormFieldLayout/styles.ts
@@ -75,10 +75,16 @@ const generateStyle = (
   props: FormFieldLayoutProps,
   styleProps: FormFieldStyleProps
 ): FormFieldLayoutStyle => {
-  const { inline, layout, vAlign, labelAlign, margin } = props
+  const { inline, layout, vAlign, labelAlign, margin, messages } = props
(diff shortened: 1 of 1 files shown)


matyasf (approved):
Looks good. The SR messages are still read and I could not notice any DOM difference here: https://instructure.design/pr-preview/pr-2106/#FormField

ToMESSKa (changes requested):
@HerrTopi If I set `messages={[{ text: "announcement", type: 'screenreader-only' }] }` in the first example in DateTimeInput, the space between the input field and the green message disappears, can you check this please?

ToMESSKa:
@HerrTopi I also noticed that if I add the screenreader-only message to a checkbox in the /#Checkbox page, the layout slightly changes, for example adding the message to 'Science 101' results in a bigger gap between 'English 203' and 'Science 101' then the gap between  'Science 101'  and ' 'History 111'  . This might worth checking out.
image

PR #2107, opened 2025-08-15
fix(ui-modal): fix Modal.body throwing error when 'as' is set to a React component

matyasf:
The new code now digs recursively in the 'ref' prop until it finds the actual DOM element and returns that one

To test:

1. Create a Modal.Body component with a custom 'as' prop set to a React component e.g. `<Flex>`
2. Create a Modal.Body component with a custom 'as' prop set to a native element, e.g. `<section>`

Fixes INSTUI-4671


git-nandor (approved):

HerrTopi (changes requested):
Yeah... Let's rethink "as" as soon as possible :D

HerrTopi (approved):

PR #2108, opened 2025-08-22
fix(ui-dialog,ui-a11y-utils): fix focus getting stuck in some cases if something is removed from the middle of the focus stack

matyasf:
The issue occurred in the following conditions:

1. A non-keyboard focusable element is added to the top of the stack,e.g. an Overlay with just text
2. The keyboard focusable element is removed from the stack just below it. In this case the focus region stayed the now removed element getting the focus stuck

To test:
- Run the code in the Cypress test locally
- Test out ALL the examples that use popovers/modals. Play with their popover settings like `shouldReturnFocus`, `shouldCloseOnDocumentClick`, `shouldContainFocus`

Fixes INSTUI-4685

### DIFF
--- a/packages/ui-a11y-utils/src/FocusRegion.ts
+++ b/packages/ui-a11y-utils/src/FocusRegion.ts
@@ -67,13 +67,15 @@ class FocusRegion {
       shouldCloseOnEscape: true,
       isTooltip: false
     }
+    this._id = uid()
+    // eslint-disable-next-line no-param-reassign
+    options.regionId = this._id
     this._contextElement = element
     this._screenReaderFocusRegion = new ScreenReaderFocusRegion(
       element,
       options
     )
     this._keyboardFocusRegion = new KeyboardFocusRegion(element, options)
-    this._id = uid()
   }
 
   updateElement(element: Element | Node, options?: FocusRegionOptions) {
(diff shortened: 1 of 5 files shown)


packages/ui-a11y-utils/src/FocusRegion.ts
```diff
@@ -211,17 +213,15 @@ class FocusRegion {
   }
 
   deactivate({ keyboard = true }: { keyboard?: boolean } = {}) {
-    if (this._active) {
```
matyasf, line 214: Removing this line is the fix. The issue was that when the non-top region (the `Tray` in the example) is removed then the code below is never running since its not the active region.

packages/ui-a11y-utils/src/FocusRegionOptions.ts
```diff
@@ -86,4 +86,8 @@ export type FocusRegionOptions = {
    * Whether or not the element is a Tooltip
    */
   isTooltip?: boolean
+  /**
+   * The ID of the `FocusRegion` this belongs to. Used only for debugging.
+   */
+  regionId?: string
```
matyasf, line 92: This just helps debug similar issues since it lets you see which focus region a component belongs to

joyenjoyer (approved):

PR #2110, opened 2025-08-27
fix(ui-buttons): disabled button with href should not receive focus

joyenjoyer:
INSTUI-4687

Test:
- add disabled and href properties to Button
- observe if it is focusable (**it should NOT be**), try it in different browsers
- try it with ScreenReaders (it should be read as a disabled/dimmed link)


git-nandor (approved):

ToMESSKa (changes requested):
@joyenjoyer In Safari (Version 18.6 (20621.3.11.11.3)), Button is still focusable (it has a focus ring) when disabled with a href. Can you please check this?

PR #2112, opened 2025-08-28
fix(ui-tabs,shared-types): color tabs.panel focus outline from the th…

HerrTopi:
…eme and make it overridable by brand overrides

Test plan: check tabs.panel's focus outline. It should be the brand blue now. Check if the ic-brand-primary is overridden, it changes color

INSTUI-4690


matyasf (approved):
works, I could test it in https://instructure.design/#using-theme-overrides/#Branding%20(user%20customizable%20theming)

image

ToMESSKa (approved):

PR #2114, opened 2025-08-28
fix(ui-buttons): fix secondary ai iconbutton when shape is circle

HerrTopi:
Test plan: check if the circle shaped ai  iconbuttons look as the supposed to look

INSTUI-4691


matyasf (approved):

matyasf (changes requested):
Looks good! Can you please add this to the VRT? https://github.com/instructure/instructure-ui/blob/master/regression-test/src/app/button/page.tsx *(I think you'll need to rebase since I merged it today)

ToMESSKa (approved):

matyasf (approved):

PR #2117, opened 2025-08-29
[v11] INSTUI-4594 feat(many): remove global themes

matyasf:
Remove the global theme registry, update the documentation, examples, tests.
This breaking change was needed because global theme objects would make running different versions of InstUI hard when using module federation


matyasf:
merged with cherry picking

PR #2125, opened 2025-09-03
fix(ui-a11y-utils): fix focus region skipping items after Drilldown/scrollbars

matyasf:
The core of the issue is that findFocusable cannot detect everything that can be focused reliably, e.g. browsers now focus on DOM elements that have scrollbars. The deleted code added the element with the scrollbar to the end of the focus  region array, causing anything after it skipped when tabbing.
The downside of deleting this code is that focus order breaks when the last element in a Popover/Modal is something that findFocusable does not detect e.g. a Menu (without a trigger,
 this is not detected because its tabIndex changes dynamically)

To test: Test modals/popovers with keyboard navigation.
Test this code:

```jsx
  constructor(props) {
    super(props)
    this.state = {
      open: false
    }
  }

  handleButtonClick = () => {
    this.setState(function (state) {
      return { open: !state.open }
    })
  }

  render() {
    return (
      <div style={{ padding: '0 0 11rem 0', margin: '0 auto' }}>
        <Button onClick={this.handleButtonClick}>
          {this.state.open ? 'Close' : 'Open'} the Modal
        </Button>
        <Modal
          as="form"
          open={this.state.open}
          onDismiss={() => {
            this.setState({ open: false })
          }}
          size="auto"
          label="Hello World"
          shouldCloseOnDocumentClick
        >
          <Modal.Header>
            <CloseButton placement="end" onClick={this.handleButtonClick} screenReaderLabel="Close"/>
          </Modal.Header>
          <Modal.Body>
            <button>a button</button>
              <Drilldown rootPageId='root' width="20rem" height="5rem">
  <Drilldown.Page id="root">
    <Drilldown.Option id="option1">Option</Drilldown.Option>
    <Drilldown.Option id="option2">option2</Drilldown.Option>
    <Drilldown.Option id="option3"><Pill>Pill</Pill> Option</Drilldown.Option>
    <Drilldown.Option id="option4" disabled>Option</Drilldown.Option>
  </Drilldown.Page>
           </Drilldown>
            after
             <button>a button</button>
          </Modal.Body>
          <Modal.Footer>
            <Button onClick={this.handleButtonClick} margin="0 x-small 0 0">Close</Button>
            <Button color="primary" type="submit">Submit</Button>
          </Modal.Footer>
        </Modal>
      </div>
    )
  }
}

render(<Example />)
```

Fixes INSTUI-4694


regression-test/src/app/layout.tsx
```diff
@@ -38,6 +38,9 @@ export default function RootLayout({
     // we need to make a new Map to reset counting on the server side
     // on each page refresh TODO fix
     <html lang="en">
+      <head>
+        <title>Component visual and regression test suite</title>
+      </head>
```
matyasf, line 43: just a small fix that reduces small issues that Axe complains about

regression-test/src/app/calendar/page.tsx
```diff
@@ -30,9 +30,6 @@ const Calendar = cl as any
 export default function CalendarPage() {
   return (
     <main className="flex gap-8 p-8 flex-col items-start axe-test">
-      {/* Default config */}
-      <Calendar />
-
```
matyasf, line 35: unrelated regression test that was not needed because it just broke every time since it always displays the current date

joyenjoyer (approved):

PR #2131, opened 2025-09-04
fix(ui-color-picker): fix misaligned popover button

HerrTopi:
Test plan: check if the popover button is aligned correctly

INSTUI-4695


joyenjoyer (approved):

balzss (commented):
this solution breaks if messages are present

PR #2132, opened 2025-09-04
fix(ui-view,ui-buttons): clarify typing of elementRefs

HerrTopi:
Test plan: Check and think about what could go wrong since this change narrows the type of all View elements to HTMLElement from Element

INSTUI-4696


matyasf (approved):
looks good!

balzss (approved):

PR #2137, opened 2025-09-09
feat(ui-tree-browser): add animation prop to TreeBrowser

ToMESSKa:
INSTUI-4683

### DIFF
--- a/packages/ui-tree-browser/src/TreeBrowser/TreeBrowserContext.ts
+++ b/packages/ui-tree-browser/src/TreeBrowser/TreeBrowserContext.ts
@@ -1,0 +1,49 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015 - present Instructure, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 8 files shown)


packages/ui-tree-browser/src/TreeBrowser/TreeButton/styles.ts
```diff
@@ -226,14 +228,16 @@ const generateStyles = (
...
       zIndex: 1,
-      transform: 'translate3d(-2%, 0, 0)',
-      opacity: 0.01,
-      transformOrigin: 'left center',
-      animationName: transform,
-      animationDuration: '0.2s',
-      animationFillMode: 'forwards',
-      animationTimingFunction: 'ease-out',
-      animationDelay: '0.2s',
+      ...(state.animation !== false && {
```
HerrTopi: just a nitpic: you don't have to check with `!==`, since animation is a `boolean` itself. It would look better to use just the value itself: `state.animation && {...`

Another nitpic of mine is `false && {}` fill resolve to `false`. Even so spreading `false` into an object works for some reason (that's js for you :D ) it doesn't make much sense. You could write these expressions like:
```suggestion
      ...(state.animation ? {<animation code>} : {}
```
This way it is clearly stated what you indend in both branches

packages/ui-tree-browser/src/TreeBrowser/TreeCollection/styles.ts
```diff
@@ -115,12 +116,14 @@ const generateStyles = (
           bottom: '1.1875rem',
           insetInlineStart: '0',
           insetInlineEnd: 'auto',
-          transform: 'scaleY(0.01)',
-          transformOrigin: 'center top',
-          animationName: list,
-          animationDuration: '0.2s',
-          animationFillMode: 'forwards',
-          animationTimingFunction: 'ease-out',
+          ...(state.animation !== false && {
```
HerrTopi: Same as prev comments

packages/ui-tree-browser/src/TreeBrowser/TreeButton/styles.ts
ToMESSKa: @HerrTopi thanks, I made the changes you suggested

packages/ui-tree-browser/src/TreeBrowser/TreeBrowserContext.ts
```diff
@@ -0,0 +1,49 @@
...
+ * ---
+ * category: utilities/react
+ * ---
+ * React context created by the `TreeBrowser` component to hold its data which are
+ * read by its children.
+ *
+ * @module
+ */
+const TreeBrowserContext = createContext<TreeBrowserContextType>({
+  animation: false
```
matyasf: I think this should be `true` for that very edge case if someone uses these components outside of a TreeBrowser

ToMESSKa: @matyasf thanks, I fixed that.

HerrTopi (approved):

PR #2141, opened 2025-09-12
fix(ui-utils,ui-drilldown): make NVDA read options in Drilldown.Group…

ToMESSKa:
… correctly in Firefox

INSTUI-4330

**ISSUE:**

NVDA was not properly announcing options within Drilldown.Group components in Firefox. The issues were:
  * all options were announced simultaneously on focus
  * the `checked` state of the selected option was not read
  * options were not navigable with keyboard

**TEST PLAN:**

- with Firefox (!) and NVDA, inspect the 'Single-select Group' example
- go through the option, they should be navigable with keyboard and the checked option should be announced as checked and the options should not be announced
- other examples with a Drilldown.Group should behave the same like above (e.g.: 'Video Settings' example, 'Option Groups' example)
- on Mac with Chrome, Firefox and Safari, examples with a Drilldown.Group should have a role='group' instead of 'menu'
- because I could not define a screenreader-specific condition, the role is also changed to 'menu' in Firefox+JAWS. This results in the group titles not always being read in Firefox+JAWS, but it is still a less severe issue than the one with Firefox+NVDA.

### DIFF
--- a/packages/ui-drilldown/src/Drilldown/DrilldownGroup/index.tsx
+++ b/packages/ui-drilldown/src/Drilldown/DrilldownGroup/index.tsx
@@ -29,6 +29,7 @@ import { optionsThemeGenerator } from '@instructure/ui-options'
 
 import { propTypes, allowedProps } from './props'
 import type { DrilldownGroupProps } from './props'
+import { isMac, isFirefox } from '@instructure/ui-utils'
 
 /**
 ---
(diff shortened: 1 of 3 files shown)


matyasf (approved):
looks (sounds) good!

joyenjoyer (approved):

PR #2148, opened 2025-09-18
chore(many): make documentation  preview builds show console messages

ToMESSKa:
INSTUI-4522

**ISSUE:**
- console messages which are displayed in development are not shown in the preview build

**TEST PLAN:**

- test code examples below, all of them should emit a console messages in the browser's console

```
<Alert screenReaderOnly> This is alert </Alert>
```

```
<Heading variant="titlePageDesktop" as="div" > titlePageDesktop </Heading>
```

```
function TestComponent(props) {
  const filteredProps = View.omitViewProps(props, TestComponent)
  
  return (
    <View {...filteredProps}>
      Content
    </View>
  )
}

render(<TestComponent invalidProp="value" display="block" />)
```

```
<InstUISettingsProvider dir="auto">
  <div>LTR text</div>
</InstUISettingsProvider>
```

- based on the logic, the console messages should not appear in production if it's not a Github preview
- these console messages should appear in the local development build too

### DIFF
--- a/packages/__docs__/webpack.config.mjs
+++ b/packages/__docs__/webpack.config.mjs
@@ -30,9 +30,12 @@ import { globbySync } from 'globby'
 import { merge } from 'webpack-merge'
 import { processSingleFile } from './lib/build-docs.mjs'
 import resolve from './resolve.mjs'
+import webpack from 'webpack'
+import TerserPlugin from 'terser-webpack-plugin'
 
 const ENV = process.env.NODE_ENV || 'production'
 const DEBUG = process.env.DEBUG || ENV === 'development'
+const GITHUB_PULL_REQUEST_PREVIEW = process.env.GITHUB_PULL_REQUEST_PREVIEW || 'false'
 
 const outputPath = resolvePath(import.meta.dirname, '__build__')
 const resolveAliases = DEBUG ? { resolve } : {}
@@ -78,6 +81,9 @@ const config = merge(baseConfig, {
     new HtmlWebpackPlugin({
       template: './src/index.html',
       chunks: ['main'],
+    }),
+    new webpack.DefinePlugin({
+      'process.env.GITHUB_PULL_REQUEST_PREVIEW': JSON.stringify(GITHUB_PULL_REQUEST_PREVIEW),
     }),
   ],
   optimization: {
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -12,6 +12,8 @@ on:
 jobs:
   deploy-preview:
     runs-on: ubuntu-latest
+    env:
+      GITHUB_PULL_REQUEST_PREVIEW: 'true'
     steps:
       - uses: actions/checkout@v4
       - name: Install Node 22
--- a/packages/console/src/console.ts
+++ b/packages/console/src/console.ts
@@ -55,7 +55,13 @@ function logMessage(
   message: string,
   ...args: unknown[]
 ) {
-  if (process.env.NODE_ENV !== 'production' && !condition) {
+  const isGitHubPullRequestPreview =
+    typeof process !== 'undefined' &&
+    process?.env?.GITHUB_PULL_REQUEST_PREVIEW === 'true'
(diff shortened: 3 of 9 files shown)


packages/__docs__/buildScripts/utils/getClientProps.mts
```diff
@@ -113,7 +113,7 @@ export function getClientProps(docs: ProcessedFile[], library: LibraryOptions) {
 }
 
 function warning(message: string, ...args: any[]) {
-  if (process.env.NODE_ENV !== 'production' && typeof console !== 'undefined') {
+  if ((process.env.NODE_ENV !== 'production' || process.env.GITHUB_PULL_REQUEST_PREVIEW === 'true') && typeof console !== 'undefined') {
```
ToMESSKa: Aside from the babel.configs, I also added the option to display console messages in the preview build where I noticed in the code that they are shown when the app is not in production (like this) I'm not sure if these are needed, so I can revert them.

packages/__docs__/buildScripts/utils/getClientProps.mts
```diff
@@ -113,7 +113,7 @@ export function getClientProps(docs: ProcessedFile[], library: LibraryOptions) {
...
     console.warn(`Warning: ${message}`, ...args)
```
ToMESSKa: Aside from the babel.configs, I also added the option to display console messages in the preview build where I noticed in the code that they are shown when the app is not in production (like this) I'm not sure if these are needed, so I can revert them. (see View, InstUISettingsProvider, getTheme, useTheme, textDirectionContextConsumer)

matyasf: I think these are needed, we do want to see these messages. (expect here, this file is not executed in the browser, but during build time... I dont event understand why we have the condition here, this should be always shown)

packages/ui-file-drop/babel.config.js
```diff
@@ -29,7 +29,9 @@ module.exports = {
       {
         coverage: Boolean(process.env.COVERAGE),
         esModules: Boolean(process.env.ES_MODULES),
-        removeConsole: process.env.NODE_ENV === 'production',
+        removeConsole:
+          process.env.NODE_ENV === 'production' &&
+          process.env.GITHUB_PULL_REQUEST_PREVIEW !== 'true',
```
matyasf: I think we can so this in a much easier way. You don't need to change any of the `babel.config.js` files, just `ui-babel-preset/lib/index.js` line 111:

```if (opts.removeConsole) {``` to

```if (process.env.GITHUB_PULL_REQUEST_PREVIEW !== 'true' && opts.removeConsole) {```

This should keep console.log in our preview builds

File: packages/__docs__/webpack.config.mjs [hunk 4]
```diff
@@ -79,6 +82,9 @@ const config = merge(baseConfig, {
       template: './src/index.html',
       chunks: ['main'],
     }),
+    new webpack.DefinePlugin({
+      'process.env.GITHUB_PULL_REQUEST_PREVIEW': JSON.stringify(GITHUB_PULL_REQUEST_PREVIEW),
+    }),
```
matyasf, line 87: just curious, why is this needed?

ToMESSKa: @matyasf I tried to omit this plugin, but it results in process.env.GITHUB_PULL_REQUEST_PREVIEW becoming undefined in the browser. This makes environment variable values available in client-side.

packages/ui-file-drop/babel.config.js
ToMESSKa: @matyasf Good point, thanks, I fixed this

packages/__docs__/buildScripts/utils/getClientProps.mts
ToMESSKa: @matyasf I removed this one from ClientProps.mts

matyasf (approved):
nice work, this will help a lot with quick debugging

PR #2150, opened 2025-09-19
fix(ui-drawer-layout,ui-a11y-utils): fix DrawerLayout not closing on ESC keypress

matyasf:
The issue was that DrawerLayout updates the FocusRegion when opening multiple times by changing the value of `shouldCloseOnEscape` (by calling `updateElement`), but FocusRegion is not adding/removing event listeners when its options change. Also add unit tests for FocusRegion (AI generated)

To test:
check if the DrawerLayout examples close when ESC is pressed. They should not close if `shouldCloseOnEscape` is set to `false`

fixes: INSTUI-4731

### DIFF
--- a/packages/ui-a11y-utils/src/__tests__/FocusRegion.test.tsx
+++ b/packages/ui-a11y-utils/src/__tests__/FocusRegion.test.tsx
@@ -1,0 +1,273 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015 - present Instructure, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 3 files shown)


packages/ui-a11y-utils/src/__tests__/FocusRegion.test.tsx
```diff
@@ -0,0 +1,301 @@
...
+      })
+
+      focusRegion.activate()
+
+      // Create a tooltip element
+      const tooltip = document.createElement('div')
+      tooltip.setAttribute('role', 'tooltip')
+      const tooltipContent = document.createElement('span')
+      tooltip.appendChild(tooltipContent)
+      document.body.appendChild(tooltip)
```
ToMESSKa: I think this test should use the InstUI Tooltip component because I don't think a plain div with a role='tooltip' would cause any issue because it does not have its own FocusRegion like Tooltip does.

But it think there is already a similar cypress test for testing this specific scenario, so maybe this test is not needed:
https://github.com/instructure/instructure-ui/pull/1905/files#diff-3321118c533c8df390253811ae334fb1a5fe9b302f85934a7afd42c498f2c196

packages/ui-drawer-layout/src/DrawerLayout/DrawerTray/props.ts
```diff
@@ -118,7 +118,11 @@ type PropsPassedToDialog = {
   shouldReturnFocus?: boolean
 
   shouldCloseOnDocumentClick?: boolean
-
+  /**
+   * Should the `<DrawerLayout.Tray />` close when ESC is pressed?
```
ToMESSKa: I don't think the question mark is needed here

ToMESSKa (changes requested):

packages/ui-a11y-utils/src/__tests__/FocusRegion.test.tsx
matyasf: you are right, I've deleted it

PR #2153, opened 2025-09-23
[v11] remove not needed dependency, remove not needed warning

matyasf:
- some `instructure/theme-registry` dependencies were still there, this package was removed
-  findDOMNode tries to access `elem?.props?.ref` after it has tried `elem.ref` because accessing`elem?.props?.ref` triggers a warning
- downgrade docs app React 18 because this is what most use. This also required the downgrade of `@testing-library/react`

to test: check things that display a popup that they work well and there are no console errors/warnings.

### DIFF
--- a/packages/ui-dom-utils/src/findDOMNode.ts
+++ b/packages/ui-dom-utils/src/findDOMNode.ts
@@ -30,10 +30,14 @@ import ReactDOM from 'react-dom'
  * React 18 (element.ref) and React 19+ (element.props.ref)
  */
 function getElementRef(elem: { ref?: any; props?: { ref: any } }) {
-  if (elem?.props?.ref !== undefined) {
-    return elem.props.ref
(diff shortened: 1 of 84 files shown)


packages/ui-dom-utils/src/findDOMNode.ts
```diff
@@ -30,10 +30,14 @@ import ReactDOM from 'react-dom'
...
+  // Note that this is added by InstUI components like Position
+  if (elem.ref) {
+    return elem.ref
   }
-  return elem.ref
+  // note that trying to access this will display a warning in React 18.
+  // this is OK because without it React itself would show an error because
+  // its using findDOMNode
+  return elem?.props?.ref
 }
```
matyasf, line 41: This is the main change. The issue was that BreadCrumb example was showing "ContextView: `ref` is not a prop" warnings with React 18

PR #2154, opened 2025-09-24
feat(ui-table): add column ordering to caption

joyenjoyer:
INSTUI-4699

**Test plan**:

1. Move to the Table docs page
2. Turn on NVDA/VoiceOver/Jaws
3. Focus the Table example that can be ordered by columns
4. Observe whether the Screenreader announce ordering - it has to announce the right column, right order
5. Try changing the ordering and observe whether it announce the changed column and order

### DIFF
--- a/packages/ui-table/src/Table/README.md
+++ b/packages/ui-table/src/Table/README.md
@@ -582,10 +582,7 @@ By default, the options in the `Select` for sorting in stacked layout are genera
                 </View>
               )}
 
-              <Table
-                caption={`${caption}: sorted by ${sortBy} in ${direction} order`}
-                {...props}
-              >
+              <Table {...props}>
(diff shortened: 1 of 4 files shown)


git-nandor (approved):

ToMESSKa (changes requested):
@joyenjoyer Something I noticed, in the screenshot below it should announce 'Sorted by Rating' (the little arrow is even blue), but it still says 'Sorted by Rank'.
Steps to reproduce:
1. Turn on VoiceOver
2. Navigate to the column Rating and press 'Space' to order it
3. Navigate out of the Table then back again
4. It says 'Sorted by Rank'

What I noticed is that if I sort the Table if before turning on VoiceOver, it announces the correct column.
But I also tried it in NVDA and it also mistakenly announces rank when it should be rating.
image

PR #2155, opened 2025-09-24
Set table caption to required

joyenjoyer:
INSTUI-4746


packages/ui-codemods/lib/__node_tests__/codemods.test.ts
```diff
@@ -49,4 +52,33 @@ describe('test codemods', () => {
...
+      'utf-8'
+    )
+    const withCaptionInput = await fs.readFile(
+      require.resolve(
+        './__testfixtures__/warnTableCaptionMissing/withCaption.input.tsx'
+      ),
+      'utf-8'
+    )
+    // Test: should warn if Table is missing caption
+    const spy = vi.spyOn(codemodHelpers, 'printWarning')
```
git-nandor: I think next time you can expect the warning messages to be handled more simply inside __textfixtures__, similar to how it’s done in renameCanvasThemesCodemod with the .warning.input.tsx and .warning.output.tsx naming conventions.

packages/ui-codemods/lib/__node_tests__/__testfixtures__/warnTableCaptionMissing/withCaption.input.tsx
```diff
@@ -0,0 +1,18 @@
...
+      <thead>
+        <tr>
+          <th>Header</th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr>
+          <td>Data</td>
+        </tr>
+      </tbody>
```
matyasf: this is a very unrealistic test case, please use e.g. a basic `<Table>` example

packages/ui-codemods/lib/__node_tests__/codemods.test.ts
matyasf: Yes, please use here the `.warning` type tests.
Also do not write any test logic in this file, use the same pattern as other tests

packages/ui-codemods/lib/warnTableCaptionMissing.ts
```diff
@@ -0,0 +1,77 @@
...
+  findImport,
+  findElement,
+  findAttribute,
+  printWarning
+} from './utils/codemodHelpers'
+import type { InstUICodemod } from './utils/instUICodemodExecutor'
+
+const TABLE_IMPORT_PATHS = ['@instructure/ui-table', '@instructure/ui']
+
+const warnTableCaptionMissing: InstUICodemod = (
```
matyasf: please use here the same format as other codemods. it needs to export 2 functions, the transformer is not needed

PR #2159, opened 2025-09-25
fix(ui-drilldown): fix Drilldown for prevent option selection when disabled

git-nandor:
INST-UI4752

Add fix for prevent option selection when Drilldown or its sub-components get disabled prop.

Test plan:
Go to Drilldown doc page and at the Disabled prop section ensure that its work as expected in all sub-component level

### DIFF
--- a/packages/ui-drilldown/src/Drilldown/DrilldownGroup/__tests__/DrilldownGroup.test.tsx
+++ b/packages/ui-drilldown/src/Drilldown/DrilldownGroup/__tests__/DrilldownGroup.test.tsx
@@ -234,6 +234,26 @@ describe('<Drilldown.Group />', () => {
       expect(option_1).not.toHaveAttribute('aria-disabled')
       expect(option_2).not.toHaveAttribute('aria-disabled')
     })
+
+    it('should not allow selection if the Drilldown.Group is disabled', async () => {
+      render(
+        <Drilldown rootPageId="page0">
+          <Drilldown.Page id="page0">
+            <Drilldown.Group id="group0" selectableType="multiple" disabled>
+              <Drilldown.Option id="opt1">Disabled Option</Drilldown.Option>
+            </Drilldown.Group>
+          </Drilldown.Page>
+        </Drilldown>
+      )
+      const optionItemContainer = screen.getByLabelText('Disabled Option')
+      const optionContent = screen.getByText('Disabled Option')
+
+      expect(optionItemContainer).toHaveAttribute('aria-checked', 'false')
+
+      await userEvent.click(optionContent)
+
+      expect(optionItemContainer).toHaveAttribute('aria-checked', 'false')
+    })
   })
 
   describe('role prop', () => {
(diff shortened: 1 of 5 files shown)


packages/ui-drilldown/src/Drilldown/DrilldownGroup/__tests__/DrilldownGroup.test.tsx
```diff
@@ -234,6 +234,26 @@ describe('<Drilldown.Group />', () => {
       expect(option_1).not.toHaveAttribute('aria-disabled')
       expect(option_2).not.toHaveAttribute('aria-disabled')
     })
+
+    it('should not allow selection if the Drilldown.Group is disabled', async () => {
+      render(
+        <Drilldown disabled rootPageId="page0">
+          <Drilldown.Page id="page0">
+            <Drilldown.Group id="group0" selectableType="multiple">
+              <Drilldown.Option id="opt1">Disabled Option</Drilldown.Option>
```
ToMESSKa, line 243: Did you mean to put the disabled prop on Drilldown.Group? Because now it's on the Drilldown.

packages/ui-drilldown/src/Drilldown/index.tsx
```diff
@@ -785,13 +785,19 @@ class Drilldown extends Component<DrilldownProps, DrilldownState> {
...
+      return
+    }
+
+    const isOptionDisabled =
+      this.props.disabled ||
+      this.currentPage?.disabled ||
+      selectedOption.groupProps?.disabled ||
+      selectedOption.props.disabled
+
+    if (isOptionDisabled) {
```
ToMESSKa, line 802: If I set disabled={true} only in the second page then I cannot go back to the first page from the second one. The docs says 'The only option that can not be disabled is the Back navigation.', so the back button should be an exception in my opinion.

packages/ui-drilldown/src/Drilldown/index.tsx
```diff
@@ -785,13 +785,19 @@ class Drilldown extends Component<DrilldownProps, DrilldownState> {
...
-      !id ||
-      !selectedOption ||
-      selectedOption.props.disabled ||
-      (event.target as HTMLElement).getAttribute('disabled') ||
-      (event.target as HTMLElement).getAttribute('aria-disabled')
-    ) {
+    if (!id || !selectedOption) {
+      event.preventDefault()
+      event.stopPropagation()
+      return
```
ToMESSKa, line 791: Is there a specific reason why there are two return statements now instead of handing the isOptionDisabled condition in line 788?

git-nandor: I think it’s clearer if they are kept separate, it becomes quite hard to follow if I combine them, and they serve different purposes. The first case is more of a security check, while the second one is part of the business logic.

ToMESSKa (approved):

joyenjoyer (approved):

PR #2177, opened 2025-10-06
feat: add bootstrap-lite

joyenjoyer:


scripts/README-bootstrap-lite.md
```diff
@@ -0,0 +1,136 @@
...
+- ❌ Clean step (no file deletion)
+- ❌ Babel builds (no transpilation to `lib/` and `es/` directories)
+- ❌ Icon generation
+- ❌ Token generation
+
+## Performance
+
+- **bootstrap:lite**: ~1-2 minutes ⚡ (or ~1 second with build cache!)
+- **bootstrap:lite --force**: ~1-2 minutes ⚡ (bypasses cache, rebuilds everything)
+- **bootstrap**: ~5-10 minutes 🐌
```
joyenjoyer, line 59: I will correct (or remove) these estimations.

balzss (commented):
after reading through the generated script, it seems like it basically does two things: builds ts and lints code with lots of comments and logging. do we need such a script? isn't `npm run build:types && npm run lint` basically the same thing? also I don't understand why there is a markdown file AND a lengthy description command in the source script itself

PR #2184, opened 2025-10-08
docs(many): remove class based examples

matyasf:
keep only functional examples
remove code that rendered both
rearrange components a bit: move ScreenReaderContent/...Content to utilites/a11y, move Options and Selectable to Utilities

To test: Check examples in docs, they should all work like before. Check the new sidebar order

INSTUI-4753


packages/__docs__/src/compileMarkdown.tsx
```diff
@@ -291,7 +279,8 @@ const renderer = (title?: string) => ({
 })
 
 function compileMarkdown(content: string, title?: string) {
-  return <Markdown renderer={renderer(title)}>{content}</Markdown>
+  const ti = title ? title : 'Untitled'
+  return <Markdown renderer={renderer(ti)}>{content}</Markdown>
```
balzss: nitpick: just use `renderer(title ?? 'Untitled')`

balzss (approved):
tested some of the components manually randomly, everything works. the code also looks ok. nice work

docs/contributor-docs/writing-docs.md
```diff
@@ -233,109 +189,57 @@ render() {
...
+The above code will display like this:
 
-````md
+```js
 ---
-type: code
+type: example
 ---
+class Example extends React.Component {
+constructor(props) {
```
ToMESSKa: maybe replacing this with a functional example too to reflect that we no longer use class-based examples?

docs/guides/accessing-the-dom.md
ToMESSKa, line 45: How about this example?

packages/emotion/src/InstUISettingsProvider/index.tsx
```diff
@@ -61,7 +61,7 @@ type InstUIProviderProps = {
 
 /**
  * ---
- * category: components/utilities
+ * category: components
  * ---
```
ToMESSKa, line 65: Was moving InstUISettingsProvider to components done on purpose?

matyasf: Yes, this is a component that should be used by everyone, e.g. we mention it in our "Getting Started" guide, so it should not be tucked away in "utils"

docs/guides/accessing-the-dom.md
matyasf: Here I think its good that we have both a functional and class based example for legacy codebases

docs/contributor-docs/writing-docs.md
matyasf: done

packages/__docs__/src/compileMarkdown.tsx
matyasf: done

PR #2189, opened 2025-10-09
fix(ui-drawer-layout,ui-a11y-utils): fix Tray closing immediately after opening and calling onDismiss

matyasf:
Fixes INSTUI-4828

To test:

Open the Tray by clicking the button and close it by clicking outside it multiple times, it should not auto-close
```
function App(props) {
  const [trayOpen, setTrayOpen] = useState(false);

  const handleOpen = () => {
    console.log('Opening tray, current state:', trayOpen);
    setTrayOpen(true);
  };

  const handleClose = () => {
    console.log('Closing tray, current state:', trayOpen);
    setTrayOpen(false);
  };

  const handleDismiss = () => {
    console.log('onDismiss called, current state:', trayOpen);
    setTrayOpen(false);
  };

  return (
    <div>
      <button
        onClick={handleOpen}
        data-testid='test-open-button'
        style={{ padding: '10px', margin: '10px' }}
      >
        Open Test Tray
      </button>

      <Tray
        label='Test Tray'
        onDismiss={handleDismiss}
        open={trayOpen}
        placement='end'
        shouldCloseOnDocumentClick
      >
        <div style={{ padding: '20px', height: '100vh' }}>
          <h3>Test Tray</h3>
          <p>
            This is a test tray to validate if the onDismiss function works
            correctly.
          </p>
          <p>Current state: {trayOpen ? 'Open' : 'Closed'}</p>
          <button
            onClick={handleClose}
            data-testid='test-close-button'
            style={{ padding: '10px', margin: '10px' }}
          >
            Close
          </button>
        </div>
      </Tray>
    </div>
  );
}
render(<App />)
```

The previous fix should work too (you can press ESC to close overlaying DrawerTray):
```
  state = {
    open: false
  }
  handleTrayDismiss = () => {
    this.setState({ open: false })
  }
  render() {
    return (
      <View height="25rem" as="div" background="primary" position="relative">
        <DrawerLayout minWidth="4444px">
          <DrawerLayout.Tray
            id="DrawerLayoutTrayExample1"
            open={this.state.open}
            placement="start"
            label="Drawer Tray Start Example"
            onDismiss={this.handleTrayDismiss}
          >
            <View
              as="div"
              maxWidth="16rem"
              textAlign="center"
              margin="large auto"
              padding="small"
            >
              <CloseButton
                placement="end"
                offset="small"
                onClick={this.handleTrayDismiss}
                screenReaderLabel="Close"
              />
                Hello from start tray with a small amount of placeholder
                content
            </View>
          </DrawerLayout.Tray>
          <DrawerLayout.Content label="Drawer content example">
            <div style={{ background: 'white', height: '100%' }}>
              <View as="div" padding="x-large">
                      <Button
                        onClick={() => {
                          this.setState({ open: true })
                        }} >
                        Expand tray
                      </Button>
                lorem ipsum
              </View>
            </div>
          </DrawerLayout.Content>
        </DrawerLayout>
      </View>
    )
  }
}

render(<Example />)
```

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -5,6 +5,7 @@
 InstUI is a React component library and design system. **Lerna 8 monorepo** with 100+ packages.
 
 **Key Resources:**
+
 - Website: https://instructure.design
 - AI docs: https://instructure.design/llms.txt
 - Component docs: https://instructure.design/markdowns/[ComponentName].md
@@ -27,6 +28,7 @@ InstUI is a React component library and design system. **Lerna 8 monorepo** with
 ```
 
 **Important file locations:**
+
 - Component source: `/packages/ui-*/src/`
 - Component tests: Co-located with components
 - Theme types: `/packages/shared-types/src/ComponentThemeVariables.ts`
@@ -76,11 +78,13 @@ npm run clean-node && npm install        # Nuclear option (removes all node_modu
 ## Finding Component Information
 
 **To find available components and their packages:**
+
 - Check `/packages/__docs__/src/components.ts` - lists all components with their package locations
 
 **Each component has two READMEs:**
 
 1. **Component README**: `/packages/[package]/src/[Component]/README.md` ⭐
+
    - What the component does and usage examples
    - **Check this first** - it has the detailed information
    - These READMEs are also used to render the component documentation pages on instructure.design
@@ -89,6 +93,7 @@ npm run clean-node && npm install        # Nuclear option (removes all node_modu
    - Package overview, installation, exports
 
 **For complete API details of a component:**
+
 - Props: Check `props.ts` files in component source
 - Theme variables: Check `theme.ts` files in component source
 
@@ -108,6 +113,7 @@ npm run clean-node && npm install        # Nuclear option (removes all node_modu
 A change is **breaking** if it requires consumers to modify their code:
 
 **Breaking changes (avoid):**
+
 - ❌ Removing or renaming a prop
 - ❌ Changing a prop's type or behavior
 - ❌ Removing or renaming a component
@@ -116,6 +122,7 @@ A change is **breaking** if it requires consumers to modify their code:
 - ❌ Removing exported utilities or functions
 
 **Not breaking changes (preferred):**
+
 - ✅ Adding new optional props
 - ✅ Adding new components
 - ✅ Bug fixes that restore intended behavior
@@ -129,10 +136,16 @@ When a breaking change is explicitly requested, document it clearly in the commi
 ## Testing Requirements
 
 All components **MUST**:
+
 1. Have unit tests (Vitest + React Testing Library in `*.test.tsx`)
 2. Have visual regression tests in `/regression-test`
 3. Pass accessibility audits (WCAG 2.1 AA)
 4. Support RTL languages
+
+### Writing tests
+
+- Try to use `vitest`'s fake timers in unit tests
+- When simulating mouse events (e.g., `fireEvent.click`), use `{ button: 0, detail: 1 }` if the click events don't seem to work. This is needed because `FocusRegion.ts/handleDocumentClick` uses `event.detail` and `event.button` to determine if the event was a mouse/touch click.
 
 ### Running Tests
 
@@ -158,11 +171,13 @@ See `/regression-test/README.md` for detailed instructions.
 InstUI uses custom markdown with special code blocks (using [gray-matter](https://github.com/jonschlinkert/gray-matter) YAML syntax) for interactive examples.
 
 **Code block types:**
+
 - `type: code` - Syntax highlighting only
 - `type: embed` - Rendered component only
 - `type: example` - Interactive (rendered + editable) ⭐ Most common
 
 **IMPORTANT:**
+
 - Always write functional component examples with hooks
 - All InstUI components are available without imports in examples
 
--- a/packages/ui-a11y-utils/src/FocusRegion.ts
+++ b/packages/ui-a11y-utils/src/FocusRegion.ts
@@ -85,10 +85,12 @@ class FocusRegion {
     this._contextElement = element
     if (options) {
       this._options = options
-      this.addOrRemoveListeners(
-        options.shouldCloseOnDocumentClick,
-        options.shouldCloseOnEscape
-      )
+      if (this._active) {
+        this.addOrRemoveListeners(
+          options.shouldCloseOnDocumentClick,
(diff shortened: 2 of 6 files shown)


packages/ui-a11y-utils/src/FocusRegion.ts
```diff
@@ -85,10 +85,12 @@ class FocusRegion {
...
-      this.addOrRemoveListeners(
-        options.shouldCloseOnDocumentClick,
-        options.shouldCloseOnEscape
-      )
+      if (this._active) {
+        this.addOrRemoveListeners(
+          options.shouldCloseOnDocumentClick,
+          options.shouldCloseOnEscape
+        )
+      }
```
matyasf, line 93: This is the fix. The issue was that Dialog called this in ` componentDidUpdate` and it added the event listeners like `onDismiss` even if the region was inactive

packages/ui-drawer-layout/src/DrawerLayout/__tests__/DrawerLayout.test.tsx
```diff
@@ -50,4 +54,48 @@ describe('<DrawerLayout />', () => {
     expect(contentWrapper).toHaveTextContent('Hello from content')
     expect(contentWrapper).toHaveAttribute('aria-label', 'Test DrawerContent')
   })
+
+  it('should close the tray when ESC is pressed in overlay mode', async () => {
```
matyasf, line 58: This is a test for the bugfix that introduced this bug https://github.com/instructure/instructure-ui/pull/2150

packages/ui-a11y-utils/src/__tests__/FocusRegion.test.tsx
```diff
@@ -252,6 +252,43 @@ describe('FocusRegion', () => {
       // The internal state should be updated
       expect(focusRegion).toBeDefined()
     })
+
+    it.only('should not add event listeners when updateElement is called on inactive region', async () => {
```
matyasf: I tried to add a test also to Tray that fails without my fix, but could not, its timing code is so complex :(

packages/ui-a11y-utils/src/__tests__/FocusRegion.test.tsx
```diff
@@ -68,7 +68,6 @@ describe('FocusRegion', () => {
       const outsideElement = document.createElement('div')
       document.body.appendChild(outsideElement)
 
-      fireEvent.mouseDown(outsideElement, { button: 0 })
```
matyasf, line 71: These were not needed, just AI adding extra for nothing in a previous commit

packages/ui-a11y-utils/src/__tests__/FocusRegion.test.tsx
```diff
@@ -252,6 +250,42 @@ describe('FocusRegion', () => {
       // The internal state should be updated
       expect(focusRegion).toBeDefined()
     })
+
+    it('should not add event listeners when updateElement is called on inactive region', async () => {
```
matyasf, line 254: low level test for the bug

packages/ui-tray/src/Tray/__tests__/Tray.test.tsx
```diff
@@ -22,22 +22,31 @@
...
 
 describe('<Tray />', async () => {
-  it('should render nothing and have a node with no parent when closed', async () => {
+  beforeAll(() => {
+    vi.useFakeTimers()
+  })
+
+  afterAll(() => {
+    vi.useRealTimers()
+  })
```
matyasf, line 40: We should always use these! it made this test file run from ~3secs to 220ms

packages/ui-tray/src/Tray/__tests__/Tray.test.tsx
```diff
@@ -48,31 +57,30 @@ describe('<Tray />', async () => {
...
     render(
       <Tray label="Tray Example" open onOpen={onOpen}>
         Hello Tray
       </Tray>
     )
-
-    await waitFor(() => {
-      expect(onOpen).toHaveBeenCalled()
+    act(() => {
+      vi.runAllTimers()
```
matyasf, line 78: This ugly code is needed with fake timers, it runs the fake timers until they all expire and executes any events that were fired during the run

packages/ui-tray/src/Tray/__tests__/Tray.test.tsx
```diff
@@ -192,12 +200,79 @@ describe('<Tray />', async () => {
...
+              vi.runAllTimers()
             })
+            expect(onExited).toHaveBeenCalled()
           })
         }
       })
     }
   })
+
+  it('should open, close via dismiss, and reopen with shouldCloseOnDocumentClick enabled', async () => {
```
matyasf, line 213: This is the high level test for the bug. It was a nightmare to write because it looked like the click events were not fired/processed.
The solution was to fire them with extra options: `fireEvent.click(src, { button: 0, detail: 1 })`. These are needed because `FocusRegion` requires them in  `handleDocumentClick` because a more straightforward solution (`event.pointerType==='mouse'`) is [not supported by Safari](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/pointerType), see https://github.com/instructure/instructure-ui/pull/1399

EDIT: its getting support in Safari, see https://bugs.webkit.org/show_bug.cgi?id=279540

CLAUDE.md
```diff
@@ -129,11 +136,17 @@ When a breaking change is explicitly requested, document it clearly in the commi
...
+
 1. Have unit tests (Vitest + React Testing Library in `*.test.tsx`)
 2. Have visual regression tests in `/regression-test`
 3. Pass accessibility audits (WCAG 2.1 AA)
 4. Support RTL languages
 
+### Writing tests
+
+- Try to use `vitest`'s fake timers in unit tests
+- When simulating mouse events (e.g., `fireEvent.click`), use `{ button: 0, detail: 1 }` if the click events don't seem to work. This is needed because `FocusRegion.ts/handleDocumentClick` uses `event.detail` and `event.button` to determine if the event was a mouse/touch click.
```
matyasf, line 148: Claude could not figure this out outside of `FocusRegion.test.ts` and I've tried at least for 2 hours. This tip will help it hopefully in the future

packages/ui-tray/src/Tray/__tests__/Tray.test.tsx
balzss: maybe we could add this to our global vitest config

PR #2210, opened 2025-10-21
INSTUI-4835 build(ui-scripts): fix build scripts

matyasf:
- now bootstrap builds the new theme tokens
- clean scripts also delete theming tokens
- Update `Avatar`'s regression test for the new theme

to test:
- Check if the new regression test in Chromatic looks as it should.
- Run clean inside packages, e.g. ui-themes, it should delete the generated theming TS files.
- build should pass on CI

### DIFF
--- a/packages/ui-scripts/lib/build/buildThemes/createFile.js
+++ b/packages/ui-scripts/lib/build/buildThemes/createFile.js
@@ -22,9 +22,7 @@
  * SOFTWARE.
  */
 
-import fs from 'fs'
-import { execSync } from 'child_process'
-
+import { promises } from 'fs'
 const license = `/*
  * The MIT License (MIT)
  *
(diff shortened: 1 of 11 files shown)


packages/ui-scripts/lib/build/clean.js
```diff
@@ -24,6 +24,8 @@
 
 import fs from 'fs'
 
+// TODO merge this with/scripts/clean.js
```
matyasf, line 27: This script is running when `clean` is executed in a package. We should parametrize the `scripts/clean.js` script and delete this one

packages/ui-scripts/lib/build/buildThemes/createFile.js
```diff
@@ -58,9 +57,6 @@ const createFile = (filePath, fileContent) => {
 
   fs.mkdirSync(filePath.split('/').slice(0, -1).join('/'), { recursive: true })
   fs.writeFileSync(filePath, `${license} ${fileContent}`)
-  // eslint-disable-next-line no-console
-  console.log(`Creating theme file: ${filePath}`)
-  execSync(`npx prettier --write ${filePath} --ignore-path .prettierignore`)
```
matyasf, line 63: I've replaced `prettier` with `dprint`, now it runs under 1 second

packages/ui-scripts/lib/build/buildThemes/createFile.js
```diff
@@ -51,16 +49,19 @@ const license = `/*
 
 `
 
-const createFile = (filePath, fileContent) => {
-  if (fs.existsSync(filePath)) {
-    fs.unlinkSync(filePath)
+const createFile = async (filePath, fileContent) => {
```
matyasf, line 52: I've rewritten the whole thing to `async` because I could not find the bug...  its not here, but still async code is better

scripts/bootstrap.js
```diff
@@ -64,13 +66,7 @@ function buildProject() {
...
 }
 function bootstrap() {
-  try {
-    fork(path.resolve('scripts/clean.js'), opts)
-  } catch (error) {
-    console.error('clean failed with error:', error)
-    process.exit(1)
-  }
-
+  execSync(path.resolve('scripts/clean.js'), opts)
```
matyasf, line 70: This was the issue. `fork` is async and we did not wait for its completion, so it deleted the built themes

PR #2214, opened 2025-10-22
docs app displays new theme variables, better theme type parsing

matyasf:
- if a component uses an old theme it works too
- Its not working yet for subcomponents.
- Add a utility function to convert box shadow objects to strings
- Update theme type generator script to make it possible to customize types
- Nested variables, like box shadows are deconstructed recursively and displayed in separate rows.

To test:
- Check that the theme variables have the correct value in new components, e.g. `Avatar`
- Old components like `Menu` should show everything correctly too
- check the generated codein `ui-themes/src/themes/newThemes` 

Completes INSTUI-4707

### DIFF
--- a/packages/__docs__/buildScripts/utils/getPathInfo.mts
+++ b/packages/__docs__/buildScripts/utils/getPathInfo.mts
@@ -45,8 +45,8 @@ export function getPathInfo(
   const themePath = resourcePath.replace('index.tsx', 'theme.ts')
 
   if (fs.existsSync(themePath)) {
-    pathInfo.themePath = pathInfo.srcPath.replace('index.tsx', 'theme.ts')
-    pathInfo.themeUrl = pathInfo.srcUrl.replace('index.tsx', 'theme.ts')
+    pathInfo.themePath = pathInfo.srcPath.replace('index.tsx', 'styles.ts')
+    pathInfo.themeUrl = pathInfo.srcUrl.replace('index.tsx', 'styles.ts')
   }
   return pathInfo
 }
--- a/packages/__docs__/src/ComponentTheme/index.tsx
+++ b/packages/__docs__/src/ComponentTheme/index.tsx
@@ -34,53 +34,69 @@ import generateStyle from './styles'
 import { allowedProps } from './props'
 import type { ComponentThemeProps } from './props'
 
+type ThemeEntry = { name: string; value: string | number }
+
 @withStyle(generateStyle, null)
 class ComponentTheme extends Component<ComponentThemeProps> {
   static allowedProps = allowedProps
 
-  renderValueCell(
-    value: undefined | string | object | number,
-    colorPrimitives: object
-  ) {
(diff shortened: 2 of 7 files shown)


packages/__docs__/src/Document/index.tsx
```diff
@@ -76,8 +76,19 @@ class Document extends Component<DocumentProps, DocumentState> {
...
+        // @ts-ignore todo type
+        generateTheme = this.props.themeVariables.components[doc.id]
+        this.setState({ componentTheme: generateTheme })
+        return
+      } else {
+        // old theme
+        generateTheme = doc?.componentInstance?.generateComponentTheme
+      }
     } else {
+      // TODO make it work for new themes
```
matyasf, line 91: Child components are not working yet, we dont have anything yet to test it with

packages/__docs__/buildScripts/utils/getPathInfo.mts
```diff
@@ -45,8 +45,8 @@ export function getPathInfo(
   const themePath = resourcePath.replace('index.tsx', 'theme.ts')
 
   if (fs.existsSync(themePath)) {
-    pathInfo.themePath = pathInfo.srcPath.replace('index.tsx', 'theme.ts')
-    pathInfo.themeUrl = pathInfo.srcUrl.replace('index.tsx', 'theme.ts')
+    pathInfo.themePath = pathInfo.srcPath.replace('index.tsx', 'styles.ts')
+    pathInfo.themeUrl = pathInfo.srcUrl.replace('index.tsx', 'styles.ts')
   }
```
matyasf, line 50: This is used for the text at the top of the themes section. Showing the theme object is not that useful, the styles file is much better, since it contains the actual logic on what style is used when

packages/ui-scripts/lib/build/buildThemes/generateComponents.js
```diff
@@ -63,74 +63,113 @@ export const resolveReferences = (semantics, key) => {
   if (isReference(value)) {
     return formatReference(value)
   }
-
+  if (value === '') {
+    // token studio returns "" if a value is not set, but defaults to 0
+    return `0,\n`
```
matyasf, line 73: This is not perfect, there might be a case when we need to default to "". Designers are trying to give everything a value, but we need to look out for such cases

packages/__docs__/src/ComponentTheme/index.tsx
```diff
@@ -34,53 +34,69 @@ import generateStyle from './styles'
...
   }
 
-  renderRows() {
-    const { componentTheme, themeVariables } = this.props
-    const colorPrimitives = themeVariables.colors.primitives
+  /**
+   * Theme objects can be nested, so we need to walk the tree to get all the
+   * keys, e.g., box shadow values
+   */
+  themeToArray(
```
joyenjoyer, line 67: nice, I like the recursion here

joyenjoyer (approved):
great job!

PR #2223, opened 2025-10-31
INSTUI-4842 [v12] better typing for themes

matyasf:
- Use [types from Token studio](https://github.com/tokens-studio/types) from typing their converted JSONs instead of types from semantics/primitives
- Add some simple typing for the whole theme object

To test: 
run `build:themes` and check out the generated code (especially the types) in `ui-themes/src/themes/newThemes`

### DIFF
--- a/packages/ui-avatar/src/Avatar/styles.ts
+++ b/packages/ui-avatar/src/Avatar/styles.ts
@@ -50,7 +50,7 @@ const generateStyle = (
   componentTheme: NewComponentTypes['Avatar'],
   params: StyleParams,
   //TODO type themes properly
-  theme: any
+  theme: any // TODO BaseTheme is is not good, it accesses theme.semantics.spacing
 ): AvatarStyle => {
   const {
     loaded,
--- a/packages/ui-scripts/lib/build/buildThemes/generateComponents.js
+++ b/packages/ui-scripts/lib/build/buildThemes/generateComponents.js
@@ -22,8 +22,6 @@
  * SOFTWARE.
  */
 
-const boxShadowType = 'BoxShadow'
-
 const isReference = (expression) =>
   expression[0] === '{' && expression[expression.length - 1] === '}'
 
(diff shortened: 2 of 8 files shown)


packages/ui-scripts/lib/build/buildThemes/generateComponents.js
```diff
@@ -75,68 +73,69 @@ export const resolveReferences = (semantics, key) => {
...
+    // Add composition token support if needed
     switch (tokenObject.type) {
       case 'border':
-        // This type is coming from Token Studio
-        ret += '{ style: "solid" | "dashed" }'
+        ret += 'TokenBorderValue'
         break
-      case 'borderWidth':
+      // the following types are coming from SingleXYToken in @token-studio/types
+      // we could add them as imports from @token-studio/types if needed
```
matyasf, line 93: I decided to just hardcode these simple types here, because importing them all if they are required would be lots of extra code

packages/ui-scripts/lib/build/buildThemes/setupThemes.js
```diff
@@ -152,12 +141,22 @@ const setupThemes = async (targetPath, input) => {
...
+        if (componentTypes.includes('TokenBoxshadowValueInst')) {
+          importBoxShadow = `import type { TokenBoxshadowValueInst } from '../commonTypes'`
+        }
+        let importBorder = ''
+        if (componentTypes.includes('TokenBorderValue')) {
+          importBorder = `import type { TokenBorderValue } from '@tokens-studio/types'`
+        }
+        let importTypography = ''
+        if (componentTypes.includes('TokenTypographyValueInst')) {
+          importTypography = `import type { TokenTypographyValueInst } from '../commonTypes'`
```
matyasf, line 153: This is a bit ugly, but simple. If we want we can make this nicer and dynamic, IMO its not needed because I dont expect Token Studio to change its types often

packages/ui-scripts/lib/build/buildThemes/setupThemes.js
```diff
@@ -249,18 +230,52 @@ const setupThemes = async (targetPath, input) => {
...
+    paragraphSpacing?: string
+    paragraphIndent?: string
+    textCase?: TokenTextCaseValue
+    textDecoration?: TokenTextDecorationValue
+  }
+
+  export type BaseTheme<P extends Record<string, any>, S extends Record<string, any>> = {
+    primitives: P
+    semantics: S
+    components: ComponentTypes
```
matyasf, line 276: This type will likely change in the future, the structure of `primitives` and `semantics` is fixed

packages/ui-avatar/src/Avatar/styles.ts
```diff
@@ -50,7 +50,7 @@ const generateStyle = (
   componentTheme: NewComponentTypes['Avatar'],
   params: StyleParams,
   //TODO type themes properly
-  theme: any
+  theme: any // TODO BaseTheme is is not good, it accesses theme.semantics.spacing
```
HerrTopi, line 53: This is the new theme already. I'm not sure what you mean about BaseTheme being not good. Can you elaborate?

Also typo (is is)

HerrTopi: I figured out what you meant. This theme should be the whole new theme in time, but as of right now, any will be enough

git-nandor (approved):

PR #2228, opened 2025-11-04
Separate focus logic from view

HerrTopi:
This pr adds a new util to emotion's `styleUtils`, `calcFocusOutlineStyles` and injects the `sharedTokens` to the `generateStyles` function. Through that one can access `margin`, `boxShadow` and `focusOutline`.

TEST_PLAN:

- Check the margin prop of avatar
- set tabIndex to 0 in avatar's outermost div and follow the instructions of the `/global-style-utils`docs
- read through said docs and check if it's correct (written mostly by ai and checked by @HerrTopi )

NOTE: `calcFocusOutlineStyles` is much simpler than the original `View` calculations. This is by design. If most cases can be covered by this simplified version, it's better to keep this way. If new requirements arise, adjustments are possible

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -16,8 +16,8 @@ jobs:
         run: npm ci
       - name: Bootstrap project
         run: npm run bootstrap
-      - name: Lint commits
-        run: npm run lint:commits
+      - name: Run commitlint for the commits in the PR
+        uses: wagoid/commitlint-github-action@v6
       - name: Lint code
         run: npm run lint:changes
 
--- a/packages/emotion/src/styleUtils/calcFocusOutlineStyles.ts
+++ b/packages/emotion/src/styleUtils/calcFocusOutlineStyles.ts
@@ -1,0 +1,103 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015 - present Instructure, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+import { alpha } from '@instructure/ui-color-utils'
+
+/**
+ * Generates consistent focus outline styles.
+ *
+ * This function creates CSS-in-JS styles for focus indicators.
+ *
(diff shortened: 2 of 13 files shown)


docs/contributor-docs/global-style-utils.md
```diff
@@ -0,0 +1,201 @@
...
+      // Use sharedTokens for focus outline styles
+      ...calcFocusOutlineStyles(sharedTokens.focusOutline, {
+        focusColor: 'info'
+      })
+    }
+  }
+}
+```
+
+### Available SharedTokens Properties
```
matyasf: I would call this an example, we do not want to maintain this doc when we change something in the shared tokens

docs/contributor-docs/global-style-utils.md
```diff
@@ -0,0 +1,201 @@
...
+
+- **`sharedTokens.margin`**: Spacing token map for use with `calcMarginFromShorthand`
+- **`sharedTokens.focusOutline`**: Focus outline theme configuration for use with `calcFocusOutlineStyles`
+- **Other shared design tokens**: Additional properties available across all components
+
+This integration ensures that all components have consistent access to the design system's shared tokens without needing to manually pass theme data to each utility function.
```
matyasf: I would remove this, its just AI bloat

docs/contributor-docs/global-style-utils.md
```diff
@@ -0,0 +1,201 @@
...
+calcMarginFromShorthand('16px medium auto small', sharedTokens.margin)
+// Returns: '16px 1rem auto 0.5rem'
+```
+
+### How it works
+
+1. **Splits the input**: The function splits the input string by spaces to handle CSS shorthand format (top, right, bottom, left)
+2. **Resolves tokens**: For each value, it checks if it's a theme token path (e.g., 'between.cards.md')
+3. **Maps values**: Theme tokens are resolved using the spacing map, while literal values (like '16px', 'auto') are passed through unchanged
+4. **Joins result**: All resolved values are joined back with spaces to create a valid CSS margin value
```
matyasf: I would remove this (line 50-99) and add it to the JSDoc of `calcMarginFromShorthand`. Here just mention what it does with an example

packages/emotion/src/styleUtils/calcFocusOutlineStyles.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  const outlineStyle = {
+    outlineColor: focusColorVariants[focusColor!],
+    outlineStyle: 'solid',
+    outlineWidth: theme.width,
+    outlineOffset: theme[focusPosition]
+  }
+  return {
+    ...(shouldAnimateFocus && {
+      transition: 'outline-offset 0.25s'
```
matyasf: It looks worse than before IMO (I've tried it on Avatar) I think the reason is that the old one also animates color (alpha) while zooming in/out. Can you add this back?

packages/emotion/src/styleUtils/calcFocusOutlineStyles.ts
```diff
@@ -0,0 +1,77 @@
...
+    onColor: string
+    successColor: string
+    dangerColor: string
+  },
+  params?: {
+    focusColor?: 'info' | 'inverse' | 'success' | 'danger'
+    focusPosition?: 'offset' | 'inset'
+    shouldAnimateFocus?: boolean
+    focusWithin?: boolean
+  }
```
matyasf, line 63: The one in `View` has an extra param, `withFocusOutline` This is an (ugly) enum, that has 3 states: force showing focus ring, do not show focus ring, auto - show when in focus. (its a boolean now, maybe an enum would be nicer)
As I see from our code, we are using this functionality, can you add this param?

docs/contributor-docs/global-style-utils.md
```diff
@@ -0,0 +1,201 @@
...
+
+## calcFocusOutlineStyles
+
+The `calcFocusOutlineStyles` utility generates consistent focus outline styles.
```
matyasf: I would remove everything below this point and add it to the JSDoc of calcFocusOutlineStyles. Here just mention what it does with an example, otherwise it will be hard to maintain

.github/workflows/pr-validation.yml
```diff
@@ -16,8 +16,8 @@ jobs:
         run: npm ci
       - name: Bootstrap project
         run: npm run bootstrap
-      - name: Lint commits
-        run: npm run lint:commits
+      - name: Run commitlint for the commits in the PR
+        uses: wagoid/commitlint-github-action@v6
```
matyasf, line 20: Why was this changed?

balzss: because it got confused on what commits to check. it worked fine when a pr was compared against master but in cases like this where the target branch is something else it checked all commit since the branch out from master. this meant it checked invalid commit messages in the v12 branch and threw errors. after some research this seemed the easiest solution

PR #2229, opened 2025-11-05
[v12] refactor(many): type new themes properly

matyasf:
- Type the new theme objects. Its a bit chaotic now because we will support old themes, a cleanup will make it much nicer when we delete them
- remove require path from docs build, because it was used to import from the deprecated 'canvas-theme', 'canvar-high-contrast-theme' packages
- remove functional theme support, it was used only for Avatar
- `useStyle` no longer needs `generateComponentTheme` since this is not used by new themes


packages/__docs__/buildScripts/build-docs.mts
```diff
@@ -259,22 +259,10 @@ function tryParseReadme(dirName: string) {
 
 function parseThemes() {
   const parsed: MainDocsData['themes'] = {}
-  parsed[canvas.key] = {
-    resource: canvas,
-    requirePath: '@instructure/canvas-theme'
```
matyasf, line 264: The `requirePath` is only used in the `Themes` docs pages to show how to import them. This is not needed, because now we import everything from `ui-themes`

packages/__docs__/src/Document/index.tsx
```diff
@@ -88,20 +87,13 @@ class Document extends Component<DocumentProps, DocumentState> {
...
       generateTheme = doc?.children?.find(
         (value) => value.id === this.state.selectedDetailsTabId
       )?.componentInstance?.generateComponentTheme
     }
-    const generateThemeFunctional =
-      functionalComponentThemes[
-        doc.id as keyof typeof functionalComponentThemes
-      ]
     if (typeof generateTheme === 'function' && themeVariables) {
+      // @ts-ignore todo type
```
matyasf, line 95: Typing is still not perfect, but it will be much simpler to do when we can delete the old themes

PR #2235, opened 2025-11-06
fix(ui-icons): fix ai info icon

joyenjoyer:
INSTUI-4847

Test:
Use the AI Info icon and check whether changing the color prop impacts it.

`      <IconAIInfoLine size="large" color="alert"/>
`


joyenjoyer:
> The new icons are the wrong size, they should be 1920x1920

Fixed. Please review again.

matyasf (approved):

PR #2237, opened 2025-11-11
fix(ui-color-picker): fix popover scrolling + docs routing

balzss:
## Summary
Fixed multiple ColorPicker issues related to popover scrolling and mixer button alignment.

### Issue 1: Popover Scrolling
Fixed ColorPicker popover scrolling when content exceeds viewport height. The popover now dynamically calculates available space based on placement (above/below trigger) and becomes scrollable when needed. A smooth fade-in transition prevents visual jump on opening.

### Issue 2: Mixer Button Alignment
Fixed mixer button alignment and visual jump issues:
- **Alignment calculation**: Was returning 0 due to timing - now uses `requestAnimationFrame` to defer measurement until CSS-in-JS styles are applied
- **Visual jump**: Button now starts bottom-aligned and dynamically switches to top-aligned after calculation completes, minimizing the visible jump

## Key Changes
- Dynamic max-height calculation for both upward/downward popover placement
- Double `requestAnimationFrame` for popover to wait for Emotion CSS-in-JS injection and child component mounting
- Single `requestAnimationFrame` for mixer button alignment to wait for layout completion
- Opacity transition with state flag to prevent visual jump in popover
- Dynamic `alignSelf` switching for mixer button to minimize visual jump
- State reset on close to fix alternating open/close behavior
- Added visual regression tests
- Fixed docs dev server client-side routing (webpack historyApiFallback)

## Test Plan
- [ ] Open ColorPicker with tall content near bottom/top of viewport
- [ ] Verify popover is scrollable and opens smoothly without visual jump
- [ ] Test multiple open/close cycles
- [ ] Verify mixer button aligns correctly with input field
- [ ] Verify minimal visual jump when mixer button appears
- [ ] Verify docs dev server routing works on refresh

Refs INSTUI-4849

### DIFF
--- a/packages/ui-color-picker/src/ColorPicker/index.tsx
+++ b/packages/ui-color-picker/src/ColorPicker/index.tsx
@@ -129,7 +129,13 @@ class ColorPicker extends Component<ColorPickerProps, ColorPickerState> {
 
   handleInputContainerRef = (el: Element | null) => {
     this.inputContainerRef = el
-    this.setLabelHeight()
+
+    if (el) {
+      // Defer measurement until after layout is complete and CSS-in-JS styles are applied
+      requestAnimationFrame(() => {
+        this.setLabelHeight()
+      })
+    }
   }
 
   popoverContentRef: HTMLDivElement | null = null
@@ -488,14 +494,15 @@ class ColorPicker extends Component<ColorPickerProps, ColorPickerState> {
       }
       isShowingContent={this.state.openColorPicker}
       onShowContent={() => {
-        this.setState({ openColorPicker: true, mixedColor: this.state.hexCode })
-      }}
-      onHideContent={() => {
         this.setState({
-          openColorPicker: false,
+          openColorPicker: true,
+          mixedColor: this.state.hexCode,
           calculatedPopoverMaxHeight: undefined,
           isHeightCalculated: false
         })
+      }}
+      onHideContent={() => {
+        this.setState({ openColorPicker: false })
       }}
       on="click"
       screenReaderLabel={this.props.popoverScreenReaderLabel}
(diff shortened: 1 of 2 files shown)


ToMESSKa (changes requested):
@balzss When I first open the ColorPicker popover, the top cannot be seen. If scroll down and up again, the popover shirks and the scrollbar appears.

ToMESSKa (changes requested):
If I open the ColorPicker, select a color and click 'Add' and then open the ColorPicker again, the viewport size is not respected and scrolling isn't avaliable.

HerrTopi (approved):
Lgtm. At least good enough. If you select a color and close the popover, it wont scroll if opened again. If closed and open again, it works as expected. Good for now

regression-test/cypress/e2e/spec.cy.ts
```diff
@@ -142,6 +142,17 @@ describe('visual regression test', () => {
...
+  it('ColorPicker', () => {
+    cy.visit('http://localhost:3000/colorpicker')
+    cy.wait(300)
+    cy.injectAxe()
+    // TODO: Fix ARIA violations before enabling a11y check
+    // - aria-allowed-attr (critical): 1 node
+    // - aria-prohibited-attr (serious): 3 nodes
+    // ColorMixer has aria-disabled on plain div without proper role
+    // cy.checkA11y('.axe-test', axeOptions, terminalLog)
+  })
```
matyasf, line 154: These issues existed before this PR, right? (I dont see them when I run in the example page with a ColorPicker open)

balzss: i think they existed but we didn't have colorpicker examples before this pr

PR #2240, opened 2025-11-12
[V12] feat(ui-metric): rework Metric

ToMESSKa:
INSTUI-4838

**ISSUE:**
- this the V12 rework of the Metric component

**TEST PLAN:**
- compare the component to the Figma file: https://www.figma.com/design/5WFqF3HVlNdfcJooCaiFoh/InstUI-Component-Library?m=auto&node-id=2511-4617&t=ZNS56TkfFeY065IF-1 (the font might be different because the new font is not installed yet, I talked to Topi about this)
- check it all the removed/renamed variables and breaking changes are listed in the README..md files


packages/ui-metric/src/Metric/README.md
```diff
@@ -2,6 +2,11 @@
 describes: Metric
 ---
 
+### Upgrade Guide for V12
+
+> - theme variable `padding` is now removed, left-right padding can be set now with the `paddingHorizontal` theme variable
+> - theme variable `fontFamily` is now removed, font can be set independently on the value and label elements using the `valueFontFamily` and `labelFontFamily` variables respectively
+
```
ToMESSKa: I put the upgrade guides in the README.md as suggested by Topi, but I can copy these in the Upgrade Guide for Version 12 file started by Matyi

matyasf: I think these should all the in the same place, the upgrade guide. We dont want to instruct users to open 50+ Readme files.

packages/ui-metric/src/Metric/index.tsx
```diff
@@ -22,74 +22,65 @@
...
-  static defaultProps = {
-    textAlign: 'center',
-    isGroupChild: false
-  }
+  const elementRef = useRef<HTMLDivElement | null>(null)
 
-  ref: Element | null = null
+  const styles = useStyle({
+    generateStyle,
+    generateComponentTheme,
```
matyasf: `generateComponentTheme` is no longer used, you can delete it from here and the file too.

packages/ui-metric/src/Metric/styles.ts
```diff
@@ -22,44 +22,56 @@
...
  */
 
-import type { MetricTheme } from '@instructure/shared-types'
 import type { MetricProps, MetricStyle } from './props'
+import { NewComponentTypes } from '@instructure/ui-themes'
+
+type StyleParams = {
+  textAlign: MetricProps['textAlign']
+  isGroupChild: MetricProps['isGroupChild']
+  themeOverride?: MetricProps['themeOverride']
```
matyasf: You dont need to supply these 2, they are not used.

packages/ui-metric/src/Metric/index.tsx
```diff
@@ -22,74 +22,65 @@
...
+      themeOverride
+    },
+    componentId: 'Metric',
+    displayName: 'Metric'
+  })
 
-  handleRef = (el: Element | null) => {
-    this.ref = el
-  }
+  useImperativeHandle(ref, () => elementRef.current as HTMLDivElement)
```
matyasf: Is this needed for `ref`? Wont this value returned by default?

packages/ui-metric/src/MetricGroup/index.tsx
```diff
@@ -22,68 +22,56 @@
...
-
-  componentDidUpdate() {
-    this.props.makeStyles?.()
-  }
+    const styles = useStyle({
+      generateStyle,
+      generateComponentTheme,
+      componentId: 'MetricGroup',
+      displayName: 'MetricGroup'
+    })
```
matyasf: hmm this is not nice, we have to improve `useStyle` so it can nicely handle style.ts files which have no theme (in a different PR). Can you please put a TODO comment to `useStyle` that it should be improved to handle `generateStyle` functions, that dont use `params` or dont have a theme?

packages/ui-metric/src/Metric/README.md
ToMESSKa: @matyasf I moved these

packages/ui-metric/src/Metric/index.tsx
ToMESSKa: @matyasf I deleted it

packages/ui-metric/src/Metric/styles.ts
ToMESSKa: @matyasf I deleted them

packages/ui-metric/src/Metric/index.tsx
ToMESSKa: @matyasf You're right, the useImperativeHandle was not needed here.

packages/ui-metric/src/MetricGroup/index.tsx
ToMESSKa: @matyasf I added that comment

PR #2241, opened 2025-11-12
Add a way to use different token sets in components

HerrTopi:


packages/emotion/src/useStyle.ts
```diff
@@ -98,18 +102,21 @@ const useStyle = <
...
   ) {
     baseComponentTheme =
-      theme.newTheme.components[componentId as keyof NewComponentTypes]
+      theme.newTheme.components[
+        componentWithTokensId as keyof NewComponentTypes
+      ]
   }
   const themeOverride = getComponentThemeOverride(
     theme,
-    displayName ? displayName : componentId || '',
```
matyasf, line 111: Why did you remove `displayName` here? I think the correct logic would be to use `useTokensFrom`, if that doesn't exist then `displayName` if that doesnt either then `componentId`, finally `''`, otherwise we might break old theme overrides (IMO we should keep them working as long as we can)

packages/emotion/src/useStyle.ts
```diff
@@ -48,6 +48,7 @@ type UseStyleParamsWithTheme<
   generateComponentTheme: GenerateComponentTheme
   componentId: string
   displayName?: string
+  useTokensFrom?: string
```
matyasf: I would not add this to the old theme type, its misleading

packages/emotion/src/useStyle.ts
```diff
@@ -60,6 +61,7 @@ type UseStyleParamsWithoutTheme<
   generateComponentTheme?: undefined
   componentId?: undefined
   displayName?: undefined
+  useTokensFrom?: undefined
```
matyasf: I would not add this to the old theme type, its misleading

packages/emotion/src/useStyle.ts
```diff
@@ -98,18 +102,21 @@ const useStyle = <
     typeof generateComponentTheme === 'function'
       ? generateComponentTheme(theme as Theme)
       : {}
+  const componentWithTokensId = useTokensFrom ?? componentId
```
matyasf, line 104: Add it here as a cast, e.g. `(useStyleParams as UseStyleParamsNew<P>).useTokensFrom`

matyasf (changes requested):

PR #2243, opened 2025-11-13
Use the new theme in RadioInput

matyasf:
- Also update its readme with a more comprehensive example
- delete old `theme.ts`
- convert RadioInput to functional component
- converts its DOM to a much flatter stucture, style the actual `<input>` element instead of hiding it and making a fake element
- add `useDeterministicID` which can be used in functional components (= hook form of `withDeterministicId`)

To test:
- check out the examples for RadioInput, RadioInputGroup
- Wrap them with e.g. `<InstUISettingsProvider theme={rebrandLight}>` to see how they look with other themes

### DIFF
--- a/.husky/prepare-commit-msg
+++ b/.husky/prepare-commit-msg
@@ -6,7 +6,9 @@ if [ "$2" = "commit" ] && [ -n "$3" ]; then
 fi
 
 # Check if a rebase is in progress
-if [ -d ".git/rebase-merge" ] || [ -d ".git/rebase-apply" ]; then
+if [ -d "$(git rev-parse --git-path rebase-merge)" ] ||
+   [ -d "$(git rev-parse --git-path rebase-apply)" ]; then
+  echo "Skipping prepare-commit-msg git hook during rebase"
   exit 0
 fi
 
--- a/packages/ui-radio-input/src/RadioInput/styles.ts
+++ b/packages/ui-radio-input/src/RadioInput/styles.ts
@@ -120,6 +120,7 @@ const generateStyle = (
   const inputVariants = {
     simple: {
       base: {
+        all: 'initial',
         boxSizing: 'border-box',
         appearance: 'none',
         display: 'block',
@@ -140,6 +141,8 @@ const generateStyle = (
             : componentTheme.borderSelectedColor,
           boxShadow: inputColors.checkedBoxShadow
         },
+        marginTop: componentTheme.controlVerticalMargin,
+        marginBottom: componentTheme.controlVerticalMargin,
         ...focusOutline
       },
       small: {
--- a/packages/ui-radio-input/src/RadioInput/props.ts
+++ b/packages/ui-radio-input/src/RadioInput/props.ts
@@ -41,6 +41,9 @@ type RadioInputOwnProps = {
    * [with the same name](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#value)
    */
   value?: string | number
+  /**
+   * The id of the input element. If not provided, a unique id will be generated.
+   */
   id?: string
   /**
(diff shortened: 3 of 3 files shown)


packages/ui-radio-input/src/RadioInput/index.tsx
```diff
@@ -22,151 +22,199 @@
...
+          {...passedProps}
+          id={id}
+          ref={handleInputRef}
+          value={value}
+          name={name}
+          checked={checked}
+          type="radio"
+          css={styles?.input}
+          readOnly={readOnly}
+          disabled={disabled}
```
matyasf, line 199: I'm not sure, that this is a good change. Why did the old code just make `readonly` disabled?

packages/ui-radio-input/src/RadioInput/styles.ts
```diff
@@ -22,185 +22,158 @@
...
+        componentTheme.borderSelectedColor
+      }`
+    }
   }
 
-  const facadeVariants = {
+  const inputColors = getInputColors()
+
+  const focusOutline = calcFocusOutlineStyles(sharedTokens.focusOutline)
+  focusOutline.transition += ', box-shadow 0.2s'
```
matyasf, line 118: This trick was needed so the checkmark is animated

packages/ui-react-utils/src/DeterministicIdContext/useDeterministicId.tsx
```diff
@@ -0,0 +1,65 @@
...
+ * @returns A function that generates deterministic IDs. The function accepts an optional instanceName parameter.
+ *
+ * @example
+ * ```tsx
+ * const MyComponent = () => {
+ *     const [deterministicId, setDeterministicId] = useState()
+ *     const getId = useDeterministicId('MyComponent')
+ *     useEffect(() => {
+ *       setDeterministicId(getId())
+ *     }, [])
```
matyasf, line 49: Be careful to always call this in a hook, otherwise it will regenerate the ID on each render

CLAUDE.md
```diff
@@ -53,6 +53,7 @@ pnpm run build:types  # Build TypeScript declarations
 
 # Testing
 pnpm run test:vitest  # Unit tests
+pnpm run test:vitest ui-radio-input # Run tests for a single package
```
matyasf, line 56: Claude always tried stupid ways to run test for a single package, hopefully this will help it

.husky/prepare-commit-msg
```diff
@@ -6,7 +6,9 @@ if [ "$2" = "commit" ] && [ -n "$3" ]; then
 fi
 
 # Check if a rebase is in progress
-if [ -d ".git/rebase-merge" ] || [ -d ".git/rebase-apply" ]; then
+if [ -d "$(git rev-parse --git-path rebase-merge)" ] ||
+   [ -d "$(git rev-parse --git-path rebase-apply)" ]; then
+  echo "Skipping prepare-commit-msg git hook during rebase"
```
matyasf, line 11: This is an unrelated fix, rebase triggering the commitizen prompts in git worktrees

adamlobler:
I noticed that the large radio button and its label are not aligned in the rebrand., there’s extra margin at the top. I’m not sure if this helps, but in the design I set the controlSize token as the label’s minHeight so it always aligns with the radio size.
Screenshot 2025-11-14 at 11 10 43

packages/ui-radio-input/src/RadioInput/index.tsx
HerrTopi: I would take the opportunity and deprecate readOnly alltogether. It's not something design needs and wasn't supported up until now.

packages/ui-radio-input/src/RadioInput/styles.ts
HerrTopi: Is this something that could be useful in certain cases? Shouldn't we add to the util as an option?

PR #2245, opened 2025-11-14
[v12] feat(ui-link): Link rework

git-nandor:
INSTUI-4795

## Summary
Migrated Link component from the old theming system.

- Refactored component implementation from class-based to functional component with hooks
- Removed deprecated theme.ts file and restructured styles.ts to use new theme access patterns
- Updated component API: separated variant and size props for better composability
- Deprecated legacy variant values (inline-small, standalone-small) in favor of combined variant + size props
- Updated Breadcrumb's BreadcrumbLink to align with new Link component type signatures
- Enhanced README documentation with new size and variant usage examples

## Test plan
On the documentation page, verify that everything displays and works correctly,
and check that it matches the [Figma design](https://www.figma.com/design/5WFqF3HVlNdfcJooCaiFoh/InstUI-Component-Library?node-id=2943-9029&p=f&m=dev).

##  
Co-Authored-By: Claude <noreply@anthropic.com>


packages/ui-link/src/Link/props.ts
```diff
@@ -110,27 +110,38 @@ type LinkOwnProps = {
...
    * Fires when the Link gains focus
    */
-  onFocus?: (event: React.FocusEvent<ViewOwnProps>) => void
+  onFocus?: (event: React.FocusEvent) => void
 
   /**
    * Fires when the Link is hovered
    */
-  onMouseEnter?: (event: React.MouseEvent<ViewOwnProps>) => void
+  onMouseEnter?: (event: React.MouseEvent) => void
```
matyasf, line 128: Please write in the update-guide.md about this change

packages/ui-link/src/Link/index.tsx
```diff
@@ -37,226 +44,261 @@ import {
...
+      }
+      if (roleValue === 'button') {
+        // Elements with button role need explicit tabIndex
+        return 0
+      }
+      // Otherwise let the browser handle default focusability
+      return undefined
+    }, [isDisabled, roleValue])
+
+    return jsx(
```
matyasf, line 277: Please dont use this syntax (`jsx(`), we never use it in our codebase. What was your intention here? I thnk you can just use `<element` ..

packages/ui-link/src/Link/index.tsx
```diff
@@ -37,226 +44,261 @@ import {
...
-      if (child && matchComponentTypes(child, ['TruncateText'])) {
-        truncateText = true
+    const roleValue = useMemo(() => {
+      if (forceButtonRole) {
+        return onClick && element !== 'button' ? 'button' : role
       }
+      return role
+    }, [forceButtonRole, onClick, element, role])
+
+    const styles = useStyle({
```
matyasf, line 170: This function also needs `themeOverride`. Pease check our a good example, e.g. `RadioInput` how it works

packages/ui-link/src/Link/index.tsx
```diff
@@ -37,226 +44,261 @@ import {
...
+      size: sizeProp,
+      ...rest
+    } = props
+
+    // This block handle deprecated variant values by mapping them to new variant + size props
+    let variant: 'inline' | 'standalone' | undefined = variantProp as
+      | 'inline'
+      | 'standalone'
+      | undefined
+    let size: 'small' | 'medium' | 'large' | undefined = sizeProp
```
matyasf, line 95: This is not needed

packages/ui-link/src/Link/index.tsx
```diff
@@ -37,226 +44,261 @@ import {
...
+    const styles = useStyle({
+      generateStyle,
+      params: {
+        ...props,
+        variant,
+        size,
+        display: displayValue,
+        margin,
+        containsTruncateText,
+        hasVisibleChildren: hasVisibleChildrenValue
```
matyasf, line 179: Its hard to read what is happening here. Please dont spread props here, write out all the variables you're sending over

packages/ui-link/src/Link/index.tsx
```diff
@@ -37,226 +44,261 @@ import {
...
-    }
-  }
+    const interaction = getInteraction({
+      props,
+      interactionTypes: ['disabled']
+    })
 
-  get containsTruncateText() {
-    let truncateText = false
+    const element = getElementType(LinkComponentForGetElementType, props)
```
matyasf, line 161: This is ugly :( But I see that this was the quickest way to do this refactor

packages/ui-link/src/Link/index.tsx
```diff
@@ -37,226 +44,261 @@ import {
...
-import type { LinkProps, LinkState, LinkStyleProps } from './props'
 
-import type { ViewOwnProps } from '@instructure/ui-view'
+import type { LinkProps } from './props'
 
 /**
 ---
 category: components
 ---
 **/
```
matyasf, line 56: This comment needs to be above the function, otherwise it will be in the wrong place in the docs

packages/ui-link/src/Link/README.md
matyasf, line 145: This is not 100% true now. As I see we only show the underline for inline variants?

packages/ui-link/src/Link/styles.ts
```diff
@@ -22,58 +22,94 @@
...
- * @param  {Object} state the state of the component, the style is applied to
+ * @param  {Object} params the props of the component, the style is applied to
+ * @param  {Object} sharedTokens Shared token object that stores common values for the theme.
  * @return {Object} The final style object, which will be used in the component
  */
 const generateStyle = (
-  componentTheme: LinkTheme,
-  props: LinkProps,
-  state: LinkStyleProps
+  componentTheme: NewComponentTypes['Link'],
```
matyasf, line 58: As I see you're not using the `inlineLink` token from the theme, is this OK?

packages/ui-link/src/Link/styles.ts
```diff
@@ -22,58 +22,94 @@
...
+  params: StyleParams,
+  sharedTokens: SharedTokens
 ): LinkStyle => {
   const {
     isWithinText,
     renderIcon,
     iconPlacement = 'start', // TODO workaround needed for react 19 where defaultprops doesn't apply for some reasong
```
matyasf, line 65: This TODO is only from React components.

packages/ui-link/src/Link/index.tsx
```diff
@@ -37,226 +44,261 @@ import {
...
-  get display() {
-    if (this.props.display) {
-      return this.props.display // user-entered display property
+      // Forward ref to the DOM element for compatibility with Position/Popover/Tooltip
+      if (typeof ref === 'function') {
+        ref(el)
+      } else if (ref) {
+        // eslint-disable-next-line no-param-reassign
+        ;(ref as React.MutableRefObject<Element | null>).current = el
+      }
```
matyasf, line 198: Why is this part needed?

packages/ui-link/src/Link/index.tsx
```diff
@@ -37,226 +44,261 @@ import {
...
+      const iconSize = linkSizeToIconSize[size || 'medium']
+
+      // Lucide icons - pass size prop to control icon size
+      // Wrap with span for gap/spacing but without fontSize override
+      const isLucideIcon =
+        renderIcon &&
+        isValidElement(renderIcon) &&
+        (renderIcon as any).type?.displayName?.startsWith('wrapLucideIcon')
+
+      if (isLucideIcon && isValidElement(renderIcon)) {
```
matyasf, line 245: You are already checking for `isValidElement` 3 lines above

matyasf (changes requested):
See my comments. This is becoming so complex that we should definitely not try to convert to functional code in the future (also its hard to review)

PR #2257, opened 2025-11-20
fix(ui-modal): adjust scrollbar detection tolerance in ModalBody

matyasf:
Increase epsilon tolerance from 0.05 to 1 pixel to avoid false positives when detecting scrollbars due to browser rounding errors. This prevents unnecessary tabIndex=0 from being applied when no scrollbar is present.

Fixes INSTUI-4863

To test:
Set checkbox sizes with this code:
```
<Modal open={true} label="Modal">
  <Modal.Header> 
    <Heading>Modal</Heading>
  </Modal.Header>
  <Modal.Body>
    <Checkbox label="Modal" size="small" /> 
  </Modal.Body>
</Modal>
```

There should be no scrollbar


ToMESSKa:
@matyasf Test plan says "There should be no scrollbar", but there is no scrollbar in the latest version with that code either. Shouldn't we check for the modal body not getting focused instead? I think the fix is about that.

Issue #2249, opened 2025-11-18
List indentation differs for ordered and unordered lists


buckett:
### Background Information

**Package Version(s)**: 10.26.

**Browser**: Chrome

**OS:** Mac OS

**Device:** Macbook

**Component**: List

### Describe the Bug

The unordered list is indented less than ordered list. I'm not sure if this is a "bug" or something that is expected. There is code to produce this behaviour so I'm assuming it's supposed to be like this, but if that's the case I think it would be good to have the documentation page show this behaviour.

We wanted 2 lists to align vertically (one unordered, one ordered) and when displaying in Canvas would see a significant difference in alignment.

### Steps To Reproduce

- https://instructure.design/List
- Paste this into the first editable section:
```jsx
<div>
  <List margin="0 0 medium" themeOverride={{listPadding: "10rem"}}>
    <List.Item>List item 1</List.Item>
    <List.Item>List item 2</List.Item>
    <List.Item>List item 3</List.Item>
  </List>
  <List as="ol" margin="0 0 medium" themeOverride={{listPadding: "10rem"}}>
    <List.Item>List item 1</List.Item>
    <List.Item>List item 2</List.Item>
    <List.Item>List item 3</List.Item>
  </List>
</div>
```
### Expected Behavior

The two lists align.

### Screenshots

<img width="1028" height="277" alt="Image" src="https://github.com/user-attachments/assets/33ea3538-8d25-4d34-89da-a455e0430036" />

### Additional Information

**Current Workaround(s):**

Double the list padding for the ordered list on the component.

```
<List as="ol" themeOverride={(componentTheme) => ({
            listPadding: `calc(${componentTheme.listPadding} * 2)`
                })}>
```

**Products Affected:**

Locally developed LTI tools.

**Are you willing to submit a PR to fix?**

- [ ] Yes, I'm willing to submit a PR

**Requested Priority:** Low

PR #2259, opened 2025-11-20
fix(ui-list): align ordered list start position with unordered lists

matyasf:
The previous implementation divided `listPadding` by 2, but this did not correctly align ordered and unordered lists. Changed to subtract a fixed `1.3rem` offset from the `listPadding` to ensure both list types start at the same vertical position.

To test: Compare ordered and unordered lists side-by-side as shown in the steps to reproduce example. Both list types should now start at the same horizontal position. Test with different override values

Fixes https://github.com/instructure/instructure-ui/issues/2249

### DIFF
--- a/packages/ui-list/src/List/styles.ts
+++ b/packages/ui-list/src/List/styles.ts
@@ -51,7 +51,9 @@ const generateStyle = (
 
       ...(ordered && {
         listStyleType: 'none',
-        paddingInlineStart: `calc(${componentTheme.listPadding} / 2)`,
+        // 1.3 rem is removed to have the ordered list start at the same point
+        // as the unordered vertically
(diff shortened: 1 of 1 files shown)


packages/ui-list/src/List/styles.ts
```diff
@@ -51,7 +51,9 @@ const generateStyle = (
 
       ...(ordered && {
         listStyleType: 'none',
-        paddingInlineStart: `calc(${componentTheme.listPadding} / 2)`,
+        // 1.3 rem is removed to have the ordered list start at the same point
```
balzss, line 54: this is good for now, but later should be updated to something more "proper"

balzss (approved):

PR #2262, opened 2025-11-21
INSTUI-4813 refactor(ui-text): migrate to new theming system

matyasf:
Migrates Text component from legacy theme.ts to new token-based theming system:

- Remove theme.ts file and use direct token references in styles
- Update color mappings to use new semantic tokens (baseColor, mutedColor, etc.)
- Add fontFamily to variant definitions from typography tokens
- Update theme variable references throughout styles.ts

BREAKING CHANGE: Removed 'alert' color prop value. Use 'primary' color instead.

Also improves documentation:
- Clarify variant usage and recommendations in README
- Add deprecation notes for size, weight, and lineHeight legacy values
- Update color examples to include inverse colors and remove alert
- Add JSDoc comments for deprecated prop values
- Improve code examples with clearer formatting

To test: Compare with Figma design


docs/guides/upgrade-guide.md
```diff
@@ -16,6 +16,11 @@ InstUI has switched to a new icon set, [Lucide](https://lucide.dev/icons/). We a
 
 ## Removal of deprecated props/components/APIs
 
+### Text
+
+- `alert` color has been removed. Please use `primary` instead.
+- Some prop values have been deprecated, see [Text](/Text) for more details.
```
matyasf, line 40: This link is not working for some reason locally, it's a docs issue

packages/ui-text/src/Text/styles.ts
```diff
@@ -22,89 +22,100 @@
...
     },
     contentImportant: {
-      fontStyle: 'normal',
-      fontWeight: componentTheme.weightImportant,
-      fontSize: componentTheme.content,
-      lineHeight: componentTheme.lineHeight150
+      fontFamily: componentTheme.contentImportant.fontFamily,
+      fontWeight: componentTheme.contentImportant.fontWeight,
+      fontSize: componentTheme.contentImportant.fontSize,
+      lineHeight: componentTheme.contentImportant.fontSize
```
git-nandor: I think lineHeight should be used here instead of fontSize.

matyasf: good catch, fixed

packages/ui-text/src/Text/styles.ts
adamlobler: I realized that the text colors don’t change between themes when the color prop isn’t set. It gets a default color from the body somehow, so once we rewrite this it should work, but it might be safer to define the default color here. I don't have time to test it, but I would add the `componentTheme.baseColor` token to the false part of the ternary. What do you think?

adamlobler (changes requested):

packages/ui-text/src/Text/styles.ts
matyasf: good catch, I've fixed it

joyenjoyer (approved):
great job

git-nandor (approved):

PR #2265, opened 2025-11-25
refactor(ui-radio-input,ui-avatar,emotion): split useStyle into old and new theming engine part

matyasf:
also make `themeOverride` prop always avaiable in the props

### DIFF
--- a/packages/emotion/src/useStyleRework.ts
+++ b/packages/emotion/src/useStyleRework.ts
@@ -1,0 +1,92 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015 - present Instructure, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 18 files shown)


packages/emotion/src/useStyleRework.ts
```diff
@@ -0,0 +1,100 @@
+/*
```
matyasf, line 1: This uses the same naming convention as https://github.com/instructure/instructure-ui/pull/2254

packages/ui-avatar/src/Avatar/props.ts
```diff
@@ -110,7 +109,7 @@ type PropKeys = keyof AvatarOwnProps
 type AllowedPropKeys = Readonly<Array<PropKeys>>
 
 type AvatarProps = AvatarOwnProps &
-  WithStyleProps<AvatarTheme, AvatarStyle> &
+  ThemeOverrideProp &
   OtherHTMLAttributes<AvatarOwnProps>
```
matyasf: This is the syntax that we should use for funcitonal components

HerrTopi (approved):

PR #2266, opened 2025-11-26
Add lucide icons to avatar

joyenjoyer:
Closes INSTUI-4864

Test: 
Try all these icon adding options with Avatar and check whether size and color passed to the icon correctly. Also check how changing theme impact the rendered component.
```
renderIcon={AlarmClockInstUIIcon} // reference to icon
renderIcon={<AlarmClockInstUIIcon />} // icon as React component
renderIcon={()=><AlarmClockInstUIIcon />} // icon as React component returned by an arrow function

### CODE REVIEW

File: packages/__docs__/src/Nav/index.tsx [hunk 1]
```diff
@@ -499,6 +499,7 @@ class Nav extends Component<NavProps, NavState> {
     const themes = this.renderThemes()
     const icons = (
       <NavToggle
+        key="icons"
```
joyenjoyer: Fixed these annoying key warning in the docs app.

packages/ui-avatar/src/Avatar/styles.ts
```diff
@@ -113,38 +111,31 @@ const generateStyle = (
   const colorVariants = {
     accent1: {
       text: componentTheme.blueTextColor,
-      background: componentTheme.blueBackgroundColor,
-      icon: iconTokens.accentBlueColor
```
joyenjoyer, line 117: The icon color should not be here.

packages/ui-avatar/src/Avatar/index.tsx
```diff
@@ -124,6 +122,23 @@ const Avatar = forwardRef<HTMLDivElement, AvatarProps>(
       />
     )
 
+    const getIconColor = () => {
```
joyenjoyer: This function matches the right icon color to Avatar's color property.

packages/ui-avatar/src/Avatar/README.md
```diff
@@ -46,16 +46,47 @@ type: example
...
 ---
 <View display="block" padding="small medium" background="primary">
-  <Avatar size="xx-small" color="ai" name="AI Assistant" renderIcon={IconAiSolid}  />
-  <Avatar size="x-small" color="ai" name="AI Assistant" renderIcon={IconAiSolid}  />
-  <Avatar size="small" color="ai" name="AI Assistant" renderIcon={IconAiSolid}  />
-  <Avatar size="medium" color="ai" name="AI Assistant" renderIcon={IconAiSolid}  />
-  <Avatar size="large" color="ai" name="AI Assistant" renderIcon={IconAiSolid}  />
-  <Avatar size="x-large" color="ai" name="AI Assistant" renderIcon={IconAiSolid}  />
-  <Avatar size="xx-large" color="ai" name="AI Assistant" renderIcon={IconAiSolid}  />
+  <Avatar size="xx-small" color="ai" name="AI Assistant" renderIcon={IconAiSolid} />
```
joyenjoyer, line 49: This has to be replaced later to the appropriate Lucide equivalent icon. (Alex is still working on the mapping)

packages/ui-avatar/src/Avatar/styles.ts
```diff
@@ -162,6 +153,20 @@ const generateStyle = (
     return 'solid'
   }
 
+  const iconColorMap = {
```
joyenjoyer: IconColorMap is placed here because the avatar theme can be directly accesse, no `useTheme()` call is needed.

packages/ui-react-utils/src/renderLucideIconWithProps.ts
```diff
@@ -0,0 +1,111 @@
...
+ * Check if an element/component is a Lucide icon by checking its displayName
+ * @param element - Element or component to check
+ * @returns True if it's a Lucide icon (displayName starts with 'wrapLucideIcon(')
+ */
+const isLucideIcon = (element: any): boolean => {
+  if (React.isValidElement(element)) {
+    const displayName = (element.type as any)?.displayName
+    return displayName?.startsWith('wrapLucideIcon(')
+  }
+  return element?.displayName?.startsWith('wrapLucideIcon(')
```
HerrTopi: I don't understand this. If not valid element, why return this line? Why not return false? Can a valid lucide icon be a not valid element?

packages/ui-react-utils/src/renderLucideIconWithProps.ts
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * ---
+ * category: utilities/react
+ * ---
+ * Renders a Lucide icon with props, handling multiple input formats.
+ * Only applies props if the icon is a Lucide icon (detected via displayName).
+ * For non-Lucide icons, renders without applying props.
+ *
+ * Supported use cases:
+ * 1. Component/function with Lucide displayName: `renderIcon={UserIcon}` - Calls with props
```
HerrTopi: What does Component/function with Lucide displayName mean?

packages/ui-react-utils/src/renderLucideIconWithProps.ts
matyasf: If its not a valid React element, it cannot be a Lucide icon, right?

packages/ui-react-utils/src/renderLucideIconWithProps.ts
```diff
@@ -0,0 +1,111 @@
...
+ * 2. JSX element: `renderIcon={<UserIcon />}` - Clones and overrides props
+ * 3. Arrow function returning Lucide: `renderIcon={() => <UserIcon />}` - Detects Lucide from result
+ *
+ * @module renderLucideIconWithProps
+ * @param elementToRender - The element to render (component/JSX/function)
+ * @param propsToApply - Props to pass to or override on the element (only for Lucide icons)
+ * @returns Rendered React element or null
+ */
+function renderLucideIconWithProps<P extends Record<string, unknown>>(
+  elementToRender: Renderable<P> | React.ReactElement | undefined,
```
matyasf: `Renderable` type entails `ReactElement` and `undefined` already

packages/ui-react-utils/src/renderLucideIconWithProps.ts
```diff
@@ -0,0 +1,111 @@
...
+): React.ReactElement | null {
+  if (!elementToRender) return null
+
+  // Check once if the input is a Lucide icon
+  const isInputLucide = isLucideIcon(elementToRender)
+
+  // Use case 2: JSX element like <UserIcon />
+  if (React.isValidElement(elementToRender)) {
+    return isInputLucide
+      ? React.cloneElement(elementToRender, propsToApply as any)
```
matyasf: the `as any` cast is not needed here

packages/ui-react-utils/src/renderLucideIconWithProps.ts
```diff
@@ -0,0 +1,111 @@
...
+  const result = callRenderProp(
+    elementToRender,
+    isInputLucide ? propsToApply : ({} as P)
+  )
+
+  // Check if the result is a React element that needs props applied
+  if (React.isValidElement(result)) {
+    // Use case 3: Arrow function returning Lucide icon (e.g., () => <UserIcon />)
+    // If function wasn't directly Lucide but result is, apply props now
+    if (!isInputLucide && isLucideIcon(result)) {
```
matyasf: Why is the `!isInputLucide` check needed here? Can there be a case where the input and the result are also Lucide icons? If this is the case the isValidElement check above is also not needed

packages/ui-react-utils/src/renderLucideIconWithProps.ts
```diff
@@ -0,0 +1,111 @@
...
+      })
+    }
+
+    // Use case 1: Function with Lucide displayName that ignored props
+    // Force props application by cloning the result
+    if (isInputLucide) {
+      return React.cloneElement(result, {
+        ...(result.props as Record<string, unknown>),
+        ...propsToApply
+      })
```
matyasf: When can this use case happen? isnt this covered by line 70?

matyasf (changes requested):
see my comments

packages/ui-react-utils/src/renderLucideIconWithProps.ts
joyenjoyer: The check is needed because React elements and component references have displayName in different places.
```
For JSX element  -> <UserIcon /> -> displayName is at: element.type.displayName
Component reference -> UserIcon -> displayName is at: element.displayName

Comment (matyasf) [reply packages/ui-react-utils/src/renderLucideIconWithProps.ts]: hmm I would mention this in its JSDOC

File: packages/ui-avatar/src/Avatar/__tests__/Avatar.test.tsx [hunk 12]
```diff
@@ -107,249 +99,30 @@ describe('<Avatar />', () => {
...
-        <Avatar name="avatar name" size="x-small" renderIcon={UserInstUIIcon} />
-      )
-
-      const avatarSvg = container.querySelector('svg')
-      expect(avatarSvg).toBeInTheDocument()
+      const svg = container.querySelector('svg')
+      expect(svg).toBeInTheDocument()
     })
 
-    it('should accept a JSX element and clone it with size/color props', async () => {
```
joyenjoyer: I removed all the MockIcon tests from here. Proper context handling is being tested in the IconPropsProvider folder with unit tests. Only basic rendering tests remain here.

packages/ui-avatar/src/Avatar/index.tsx
```diff
@@ -138,13 +148,15 @@ const Avatar = forwardRef<HTMLDivElement, AvatarProps>(
...
-        // TODO we should never do this, do not create a fake style
-        const iconColor = styles?.iconColor
+        const iconColor = hasInverseColor ? 'onColor' : ICON_COLOR_MAP[color]
 
-        return renderLucideIconWithProps(renderIcon, {
-          size: iconSize,
-          color: iconColor
-        })
+        return (
+          <IconPropsProvider size={iconSize} color={iconColor}>
```
joyenjoyer: I removed `renderLucideIconWithProps` since it was too complex to understand and added `IconPropsProvider` instead. Lucide icons consume `size` and `color` through it.

This works with all the wanted patterns:
```
renderIcon={AlarmClockInstUIIcon}
renderIcon={<AlarmClockInstUIIcon />}
renderIcon={()=><AlarmClockInstUIIcon />}

File: packages/ui-avatar/src/Avatar/props.ts [hunk 14]
```diff
@@ -101,7 +101,7 @@ type AvatarOwnProps = {
    * An icon, or function that returns an icon that gets displayed. If the `src` prop is provided, `src` will have priority.
    * When using Lucide icons, Avatar will automatically pass the appropriate size and color props based on the Avatar's size and color.
    */
-  renderIcon?: Renderable<{ size?: string | number; color?: string }>
+  renderIcon?: Renderable
```
joyenjoyer, line 104: Removed unnecessarily complex typing

packages/ui-icons-lucide/src/wrapLucideIcon/index.tsx
```diff
@@ -50,86 +50,30 @@ export function wrapLucideIcon(Icon: LucideIcon): LucideIcon {
...
     } = props
-    // TODO we should never do this
-    const theme = useTheme() as Theme
-    const iconTheme = theme?.newTheme?.components?.Icon
+
+    // Get icon props from context (if available)
+    const contextProps = useIconProps()
+
+    // Merge props: direct props take precedence over context props
+    const finalSize = size ?? contextProps?.size
```
joyenjoyer: Followed the React best practice here, component props always override context values by default.

packages/ui-react-utils/src/IconPropsProvider/IconPropsContext.ts
```diff
@@ -0,0 +1,30 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+import React from 'react'
+import type { IconPropsContextValue } from './IconPropsProvider'
```
joyenjoyer, line 26: I followed the same file structuring I saw with DeterministicIdProvider

packages/ui-icons-lucide/src/wrapLucideIcon/props.ts
```diff
@@ -24,40 +24,91 @@
...
+  | 'mutedColor'
+  | 'successColor'
+  | 'errorColor'
+  | 'warningColor'
+  | 'infoColor'
+  | 'onColor'
+  | 'inverseColor'
+  | 'disabledBaseColor'
+  | 'disabledOnColor'
+  | 'dark'
```
joyenjoyer, line 67: Designers confirmed that `dark` is a valid color, they will use it in the future.

packages/ui-react-utils/src/__tests__/IconPropsProvider.test.tsx
```diff
@@ -0,0 +1,161 @@
...
+ * SOFTWARE.
+ */
+
+import { render, screen } from '@testing-library/react'
+import '@testing-library/jest-dom'
+
+import { IconPropsProvider, useIconProps } from '../IconPropsProvider'
+
+// Test component that exposes context values
+const TestComponentWithHook = () => {
```
joyenjoyer, line 31: Mocks are perfectly valid here, since these are unit tests.

packages/ui-icons-lucide/src/wrapLucideIcon/props.ts
matyasf: The old icons use `'x-small' | 'small' | 'medium' | 'large' | 'x-large'` sizes. How will we handle the migration?

matyasf: The old icons use other values (`'inherit',  'primary', 'secondary',...` How do we handle the migration?

packages/__docs__/src/Icons/LucideIconsGallery.tsx
matyasf: `currentColor` is not valid AFAIK, a better example would be `successColor`

packages/__docs__/src/Icons/LucideIconsGallery.tsx
```diff
@@ -299,7 +299,6 @@ const LucideIconsGallery = () => {
                 <li>
                   <code>strokeWidth</code>: Number - e.g., <code>2</code>
```
matyasf: a better example is `sm`

packages/__docs__/src/Icons/LucideIconsGallery.tsx
matyasf: here a semantic example would be better

packages/__docs__/src/Icons/LucideIconsGallery.tsx
```diff
@@ -299,7 +299,6 @@ const LucideIconsGallery = () => {
...
                 </li>
-                <li>Plus all standard SVG props (className, style, etc.)</li>
```
matyasf, line 302: You are adding SVG props to the Icon (with `passthroughProps(rest)` ), so they are allowed (also this is OK because we do the same with the old icons). But this util removes `className`, `children` and a couple of other props

packages/ui-icons-lucide/src/wrapLucideIcon/props.ts
```diff
@@ -105,6 +159,19 @@ type LucideIconWrapperProps = Omit<
     themeOverride?: ThemeOverrideValue
   } & OtherHTMLAttributes<InstUIIconOwnProps>
```
matyasf: Here you could omit the props removed by `passtroughProps`, I think the syntax is `Omit<OtherHTMLAttributes<InstUIIconOwnProps>, 'children', ...>

packages/ui-avatar/package.json
matyasf: is `ui-icons` still needed?

packages/ui-react-utils/src/IconPropsProvider/IconPropsProvider.tsx
```diff
@@ -0,0 +1,49 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+import { IconPropsContext } from './IconPropsContext'
+
+type IconPropsContextValue = {
+  size?: string | number
+  color?: string
```
matyasf, line 29: These have the same type as the props with the same name from `wrapLucideIcon/props.ts` I think. But in this case we will have to pull in the icons package as a dependency, which might create an impossible build graph... So it might be better to move these where the icons are, we will use them with icons anyway

packages/ui-icons-lucide/src/wrapLucideIcon/props.ts
matyasf: I would add these to the props too, ask for a mapping from design, and do it in the code in index.ts/styles.ts

matyasf: I would add these to the props too, ask for a mapping from design, and do it in the code in index.ts/styles.ts

packages/ui-icons-lucide/src/wrapLucideIcon/props.ts
matyasf: e.g.
```
type LucideIconWrapperProps = Omit<LucideProps, 'size' | 'color' | 'strokeWidth' | 'rotate' | 'children' | 'style' | others> &
  InstUIIconOwnProps & {
    themeOverride?: ThemeOverrideValue
  }
```

matyasf (changes requested):
Looks good!
I've found just one small visual bug: In the themeOverride prop example the examples are not aligned vertically. 

Image

According to AI its because "The children elements all have display: inline-flex and vertical-align: baseline. While they all have the same height (48px), vertical-align: baseline aligns the baselines of the elements. If there is text content or different line-height values within the inline-flex items, or if the parent div has a specific line-height or font-size that interacts with the baseline, it can cause vertical misalignment"

and the fix is pretty simple, just add `verticalAlign:'middle` to the `avatar` CSS class

packages/ui-icons-lucide/src/wrapLucideIcon/index.tsx
HerrTopi: In general I would agree, but this code design explicitly wants to force props if applicable. For example we don't want to have misconfigured Avatars with incorrects sizes or colors. I'd default to context here and if the need arises, we could have a prop like "forceSize" or something later

packages/ui-icons-lucide/src/wrapLucideIcon/index.tsx
```diff
@@ -49,84 +52,30 @@ export function wrapLucideIcon(Icon: LucideIcon): LucideIcon {
...
-        // Semantic color token from theme (exclude size/strokeWidth/dark properties)
-        colorValue = color
-      } else {
-        // Custom CSS color (e.g., "#ff0000", "rgb(255, 0, 0)")
-        customColor = color
-      }
-    }
-
     const styles = useStyle({
       componentId: 'Icon' as const,
```
HerrTopi: Nitpic but this const doesn't do anything here, remove it please so it wont clutter the code

packages/ui-icons-lucide/src/wrapLucideIcon/index.tsx
```diff
@@ -49,84 +52,30 @@ export function wrapLucideIcon(Icon: LucideIcon): LucideIcon {
...
       generateStyle,
       themeOverride,
       params: {
-        size: semanticSize as InstUIIconOwnProps['size'],
-        color: colorValue,
+        size: finalSize as LucideIconWrapperProps['size'],
```
HerrTopi: This cast isnt't needed

packages/ui-icons-lucide/src/wrapLucideIcon/index.tsx
```diff
@@ -49,84 +52,30 @@ export function wrapLucideIcon(Icon: LucideIcon): LucideIcon {
...
+        strokeWidth,
+        color: finalColor,
```
HerrTopi, line 78: What's the reason behind `displayName: `LucideIcon(${Icon.displayName || Icon.name})  ` aremt't we sure that displayName (or name) will be available?

Line 83. For some reason I can't comment there

HerrTopi: Line 86:  const accessibilityProps: Record<string, string | boolean> = {}

| boolean is not needed

packages/ui-icons-lucide/src/wrapLucideIcon/index.tsx
```diff
@@ -143,25 +92,72 @@ export function wrapLucideIcon(Icon: LucideIcon): LucideIcon {
...
+          strokeWidth={styles.numericStrokeWidth}
           absoluteStrokeWidth={absoluteStrokeWidth}
           {...accessibilityProps}
-          {...passthroughProps(rest)}
         />
       </span>
     )
   }
 
   WrappedIcon.displayName = `wrapLucideIcon(${Icon.displayName || Icon.name})`
```
HerrTopi: Why is this displayname different from the one in line 83? It should be the same for consistency and theme override's sake

packages/ui-icons-lucide/src/wrapLucideIcon/index.tsx
```diff
@@ -143,25 +92,72 @@ export function wrapLucideIcon(Icon: LucideIcon): LucideIcon {
...
 
-  return WrappedIcon as LucideIcon
+  return WrappedIcon
 }
 
 export type { LucideIconWrapperProps, InstUIIconOwnProps }
```
HerrTopi: I don't understand this part. Why re-export these from here? These are already exported from their respected files.

HerrTopi: Same for export { default as generateStyle } from './styles'

packages/ui-icons-lucide/src/wrapLucideIcon/props.ts
```diff
@@ -24,37 +24,97 @@
...
+  | 'actionSecondaryOnColorHoverColor'
+  | 'actionSecondaryOnColorActiveColor'
+  | 'actionSecondaryOnColorDisabledColor'
+  | 'actionSuccessSecondaryHoverColor'
+  | 'actionSuccessSecondaryActiveColor'
+  | 'actionDestructiveSecondaryHoverColor'
+  | 'actionDestructiveSecondaryActiveColor'
 
 type InstUIIconOwnProps = {
   /**
```
HerrTopi, line 129: Again, I can't comment on unchanged parts, but:
StorkeWidth is a not allowed prop, all stroke width have to be  calculated from size and are not changable.
Size van not be number, only the preset tokens. 
Same for color. No inherit, no string, only the presets
Ask about these from desing

HerrTopi: Calculations and  logic are tied to these things in styles.ts, so I won'r review it in depth because it needs to be drastically changed before.

adamlobler (changes requested):
For the Avatar, the size mapping looks okay in the code, but I’ve noticed something strange. In the design, for example, an xs icon has a 12px width and a 1px stroke width, but when we use the same icon size and stroke width in the code, they look different. This is most obvious for the xs size, but I think it affects all sizes.

Image

Can you check what might be wrong here? In the design, the pixel grid is visible, so we can easily verify the stroke width, and it is exactly 1px there.

adamlobler:
> > @joyenjoyer I think I’ve already found the solution. We use a 24px viewBox for all sizes. For example, a 12px icon has a 24px viewBox and a 1px stroke width. Because the viewBox is twice the size of the icon, the stroke width becomes half of the original size.
> > So, in this case, we would need 1px * (24 / 12) = 2px stroke width to achieve a real 1px-wide stroke. We can apply the same logic to all sizes. For example: a 20px icon with a 1.5px original stroke width and a 24px viewBox would need 1.5 * (24 / 20) = 1.8px stroke width.
> > Or we could figure out how to use a suitable viewBox size for each icon size.
> 
> In my understanding, this can be solved in Token Studio changing strokeWidth values, right?

No, it should be fixed in code, because now in code the strokewidth is not 1px when you set it to 1px because of the different viewbox size in the SVG itself

packages/ui-icons-lucide/src/wrapLucideIcon/styles.ts
```diff
@@ -34,28 +35,120 @@ type StyleParams = {
...
+      return px(componentTheme[propName])
+    }
+  }
+  return undefined
+}
+
+/**
+ * Determine semantic color value from theme
+ */
+const determineColorValue = (
```
HerrTopi: This method does nothing really. It can be the following:

const colorValue = componentTheme[color]

(with some tweaking to remove non eligable props)

packages/ui-icons-lucide/src/wrapLucideIcon/index.tsx
```diff
@@ -22,119 +22,74 @@
...
       title,
       elementRef,
       themeOverride,
-      absoluteStrokeWidth,
-      className,
-      style,
+      // Default to true: ensures stroke width renders at actual pixel values
+      // regardless of icon size (fixes viewBox scaling issue where 24x24 viewBox
+      // causes strokes to scale proportionally with icon size)
+      absoluteStrokeWidth = true,
```
matyasf: What is the ViewBox scaling issue? Also why do we need this prop to be settable externally? Could not just we use always `true`?

packages/ui-icons-lucide/src/wrapLucideIcon/index.tsx
```diff
@@ -22,119 +22,74 @@
...
     } = props
 
-    const theme = useTheme() as Theme
-    const iconTheme = theme?.newTheme?.components?.Icon
+    // Get icon props from context (if available)
+    const contextProps = useContext(IconPropsContext)
+
+    // Merge props: context props take precedence over direct props
+    const finalSize = contextProps?.size ?? size
+    const finalColor = color ?? contextProps?.color
```
matyasf: Both should be read from context first. Or is there a reason for this discrepancy?

packages/ui-icons-lucide/src/wrapLucideIcon/props.ts
```diff
@@ -24,51 +24,102 @@
...
+   * Semantic size token (also determines stroke width automatically)
    */
-  strokeWidth?: IconStrokeWidthToken | number | string
+  size?: IconSizeToken
   /**
-   * Icon theme color token or CSS color value
+   * Icon theme color token
    */
-  color?: 'inherit' | IconColorToken | string
+  color?: 'inherit' | IconColorToken
```
matyasf: this should also contain all the other legacy tokens too (and they need some mapping for to the new colors -- ask design for the mapping)

packages/ui-icons-lucide/src/wrapLucideIcon/styles.ts
```diff
@@ -34,28 +35,120 @@ type StyleParams = {
...
+
+    // Fall back to componentTheme (for other tokens like xs, sm, md, lg, xl, 2xl)
+    const propName = `size${size.charAt(0).toUpperCase()}${size.slice(
+      1
+    )}` as keyof typeof componentTheme
+    if (propName in componentTheme) {
+      return px(componentTheme[propName])
+    }
+  }
+  return undefined
```
matyasf: if its some not supported value (e.g. a number) we should show a warning

packages/ui-icons-lucide/src/wrapLucideIcon/styles.ts
```diff
@@ -34,28 +35,120 @@ type StyleParams = {
...
+  if (typeof size === 'string') {
+    // Check SVGIcon tokens first (for legacy size tokens)
+    if (size in svgIconSizeMapping) {
+      return px(svgIconSizeMapping[size as keyof typeof svgIconSizeMapping])
+    }
+
+    // Fall back to componentTheme (for other tokens like xs, sm, md, lg, xl, 2xl)
+    const propName = `size${size.charAt(0).toUpperCase()}${size.slice(
+      1
+    )}` as keyof typeof componentTheme
```
matyasf: We should not rely on such naming conventions, it creates the possibility of nasty bugs. Please just use a hashmap like 
```
{
  xs: sizeXs,
...
}
```

packages/ui-icons-lucide/src/wrapLucideIcon/styles.ts
```diff
@@ -34,28 +35,120 @@ type StyleParams = {
...
+/**
+ * Convert semantic size token to stroke width value for Lucide
+ * Derives stroke width from size
+ */
+const convertSizeToStrokeWidth = (
+  size: LucideIconWrapperProps['size'],
+  componentTheme: NewComponentTypes['Icon']
+) => {
+  if (typeof size === 'string') {
+    const propName = `strokeWidth${size.charAt(0).toUpperCase()}${size.slice(
```
matyasf: again, please just make a hashmap

packages/ui-icons-lucide/src/wrapLucideIcon/props.ts
```diff
@@ -96,15 +147,38 @@ type InstUIIconOwnProps = {
...
+ * children, style, and className are explicitly omitted.
  */
 type LucideIconWrapperProps = Omit<
-  LucideProps,
-  'size' | 'color' | 'strokeWidth' | 'rotate'
-> &
-  InstUIIconOwnProps & {
-    themeOverride?: ThemeOverrideValue
-  } & OtherHTMLAttributes<InstUIIconOwnProps>
+  Omit<LucideProps, 'size' | 'color' | 'strokeWidth' | 'rotate'> &
```
matyasf: I dont think we need to expose all Lucide props here. As I see Lucide is accepting SVG props, but we're already allowing this with `OtherHTMLAttributes` (and just this is needed for good backward compatibility with old icons)

matyasf (changes requested):

packages/ui-icons-lucide/src/wrapLucideIcon/styles.ts
```diff
@@ -135,7 +191,7 @@ const generateStyle = (
 
   if (colorValue) {
     if (colorValue === 'inherit') {
-      resolvedColor = 'inherit'
+      resolvedColor = 'currentColor'
```
joyenjoyer: https://lucide.dev/guide/basics/color

`currentColor` has to be set to inherit colors.

packages/ui-icons-lucide/scripts/generateIndex.ts
```diff
@@ -106,11 +106,8 @@ async function generateIndex() {
 import * as Lucide from 'lucide-react'
 import { wrapLucideIcon } from './wrapLucideIcon'
 
-export type { LucideProps, LucideIcon } from 'lucide-react'
-export type { LucideIconWrapperProps, InstUIIconOwnProps } from './wrapLucideIcon'
-export { wrapLucideIcon }
```
joyenjoyer, line 111: Removed unneccessary exports

packages/ui-icons-lucide/src/index.ts
```diff
@@ -25,14 +25,8 @@
 import * as Lucide from 'lucide-react'
 import { wrapLucideIcon } from './wrapLucideIcon'
 
-export type { LucideProps, LucideIcon } from 'lucide-react'
```
joyenjoyer, line 28: Removed unneccessary exports

packages/ui-icons-lucide/src/wrapLucideIcon/styles.ts
```diff
@@ -125,19 +71,24 @@ const convertSizeToStrokeWidth = (
...
-    return px(componentTheme[themeKey as keyof NewComponentTypes['Icon']])
+  const strokeWidthTokenMap = {
+    xs: componentTheme.strokeWidthXs,
+    sm: componentTheme.strokeWidthSm,
+    md: componentTheme.strokeWidthMd,
+    lg: componentTheme.strokeWidthLg,
+    xl: componentTheme.strokeWidthXl,
+    '2xl': componentTheme.strokeWidth2xl,
+    // Legacy SVGIcon stroke tokens (DEPRECATED)
+    'x-small': '0.0859375rem',
```
joyenjoyer, line 82: strokeWidth values for legacy sizes from design

packages/ui-icons-lucide/src/wrapLucideIcon/styles.ts
```diff
@@ -146,26 +97,113 @@ const determineColorValue = (
...
+    success: componentTheme.successColor,
+    error: componentTheme.errorColor,
+    alert: componentTheme.infoColor,
+    warning: componentTheme.warningColor,
+    brand: componentTheme.infoColor,
 
-  if (color in componentTheme) {
-    return color
+    // auto is legacy token from SVGIcon (DEPRECATED)
+    auto: 'currentColor',
```
joyenjoyer, line 203: if you want to test these values:

```
<div style={{ padding: '20px', fontFamily: 'sans-serif' }}>
      <h2>Icon Color Inheritance Test</h2>

      {/* Test 'inherit' color */}
      <div style={{ marginBottom: '30px' }}>
        <h3>color="inherit"</h3>

        <div style={{ color: 'red', fontSize: '16px', marginBottom: '10px' }}>
          <span>Red parent: </span>
          <HeartInstUIIcon color="inherit" size="md" />
          <StarInstUIIcon color="inherit" size="md" />
          <CheckInstUIIcon color="inherit" size="md" />
        </div>

        <div style={{ color: 'blue', fontSize: '16px', marginBottom: '10px' }}>
          <span>Blue parent: </span>
          <HeartInstUIIcon color="inherit" size="md" />
          <StarInstUIIcon color="inherit" size="md" />
          <CheckInstUIIcon color="inherit" size="md" />
        </div>

        <div style={{ color: 'green', fontSize: '16px', marginBottom: '10px' }}>
          <span>Green parent: </span>
          <HeartInstUIIcon color="inherit" size="md" />
          <StarInstUIIcon color="inherit" size="md" />
          <CheckInstUIIcon color="inherit" size="md" />
        </div>
      </div>

      {/* Test 'auto' color (legacy) */}
      <div style={{ marginBottom: '30px' }}>
        <h3>color="auto" (legacy - DEPRECATED)</h3>

        <div style={{ color: 'purple', fontSize: '16px', marginBottom: '10px' }}>
          <span>Purple parent: </span>
          <HeartInstUIIcon color="auto" size="md" />
          <StarInstUIIcon color="auto" size="md" />
          <CheckInstUIIcon color="auto" size="md" />
        </div>

        <div style={{ color: 'orange', fontSize: '16px', marginBottom: '10px' }}>
          <span>Orange parent: </span>
          <HeartInstUIIcon color="auto" size="md" />
          <StarInstUIIcon color="auto" size="md" />
          <CheckInstUIIcon color="auto" size="md" />
        </div>

        <div style={{ color: '#ff1493', fontSize: '16px', marginBottom: '10px' }}>
          <span>DeepPink parent: </span>
          <HeartInstUIIcon color="auto" size="md" />
          <StarInstUIIcon color="auto" size="md" />
          <CheckInstUIIcon color="auto" size="md" />
        </div>
      </div>

      {/* Compare with explicit color */}
      <div>
        <h3>Comparison: explicit color (does NOT inherit)</h3>

        <div style={{ color: 'red', fontSize: '16px', marginBottom: '10px' }}>
          <span>Red parent with baseColor icons: </span>
          <HeartInstUIIcon color="baseColor" size="md" />
          <StarInstUIIcon color="successColor" size="md" />
          <CheckInstUIIcon color="errorColor" size="md" />
        </div>
      </div>

      {/* Test in nested elements */}
      <div style={{ marginTop: '30px' }}>
        <h3>Nested inheritance test</h3>
        <div style={{ color: 'darkblue', fontSize: '16px' }}>
          <div style={{ padding: '10px', border: '1px solid currentColor' }}>
            Outer (darkblue):
            <HeartInstUIIcon color="inherit" size="lg" />
            <div style={{ color: 'crimson', padding: '10px', marginTop: '10px' }}>
              Inner (crimson):
              <StarInstUIIcon color="inherit" size="lg" />
              <HeartInstUIIcon color="auto" size="lg" />
            </div>
          </div>
        </div>
      </div>
    </div>

File: packages/__docs__/src/Icons/LucideIconsGallery.tsx [hunk 42]
```diff
@@ -42,15 +42,7 @@ import { Flex } from '@instructure/ui-flex'
...
-  (name) =>
-    name !== 'withRTL' &&
-    name !== 'wrapLucideIcon' &&
-    name !== 'LucideProps' &&
-    name !== 'LucideIcon' &&
-    name !== 'LucideIconWrapperProps' &&
-    name !== 'InstUIIconProps' &&
-    name !== 'LucideIconTheme' &&
-    name !== 'BIDIRECTIONAL_ICONS'
+  (name) => name !== 'renderIconWithProps' && name !== 'LucideIconWrapperProps'
```
joyenjoyer, line 45: This part caused the scrolling issue. These need to be excluded from rendering so that only the icons are added to the gallery.

packages/__docs__/src/Icons/LegacyIconsGallery.tsx
```diff
@@ -249,13 +249,6 @@ const LegacyIconsGallery = ({ glyphs }: LegacyIconsGalleryProps) => {
 
   return (
     <div css={{ overflowX: 'hidden' }}>
-      <Alert variant="warning" margin="0 0 medium 0">
```
joyenjoyer, line 252: Removed this alert for now, until we have codemods or aliases ready.

matyasf (approved):
looks good!
In a future ticket we should tidy up the icon gallery, it has a couple of visual issues

HerrTopi (approved):

PR #2271, opened 2025-11-27
feat(ui-number-input): rewrite NumberInput to the new theming system

matyasf:
Convert NumberInput to the new theming system. Notes:

- Converted NumberInput from class-based to functional component
- It uses the theme from `TextInput`, they are the same in Figma
- The styling of the upper label/messages are not done, they are coming in `FormFieldLayout`

Tokens are also not fully used (because they are used only in `TextInput`):
- `fontSizeSm`, `heightSm`, `paddingHorizontalSm`: This component has no `small` size
- `gapContent`:  This is a gap between the text and elements rendered after it, `NumberInput` does not have such

To test:
- check the examples in the docs, they should function exactly as before
- compare its CSS to the ones in Figma

### DIFF
--- a/packages/__docs__/resolve.mjs
+++ b/packages/__docs__/resolve.mjs
@@ -92,6 +92,7 @@ const alias = {
   ),
   '@instructure/ui-grid$': path.resolve(import.meta.dirname, '../ui-grid/src/'),
   '@instructure/ui-i18n$': path.resolve(import.meta.dirname, '../ui-i18n/src/'),
+  '@instructure/ui-icons-lucide': path.resolve(import.meta.dirname, '../ui-icons-lucide/src/'),
   '@instructure/ui-img$': path.resolve(import.meta.dirname, '../ui-img/src/'),
   '@instructure/ui-instructure$': path.resolve(import.meta.dirname, '../ui-instructure/src/'),
   '@instructure/ui-link$': path.resolve(import.meta.dirname, '../ui-link/src/'),
--- a/packages/emotion/src/styleUtils/calcFocusOutlineStyles.ts
+++ b/packages/emotion/src/styleUtils/calcFocusOutlineStyles.ts
@@ -78,6 +78,7 @@ const calcFocusOutlineStyles = (
       transition: 'outline-color 0.2s, outline-offset 0.25s'
     }),
     outlineOffset: '-0.8rem',
+    outlineStyle: 'solid',
     outlineColor: alpha(outlineStyle.outlineColor, 0),
     '&:focus': {
       ...outlineStyle,
--- a/packages/ui-icons-lucide/src/wrapLucideIcon/index.tsx
+++ b/packages/ui-icons-lucide/src/wrapLucideIcon/index.tsx
@@ -146,6 +146,7 @@ export function wrapLucideIcon(Icon: LucideIcon): LucideIcon {
     return (
       <span css={styles?.lucideIcon} className={className} style={style}>
         <Icon
+          name={Icon.displayName}
           ref={handleElementRef}
           size={numericSize}
           color={customColor}
--- a/packages/ui-number-input/src/NumberInput/README.md
+++ b/packages/ui-number-input/src/NumberInput/README.md
@@ -8,9 +8,7 @@ Note that this field **does not work
 uncontrolled** - you must pass event handlers if you want it to respond to
 user input.
 
-This example handles arrow buttons, up/down arrow keys, and typing into
-the input. It also includes an `onBlur` handler that displays an error message
-if the input is invalid or missing.
+This example handles arrow buttons, up/down arrow keys, and typing into the input. It also includes an `onBlur` handler that displays an error message if the input is invalid or missing.
 
 ```js
 ---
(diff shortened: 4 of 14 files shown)


packages/emotion/src/styleUtils/calcFocusOutlineStyles.ts
```diff
@@ -78,6 +78,7 @@ const calcFocusOutlineStyles = (
       transition: 'outline-color 0.2s, outline-offset 0.25s'
     }),
     outlineOffset: '-0.8rem',
+    outlineStyle: 'solid',
```
matyasf, line 81: This is a small fix for the focus outline not showing because CSS styles were reset by `FormFieldLayout`

packages/ui-number-input/src/NumberInput/index.tsx
```diff
@@ -58,207 +61,242 @@ category: components
...
+        textAlign,
+        interaction,
+        invalid,
+        success
+      }, // Type assertion needed because old generateStyle expects separate props/state
+      componentId: 'NumberInput',
+      displayName: 'NumberInput',
+      // @ts-ignore TODO fix typing
+      useTokensFrom: 'TextInput'
+    })
```
matyasf, line 153: For proper typing well need to merge #2265

packages/ui-number-input/src/NumberInput/props.ts
```diff
@@ -82,6 +83,7 @@ type NumberInputOwnProps = {
 
   /**
    * Whether or not to display the up/down arrow buttons.
+   * They are not visible when `readonly`
```
matyasf, line 85: these are visual changes requested by design

packages/ui-number-input/src/NumberInput/index.tsx
```diff
@@ -58,207 +61,242 @@ category: components
...
+            aria-hidden
+            css={styles?.arrow}
+            onMouseDown={handleClickUpArrow}
+            tabIndex={-1}
+            type="button"
+          >
+            {customIcons?.increase ? (
+              callRenderProp(customIcons.increase)
+            ) : (
+              <IconArrowOpenUpLine />
```
adamlobler: We should define the icon colors here for each state, now it stays black in every theme and state.

packages/ui-number-input/src/NumberInput/index.tsx
```diff
@@ -58,207 +61,242 @@ category: components
...
+            aria-hidden
+            css={styles?.arrow}
+            onMouseDown={handleClickDownArrow}
+            tabIndex={-1}
+            type="button"
+          >
+            {customIcons?.decrease ? (
+              callRenderProp(customIcons.decrease)
+            ) : (
+              <IconArrowOpenDownLine />
```
adamlobler: We should define the icon colors here for each state, now it stays black in every theme and state.

adamlobler (commented):
Looking good, I just left two minor comments related to the arrow icon colors. Other than that, the labels and message colors/sizes seem a bit strange, but I assume that will be handled in the formField component.

packages/ui-number-input/src/NumberInput/index.tsx
matyasf: I've fixed this, good catch

packages/ui-number-input/src/NumberInput/index.tsx
matyasf: I've fixed this, good catch

packages/ui-number-input/src/NumberInput/index.tsx
```diff
@@ -58,207 +61,273 @@ category: components
...
+    type ArrowButtonColors =
+      | 'actionSecondaryBaseColor'
+      | 'actionSecondaryHoverColor'
+      | 'actionSecondaryActiveColor'
+      | 'actionSecondaryDisabledColor'
+    const [upButtonState, setUpButtonState] = useState<ArrowButtonColors>(
+      'actionSecondaryBaseColor'
+    )
+    const [downButtonState, setDownButtonState] = useState<ArrowButtonColors>(
+      'actionSecondaryBaseColor'
```
matyasf, line 101: This is quite an ugly way to handle the buttons state, but the original code had this as `<button>` and I did not want to change it. Its needed because the new design adds hover/down states to the icons

packages/__docs__/resolve.mjs
```diff
@@ -92,6 +92,7 @@ const alias = {
   ),
   '@instructure/ui-grid$': path.resolve(import.meta.dirname, '../ui-grid/src/'),
   '@instructure/ui-i18n$': path.resolve(import.meta.dirname, '../ui-i18n/src/'),
+  '@instructure/ui-icons-lucide': path.resolve(import.meta.dirname, '../ui-icons-lucide/src/'),
```
matyasf, line 95: this fix recompiles + reloads docs when one changes something in this package

packages/ui-icons-lucide/src/wrapLucideIcon/index.tsx
```diff
@@ -146,6 +146,7 @@ export function wrapLucideIcon(Icon: LucideIcon): LucideIcon {
     return (
       <span css={styles?.lucideIcon} className={className} style={style}>
         <Icon
+          name={Icon.displayName}
```
matyasf, line 149: some unit tests target these by the display name

packages/ui-number-input/src/NumberInput/README.md
```diff
@@ -150,16 +148,43 @@ render(<Example />)
...
+    />
+    <NumberInput
+      interaction='readonly'
+      renderLabel='readonly'
+      placeholder="placeholder"
+    />
+    <NumberInput
+      renderLabel='has error'
+      placeholder="placeholder"
+      messages={[{ text: 'This is an error.', type: 'error' }]}
```
adamlobler, line 175: I’m not sure about this, but as far as I know, this error message is the legacy format that users shouldn’t use. If that’s true, then we should show the newError message here instead.

packages/ui-number-input/src/NumberInput/README.md
```diff
@@ -150,16 +148,43 @@ render(<Example />)
...
 ```js
 ---
-type: example
+  type: example
 ---
-<div>
-  <NumberInput renderLabel="Default-size input" /><br/>
-  <NumberInput size="large" renderLabel="Large-size input" />
-</div>
+  <Flex gap='2rem' direction='column'>
```
adamlobler: The "gap" doesn't apply the raw value here for some reason, we should use `gap='medium'` instead.

packages/ui-number-input/src/NumberInput/README.md
matyasf: This is part of FormField and not this component, we will fix it there. (and AFAIK `type: 'error` will be the new error, and we'lll remove the old type error)

packages/ui-number-input/src/NumberInput/README.md
matyasf: good catch, fixed it

PR #2273, opened 2025-11-28
feat(many): rework TextArea and dependent FormField components

ToMESSKa:
INSTUI-4812

**ISSUE:**

this PR:
- updates the texArea tokens
- collapses 'newError' and 'error' type in FormField and both types have functionality of 'newError'
- migrates the asterisk logic from components using FormField into FormFieldLayout
- updates FormField tokens
- disables 'success' and 'success' messages when the component using FormField is 'disabled' or 'readOnly'

**TEST PLAN:**

- check the figma: https://www.figma.com/design/5WFqF3HVlNdfcJooCaiFoh/InstUI-Component-Library?node-id=3437-9532&p=f&m=dev
- test the TexArea focus outline, it should have no errors even when the TextAre is getting resized
- open the example on Form Errors page and replace the type: 'error' with type: 'newError' (see the code below) and test if the 'error' and the 'newError' type has no visual differences when it comes to form errors
```
const messages = showError
    ? [{type: 'error', text: showLongError ? 'Long error....
```
- on Form Errors page the asterisk should appear when the isRequired is true
- open the FormField and FormFieldGroup pages, and test the examples, the layout should be the same as on the latest version (except the success state having an icon and the new error style should be applied instead of the old one)
- check TextInput, NumberInput, RadioInputGroup and TextArea, 'error' and 'success' messages should not appear when the component is 'disabled' or 'readOnly', but 'hint' type should
- check TextArea, the border's color should not be green/red, when it is in a success/error state and 'disabled' or 'readOnly' at the same time. In case of readOnly, the focus outline should not be green/red either


packages/ui-form-field/src/FormFieldLayout/index.tsx
```diff
@@ -22,188 +22,202 @@
...
+      const labelContent = hasVisibleLabel ? (
+        <>
+          {label}
+          {isRequired && (
+            <span
+              css={invalid ? styles?.requiredAsterisk : {}}
+              aria-hidden={true}
+            >
+              {' '}
+              *
```
ToMESSKa, line 168: The logic rendering the asterisk has been moved to FormFieldLayout for the other components to here.

packages/ui-form-field/src/FormFieldMessage/index.tsx
```diff
@@ -51,50 +51,54 @@ type: example
...
-        {this.props.variant === 'newError' && this.props.children && (
-          <span css={styles?.errorIcon}>
-            <IconWarningSolid color="error" />
+        {shouldShowIcon && (
+          <span css={styles?.icon}>
+            {isErrorVariant ? (
+              <AlertCircleInstUIIcon size="sm" color="errorColor" />
+            ) : (
+              <CheckCircle2InstUIIcon size="sm" color="successColor" />
+            )}
```
ToMESSKa, line 86: now in the case of success and error (and newError), an icon is rendered

packages/ui-radio-input/src/RadioInputGroup/styles.ts
```diff
@@ -1,46 +0,0 @@
...
- * Generates the style object from the theme and provided additional information
- * @param  {Object} componentTheme The theme variable object.
- * @return {Object} The final style object, which will be used in the component
- */
-const generateStyle = (componentTheme: RadioInputGroupTheme): RadioInputGroupStyle => {
-  const { invalidAsteriskColor } = componentTheme
-
-  return {
-    invalidAsterisk: {
-      color: invalidAsteriskColor
```
ToMESSKa, line 41: as the only theme variable invalidAsteriskColor has been removed, I removed the styles.ts too

packages/ui-text-area/src/TextArea/index.tsx
```diff
@@ -105,304 +156,276 @@ class TextArea extends Component<TextAreaProps> {
...
-          {textarea}
-          {!disabled && !readOnly ? (
-            <span
-              css={this.props.styles?.textAreaOutline}
-              aria-hidden="true"
-              ref={(e) => {
-                this._highlightRef = e
-              }}
-            />
-          ) : null}
```
ToMESSKa, line 400: As applying the focus outline was delegated to calcFocusOutlineStyles even when the TextArea is resized, this span was no loner needed was removed

packages/ui-text-area/src/TextArea/index.tsx
```diff
@@ -105,304 +156,276 @@ class TextArea extends Component<TextAreaProps> {
...
-            }
-          }, 500)
-        }
-      }
-    })
+  }, [])
 
-    this.autoGrow()
-    this.props.makeStyles?.()
-  }
```
ToMESSKa, line 131: this logic was deleted as calcFocusOutlineStyles is used for focus outline even when the TextArea is getting resized

packages/ui-text-area/src/TextArea/styles.ts
```diff
@@ -72,54 +80,61 @@ const generateStyle = (
...
-      top: '-0.25rem',
-      bottom: '-0.25rem',
-      left: '-0.25rem',
-      right: '-0.25rem',
-      border: `${componentTheme.focusOutlineWidth} ${componentTheme.focusOutlineStyle} ${componentTheme.focusOutlineColor}`,
-      borderRadius: `calc(${componentTheme.borderRadius} * 1.5)`,
-      transition: 'all 0.2s',
-      // properties to transition on :focus
-      opacity: 0,
-      transform: 'scale(0.95)'
```
ToMESSKa, line 122: This is also not needed anymore as calcFocusOutlineStyles is used

packages/ui-form-field/src/FormFieldLayout/styles.ts
```diff
@@ -65,19 +76,19 @@ const generateGridLayout = (
...
+ * @param componentTheme The theme variable object.
+ * @param params Additional parameters to customize the style.
+ * @param sharedTokens Shared token object that stores common values for the theme.
+ * @return The final style object, which will be used in the component
  */
 const generateStyle = (
-  componentTheme: FormFieldLayoutTheme,
-  props: FormFieldLayoutProps,
-  styleProps: FormFieldStyleProps
+  componentTheme: NewComponentTypes['FormFieldLayout'],
```
matyasf, line 85: You are not using `componentTheme.readOnlyTextColor`, is this intentional?

packages/ui-form-field/src/FormFieldMessage/index.tsx
```diff
@@ -22,18 +22,18 @@
  * SOFTWARE.
```
matyasf, line 22: You forgot to delete `theme.ts` from this directory

packages/ui-text-area/src/TextArea/index.tsx
```diff
@@ -105,304 +156,276 @@ class TextArea extends Component<TextAreaProps> {
...
+        if (typeof textareaRef === 'function') {
+          textareaRef(textarea)
+        }
+      }}
+      style={style}
+      id={id}
+      required={required}
+      aria-required={required}
+      aria-invalid={isInvalid ? 'true' : undefined}
+      disabled={disabled || readOnly}
```
matyasf: This is a weird pattern that is here with us likely for legacy reasons (its from this commit https://github.com/instructure/instructure-ui/commit/f494d0d6cc0aca0bc97bfe802f8b7ee44a7d350a#diff-7ece62f28a2d18273bf2d2929da636add19dd315525e0f48d50e0dc1a64c26b2 ). We should split this into 2 like
```
disabled={disabled}
readOnly={readOnly}
```
(I've done this for `NumberInput`)
but I dont know the intention why its like this. We should ask this during a11y review

packages/ui-text-area/src/TextArea/styles.ts
```diff
@@ -22,41 +22,49 @@
...
  * @param  {Object} componentTheme The theme variable object.
- * @param  {Object} props the props of the component, the style is applied to
- * @param  {Object} state the state of the component, the style is applied to
+ * @param  {Object} params Additional parameters to customize the style.
  * @return {Object} The final style object, which will be used in the component
  */
 const generateStyle = (
-  componentTheme: TextAreaTheme,
-  props: TextAreaProps
+  componentTheme: NewComponentTypes['TextArea'],
```
matyasf, line 47: `gapContent` and `gapPrimitives` are not used, is this intentional?

packages/ui-text-area/src/TextArea/styles.ts
```diff
@@ -72,54 +80,61 @@ const generateStyle = (
...
+        color: componentTheme.textDisabledColor,
+        background: componentTheme.backgroundDisabledColor,
+        borderColor: componentTheme.borderDisabledColor,
+        cursor: 'not-allowed',
+        pointerEvents: 'none',
+        opacity: 0.5
+      }),
+      ...(success && {
+        borderColor: componentTheme.successBorderColor
+      }),
```
matyasf: For some reason I dont see the border turning green when I have a success message, e.g.
```
<TextArea  messages={[{ text: 'some message', type: 'success' }]} label="Description" />
```

Also note that it should not turn green in the disabled, readonly states

packages/ui-text-area/src/TextArea/styles.ts
```diff
@@ -72,54 +80,61 @@ const generateStyle = (
...
-        transform: 'scale(1)',
-        opacity: 1
-      },
-      '&[aria-invalid], &[aria-invalid]:focus, &[aria-invalid]:focus + [class$=-textArea__outline]': {
+      ...calcFocusOutlineStyles(sharedTokens.focusOutline, {
+        focusColor
+      }),
+      '&[aria-invalid], &[aria-invalid]:focus': {
         borderColor: componentTheme.errorBorderColor
       },
```
matyasf: the border should be grey when disabled/readonly. I suggest to use here the `disabled`/`readOnly` props, that was you can make `if..else if..` switches easier (check `RadioInput`'s `style.ts` for an example

packages/ui-form-field/src/FormFieldLayout/styles.ts
ToMESSKa: @matyasf yes, this token was supposed to be the color of the 'read-only' helper text, which will not be implemented
image

packages/ui-form-field/src/FormFieldMessage/index.tsx
ToMESSKa: @matyasf thanks, I deleted that

packages/ui-text-area/src/TextArea/index.tsx
ToMESSKa: @matyasf I fixed that

packages/ui-text-area/src/TextArea/styles.ts
ToMESSKa: @matyasf I talked to Ádám, these will be deleted

packages/ui-text-area/src/TextArea/styles.ts
ToMESSKa: @matyasf In the canvas theme, the border does not turn green in case of success (see the Figma), only in the new theme. Maybe that was the reason?

packages/ui-text-area/src/TextArea/styles.ts
ToMESSKa: @matyasf I talked to Ádám about the success/error and disabled/readOnly combinations today. He said that these two variants should not exist if the component is disabled/readOnly, so I changed the logic in FormFieldLayout and passed the disabled/readOnly prop in TextInput, NumberInput, RadioInputGroup and TexArea to FormField. So there should be no error or success messages/icons in these components if they are disabled/readOnly (see the last two items of the test plan). This is kind of a breaking change so added it to the upgrade guide

packages/ui-radio-input/src/RadioInputGroup/index.tsx
```diff
@@ -53,7 +47,6 @@ category: components
 ---
 **/
 @withDeterministicId()
-@withStyle(generateStyle, generateComponentTheme)
 class RadioInputGroup extends Component<
```
joyenjoyer, line 50: Why didn't you change this to functional component?

packages/ui-text-input/src/TextInput/index.tsx
```diff
@@ -352,23 +351,7 @@ class TextInput extends Component<TextInputProps, TextInputState> {
...
-          >
-            {' '}
-            *
-          </span>
-        )}
-      </Fragment>
-    ) : (
-      rawLabel
-    )
+    const label = callRenderProp(renderLabel)
```
joyenjoyer, line 354: You may consider using `React.createElement` here instead of `callRenderProp`. We would like to deprecate it soon.

packages/ui-number-input/src/NumberInput/index.tsx
```diff
@@ -318,20 +316,7 @@ const NumberInput = forwardRef<NumberInputHandle, NumberInputProps>(
...
-          <span css={invalid ? styles?.requiredInvalid : {}} aria-hidden={true}>
-            {' '}
-            *
-          </span>
-        )}
-      </Fragment>
-    ) : (
-      rawLabel
-    )
+    const label = callRenderProp(renderLabel)
```
joyenjoyer, line 319: You may consider `React.createElement` here instead of `callRenderProp`. We would like to deprecate it soon.

packages/ui-text-input/src/TextInput/index.tsx
matyasf: hmm we should check that this is not a breaking change, e.g. make 100% sure that `React.createElement` can handle everything that `callRenderProp` can (e.g. just a string, a JSX, a React class component,...)

ToMESSKa: @joyenjoyer I talked to @matyasf about this and he suggested to leave it as it is for now.

packages/ui-number-input/src/NumberInput/index.tsx
ToMESSKa: @joyenjoyer see above

packages/ui-radio-input/src/RadioInputGroup/index.tsx
ToMESSKa: @joyenjoyer I only changed those components to functional which needed their tokens rewired as a part of this PR and needed the useStyle hook to make it work. But as we agreed, the conversion to functional components is not a priority anymore.

packages/ui-text-area/src/TextArea/styles.ts
matyasf: readonly still shows messages
image

matyasf (changes requested):
nice work, I just found 1 issue

packages/ui-text-area/src/TextArea/styles.ts
ToMESSKa: @matyasf the prop is called readOnly not readonly

matyasf: oh I see. So we are ONLY showing info messages for these, not `error` and `success`

PR #2380, opened 2026-01-22
[v12] feat(ui-toggle-details): rework ToggleDetails

ToMESSKa:
INSTUI-4815

**ISSUE:**

- ToggleDetails needs to be migrated to the new theming system

**TEST PLAN:**
- compare the component to the Figma design https://www.figma.com/design/5WFqF3HVlNdfcJooCaiFoh/InstUI-Component-Library---Beta?node-id=9871-951&p=f&m=dev


adamlobler (changes requested):
I’m not sure why, but when I set the line-height of the toggleDetails content to 150% in the inspector, it looks good. However, when it gets the same value from the toggleDetails style, it appears much smaller. It looks like the 150% is not connected to the element's font-size its 150% of some general font-size.

150% line-height set by the component style:
Image

150% line-height set in the inspector:
Image

git-nandor (changes requested):
See my comments

adamlobler (changes requested):
The line-height looks great, thanks for the modifications! 🙌 Other than the icon misalignment that Nandi mentioned, I think it’s going to be okay, the other issues are related to the button component, which we should fix there.

ToMESSKa:
> image
> The filled variant has an incorrect background color, and the icons used for the expanded and closed states don’t match the designs in Figma.

@git-nandor Both of those issues will be handled in the Button/Iconbutton reworks.

git-nandor (approved):

adamlobler (approved):

joyenjoyer (approved):

PR #2391, opened 2026-02-03
[v12] feat(ui-popover,ui-options,ui-menu): rework Menu and Options

ToMESSKa:
INSTUI-4796

**ISSUE:**

Menu and Options needs to be migrated to the new theming system

TEST PLAN:

- compare the component to the Figma design [Menu](https://www.figma.com/design/5WFqF3HVlNdfcJooCaiFoh/InstUI-Component-Library---Beta?node-id=15052-3130&p=f&m=dev) and [Options](https://www.figma.com/design/5WFqF3HVlNdfcJooCaiFoh/InstUI-Component-Library---Beta?node-id=7854-2&p=f&m=dev)
- check if the icons appear in the right color when the Menu items are hovered or when keyboard navigation is used.


packages/ui-options/src/Options/Item/styles.ts
```diff
@@ -61,7 +61,7 @@ const generateStyle = (
     },
     selected: {
       background: componentTheme.selectedBackground,
-      color: componentTheme.highlightedLabelColor
+      color: componentTheme.selectedLabelColor
```
ToMESSKa, line 64: I changed this @hajnaldo as selectedLabelColor was not used before

packages/ui-popover/src/Popover/index.tsx
```diff
@@ -601,8 +601,6 @@ class Popover extends Component<PopoverProps, PopoverState> {
             {...viewProps}
             borderColor={styles?.borderColor}
             borderRadius={styles?.borderRadius}
-            overflowX="hidden"
-            overflowY="hidden"
```
ToMESSKa, line 605: This was chaged by @adamlobler but it clipped the focus ring of the menu so I reverted this

packages/ui-options/src/Options/Separator/index.tsx
```diff
@@ -26,10 +26,9 @@ import { Component } from 'react'
 
 import { getElementType, omitProps } from '@instructure/ui-react-utils'
 
-import { withStyleRework as withStyle } from '@instructure/emotion'
+import { withStyle } from '@instructure/emotion'
 
 import generateStyles from './styles'
-import generateComponentTheme from './theme'
```
ToMESSKa, line 32: I did not delete the theme.ts files of Options yet because Drilldown uses them. I will remove them in the Drilldown rework PR.

matyasf (approved):
looks good, it looks like the design to me and all the tokens are used. FYI design might make some changes about how a selected item looks

hajnaldo (commented):
Menu looks and works as expected from design pov.

Some comments regarding the Options:
- I think the "highlighted disabled" combo is not valid, because highlighted here means hover and a disabled item cannot be hovered
- odd behavior observed on the showcase: when clicking on an Option, it stays with the highlighted color, and after leaving the option, the background changes to the selected color

HerrTopi (approved):

PR #2392, opened 2026-02-03
[v12] feat(ui-checkbox): Checkbox rework - old

git-nandor:
INSTUI-4788

## Summary
Migrated Checkbox component from the old theming system.

## Test plan
On the documentation page, verify that everything displays and works correctly.

Figma: 
- [simple](https://www.figma.com/design/5WFqF3HVlNdfcJooCaiFoh/InstUI-Component-Library---Beta?node-id=5248-5114&p=f&m=dev) variant
- [toggle](https://www.figma.com/design/5WFqF3HVlNdfcJooCaiFoh/InstUI-Component-Library---Beta?node-id=7283-6852&m=dev) (switch) variant

##  


packages/ui-checkbox/src/Checkbox/index.tsx
adamlobler, line 288: I see that these styles contain the left padding, but on the site unfortunately i can't see it:/

Image

adamlobler (changes requested):

packages/ui-checkbox/src/Checkbox/index.tsx
git-nandor: Based on our discussion, I’ve added a custom calculation to determine the correct padding values.

packages/ui-checkbox/src/Checkbox/CheckboxFacade/index.tsx
```diff
@@ -68,14 +72,24 @@ class CheckboxFacade extends Component<CheckboxFacadeProps> {
...
+      }
+      return 'inverseColor'
+    }
+
+    const iconColor = getIconColor()
+
+    if (indeterminate) {
+      return renderIconWithProps(MinusInstUIIcon, 'sm', iconColor)
+    } else if (checked) {
+      return renderIconWithProps(CheckInstUIIcon, 'sm', iconColor)
```
ToMESSKa, line 92: Image
In this the check icon size does not respect the size prop, now it's medium for all sizes

ToMESSKa, line 92: Image
Here the icon is kinda hard to see? It it because checked disabled state is an invalid design state? I cannot see something like that in the Figma. Same for the inderminate + disabled state

packages/ui-checkbox/src/Checkbox/CheckboxFacade/index.tsx
```diff
@@ -68,14 +72,24 @@ class CheckboxFacade extends Component<CheckboxFacadeProps> {
...
   renderIcon() {
-    if (this.props.indeterminate) {
-      return (
-        <SVGIcon viewBox="0 0 1920 1920" inline={false}>
-          <rect x="140" y="820" width="1640" height="280" />
-        </SVGIcon>
-      )
-    } else if (this.props.checked) {
-      return <IconCheckMarkSolid inline={false} />
+    const { disabled, readOnly, indeterminate, checked } = this.props
```
ToMESSKa, line 75: Image

Here why is the border circle congtaing the x icon is red? It is not red on v11.

packages/ui-checkbox/src/Checkbox/ToggleFacade/styles.ts
```diff
@@ -32,19 +33,21 @@ import type { ToggleFacadeProps, ToggleFacadeStyle } from './props'
...
  * @param  {Object} componentTheme The theme variable object.
  * @param  {Object} props the props of the component, the style is applied to
+ * @param  {Object} sharedTokens Shared theme token object
  * @param  {Object} state the state of the component, the style is applied to
  * @return {Object} The final style object, which will be used in the component
  */
 const generateStyle = (
-  componentTheme: ToggleFacadeTheme,
-  props: ToggleFacadeProps
+  componentTheme: NewComponentTypes['Toggle'],
```
ToMESSKa, line 41: In Toggle, these tokens are not used  
1. backgroundColor
2. backgroundHoverColor
3. padding
4. toggleWidth

ToMESSKa (changes requested):

packages/ui-checkbox/src/Checkbox/CheckboxFacade/index.tsx
ToMESSKa: Okay sorry, now I found it in the figma and it looks good

packages/ui-checkbox/src/Checkbox/CheckboxFacade/index.tsx
matyasf: FYI the same is true for the `toggle` variant

git-nandor: I double-checked, and the request from the design side is that the icons should all be shown in the same size.

git-nandor: I checked with design, and accessibility-wise there’s no strict contrast requirement for the disabled state, so they think this is fine as it is.

packages/ui-checkbox/src/Checkbox/ToggleFacade/styles.ts
git-nandor: I’m still waiting for Ádám’s feedback. The padding token will probably be removed, but the others have already been used.

packages/ui-checkbox/src/Checkbox/styles.ts
```diff
@@ -32,24 +32,49 @@ import type { ComponentTheme } from '@instructure/shared-types'
...
-      color: componentTheme.requiredInvalidColor
+      color: componentTheme.asteriskColor
     },
     indentedError: {
-      paddingLeft: componentTheme.checkErrorInsetWidth
+      paddingLeft: sizeVariants[size!].paddingLeft
     },
     indentedToggleError: {
-      paddingLeft: componentTheme.toggleErrorInsetWidth,
+      paddingLeft: sizeVariants[size!].paddingLeft
```
matyasf, line 77: Now this code causes a crash when the `size` is an empty string or a non-value value. Can you make it default to `medium` in these cases? Usually we dont crash, rather just fall back to the default when users enter an invalid value

matyasf (changes requested):
The toggle variant's checked background is a bit too long (there is a green area right to the checkmark, it does not look like this in the Figma:

Image

Also in the Figma I dont see the checked + error state, can you check with design that what you've made is how it should look like? I think there are valid use cases for something to be in this state.

packages/ui-checkbox/src/Checkbox/styles.ts
git-nandor: Thanks! Fixed in the new PR

git-nandor:
> The toggle variant's checked background is a bit too long (there is a green area right to the checkmark, it does not look like this in the Figma:
> 
> Image
> Also in the Figma I dont see the checked + error state, can you check with design that what you've made is how it should look like? I think there are valid use cases for something to be in this state.

Thanks! Fixed in the new PR, and I talked with Ádám about the checked + error state issue. From the design side, it's currently not handled and they don’t plan to create a separate design for it for now. They asked me to remove the example related to the message from the docs page. They will revisit it in the future.

PR #2401, opened 2026-02-11
[v12] feat(ui-drawer-layout): rework DrawerLayout

ToMESSKa:
INSTUI-4914

ISSUE:

- DrawerLayout needs to be migrated to the new theming system [DrawerLayout](https://www.figma.com/design/5WFqF3HVlNdfcJooCaiFoh/InstUI-Component-Library---Beta?node-id=13549-1766&p=f&m=dev)

TEST PLAN:

- compare the component to the Figma design


matyasf (changes requested):
Code looks good, but can you please fix the examples when using the dark theme? The sample content should be also light text on dark background

ToMESSKa:
> Code looks good, but can you please fix the examples when using the dark theme? The sample content should be also light text on dark background

Thanks, I updated the examples.

matyasf (approved):
Nice work

balzss (approved):

hajnaldo (commented):
The close button is still the old one in rebrand as well - this is because buttons haven't been merged yet

PR #2402, opened 2026-02-11
Inst UI 4797 modal rework - old

git-nandor:


packages/ui-modal/src/Modal/README.md
```diff
@@ -575,11 +571,11 @@ class Example extends React.Component {
         >
```
ToMESSKa, line 571: Looks good!
One tiny thing, in the new `rebrand-light` theme, the modal body's bottom corners look curved too. I think this was an issue in v11 too, but less visible because the border radius is much smaller in the canvas theme.
Image

ToMESSKa (approved):

ToMESSKa (approved):

ToMESSKa (changes requested):
see my comment

adamlobler (changes requested):
The content has the same border radius at the bottom as the modal itself, but it shouldn't have any border radius, it's an issue on the original component as well but now its more obvious.
Image

packages/ui-modal/src/Modal/README.md
adamlobler: Sorry I didn't see this one before I wrote my review comment 😅

joyenjoyer (changes requested):
Please fix what Tamas commented on, otherwise it looks great to me, don't forget to do a rebase first and squash commits

packages/ui-modal/src/Modal/README.md
git-nandor: Thanks! I fixed it in the new PR.

PR #2420, opened 2026-02-23
[v12] feat(ui-truncate-text): TruncateText rework old

git-nandor:
INSTUI-4819

## Summary
Migrated TruncateText component from the old theming system.

## Test plan
On the documentation page, verify that everything displays and works correctly.


ToMESSKa (approved):

adamlobler (changes requested):
We shouldn't use plain spans in the examples because they don't get the proper colors when we switch themes, we should use the Text component instead.

Image

Image

PR #2465, opened 2026-03-17
feat(ui-options,ui-select,ui-simple-select,ui-time-select): rework Select, SimpleSelect and TimeSelect

ToMESSKa:
INSTUI-4807

**ISSUE:**

- Select, SimpleSelect, TimeSelect needs to be migrated to the new theming system

**TEST PLAN:**

for Select, SimpleSelect, TimeSelect:
- check if all necessary the imports are renamed to latest in v2
- check if the right imports are used in `exports/b.ts` and `exports/a.ts`
- check if the `package.json` uses the correct versioning
- check if the renamed or removed tokes are documented in the upgrade guide
- check if the index.ts files use withStyle instead of withStyleLegacy


packages/ui-options/src/Options/v2/Item/styles.ts
```diff
@@ -71,7 +71,7 @@ const generateStyle = (
     },
     'selected-highlighted': {
       background: componentTheme.selectedHighlightedBackground,
-      color: componentTheme.highlightedLabelColor
+      color: componentTheme.selectedLabelColor
```
ToMESSKa, line 74: This worked in the two old canvas themes, but in the new light theme, the label would be black on a dark background (bad visibility) and in the new dark theme it would be white on a white background (bad visibilty). So I changed this. You can check this on the first example of in Select.

balzss (changes requested):
2 smaller things that i've noticed:
- with the new `light` theme, the select looks a bit too dark, i've already pinged @hajnaldo about this, she will look into this
- the SimpleSelect `Controlled` example features a scrollbar and the corners doesn't follow the border radius correctly. this might be the same issue as #2473 or very similar. can you please rebase once it's merged to see if that solves it and if it isn't, try to maybe adopt that solution for the select?

packages/ui-time-select/src/TimeSelect/v2/props.ts
```diff
@@ -0,0 +1,327 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+import { InputHTMLAttributes } from 'react'
+
+import type { FormMessage } from '@instructure/ui-form-field/v11_6'
```
matyasf: I think this needs to be `/latest`

packages/ui-simple-select/src/SimpleSelect/v2/props.ts
```diff
@@ -0,0 +1,300 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+import { InputHTMLAttributes } from 'react'
+
+import type { FormMessage } from '@instructure/ui-form-field/v11_6'
```
matyasf: I think this needs to be `/latest`

packages/ui-simple-select/src/SimpleSelect/v2/props.ts
```diff
@@ -0,0 +1,300 @@
...
+import type {
+  OtherHTMLAttributes,
+  PickPropsWithExceptions
+} from '@instructure/shared-types'
+import type {
+  PlacementPropValues,
+  PositionConstraint,
+  PositionMountNode
+} from '@instructure/ui-position'
+import type { SelectOwnProps } from '@instructure/ui-select/v11_6'
```
matyasf: I think this needs to be /latest

matyasf (changes requested):
Just a couple of import issues

packages/ui-time-select/src/TimeSelect/v2/props.ts
ToMESSKa: Changed

packages/ui-simple-select/src/SimpleSelect/v2/props.ts
ToMESSKa: Changed

packages/ui-simple-select/src/SimpleSelect/v2/props.ts
ToMESSKa: Changed

matyasf (changes requested):
> I've also noticed that group select doesn't have the correct border radius in light theme
regarding the gray color: this is a bug from much earlier in the code but it only got "visible" with the new themes. basically the issue is that the select component displays a TextInput in readonly (regardless it's only interaction prop) mode and the new light theme has a dark grey bg for that

good catch, we should do something about it. As I see it being readonly causes some issues, see INSTUI-4500 and https://github.com/instructure/instructure-ui/pull/1935 As I see it would be quite a large task to fix this properly (=do not use an `input` if its read-only` @ToMESSKa  Can you please ask design about the preferred color? Then I think this should be fixable with a theme override inside the component (or some other trick)

ToMESSKa:
> @ToMESSKa
> 
> * I've also noticed that group select doesn't have the correct border radius in `light` theme
> * regarding the gray color: this is a bug from much earlier in the code but it only got "visible" with the new themes. basically the issue is that the select component displays a TextInput in readonly (regardless it's only `interaction` prop) mode and the new light theme has a dark grey bg for that

@matyasf I put a themeOverride on the TextInput in Select (also fixed the TypeScript error in TextInput similar to this PR: https://github.com/instructure/instructure-ui/pull/2487), so now it should look good. Test if the interaction="readonly" prop is explicitly set — the color should still be grey.

packages/ui-color-picker/src/ColorPicker/v2/index.tsx
```diff
@@ -710,7 +710,11 @@ class ColorPicker extends Component<ColorPickerProps, ColorPickerState> {
           display="inline-block"
           width={width}
           placeholder={placeholderText}
-          themeOverride={{ padding: '' }}
+          themeOverride={{
+            paddingHorizontalSm: '',
+            paddingHorizontalMd: '',
```
ToMESSKa, line 715: Paddig token was removed from TextInput so changing TextInputTheme to NewComponentTypes['TextInput'] in TextInput's props.ts resulted an error here so I fixed this.

packages/ui-text-input/src/TextInput/v2/styles.ts
```diff
@@ -205,7 +205,12 @@ const generateStyle = (
       justifyContent: 'flex-start',
       flexDirection: 'row'
     },
+    beforeElement: {
+      ...(interaction === 'disabled' && { opacity: 0.5 }),
+      label: 'textInput__beforeElement'
+    },
     afterElement: {
+      ...(interaction === 'disabled' && { opacity: 0.5 }),
```
ToMESSKa, line 213: TextInput v1 disabled state worked by setting the opacity of whole input. After the v2 rework the icons put in renderBeforeInput and renderAfterInput stayed the same color in the disabled state, so I fixed this. You can test this in the second example of TextInput.

matyasf (approved):
looks good, nice work!

hajnaldo (changes requested):
Selected indicator | Options, Select, Simple Select:

Selected items should have a check icon on the right, please see an example here: 
https://www.figma.com/design/5WFqF3HVlNdfcJooCaiFoh/InstUI-Component-Library---Beta?m=auto&node-id=22455-2368&t=c1KrRz9NMggj3y4N-1

Minor docu page issue for Options:
In the first 2 examples the actionable options don't get highlighted on hover

packages/ui-select/src/Select/v2/props.ts
```diff
@@ -0,0 +1,335 @@
...
+ * SOFTWARE.
+ */
+
+import { InputHTMLAttributes } from 'react'
+
+import type {
+  OtherHTMLAttributes,
+  SelectTheme
+} from '@instructure/shared-types'
+import type { FormMessage } from '@instructure/ui-form-field'
```
balzss: shouldn't this be `@instructure/ui-form-field/latest` ?

packages/ui-select/src/Select/v2/props.ts
```diff
@@ -0,0 +1,335 @@
...
+   */
+  mountNode?: PositionMountNode
+}
+
+type PropKeys = keyof SelectOwnProps
+
+type AllowedPropKeys = Readonly<Array<PropKeys>>
+
+type SelectProps = SelectOwnProps &
+  WithStyleProps<SelectTheme, SelectStyle> &
```
balzss: q: should we use `NewComponentTypes['Select']` here?

balzss (approved):
2 minor things, other than those it looks ok

packages/ui-select/src/Select/v2/props.ts
ToMESSKa: Done

packages/ui-select/src/Select/v2/props.ts
ToMESSKa: Done

balzss (approved):

ToMESSKa:
> Selected indicator | Options, Select, Simple Select:
> 
> Selected items should have a check icon on the right, please see an example here: https://www.figma.com/design/5WFqF3HVlNdfcJooCaiFoh/InstUI-Component-Library---Beta?m=auto&node-id=22455-2368&t=c1KrRz9NMggj3y4N-1
> 
> Minor docu page issue for Options: In the first 2 examples the actionable options don't get highlighted on hover

@hajnaldo I talked to the team about the check icon, they advised against implementing it now, but we should get back to this question for a future release.
The document issue is actually on purpose, those examples are meant to represent that Options does not manage any state or react to any user interaction in itself.

PR #2473, opened 2026-03-18
fix(many): fix checkbox icons and fix border bugs in menu and in dateInput

git-nandor:
Use border-radius: inherit on Dialog, Menu, and Calendar so rounded
corners from ContextView flow through the component tree without
requiring overflow: clip (which was clipping CSS focus outlines).
Also broaden ToggleFacade icon selector from lucideIcon to icon.

Test:
check DateInput popup borders
check Menu borders
check Checkbox toggle variant icons

INSTUI-4961

### DIFF
--- a/packages/ui-dialog/src/Dialog/index.tsx
+++ b/packages/ui-dialog/src/Dialog/index.tsx
@@ -168,7 +168,8 @@ class Dialog extends Component<DialogProps> {
         {...omitProps(this.props, Dialog.allowedProps)}
         role={role}
         aria-label={this.props.label}
-        className={this.props.className}
+        className={this.props.className} // TODO in V2 remove className, there is no usage of it.
+        style={{ borderRadius: 'inherit' }} // ensure the dialog inherits border radius from View
         ref={this.getRef}
       >
         {this.props.children}
--- a/packages/ui-menu/src/Menu/v2/styles.ts
+++ b/packages/ui-menu/src/Menu/v2/styles.ts
@@ -55,7 +55,7 @@ const generateStyle = (
       padding: '0.25rem 0',
       // TODO-rework
       //background: componentTheme.background,
-      //borderRadius: componentTheme.borderRadius,
+      borderRadius: 'inherit',
       display: 'block',
       position: 'relative',
       overflow: 'hidden',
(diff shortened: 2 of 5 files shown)


packages/ui-dialog/src/Dialog/index.tsx
```diff
@@ -169,6 +169,7 @@ class Dialog extends Component<DialogProps> {
         role={role}
         aria-label={this.props.label}
         className={this.props.className}
```
matyasf: bit off, but I was really wondering why we allow `className` here. I think its a bug, since its there since the component was created, lets remove it https://github.com/instructure/instructure-ui/commit/c88cf8ec41c89f8f97f39b339fbc4e0c4ba5b793#diff-3ca2f171e2d8a67947b8794d172024b87fb9de4fc3f0ead84fb15bd31fce56ccR212

matyasf (commented):
Looks good, I could not find any visual issues

packages/ui-dialog/src/Dialog/index.tsx
balzss: might be because it's a util component, not a regular one. maybe we use that prop internally? i'd double check before removing it

packages/ui-menu/src/Menu/v2/styles.ts
```diff
@@ -55,7 +55,7 @@ const generateStyle = (
       padding: '0.25rem 0',
       // TODO-rework
       //background: componentTheme.background,
```
balzss, line 57: unrelated to the pr but i'm not sure why this is commented out

balzss (approved):

packages/ui-menu/src/Menu/v2/styles.ts
git-nandor: i'll double check it with designer

packages/ui-dialog/src/Dialog/index.tsx
balzss: i'd say let's add a todo comment and leave it as is, something like `TODO: investigate if className is actually needed here or can be removed`

packages/ui-menu/src/Menu/v2/styles.ts
balzss: if there is no straightworfard answer, let's leave it for now since it's out of the scope of this task anyway

packages/ui-dialog/src/Dialog/index.tsx
matyasf: I could not find any usges of it, but someone else might be using it. Let's add a `TODO` for now, and remove it in v2

matyasf (approved):
Looks good, please add a TODO to the strange `className` prop

PR #2487, opened 2026-03-24
Better types for themes, small fixes

matyasf:
- Fix types for the `themeOverride` prop. Until now it was using theme types from the old engine.
- Delete `TextInput/v2/theme.ts`, it was unused

To test:
Check if theme overrides are typed correctly for v1 and v2 components, e.g.

```
<Button themeOverride={{aiActiveTextColor: 'red'}}>aa</Button> // v2 Button, should show no TS error
const a = <Button themeOverride={{aiActiveBoxShadow: 'red'}}>asd</Button>  // v1 Button, should show no TS error

// v2 Avatars
const a = <Avatar name='a'
                  themeOverride={{ashBackgroundColor: 'red'}} />
const b = <Avatar name='a'
                  themeOverride={(_compTheme, theme) => {
                    return {ashTextColor: theme.newTheme.primitives.colors.green.green10}
                  }} />

```


packages/ui-select/src/Select/v2/index.tsx
```diff
@@ -811,11 +811,10 @@ class Select extends Component<SelectProps> {
         suppressHydrationWarning
         {...(interaction === 'enabled' &&
           !isEditable && {
-          themeOverride: (componentTheme) => ({
-            //@ts-expect-error TODO-theme-types
```
matyasf, line 815: This is the only change here, dont know why this file was reformatted

HerrTopi (approved):

ToMESSKa (changes requested):
In the props.ts for Pagination and Table (and subcomponents) I see `WithStyleProps<null,` instead of `WithStyleProps<ReturnType<NewComponentTypes['XXX']`. Is that supposed to be like that?

Also for these components, the the themeOverride form ( e.g. <Avatar themeOverride={{borderColor: 'red'}}>aa</Avatar>) gives some TS error. I see that these use `themeOverride?: ThemeOverrideValue` where other components use `WithStyleProps<ReturnType<NewComponentTypes['XXX']`
  - AiInformation                                                           
  - DataPermissionLevels                                                           
  - NutritionFacts                                    
  - Avatar                                                                       
  - FormFieldLayout                                                                         
  - NumberInput                                                                    
  - RadioInput                                                                     
  - Spinner                                                                        
  - TextArea

matyasf:
> In the props.ts for Pagination and Table (and subcomponents) I see `WithStyleProps<null,` instead of `WithStyleProps<ReturnType<NewComponentTypes['XXX']`. Is that supposed to be like that?

Only if the component has no theme (= `generateStyle` has no `componentTheme` parameter). I've fixed it where it was needed, good catch

> Also for these components, the the themeOverride form ( e.g. <Avatar themeOverride={{borderColor: 'red'}}>aa) gives some TS error. I see that these use `themeOverride?: ThemeOverrideValue` where other components use `WithStyleProps<ReturnType<NewComponentTypes['XXX']`
> 
> * AiInformation
> * DataPermissionLevels
> * NutritionFacts
> * Avatar
> * FormFieldLayout
> * NumberInput
> * RadioInput
> * Spinner
> * TextArea

I've fixed these too, `useStyle` was not typed properly

ToMESSKa (approved):

PR #2495, opened 2026-03-30
fix(ui-scripts,emotion): fix primitives and semantics override functionality

HerrTopi:
The PR converts the theme system from static token objects to lazy functions (semantics(primitives), components(semantics), sharedTokens(semantics)), enabling override-by-recomputation. It also
  introduces frozen themes so older component version's themes stay pinned to a snapshot.

This solution stores the primitives, semantics, sharedTokens (every part of every theme, except the unaffected components) and the affected components, for example `packages/ui-themes/src/themes/frozenThemes/v11_7/dark/primitives.ts`

TEST_PLAN:
- Check if overrides in 11.6 have not changed.
- Check the /v11_7/new-theme-overrides to see if it works as it should
- Check the code logic especially for InstUISettingsProvider and withStyle
 
Notes:
-  To test the frozen theme system, please check out [https://github.com/instructure/instructure-ui/pull/2519](This pr). It's the same as this but with added frozen functionality for Alert. It also contains some examples. That's only a show-case pr, will be abandoned when this will be merged

TODO for later: Find ALL v2 components that use `themeOverride` internally and migrate to this new system. (e.g. ColorPreset, Select)


docs/guides/new-theme-overrides.md
```diff
@@ -0,0 +1,399 @@
...
+      Success uses dark green (via semantic override).
+    </Alert>
+    <Alert variant="warning" margin="small">
+      Warning uses dark orange (via direct component override).
+    </Alert>
+  </InstUISettingsProvider>
+</InstUISettingsProvider>
+```
+
+### 8. Per-component `themeOverride` prop
```
matyasf: I'd rename this feature to "Componet's themeOverride prop". Now its unclear what this refers to, first I thought that its about the component level override in InstUISettingsProvider

packages/emotion/src/getComponentThemeOverrideLegacy.ts
```diff
@@ -0,0 +1,92 @@
...
+ * @param themeOverride - The theme override object
+ * @param componentTheme - The component's default theme
+ * @returns The calculated theme override object
+ */
+const getComponentThemeOverride = (
+  theme: ThemeOrLegacyOverride,
+  displayName: string,
+  componentId?: string,
+  // ThemeOverrideProp is the old type, ThemeOverrideValue is the new one
+  themeOverride?: ThemeOverrideProp['themeOverride'] | ThemeOverrideValue,
```
matyasf: If this is legacy why does it handle the new themes?
Am I right that code that handles new themes needs to be deleted from this file?

packages/emotion/src/useStyle.ts
```diff
@@ -23,7 +23,9 @@
  */
 
 import { useTheme } from './useTheme'
-import { getComponentThemeOverride } from './getComponentThemeOverride'
+// TODO: rework useStyle to use the new getComponentThemeOverride signature
```
matyasf: and to support frozen themes

packages/emotion/src/withStyle.tsx
```diff
@@ -231,10 +262,6 @@ const withStyle = decorator(
           {...props}
           makeStyles={makeStyleHandler}
           styles={styles}
-          // passing themeOverrides is needed for components like Button
-          // that have no makeStyles of their own and only pass themeOverrides
-          // to the underlying component (e.g.: BaseButton)
-          themeOverride={themeOverride}
```
matyasf, line 239: Deleting this might be an issue. As the comment says, how do we handle the case in `Button`?

packages/emotion/src/withStyle.tsx
```diff
@@ -161,7 +162,26 @@ const withStyle = decorator(
...
+      const themeInContext = useTheme() as Theme
+      const themeKey = themeInContext.key
+
+      // this is for the new overrides. theme.themeOverride has all the overrides
+      const themeOverride = themeInContext.themeOverride
+
+      // if a new theme has been added to the lib since this component version has been frozen, it can't be used with this
+      // theme, so we throw an error. Solution: upgrade to the latest version, it will support it
+      if (frozenTheme && !frozenTheme[themeKey]) {
+        throw new Error(
```
matyasf: This crashes the whole React render tree. I'd just emit an `error` to the console.

packages/ui-alerts/src/Alert/v2/index.tsx
```diff
@@ -49,14 +49,15 @@ import generateStyle from './styles'
...
+import frozenThemes from '@instructure/ui-themes/v11_7'
 
 /**
 ---
 category: components
 ---
 **/
 @withDeterministicId()
-@withStyle(generateStyle)
+@withStyle(generateStyle, null, frozenThemes)
```
matyasf: I like this solution, its apparent that something is different

packages/ui-alerts/src/Alert/v2/styles.ts
```diff
@@ -86,11 +84,12 @@ const generateStyle = (
...
-        boxShadow: boxShadowObjectsToCSSString(
-          sharedTokens.boxShadow.elevation4
-        )
-      })
+      ...(hasShadow &&
+        {
+          // boxShadow: boxShadowObjectsToCSSString(
+          //   sharedTokens.boxShadow.elevation4
+          // )
+        })
```
matyasf: I guess this is removed unintentionally?

packages/ui-themes/package.json
```diff
@@ -36,17 +36,24 @@
...
-    "./types/*": "./types/*",
-    "./package.json": "./package.json",
-    "./src/*": "./src/*"
+    "./v11_7": {
+      "src": "./src/themes/frozenThemes/v11_7/index.ts",
+      "types": "./types/themes/frozenThemes/v11_7/index.d.ts",
+      "import": "./es/themes/frozenThemes/v11_7/index.js",
+      "require": "./lib/themes/frozenThemes/v11_7/index.js",
+      "default": "./es/themes/frozenThemes/v11_7/index.js"
+    }
```
matyasf: I like this generally, good idea

packages/ui-themes/src/themes/frozenThemes/v11_7/index.ts
```diff
@@ -0,0 +1,35 @@
...
+import legacyCanvasHighContrast from './legacyCanvasHighContrast'
+import light from './light'
+import dark from './dark'
+
+export default {
+  canvas: legacyCanvas,
+  'canvas-high-contrast': legacyCanvasHighContrast,
+  light,
+  dark
+}
```
matyasf: Its a bit weird, that we always export the same stuff from `/vXY` and `/` except here.

Its OK that we don't export legacy themes, but I would export these the same way as the main index file:

```
export {
  dark,
  light,
  canvas,
  canvasHighContrast
}
export default canvas
```

`boxShadowObjectsToCSSString` would also be better in some of our utility packages, but that might create lots of extra dependencies, so lets leave it for now here.

matyasf (commented):
`ui-themes` will need a new folder structure, its a chaos now 

Image

.fallow/cache.bin
matyasf, line 1: this needs to be deleted

matyasf:
TODO:
 - there should be just 1 colors.ts: import { colors } from '../canvas/colors'
- comment in the new themes:
  // This is needed so 11.6 component don't break when using this theme.
  // They will default to the 'canvas' theme
  ...sharedThemeTokens,
  colors
- update description "Only usable with v11.7 or newer components"
- sharedThemeTokens -> legacySharedThemeTokens
- newThemes -> newThemeTokens

packages/emotion/src/withStyle.tsx
HerrTopi: it was but not anymore: in ...props the raw themeOverride is passed down to the children so it can calc its own overrides

ToMESSKa:
image
This pill is not purple

ToMESSKa:
image

This button is not deep pink

docs/guides/new-theme-overrides.md
```diff
@@ -0,0 +1,648 @@
...
+    variant="info"
+    margin="small"
+    themeOverride={{
+      infoBorderColor: 'crimson',
+      infoIconBackground: 'crimson'
+    }}
+  >
+    This specific Alert has crimson info styling.
+  </Alert>
+</InstUISettingsProvider>
```
ToMESSKa: This example might give someone the false impression that InstUISettingsProvider wrapper is needed around the component when using the `themeOverride` prop (which is not)?

ToMESSKa (changes requested):
see my comments

packages/emotion/src/EmotionTypes.ts
```diff
@@ -147,7 +147,7 @@ export interface StyleObject {
 
 export type {
   BaseThemeOrOverride,
-  ThemeOrOverride,
```
matyasf, line 150: Technically this is a breaking change, but I think we can leave it in

packages/emotion/src/useStyle.ts
```diff
@@ -70,38 +65,62 @@ const useStyle = <P extends GenerateStyleParams>(useStyleParams: {
...
+  // resolving the theming functions and applying the overrides
+  const primitiveOverrides = themeOverrideFromProvider?.primitives
+  const semanticsOverrides = themeOverrideFromProvider?.semantics
+  // @ts-ignore TODO-theme-types: fix typing
+  const sharedTokensOverrides = themeOverrideFromProvider?.sharedTokens
+  const componentOverridesFromSettingsProvider =
+    // @ts-ignore TODO-theme-types: fix typing
+    themeOverrideFromProvider?.components?.[
+    componentWithTokensId as keyof NewComponentTypes
+    ]
```
matyasf, line 84: Can you please write unit tests for this one (separate PR is OK too)

packages/emotion/src/withStyle.tsx
```diff
@@ -163,7 +164,26 @@ const withStyle = decorator(
...
+          `The version of ${displayName} you are using does not support the currently applied "${themeKey}" theme. ` +
+          `Please upgrade to the latest version of ${displayName}.`
+        )
+      }
+
+      // if this component is an older component version but still uses the new theming system, it'll have a frozenTheme
+      // object passed to it. We use that instead of the current theme so backward compatibility stays intact
+      const theme = frozenTheme
+        ? { newTheme: frozenTheme[themeKey] }
+        : themeInContext
```
matyasf, line 186: Can you write some tests for themes and overrides? (separate PR is OK too)

matyasf (approved):
Looks good, please add a task to write unit tests for this functionality

docs/guides/new-theme-overrides.md
```diff
@@ -0,0 +1,648 @@
...
+  <InstUISettingsProvider
+    themeOverride={{
+      components: {
+        Alert: {
+          infoBorderColor: 'purple',
+          infoIconBackground: 'purple'
+        },
+        Pill: {
+          primaryBackground: 'purple',
+          primaryBorderColor: 'purple'
```
git-nandor: The variable names for the pill are incorrect. Please use backgroundColor and baseBorderColor instead.

docs/guides/using-theme-overrides.md
```diff
@@ -122,18 +122,42 @@ type: example
...
+      themeOverride={{
+        components:{
           Alert: {
             infoIconBackground: "darkblue",
             infoBorderColor: "darkblue"
+          },
+          Pill:{
+
+          }
+        }
```
git-nandor, line 134: Is the double InstUISettingsProvider wrapper intentional?
Was it intentional to modify this old documentation page?

docs/guides/using-theme-overrides.md
git-nandor: The text input background doesn't change in either 11_6 or 11_7 versions!

git-nandor (changes requested):

docs/guides/new-theme-overrides.md
HerrTopi: Good catch. removed.

packages/emotion/src/useStyle.ts
HerrTopi: Added: https://instructure.atlassian.net/browse/INSTUI-5003

packages/emotion/src/withStyle.tsx
HerrTopi: Added : https://instructure.atlassian.net/browse/INSTUI-5003

docs/guides/using-theme-overrides.md
HerrTopi: Unrelated to this PR but a real bug, fixed

ToMESSKa (approved):

PR #2498, opened 2026-04-01
feat(ui-img): rework Img

ToMESSKa:
BREAKING CHANGE: contains breaking changes due to component using the new theming system

INSTUI-4970

**ISSUE:**

- Img needs to be migrated to the new theming system

**TEST PLAN:**


packages/ui-img/src/Img/v2/props.ts
```diff
@@ -0,0 +1,88 @@
...
+  height?: string | number
+  width?: string | number
+}
+
+type PropKeys = keyof ImgOwnProps
+
+type AllowedPropKeys = Readonly<Array<PropKeys>>
+
+type ImgProps = ImgOwnProps &
+  WithStyleProps<ImgTheme, ImgStyle> &
```
matyasf: Here use `WithStyleProps<NewComponentTypes['Img']` please

matyasf (changes requested):
nics work, just 1 small change

git-nandor (approved):

packages/ui-img/src/Img/v2/props.ts
ToMESSKa: I changed this

matyasf (approved):

hajnaldo (approved):

PR #2500, opened 2026-04-02
feat(ui-expandable): rework Expandable

git-nandor:
INSTUI-4988

Rework Expandable

Test:
Check doc Expandable page.
Ensure correct file structure.


packages/ui-expandable/package.json
```diff
@@ -46,17 +46,31 @@
...
+    ".": {
+      "src": "./src/exports/a.ts",
+      "types": "./types/exports/a.d.ts",
+      "import": "./es/exports/a.js",
+      "require": "./lib/exports/a.js",
+      "default": "./es/exports/a.js"
+    },
+    "./v11_6": {
+      "src": "./src/exports/a.ts",
+      "types": "./types/exports/a.d.ts",
```
ToMESSKa, line 63: I feel completely clueless here but I was wondering, shouldn't be there a 11_7 as well?

ToMESSKa (changes requested):

balzss (approved):
looks ok other than the thing mentioned by @ToMESSKa

packages/ui-expandable/package.json
git-nandor: I spoke with Topi, and yes, we also need v11_7, but the bump script will handle this automatically.

ToMESSKa (approved):

PR #2503, opened 2026-04-07
Move tests outside versions

balzss:
INSTUI-4975

## Summary

- Move all component test files from inside versioned directories (`v1/__tests__/`, `v2/__tests__/`) to component-level `__tests__/` directories
- Update test imports to use `@instructure/ui-<package>/latest` instead of relative versioned paths (e.g. `../v2`)
- Add `create-component-version` script to automate creating new component versions with proper git staging, export ordering, and blame preservation
- Move utility test files (`accepts`, `getEventFiles`, `cleanString`, `measureText`, `cleanData`) out of versioned dirs

## Test plan

- [ ] `pnpm run test:vitest` passes (all 2100+ tests)
- [ ] Run `pnpm run create-component-version` on a test component and verify:
  - [ ] `package.json` is staged
  - [ ] `./latest` appears last in the exports object
  - [ ] The frozen (old) version preserves original git blame


ToMESSKa:
- `ui-file-drop/src/FileDrop/v2/utils/__tests__` and `ui-truncate-text/src/TruncateText/v2/utils/__tests__` still have some tests files, is that intentional?

ToMESSKa:
- the test count is smaller than on the master
Image

packages/ui-breadcrumb/src/Breadcrumb/__tests__/Breadcrumb.test.tsx
```diff
@@ -27,7 +27,7 @@ import { vi } from 'vitest'
 import type { MockInstance } from 'vitest'
 import '@testing-library/jest-dom'
 import { runAxeCheck } from '@instructure/ui-axe-check'
-import { Breadcrumb } from '../index'
+import { Breadcrumb } from '../v2'
```
ToMESSKa: Is there a reason why it is not is it not `import { Breadcrumb } from '@instructure/ui-breadcrumb/latest'` like elsewhere?

packages/ui-radio-input/src/RadioInput/__tests__/RadioInput.test.tsx
```diff
@@ -29,7 +29,8 @@ import { vi } from 'vitest'
 
 import '@testing-library/jest-dom'
 import { runAxeCheck } from '@instructure/ui-axe-check'
-import { RadioInput, type RadioInputHandle } from '../index'
+import { RadioInput } from '@instructure/ui-radio-input/latest'
+import { type RadioInputHandle } from '../v2/index'
```
ToMESSKa: see my question on Breadcrumb

ToMESSKa:
the `pnpm run create-component-version` does not stage the package.json but it stages the rest after running it?

image

ToMESSKa:
Maybe a nitpick, but the latest isn't the last in order here

image

ToMESSKa:
Also I created a v3 for Avatar and committed but the the whole v3 files are showing me as the author? Maybe I misunderstood something,

ToMESSKa (changes requested):

ToMESSKa:
> * the test count is smaller than on the master
> 
> Image

The test count still do not match

ToMESSKa (changes requested):
see my comment

balzss:
> > * the test count is smaller than on the master
> > 
> > Image
> 
> The test count still do not match

you are right, the test count didn't match. there were a couple tests removed accidentally, those were added back, the rest are tests that were duplicated (present in both v1 and v2 folders)

packages/ui-scripts/lib/commands/create-component-version.ts
```diff
@@ -0,0 +1,420 @@
+/*
```
matyasf, line 1: FYI In the future we should create TypeScript files in `ui-scripts`, since newer Node versions can run them natively

packages/ui-scripts/lib/commands/create-component-version.js
```diff
@@ -0,0 +1,420 @@
...
+  info(`\nNext steps:`)
+  info(`  1. Make your breaking changes in ${relPath}/${newVersion}/`)
+  info(`  2. Run tests: pnpm run test:vitest ${path.basename(pkgDir)}`)
+  info(`  3. Update the README in ${relPath}/${newVersion}/README.md`)
+}
+
+/**
+ * Freeze the old version: pin /latest imports to the current version.
+ */
+function freezeVersion(repoRoot, versionDir, version) {
```
matyasf: `version` is unused

packages/ui-scripts/lib/commands/create-component-version.js
```diff
@@ -0,0 +1,420 @@
...
+  // Find the new export letter
+  const exportsDir = path.join(pkgDir, 'src', 'exports')
+  const exportFiles = fs
+    .readdirSync(exportsDir)
+    .filter((f) => /^[a-z]\.ts$/.test(f))
+    .sort()
+  const newLetter = exportFiles[exportFiles.length - 1].charAt(0)
+
+  // Determine new library version key
+  const currentVersion = pkgJson.version.replace(/\.\d+$/, '').replace('.', '_')
```
matyasf: `currentVersion` is unused

packages/ui-scripts/lib/commands/create-component-version.js
```diff
@@ -0,0 +1,420 @@
...
+}
+
+/**
+ * Update package.json exports: add new version entry and update /latest.
+ */
+function updatePackageJsonExports(repoRoot, pkgDir) {
+  const pkgJsonPath = path.join(pkgDir, 'package.json')
+  const pkgJson = JSON.parse(fs.readFileSync(pkgJsonPath, 'utf-8'))
+
+  if (!pkgJson.exports || !pkgJson.exports['./latest']) return
```
matyasf: Should't we throw an error in this case?

matyasf (changes requested):
looks good, just some small changes needed

matyasf (approved):
looks good!

ToMESSKa (approved):

PR #2506, opened 2026-04-07
fix(ui-calendar): DateInput2 days missing aria-selected

git-nandor:
INSTUI-4989

When a user selects a date, screen readers had no way to know which day was currently selected — aria-selected

## Solution summary

Every ARIA role that supports aria-selected requires a specific ancestor context — there's no standalone option
The only axe-clean alternatives for table mode without structural changes:

- aria-pressed on role="button" — valid, but "pressed" implies a toggle, not date selection. Wrong semantic.
- selectedLabel text ← what's implemented. Axe-clean, universally supported, correct semantic.

## Changes

- Fixes missing `aria-selected` on calendar days in `DateInput`, `DateInput2`, and `DateTimeInput` when using the default table layout
- In `listbox` mode (`role="listbox"`), days correctly use `role="option"` with `aria-selected="true/false"`; in table mode, days use `role="button"` where `aria-selected` is not ARIA-valid, so selection is now announced via a new `selectedLabel` prop that appends text to the accessible label (e.g. `"15 April 2026, Selected"`)
- Adds a `selectedLabel` prop to `Calendar`, `Calendar.Day`, `DateInput` (v1/v2), `DateInput2`, and `DateTimeInput` — passed through `screenReaderLabels.selectedLabel` in the DateInput variants;
- Adds a TODO comment documenting the future WAI-ARIA grid pattern implementation path

## Test plan
On the docs page, try this code sample and check what VoiceOver announces:
```
const Example = () => {
    const [selectedDate, setSelectedDate] = useState('')
    const [visibleMonth, setVisibleMonth] = useState('2025-05')
    const [useListbox, setUseListbox] = useState(false)

    return (
      <div>
        <Checkbox
          label={`role="${useListbox ? 'listbox' : 'table (default)'}" — click to toggle`}
          checked={useListbox}
          onChange={() => {
            setUseListbox((prev) => !prev)
            setSelectedDate('')
          }}
        />
        <View as="div" margin="small 0">
          <Text size="small" color="secondary">
            {useListbox
              ? 'Listbox mode: days have role="option" and aria-selected="true/false". Selection is announced via aria-selected.'
              : 'Table mode: days have role="button" with no aria-selected. Selection is announced via the accessible label (selectedLabel).'}
          </Text>
        </View>
        <Calendar
          role={useListbox ? 'listbox' : undefined}
          visibleMonth={visibleMonth}
          currentDate="2025-05-01"
          selectedDate={selectedDate}
          selectedLabel="selected"
          onRequestRenderNextMonth={(_e, requestedMonth) =>
            setVisibleMonth(requestedMonth)
          }
          onRequestRenderPrevMonth={(_e, requestedMonth) =>
            setVisibleMonth(requestedMonth)
          }
          onDateSelected={(date) => setSelectedDate(date)}
        />
      </div>
    )
  }

  render(<Example />)
```
  
  
### DIFF
--- a/packages/ui-calendar/src/Calendar/v2/Day/props.ts
+++ b/packages/ui-calendar/src/Calendar/v2/Day/props.ts
@@ -58,6 +58,11 @@ type CalendarDayOwnProps = {
    */
   isSelected?: boolean
   /**
+   * Screen reader label appended to the date label when the day is selected.
+   * Used to announce the selected state to assistive technologies.
+   */
+  selectedLabel: string
+  /**
    * Is the `<Calendar.Day />` today
    */
   isToday?: boolean
@@ -117,6 +122,7 @@ const allowedProps: AllowedPropKeys = [
   'label',
   'interaction',
   'isSelected',
+  'selectedLabel',
   'isToday',
   'isOutsideMonth',
   'onClick',
--- a/packages/ui-date-input/src/DateInput/v2/props.ts
+++ b/packages/ui-date-input/src/DateInput/v2/props.ts
@@ -33,14 +33,14 @@ type DateInputOwnProps = {
    */
   renderLabel: Renderable
   /**
-   * Accessible labels for the calendar button, month navigation buttons, and date picker dialog.
+   * Accessible labels for the calendar button, month navigation buttons, date picker dialog, and selected date state.
    */
   screenReaderLabels: {
     calendarIcon: string
     prevMonthButton: string
     nextMonthButton: string
-    // TODO: Make this field required in the next version. Currently optional to avoid breaking change.
-    datePickerDialog?: string
+    datePickerDialog: string
+    selectedLabel: string
   }
   /**
    * Specifies the input value.
--- a/packages/ui-date-time-input/src/DateTimeInput/v1/props.ts
+++ b/packages/ui-date-time-input/src/DateTimeInput/v1/props.ts
@@ -250,6 +250,11 @@ type DateTimeInputProps = {
    * NOTE: this won't call onChange, so you have to reset the value manually when calling reset
    */
   reset?: (reset: () => void) => void
+  /**
+   * Screen reader label appended to the date label when the day is selected.
+   * Used to announce the selected state to assistive technologies.
+   */
+  selectedLabel?: string
 }
 
 type DateTimeInputState = {
@@ -309,7 +314,8 @@ const allowedProps: AllowedPropKeys = [
   'disabledDates',
   'disabledDateTimeMessage',
   'allowNonStepInput',
-  'reset'
+  'reset',
+  'selectedLabel'
 ]
 
 export type { DateTimeInputProps, DateTimeInputState }
--- a/docs/guides/upgrade-guide.md
+++ b/docs/guides/upgrade-guide.md
@@ -470,6 +470,38 @@ type: embed
 
 ```
 
+#### DateTimeInput v1 → v2 API changes
+
+**Removed props:**
+
+| Removed prop          | Replacement                      |
+| --------------------- | -------------------------------- |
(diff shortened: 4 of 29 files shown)


packages/ui-date-input/src/DateInput/v2/props.ts
```diff
@@ -33,14 +33,15 @@ type DateInputOwnProps = {
...
   renderLabel: Renderable
   /**
-   * Accessible labels for the calendar button, month navigation buttons, and date picker dialog.
+   * Accessible labels for the calendar button, month navigation buttons, date picker dialog, and selected date state.
    */
   screenReaderLabels: {
     calendarIcon: string
     prevMonthButton: string
     nextMonthButton: string
     // TODO: Make this field required in the next version. Currently optional to avoid breaking change.
```
git-nandor: It’s not included in v1 yet, so I left it unchanged. As I understand it, this will apply to v3.
This is also the case for DateInput2

packages/ui-calendar/src/Calendar/v2/Day/props.ts
```diff
@@ -57,6 +57,11 @@ type CalendarDayOwnProps = {
    * Is the `<Calendar.Day />` selected
    */
   isSelected?: boolean
+  /**
+   * Screen reader label appended to the date label when the day is selected.
+   * Used to announce the selected state to assistive technologies.
+   */
+  selectedLabel?: string
```
matyasf: I think this can be made mandatory in v2

packages/ui-date-input/src/DateInput/v2/props.ts
```diff
@@ -33,14 +33,15 @@ type DateInputOwnProps = {
...
     datePickerDialog?: string
+    selectedLabel?: string
```
matyasf: I think this can be mandatory in v2

packages/ui-date-input/src/DateInput/v2/props.ts
matyasf: If its just used in v11.7, then please make it mandatory

packages/ui-date-time-input/src/DateTimeInput/v1/props.ts
```diff
@@ -250,6 +250,11 @@ type DateTimeInputProps = {
    * NOTE: this won't call onChange, so you have to reset the value manually when calling reset
    */
   reset?: (reset: () => void) => void
+  /**
+   * Screen reader label appended to the date label when the day is selected.
+   * Used to announce the selected state to assistive technologies.
+   */
+  selectedLabel?: string
```
matyasf, line 257: I think this can be made mandatory

matyasf (changes requested):
see my comments

balzss (approved):
I agree with @matyasf on making the prop mandatory in v2. other than that everything looks ok

packages/ui-date-time-input/src/DateTimeInput/v1/props.ts
git-nandor: If this becomes mandatory in DateTimeInput/v1, it will also impact Calendar/v1, Day/v1, and DateInput/v1.
Should I create a DateTimeInput/v2 and make it mandatory there?

matyasf: hmm if this would be the only change, then no IMO

matyasf: https://github.com/instructure/instructure-ui/pull/2518/changes#diff-654096f083ddf568261002852e75214784f20bb79fd09eb025f0905774fe99da should be mandatory in v2, but in this PR

matyasf (approved):

PR #2508, opened 2026-04-09
feat(ui-table): add screen reader announcements for column sorting

ToMESSKa:
INSTUI-4699

**ISSUE:**
- Table is not accessible by default: screen reader announcements for column sorting use custom code in the examples when this behavior should be the default

**TEST PLAN:**
- check for v11.6 and and v11.7:
  - focus the whole table in the two Table examples that can be ordered by columns with NVDA/VoiceOver/JAWS.
  - check whether the screen reader announces ordering for the Table — it must announce the right column and right order.
  - try changing the ordering and observe whether it announces the changed column and order.
  - after changing the ordering, navigate out of the Table then back again. It should announce the right ordering and column
  - check if every changes here https://github.com/instructure/instructure-ui/pull/2154 has been implemented

This was [reported for Canvas 11.6.0 ](https://instructure.slack.com/archives/C0JCJ63TR/p1775727754648709?thread_ts=1756889154.173259&cid=C0JCJ63TR) so no backport is needed

### DIFF
--- a/packages/ui-table/src/Table/v1/index.tsx
+++ b/packages/ui-table/src/Table/v1/index.tsx
@@ -71,23 +71,51 @@ class Table extends Component<TableProps> {
   static Cell = Cell
 
   ref: Element | null = null
+  // Reference to hidden aria-live region for announcing caption changes to screen readers
+  _liveRegionRef: HTMLDivElement | null = null
+  // Timeout for delayed announcement (workaround for Safari/VoiceOver caption update bug)
+  _announcementTimeout?: ReturnType<typeof setTimeout>
 
   handleRef = (el: Element | null) => {
-    const { elementRef } = this.props
-
     this.ref = el
-
-    if (typeof elementRef === 'function') {
-      elementRef(el)
-    }
+    this.props.elementRef?.(el)
   }
 
   componentDidMount() {
     this.props.makeStyles?.()
   }
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps: TableProps) {
     this.props.makeStyles?.()
+    // Announce caption changes for screen readers (especially VoiceOver)
+    // Safari/VoiceOver has a known bug where dynamic <caption> updates aren't announced,
+    // so we use an aria-live region as a workaround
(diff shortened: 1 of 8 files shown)


packages/ui-table/src/Table/v1/index.tsx
```diff
@@ -71,23 +71,51 @@ class Table extends Component<TableProps> {
   static Cell = Cell
 
   ref: Element | null = null
+  // Reference to hidden aria-live region for announcing caption changes to screen readers
```
ToMESSKa, line 74: The thing is I could not reproduce the issue I myself reported here last November https://github.com/instructure/instructure-ui/pull/2154#pullrequestreview-3432944532 maybe there was some browers/screender update since then which fixed this (?) So I just implemented @joyenjoyer's exact fix for the multi version. Please test if that issue is not longer exists.

packages/ui-table/src/Table/v1/index.tsx
```diff
@@ -71,23 +71,51 @@ class Table extends Component<TableProps> {
...
   componentDidMount() {
     this.props.makeStyles?.()
   }
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps: TableProps) {
     this.props.makeStyles?.()
+    // Announce caption changes for screen readers (especially VoiceOver)
+    // Safari/VoiceOver has a known bug where dynamic <caption> updates aren't announced,
+    // so we use an aria-live region as a workaround
```
matyasf, line 92: Do we know the link to this bug? it would be nice to add there so we can track it.

matyasf (changes requested):

packages/ui-table/src/Table/v1/index.tsx
ToMESSKa: I asked @joyenjoyer about this but unfortunately no link is available. But he and me tested this dynamic `<caption> `bug and it still persists.

matyasf (approved):

joyenjoyer (approved):
looks good

PR #2512, opened 2026-04-10
feat(ui-top-nav-bar,ui-options,ui-buttons): rework TopNavBar

ToMESSKa:
BREAKING CHANGE: contains breaking changes due to component using the new theming system

INSTUI-4970

**ISSUE:**

TopNavBar needs to be migrated to the new theming system

**TEST PLAN:**

- check if the NewComponentTypes are used in WithStyleProps in props.ts


packages/ui-options/src/Options/v2/Item/index.tsx
```diff
@@ -88,18 +88,24 @@ class Item extends Component<OptionsItemProps> {
   ) {
     const { styles, variant, as, role, children } = this.props
 
+    let labelContent: React.ReactNode
+    if (typeof renderLabel === 'function') {
+      labelContent = createElement(
+        renderLabel as React.ComponentType<any>,
+        { variant, as, role },
+        children as React.ReactNode
```
ToMESSKa, line 96: In the first TopNavBar example, Opening the 'User settings' menu threw an "Invalid hook call." error. Here icons are passed as component references and I also passed the new lucide icons like that (`UserInstUIIcon `instead of `<UserInstUIIcon/>`) but I suppose because Lucide icons use hooks, React threw an error, because hooks can't run inside a class component's render. So I tried to use createElement here to solve this.

But this resulted the `<Drilldown.Option /> > renderBeforeLabel prop > as function should have option props as params` test to fail, because createElement also passed an extra arg (legacyContext). So a  wrapper is used in Options.Item to only pass 2 args instead of 3.

Not sure it is the best solution so open for suggestions.

packages/ui-top-nav-bar/src/TopNavBar/v1/TopNavBarActionItems/index.tsx
```diff
@@ -238,11 +238,17 @@ class TopNavBarActionItems extends Component<
...
+      renderHiddenItemsMenuTriggerLabel,
+      renderHiddenItemsMenuTriggerTooltip,
+      ...restProps
+    } = this.props
 
     return (
       <TruncateList
-        {...passthroughProps(omitProps(this.props, allowedProps))}
+        {...passthroughProps(omitProps(restProps, allowedProps))}
         key={this.state.key} // rerender if child count changes
```
ToMESSKa, line 252: This @git-nandor's fix incorporated in the PR: renderHiddenItemsMenuTrigger* props were spreading through to `<ul>` causing React unrecognized prop warnings.

packages/ui-top-nav-bar/src/TopNavBar/v1/TopNavBarMenuItems/index.tsx
```diff
@@ -261,7 +267,7 @@ class TopNavBarMenuItems extends Component<
 
     return (
       <TruncateList
-        {...omitProps(this.props, allowedProps)}
+        {...omitProps(restProps, allowedProps)}
         key={this.state.key} // rerender if child count changes
```
ToMESSKa, line 271: This @git-nandor's fix incorporated in the PR: renderHiddenItemsMenuTrigger* props were spreading through to `<ul> `causing React unrecognized prop warnings.

packages/ui-top-nav-bar/src/TopNavBar/v1/README.md
```diff
@@ -2033,7 +2033,7 @@ type: example
               renderHiddenItemsMenuTriggerLabel={(
                 hiddenChildrenCount
               ) => `${hiddenChildrenCount} More`}
-             renderHiddenItemsMenuTriggerAriaLabel={(
+             renderHiddenItemsMenuTriggerTooltip={(
                hiddenChildrenCount
```
ToMESSKa, line 2037: This @git-nandor's fix incorporated in the PR: Copy-paste error in docs; ActionItems doesn't have an AriaLabel variant, only Tooltip.

packages/ui-top-nav-bar/src/TopNavBar/v2/TopNavBarBrand/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+              ({
+                newTheme: {
+                  sharedTokens: {
+                    focusOutline: {
+                      ...currentTheme?.newTheme?.sharedTokens?.focusOutline,
+                      inset: styles?.focusOutlineInset
+                    }
+                  }
+                }
+              } as any)
```
ToMESSKa, line 108: This code previously included a themeOverride which was overriding one of BaseButton's focus ring tokens, which is deleted v2 (and repalced with calcFocusOutline). So I tried to override the sharedtoken value here to achieve that inner focus in the brand icon see below. But not sure it is the proper way to do it.

Image

packages/ui-options/src/Options/v2/Item/index.tsx
matyasf: I still get an invalid hook with this code, can you look into it? (both in the 11.6 and 11.7 previews)

```
<View display="block" width="300px" >
  <Options renderLabel={UserInstUIIcon}>
    <Options.Item>Default option</Options.Item>
  </Options>
</View>
```

packages/ui-top-nav-bar/src/TopNavBar/v2/TopNavBarBrand/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+        'Warning: `renderName` prop on [TopNavBar.Brand] is deprecated, please remove it from your code. Further info: https://instructure.design/#v9-upgrade-guide/#deprecated-properties'
+      )
+    }
+
+    // TODO: Remove this in v10!
+    if ('nameBackground' in this.props) {
+      console.error(
+        'Warning: `nameBackground` prop on [TopNavBar.Brand] is deprecated, please remove it from your code. Further info: https://instructure.design/#v9-upgrade-guide/#deprecated-properties'
+      )
+    }
```
matyasf: Can you remove this in the v2 version? (also put it into the changelog

packages/ui-top-nav-bar/src/TopNavBar/v2/TopNavBarBrand/index.tsx
matyasf: I think it looks good, I've tried to modify the override and I could see the results.

matyasf (changes requested):
see my comments.

This component has no token name changes, right? If there are please put it into the README

packages/ui-top-nav-bar/src/TopNavBar/v2/README.md
```diff
@@ -0,0 +1,2363 @@
...
+                            this.setState({ searchInputValue: '' })
+
+                            if (currentLayout === 'smallViewport' && typeof closeInPlaceDialog === 'function') {
+                              closeInPlaceDialog()
+                            }
+                          }
+                        }}
+                        themeOverride={(_theme, globalTheme) => ({
+                          focusOutlineColor: globalTheme.colors.borderLightest
+                        })}
```
matyasf: This does not work, you need to override `sharedTokens`

packages/ui-options/src/Options/v2/Item/index.tsx
ToMESSKa: I fixed this Options/v2/index.tsx and Options/v1/index.tsx. I have put the Options/v2/Item/index.tsx fix into Options/v1/Item/index.tsx

packages/ui-top-nav-bar/src/TopNavBar/v2/TopNavBarBrand/index.tsx
ToMESSKa: I removed them and updated the upgrade guide. There was no token change.

packages/ui-top-nav-bar/src/TopNavBar/v2/README.md
ToMESSKa: I fixed this, please check.

matyasf (approved):
looks good now, nice work

packages/ui-top-nav-bar/src/TopNavBar/v2/TopNavBarLayout/DesktopLayout/props.ts
```diff
@@ -0,0 +1,91 @@
...
+type DesktopLayoutChild = React.ComponentElement<
+  TopNavBarDesktopLayoutProps,
+  TopNavBarDesktopLayout
+>
+
+type DesktopLayoutOwnProps = {
+  /**
+   * Hides the separator between the action items and the user block.
+   */
+  hideActionsUserSeparator?: boolean
```
balzss: this is not an issue with your PR since this was also present in v1 but it's not used anymore, so we should remove it now

packages/ui-top-nav-bar/src/TopNavBar/v2/TopNavBarLayout/DesktopLayout/styles.ts
```diff
@@ -0,0 +1,145 @@
...
+    actionItemsContainer: {
+      label: 'topNavBarDesktopLayout__actionItemsContainer',
+      display: 'flex',
+      flexGrow: 0,
+      flexShrink: 0,
+      marginBlock: 0,
+      marginInline: componentTheme.desktopActionItemsContainerInlineMargin
+    },
+    spacer: {
+      flexGrow: '1'
```
balzss: nitpick: this should be a number, not string

packages/ui-top-nav-bar/src/TopNavBar/v2/TopNavBarLayout/SmallViewportLayout/styles.ts
```diff
@@ -0,0 +1,185 @@
...
+        height: '100%',
+        position: 'absolute'
+      },
+      // removes the focus ring on the drilldown,
+      // not needed on the mobile menu
+      [`#${drilldownId}`]: {
+        height: '100%',
+
+        '&::before': {
+          // display: 'none'
```
balzss: let's remove this if not needed

balzss (approved):
the code looks good and the component works as expected. great work, only left some minor comments but if those are fixed, this can be merged

hajnaldo (commented):
**TopNavBar MenuItems:**
Border radius seems to be the legacy one in light and dark (4px), should be 16px
➡️ Is there a token for this specific radius? - This is visible under the MenuItems example.

ToMESSKa:
> **TopNavBar MenuItems:** Border radius seems to be the legacy one in light and dark (4px), should be 16px ➡️ Is there a token for this specific radius? - This is visible under the MenuItems example.

@hajnaldo I should be fixed now, can you check it please?

packages/ui-top-nav-bar/src/TopNavBar/v2/TopNavBarLayout/SmallViewportLayout/styles.ts
ToMESSKa: Removed

packages/ui-top-nav-bar/src/TopNavBar/v2/TopNavBarLayout/DesktopLayout/styles.ts
ToMESSKa: Fixed this

packages/ui-top-nav-bar/src/TopNavBar/v2/TopNavBarLayout/DesktopLayout/props.ts
ToMESSKa: Removed it

hajnaldo (approved):

PR #2514, opened 2026-04-14
fix(ui-motion): use elementRef instead of ref in BaseTransition renderChildren

Gazuru:
## Summary

- Silences the React 18 `Alert: 'ref' is not a prop` warning that appears when wrapping `<Alert>` (or any InstUI class component) in `<Transition>`.
- `BaseTransition` now routes refs by child kind: host elements get `ref`, forwardRef components get `ref` + `elementRef`, anything else gets only `elementRef`. No more passing `ref` to non-forwardRef components.
- `Alert.handleRef` chains the consumer-supplied `elementRef` so the DOM node still reaches the wrapping `Transition` and the fade in/out works end-to-end.

Fixes LX-4024.

### DIFF
--- a/packages/ui-motion/src/Transition/BaseTransition/index.ts
+++ b/packages/ui-motion/src/Transition/BaseTransition/index.ts
@@ -24,11 +24,12 @@
 
 import { Component, ReactElement } from 'react'
 
-import { getClassList, findDOMNode } from '@instructure/ui-dom-utils'
+import { getClassList } from '@instructure/ui-dom-utils'
 import {
   ensureSingleChild,
   safeCloneElement
 } from '@instructure/ui-react-utils'
+import { createChainedFunction } from '@instructure/ui-utils'
 
 import { allowedProps } from './props'
 import type {
@@ -37,6 +38,24 @@ import type {
   BaseTransitionStateValue,
   BaseTransitionStatesType
 } from './props'
+
+const REACT_FORWARD_REF_TYPE = Symbol.for('react.forward_ref')
+const REACT_MEMO_TYPE = Symbol.for('react.memo')
+
+type ElementRefCallback = (el: Element | null) => void
+type ChildElementType = ReactElement['type']
+
+function unwrapMemo(type: ChildElementType): ChildElementType {
+  let unwrapped = type
+  while (
+    unwrapped &&
+    typeof unwrapped !== 'string' &&
+    (unwrapped as { $$typeof?: symbol }).$$typeof === REACT_MEMO_TYPE
+  ) {
+    unwrapped = (unwrapped as unknown as { type: ChildElementType }).type
+  }
+  return unwrapped
+}
 
 const STATES: Record<BaseTransitionStateValue, BaseTransitionStatesType> = {
   EXITED: -2,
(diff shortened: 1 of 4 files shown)


packages/ui-motion/src/Transition/BaseTransition/index.ts
```diff
@@ -333,19 +333,23 @@ class BaseTransition extends Component<
...
-          {
-            'aria-hidden': !this.props.in ? true : undefined,
-            ref: (el: React.ReactInstance | null) => {
-              const ref = (findDOMNode(el) as Element) || null
-
-              this.handleRef(ref)
+    if (!this.props.children) return null
+
+    const child = ensureSingleChild(this.props.children) as ReactElement
+    const isInstUIComponent = !!(child.type as any)?.componentId
```
joyenjoyer: Not all components have `componentId`, I believe this is not a reliable way to check if it is an InstUI component or not.

ToMESSKa:
@Gazuru The `ref is not a prop ` warning disappears but the smooth fade in and fade out effect do not work with the same code example we received from you (see below) on this branch as it does on the master.
(Also I see a lot merge conflicts. I think it is easier to just reapply your changes than trying to resolve the conflicts.)

`master` vs `LX-4014-fix-ref-is-not-a-prop`

https://github.com/user-attachments/assets/4dea013e-4747-4bac-8d31-6d4f533fe853

https://github.com/user-attachments/assets/5c2302e6-c82d-4ddb-a0d6-e71c26a3deb0

 ```
const Example = () => {
    const [alerts, setAlerts] = useState([])
    const nextId = useRef(0)

    const trigger = () => {
      const id = nextId.current++
      setAlerts((prev) => [...prev, id])
    }

    const remove = (id) =>
      setAlerts((prev) => prev.filter((x) => x !== id))

    return (
      <div>
        <Button onClick={trigger}>Trigger flash alert</Button>
        <View as="div" margin="small 0 0">
          {alerts.map((id) => (
            <FlashAlert
              key={id}
              message="Module access updated successfully"
              variant="success"
              timeout={4000}
              onClose={() => remove(id)}
            />
          ))}
        </View>
      </div>
    )
  }

  const FlashAlert = ({ onClose, message, variant, timeout }) => {
    const [isOpen, setIsOpen] = useState(true)

    const closeAlert = () => {
      setIsOpen(false)
      setTimeout(onClose, 500)
    }

    return (
      <Transition transitionOnMount in={isOpen} type="fade">
        <Alert
          variant={variant}
          renderCloseButtonLabel="Close"
          onDismiss={closeAlert}
          margin="small auto"
          timeout={timeout}
        >
          <div>{message}</div>
        </Alert>
      </Transition>
    )
  }

  render(<Example />)


```

**Review by ToMESSKa — CHANGES_REQUESTED** [review]
see my comment

**Comment by ToMESSKa** [comment]
Looks good now:
- no `ref is not a prop` warning 
- the fade-in fade-out transition works

The only issue might be that as findDOMNode has been removed, **_non-InstUI class components_** wrapped in `<Transition>`  will now silently fail to expose the DOM node (and the animation breaks too). But as this pattern could be rare so it's not an issue? What do you think @balzss @joyenjoyer?

check this example in the master vs this PR
```
class PlainClassChild extends React.Component {
  render() {
    return (
      <div style={{ padding: '0.5rem', background: '#e0e7ff' }}>
        Plain class component (no elementRef, no forwardRef)
      </div>
    )
  }
}

const Example = () => {
  const [captured, setCaptured] = React.useState('(none)')
  const [isIn, setIsIn] = React.useState(true)

  return (
    <View as="div" padding="medium">
      <button onClick={() => setIsIn((v) => !v)} style={{ marginBottom: '1rem' }}>
        Toggle in: currently {String(isIn)}
      </button>

      <Transition
        type="fade"
        in={isIn}
        transitionOnMount={true}
        elementRef={(el) => {
          setCaptured(el instanceof Element ? el.tagName : String(el))
        }}
      >
        <PlainClassChild />
      </Transition>

      <p>elementRef received: <code>{captured}</code></p>
    </View>
  )
}

render(<Example />)
```

ToMESSKa (commented):
see my comment

joyenjoyer:
I created #2618 to handle this issue. Check that one.

PR #2517, opened 2026-04-15
feat(ui-color-picker): improve ColorPicker paste behavior

ToMESSKa:
INSTUI-4834

**ISSUE:**
- pasted hex color values can't replace existing values without manual clearing, e.g. pasting FF0000, then select all FF0000 and paste AABBCC resulted in nothing

**TEST PLAN:**

-  selection replacement
    - paste FF0000, then select all FF0000 and paste AABBCC, the input should become AABBCC and result in valid color
    - in FF0000, select 00 in the middle, paste FF, it input should become FFFF00 and result in valid color
  
-  cursor position
    - paste 0000, put the cursor at at the start of 0000 and paste FF, the input should become FF0000 and result in valid color
    - paste FF00, put the cursor in the middle of FF00 and paste AB, the input should become FFAB00 and result in valid color
    - put the cursor at the end of 0000 and  Paste FF, the input should become 0000FF

- '#' stripping
    - paste #FF0000, it should still become FF0000
    - paste FF0000, it should still become FF0000

- length limits
    - Paste a 7-char string like FF00001, it should be still blocked

- invalid content
    - paste ZZZZZZ, it should be still blocked

### DIFF
--- a/packages/ui-color-picker/src/ColorPicker/v1/index.tsx
+++ b/packages/ui-color-picker/src/ColorPicker/v1/index.tsx
@@ -431,7 +431,14 @@ class ColorPicker extends Component<ColorPickerProps, ColorPickerState> {
       return event.preventDefault()
     }
 
-    const newHex = `${this.state.hexCode}${toPaste}`
+    const input = event.target as HTMLInputElement
+    const currentHex = this.state.hexCode
+    const selectionStart = input.selectionStart ?? currentHex.length
+    const selectionEnd = input.selectionEnd ?? currentHex.length
+    const newHex =
+      currentHex.slice(0, selectionStart) +
+      toPaste +
+      currentHex.slice(selectionEnd)
     if (isValid(newHex)) {
       if (typeof this.props.onChange === 'function') {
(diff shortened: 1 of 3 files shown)


packages/ui-color-picker/src/ColorPicker/v1/index.tsx
```diff
@@ -431,7 +431,14 @@ class ColorPicker extends Component<ColorPickerProps, ColorPickerState> {
       return event.preventDefault()
     }
 
-    const newHex = `${this.state.hexCode}${toPaste}`
```
ToMESSKa, line 434: new pasted value previously always got appended at the end of the current one (e.g. FF0000 selected then AABBCC pasted → F0000AABBCC) and it was evaluated as invalid

joyenjoyer (changes requested):
looks good, I would add a few unit tests though

ToMESSKa:
> looks good, I would add a few unit tests though

@joyenjoyer I added some tests

HerrTopi (approved):

joyenjoyer (approved):

PR #2518, opened 2026-04-16
feat(ui-select,ui-date-time-input): rework DateTimeInput and replace DateInput v1 with DateInput v2 in DateTimeInput

ToMESSKa:
BREAKING CHANGE:  

- prevMonthLabel prop removed (use screenReaderLabels.prevMonthButton instead)
- nextMonthLabel prop removed (use screenReaderLabels.nextMonthButton instead)
- renderWeekdayLabels prop removed| dateFormat type changed: string → string | { parser: (input: string) => Date | null, formatter: (date: Date) => string }
- screenReaderLabels is a new required prop 
- dateFormat default changed: Moment's 'LL' (long month name) → locale's default date format

INSTUI-4991

**ISSUE:**

- DateTimeInput needs to be migrated to the new theming system and use DateInput v2 under the hood

**TEST PLAN:**

- check if the v2 component uses all the tokens (if there is any)
- check if the NewComponentTypes are used in WithStyleProps in props.ts
- **check if the components behaves the same after DateTimeInput v2 was implemented under the hood**

### DIFF
--- a/cypress/component/DateTimeInput.cy.tsx
+++ b/cypress/component/DateTimeInput.cy.tsx
@@ -39,8 +39,11 @@ describe('<DateInput/>', () => {
     cy.mount(
       <DateTimeInput
         description="date time description"
-        prevMonthLabel="Previous month"
-        nextMonthLabel="Next month"
+        screenReaderLabels={{
+          calendarIcon: 'Choose date',
+          prevMonthButton: 'Previous month',
+          nextMonthButton: 'Next month'
+        }}
         dateRenderLabel="date-input label"
         timeRenderLabel="time-input label"
         invalidDateTimeMessage="whoops"
@@ -56,10 +59,10 @@ describe('<DateInput/>', () => {
     cy.contains('time-input label')
     cy.contains('Thursday, January 18, 2018 1:30 PM')
 
-    cy.get('input[id^="Selectable_"]').as('dateInput')
+    cy.get('input[id^="TextInput_"]').as('dateInput')
     cy.get('input[id^="Select_"]').as('timeInput')
 
-    cy.get('@dateInput').should('have.value', 'January 18, 2018')
+    cy.get('@dateInput').should('have.value', '1/18/2018')
     cy.get('@timeInput').should('have.value', '1:30 PM')
 
     cy.get('@dateInput').clear().blur()
@@ -67,32 +70,29 @@ describe('<DateInput/>', () => {
     cy.get('@timeInput').should('have.value', '')
     cy.wrap(onChange).should('have.been.called')
 
-    cy.get('@dateInput').realClick().wait(100)
+    cy.contains('button', 'Choose date').realClick().wait(100)
 
-    cy.contains('button', '22')
-      .realClick()
-      .wait(100)
-      .then(($btn) => {
-        const selectedDateId = $btn.attr('id')!
-        const selectedDateValue = DateTime.parse(
-          selectedDateId,
+    cy.contains('button', '22').realClick().wait(100)
+
+    cy.wrap(onChange)
+      .should('have.been.called')
+      .then((spy) => {
+        const selectedDateId: string = spy.lastCall.args[1]
+        const selectedDateValue = new Date(selectedDateId).toLocaleDateString(
           'en-US',
-          'US/Eastern'
-        ).format('LL')
+          {
+            timeZone: 'US/Eastern',
+            calendar: 'gregory',
+            numberingSystem: 'latn'
+          }
+        )
 
         cy.get('@dateInput').should('have.value', selectedDateValue)
         cy.get('@timeInput').should('have.value', '4:16 PM')
 
-        cy.wrap(onChange)
-          .should('have.been.called')
-          .then((spy) => {
-            const lastCallFirstArg = spy.lastCall.args[1]
+        const lastCallDatePart = selectedDateId.split('T')[0]
 
-            const lastCallDatePart = lastCallFirstArg.split('T')[0]
-            const expectedDatePart = selectedDateId.split('T')[0]
-
-            expect(lastCallDatePart).to.equal(expectedDatePart)
-          })
+        expect(lastCallDatePart).to.include('-22')
       })
   })
 
@@ -103,8 +103,11 @@ describe('<DateInput/>', () => {
       <DateTimeInput
         description="date_time"
         dateRenderLabel="date-input"
-        prevMonthLabel="Previous month"
-        nextMonthLabel="Next month"
+        screenReaderLabels={{
+          calendarIcon: 'Open calendar',
+          prevMonthButton: 'Previous month',
+          nextMonthButton: 'Next month'
+        }}
         timeRenderLabel="time-input"
         invalidDateTimeMessage="whoops"
         locale="en-US"
@@ -112,7 +115,7 @@ describe('<DateInput/>', () => {
         onChange={onChange}
       />
     )
-    cy.get('input[id^="Selectable_"]').as('dateInput')
+    cy.get('input[id^="TextInput_"]').as('dateInput')
     cy.get('@dateInput').realClick().wait(100)
     cy.get('@dateInput').type('Not a date{enter}')
     cy.get('@dateInput').blur()
@@ -128,8 +131,11 @@ describe('<DateInput/>', () => {
     cy.mount(
       <DateTimeInput
         description="date_time"
-        prevMonthLabel="Previous month"
-        nextMonthLabel="Next month"
+        screenReaderLabels={{
+          calendarIcon: 'Open calendar',
+          prevMonthButton: 'Previous month',
+          nextMonthButton: 'Next month'
+        }}
         dateRenderLabel="date-input"
         timeRenderLabel="time-input"
         invalidDateTimeMessage="whoops"
@@ -140,9 +146,9 @@ describe('<DateInput/>', () => {
         disabledDateTimeMessage={errorMsg}
       />
     )
-    cy.get('input[id^="Selectable_"]').as('dateInput')
+    cy.get('input[id^="TextInput_"]').as('dateInput')
     cy.get('body').should('contain', errorMsg)
-    cy.get('@dateInput').clear().type(`05/18/2017{enter}`)
+    cy.get('@dateInput').clear().type(`05/18/2017`).blur()
 
     cy.get('body').should('not.contain', errorMsg)
   })
@@ -160,8 +166,11 @@ describe('<DateInput/>', () => {
     cy.mount(
       <DateTimeInput
         description="date_time"
-        prevMonthLabel="Previous month"
-        nextMonthLabel="Next month"
+        screenReaderLabels={{
+          calendarIcon: 'Open calendar',
+          prevMonthButton: 'Previous month',
+          nextMonthButton: 'Next month'
+        }}
         dateRenderLabel="date-input"
         timeRenderLabel="time-input"
         invalidDateTimeMessage="whoops"
@@ -172,9 +181,9 @@ describe('<DateInput/>', () => {
         disabledDateTimeMessage={errorMsg}
       />
     )
-    cy.get('input[id^="Selectable_"]').as('dateInput')
+    cy.get('input[id^="TextInput_"]').as('dateInput')
     cy.get('body').should('contain', errorMsgText)
-    cy.get('@dateInput').clear().type(`May 18, 2022{enter}`)
+    cy.get('@dateInput').clear().type(`May 18, 2022`).blur()
 
     cy.get('body').should('not.contain', errorMsgText)
   })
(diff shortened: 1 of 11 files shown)


packages/ui-select/src/Select/v2/index.tsx
```diff
@@ -422,13 +421,6 @@ class Select extends Component<SelectProps> {
         : (renderOptionLabel as React.ReactNode)
     }
 
-    const { isOptionContentAppliedToInput, size = 'medium' } = this.props
-    const iconSize = selectSizeToIconSize[size]
-    const checkIcon =
```
ToMESSKa, line 427: Unfortunately in my [Select rework](https://github.com/instructure/instructure-ui/pull/2465#pullrequestreview-4118725859), some draft code was left in about implementing a check icon on selected items (since then we aggreed not to implement that), so I reverted these here.

cypress/component/DateTimeInput.cy.tsx
```diff
@@ -67,13 +68,14 @@ describe('<DateInput/>', () => {
...
 
-    cy.contains('button', '22')
-      .realClick()
-      .wait(100)
-      .then(($btn) => {
-        const selectedDateId = $btn.attr('id')!
+    cy.wrap(onChange)
+      .should('have.been.called')
+      .then((spy) => {
+        const selectedDateId: string = spy.lastCall.args[1]
```
ToMESSKa, line 80: In DateInput v2 no id is set on day buttons, so I get the ISO date from the onChange spy.

cypress/component/DateTimeInput.cy.tsx
```diff
@@ -140,9 +135,9 @@ describe('<DateInput/>', () => {
         disabledDateTimeMessage={errorMsg}
       />
     )
-    cy.get('input[id^="Selectable_"]').as('dateInput')
+    cy.get('input[id^="TextInput_"]').as('dateInput')
     cy.get('body').should('contain', errorMsg)
-    cy.get('@dateInput').clear().type(`05/18/2017{enter}`)
+    cy.get('@dateInput').clear().type(`05/18/2017`).blur()
```
ToMESSKa, line 151: DateInput v1 had an Enter key handler. In v2 there's no such handler, so the .blur() is used.

docs/guides/upgrade-guide.md
```diff
@@ -263,6 +263,38 @@ type: embed
...
 ```
 
+#### DateTimeInput v1 → v2 API changes
+
+**Removed props:**
+
+| Removed prop          | Replacement                      |
+| --------------------- | -------------------------------- |
+| `renderWeekdayLabels` | Built in — no replacement needed |
+
```
git-nandor: It might be better to use V12ChangelogTable here.

packages/ui-date-time-input/src/DateTimeInput/v2/README.md
```diff
@@ -0,0 +1,261 @@
...
+  const Example = () => {
+    return (
+      <div style={{ height: '15rem' }}>
+        <DateTimeInput
+          description="Pick a date and time"
+          datePlaceholder="Choose a date"
+          dateRenderLabel="Date"
+          timeRenderLabel="Time"
+          invalidDateTimeMessage="Invalid date!"
+          calendarIcon="Open calendar" prevMonthLabel="Previous month" nextMonthLabel="Next month"
```
git-nandor: The example code would look better on the documentation page if these were broken into new lines.

git-nandor:
The commit message doesn't include the breaking prop changes. 
Could you please ask Matyi or Balázs about this?

**Overall, it looks good to me! I only have a few minor observations, which I've left in the comments.**

I found some slight differences between v11.6 and v11.7, but these seem to be specific to the v2 sub-components that DateTimeInput now uses:
- The calendar no longer opens when clicking the input and you can only navigate between days using the Tab key.
- The hover state lacks contrast, and disabled days are not visually distinct enough.
- The form field no longer gets a border when an invalid date is entered.

git-nandor (approved):

docs/guides/upgrade-guide.md
ToMESSKa: @git-nandor V12ChangelogTable is only useful for theme variable changes, but not prop and API changes, so I think I'll leave it as it is

packages/ui-date-time-input/src/DateTimeInput/v2/README.md
ToMESSKa: @git-nandor Thanks, I fixed these

ToMESSKa:
> The commit message doesn't include the breaking prop changes. Could you please ask Matyi or Balázs about this?
> 
> **Overall, it looks good to me! I only have a few minor observations, which I've left in the comments.**
> 
> I found some slight differences between v11.6 and v11.7, but these seem to be specific to the v2 sub-components that DateTimeInput now uses:
> 
> * The calendar no longer opens when clicking the input and you can only navigate between days using the Tab key.
> * The hover state lacks contrast, and disabled days are not visually distinct enough.
> * The form field no longer gets a border when an invalid date is entered.

@git-nandor the commit message includes the prop changes https://github.com/instructure/instructure-ui/pull/2518/changes/a84f23776d6f68d2a320929459ed70b50921deb9

- the first difference is due to different API between DateInput v1 and v2
- the second difference: @adamlobler can you please take a look at these when you review the component please?
- the third difference is due to different API between FormField v1 and v2

packages/ui-date-time-input/src/DateTimeInput/v2/index.tsx
```diff
@@ -0,0 +1,464 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+import { Component, SyntheticEvent } from 'react'
+import { Locale, DateTime, ApplyLocaleContext } from '@instructure/ui-i18n'
+import type { Moment } from '@instructure/ui-i18n'
```
balzss, line 27: we should not use moment at all

packages/ui-date-time-input/src/DateTimeInput/v2/index.tsx
```diff
@@ -0,0 +1,464 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+import { Component, SyntheticEvent } from 'react'
+import { Locale, DateTime, ApplyLocaleContext } from '@instructure/ui-i18n'
```
balzss, line 26: we should not use DateTime at all since it's just a moment wrapper

packages/ui-date-time-input/src/DateTimeInput/v2/index.tsx
```diff
@@ -0,0 +1,464 @@
...
+            nextMonthButton: nextMonthLabel,
+            datePickerDialog
+          }}
+          withYearPicker={withYearPicker}
+          value={this.state.dateInputText}
+          onChange={this.handleDateTextChange}
+          onRequestValidateDate={this.handleDateValidated}
+          onBlur={(e) => this.handleBlur(e)}
+          inputRef={dateInputRef}
+          placeholder={datePlaceholder ?? ''}
```
balzss: this should be just `placeholder={datePlaceholder}` because DateInput accepts null as a placeholder an uses it's internal placeholder

packages/ui-date-time-input/src/DateTimeInput/v2/index.tsx
```diff
@@ -0,0 +1,464 @@
...
+          isRequired={isRequired}
+          messages={subComponentMessages}
+          interaction={interaction}
+          locale={locale}
+          timezone={timezone}
+          disabledDates={disabledDates}
+          dateFormat={{
```
balzss: instead of providing this premade formatter and parser, let's allow the users to pass their own or just rely on the one provided by DateInput by default (so `dateFormat={props.dateFormat}`)

packages/ui-date-time-input/src/DateTimeInput/v2/props.ts
```diff
@@ -0,0 +1,307 @@
...
+ * SOFTWARE.
+ */
+
+import { SyntheticEvent } from 'react'
+import type { FormMessage } from '@instructure/ui-form-field/latest'
+import type { InteractionType } from '@instructure/ui-react-utils'
+import type { Moment } from '@instructure/ui-i18n'
+import type { Renderable } from '@instructure/shared-types'
+
+type DateTimeInputProps = {
```
balzss, line 31: we should make the api of DateTimeInput more similar to DateInput, e.g. instead of separate `prevMonthLabel` and `nextMonthLabel` props, just use `screenreaderlabel={...`

git-nandor (approved):

packages/ui-date-time-input/src/DateTimeInput/v2/props.ts
```diff
@@ -0,0 +1,307 @@
...
+  /**
+   * The format of the date shown in the `DateInput` when a date is selected.
+   * Valid formats are compatible with
+   * [Moment formats](https://momentjs.com/docs/#/displaying/format/),
+   * including localized formats.
+   *
+   * If omitted, it will use 'LL' which is a localized date with full month,
+   * e.g. "August 6, 2014"
+   **/
+  dateFormat?: string
```
balzss: this prop should be replaced with the dateFormat type accepted by DateInput

balzss (commented):
overall great work, left some comments about making the api more similar to DateInput

packages/ui-date-time-input/src/DateTimeInput/v2/index.tsx
ToMESSKa: Done.

packages/ui-date-time-input/src/DateTimeInput/v2/index.tsx
ToMESSKa: Done. This also changed the default `dateFormat` Moment's 'LL' (long month name) → locale's default date format, so I had to update the tests accordingly. (e. g. 'May 1, 2017' to '3/29/2022')

packages/ui-date-time-input/src/DateTimeInput/v2/props.ts
ToMESSKa: Done.

packages/ui-date-time-input/src/DateTimeInput/v2/props.ts
ToMESSKa: Done.

packages/ui-date-time-input/src/DateTimeInput/v2/index.tsx
```diff
@@ -0,0 +1,439 @@
...
+    if (this.props.timezone) {
+      return this.props.timezone
+    } else if (this.context && this.context.timezone) {
+      return this.context.timezone
+    }
+    return DateTime.browserTimeZone()
+  }
+
+  private formatDateInput(date: Date): string {
+    const { dateFormat } = this.props
```
ToMESSKa, line 211: I added this which is the same as the formatDate function in DateInput v2. DateTimeInput controls the DateInput `value` prop directly via state, so it needs to format the date itself.

packages/ui-date-time-input/src/DateTimeInput/v2/index.tsx
ToMESSKa: I looked at this and it would require some major rework and new functions to active the same functionalty. So we agreed to do this in another ticket.

packages/ui-date-time-input/src/DateTimeInput/v2/index.tsx
ToMESSKa: I looked at this and it would require some major rework and new functions to active the same functionalty. So we agreed to do this in another ticket.

adamlobler (changes requested):
I checked the components and other than the yearPicker issue I mentioned in the channel, I think it’s okay. I created a ticket for that earlier: https://instructure.atlassian.net/browse/INSTUI-5014

git-nandor (approved):

balzss (approved):

adamlobler (approved):

PR #2523, opened 2026-04-23
refactor(ui-text): add 'inherit' prop to Text

matyasf:
The v2 Text component uses the 'primary' color by default, v1 none (inherit). Add an 'inherit' value to v2 Text, so previous behaviour can be emulated.

To test: Check the updated docs that it looks correct


ToMESSKa (approved):

joyenjoyer (approved):
looks good

PR #2528, opened 2026-04-27
feat(ui-checkbox): add data-checked attribute to reflect checked state

ToMESSKa:
INSTUI-4902

**ISSUE:**
- the checked state of the Checkbox was only reflected as a DOM property, which MutationObserver-based tools can't observe and adding data-checked gives them an attribute to watch.

**TEST PLAN:**
- inspect the `<input> `element in Checkbox v11.7, the data-checked attribute value should reflect the checked state (true, false, mixed) and upgrade accordingly in the checkbox and toggle variant too
- the screenreader-emmitted messages should stay the same

### DIFF
--- a/packages/ui-checkbox/src/Checkbox/v2/index.tsx
+++ b/packages/ui-checkbox/src/Checkbox/v2/index.tsx
@@ -328,6 +328,7 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
       >
         <div css={styles?.container}>
           <input
+            data-checked={indeterminate ? 'mixed' : this.checked}
             {...props}
             id={this.id}
             value={value}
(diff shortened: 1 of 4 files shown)


joyenjoyer (approved):
good job, works well for me in Firefox, Chrome and Safari

packages/ui-checkbox/src/Checkbox/v2/index.tsx
```diff
@@ -328,6 +328,7 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
       >
         <div css={styles?.container}>
           <input
+            data-checked={indeterminate ? 'mixed' : this.checked}
```
matyasf, line 331: Can you please write in Checkboxes and CheckboxGroup's Readme about this?

matyasf (changes requested):
looks good, just documentation is missing

packages/ui-checkbox/src/Checkbox/v2/index.tsx
ToMESSKa: added these, check please

matyasf (approved):

PR #2529, opened 2026-04-27
Calendar accessibility fix

git-nandor:
INSTUI-5009
INSTUI-5010
INSTUI-5014

Improve date picker calendar accessibility

## Changes

- Add aria-live="polite" aria-atomic="true" to the month/year header in
  Calendar v1 and v2. Previously the container was a ```<span>``` wrapping ```<div>```
  children (invalid HTML), causing screen readers to lose track of the live
  region subtree and skip announcements on month navigation. Changed to ```<div>```.

- Add target month name to navigation button screen reader labels in Calendar
  v1 and v2. Default buttons now announce e.g. "Previous month, November 2023"
  instead of the generic "Previous month".

- Extend renderPrevMonthButton/renderNextMonthButton prop type from Renderable
  to ```Renderable<{ targetMonthSrLabel: string }>``` in Calendar v1 and v2. When
  passed as a function, the render prop receives the pre-formatted target month
  label so consumers can build descriptive accessible names for custom buttons.

- Update DateInput v2 to use function render props for navigation buttons,
  appending targetMonthSrLabel to the consumer-provided label string so the
  full announcement reads e.g. "Next month, January 2024".

- Add aria-modal to Dialog when role="dialog" and shouldContainFocus is true,
  scoping the screen reader virtual cursor to the dialog content. Guarded by
  shouldContainFocus so non-modal Dialog usages (Tray without containment,
  DrawerTray during transitions, etc.) are unaffected.
  
  ## Test plan
  

- [x]   Open a DateInput date picker with VoiceOver (macOS) or NVDA (Windows): verify the popover is announced as a dialog with its label (e.g. "Date picker, dialog")

- [x]  The calendar heading displaying the year and month should use an h2

- [x]  The last number in the yearPicker always fully visible

- [x]  With the calendar open, press Tab past the last focusable element and Shift+Tab before the first: verify focus stays inside the picker and does not escape

- [x]  Open a Modal, Tray, and Popover (non–date picker usage): verify aria-modal appears only when shouldContainFocus is true and does not appear on non-modal usages

- [x]  Navigate to the next/previous month using the arrow buttons: verify the screen reader announces the new month and year (e.g. "January 2024") via the aria-live region without requiring the user to move focus

- [x]  Check the previous and next month button accessible names: verify they include the target month (e.g. "Previous month, November 2023" / "Next month, January 2024") and not just the generic label

- [x]  Pass a function to renderPrevMonthButton / renderNextMonthButton on Calendar v2: verify targetMonthSrLabel is received 

```
  <Calendar
  renderPrevMonthButton={({ targetMonthSrLabel }) => (
    <IconButton
      size="small"
      renderIcon={<ChevronLeftInstUIIcon color="baseColor" />}
      screenReaderLabel={`Go to, ${targetMonthSrLabel}`}
    />
  )}
  renderNextMonthButton={({ targetMonthSrLabel }) => (
    <IconButton
      size="small"
      renderIcon={<ChevronRightInstUIIcon color="baseColor" />}
      screenReaderLabel={`Go to, ${targetMonthSrLabel}`}
    />
  )}
  selectedLabel="Selected"
 />
```

 

### DIFF
--- a/packages/ui-calendar/src/Calendar/v1/props.ts
+++ b/packages/ui-calendar/src/Calendar/v1/props.ts
@@ -98,19 +98,23 @@ type CalendarOwnProps = {
    */
   renderNavigationLabel?: Renderable
   /**
-   * A button to render in the navigation header. The recommendation is to
-   * compose it with the [IconButton](IconButton) component by setting the `size`
-   * prop to `small`, `withBorder` and `withBackground` to `false`, and setting
-   * `renderIcon` to [IconArrowOpenEnd](icons).
+   * A button to render in the navigation header.
+   * When passed as a function, receives `{ targetMonthSrLabel }` —
+   * a pre-formatted screen reader label for the target month (e.g. "November 2023").
+   * The recommendation is to compose it with the [IconButton](Button) component
+   * by setting the `size` prop to `small`, `withBorder` and `withBackground` to `false`,
+   * and setting `renderIcon` to [IconArrowOpenStart](icons).
    */
-  renderNextMonthButton?: Renderable
+  renderNextMonthButton?: Renderable<{ targetMonthSrLabel: string }>
   /**
-   * A button to render in the navigation header. The recommendation is to
-   * compose it with the [IconButton](Button) component by setting the `size`
-   * prop to `small`, `withBorder` and `withBackground` to `false`, and setting
-   * `renderIcon` to [IconArrowOpenStart](icons).
+   * A button to render in the navigation header.
+   * When passed as a function, receives `{ targetMonthSrLabel }` —
+   * a pre-formatted screen reader label for the target month (e.g. "November 2023").
+   * The recommendation is to compose it with the [IconButton](Button) component
+   * by setting the `size` prop to `small`, `withBorder` and `withBackground` to `false`,
+   * and setting `renderIcon` to [IconArrowOpenStart](icons).
    */
-  renderPrevMonthButton?: Renderable
+  renderPrevMonthButton?: Renderable<{ targetMonthSrLabel: string }>
   /**
    * An array of labels containing the name of each day of the week. The visible
    * portion of the label should be abbreviated (no longer than three characters).
--- a/packages/ui-calendar/src/Calendar/v2/index.tsx
+++ b/packages/ui-calendar/src/Calendar/v2/index.tsx
@@ -174,28 +174,41 @@ class Calendar extends Component<CalendarProps, CalendarState> {
 
   renderMonthNavigationButtons = () => {
     const { renderNextMonthButton, renderPrevMonthButton } = this.props
+    const { visibleMonth } = this.state
+    const prevMonthName = visibleMonth
+      .clone()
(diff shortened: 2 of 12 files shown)


packages/ui-calendar/src/Calendar/v1/props.ts
```diff
@@ -103,14 +103,14 @@ type CalendarOwnProps = {
...
    * `renderIcon` to [IconArrowOpenEnd](icons).
    */
-  renderNextMonthButton?: Renderable
+  renderNextMonthButton?: Renderable<{ targetMonthSrLabel: string }>
   /**
    * A button to render in the navigation header. The recommendation is to
    * compose it with the [IconButton](Button) component by setting the `size`
    * prop to `small`, `withBorder` and `withBackground` to `false`, and setting
    * `renderIcon` to [IconArrowOpenStart](icons).
    */
```
ToMESSKa: Can you please update the prop descriptions here too like in v2?

ToMESSKa (changes requested):
Really nice job. See my comment.

packages/ui-calendar/src/Calendar/v1/props.ts
git-nandor: Updated

ToMESSKa (approved):

packages/ui-calendar/src/Calendar/v2/index.tsx
```diff
@@ -298,12 +311,12 @@ class Calendar extends Component<CalendarProps, CalendarState> {
...
-            <span>
-              <div>{visibleMonth.format('MMMM')}</div>
+            <div aria-live="polite" aria-atomic="true">
+              <h2 css={styles?.navigationLabel}>{visibleMonth.format('MMMM')}</h2>
               {!withYearPicker ? (
-                <div>{visibleMonth.format('YYYY')}</div>
+                <h2 css={styles?.navigationLabel}>{visibleMonth.format('YYYY')}</h2>
               ) : null}
-            </span>
+            </div>
```
balzss: Two sibling `<h2>`s split a single conceptual heading. Screen readers announce them as two separate headings ("December, heading level 2. 2023, heading level 2.") and both show up in heading navigation, instead of one combined "December 2023". `aria-atomic` doesn't merge them — the heading boundaries still apply.

Suggest collapsing into a single `<h2>` (which can also host the live-region attributes):

```suggestion
            <h2
              aria-live="polite"
              aria-atomic="true"
              css={styles?.navigationLabel}
            >
              <span>{visibleMonth.format('MMMM')}</span>
              {!withYearPicker ? (
                <span>{visibleMonth.format('YYYY')}</span>
              ) : null}
            </h2>
```

If the existing two-line visual stacking matters, give the inner `<span>`s `display: block` in `navigationLabel` styles. One heading, one live region, one announcement.

git-nandor: Thanks! Modified

balzss (approved):
looks good now, great work!

PR #2537, opened 2026-05-05
refactor: update regression test dependencies

matyasf:
Also use ESLint from the root package.

To check:
- run the lint command in the regression test
- run the regression test locally

### DIFF
--- a/packages/__docs__/buildScripts/samplemedia/placeholder-image.js
+++ b/packages/__docs__/buildScripts/samplemedia/placeholder-image.js
@@ -26,7 +26,6 @@ import IconSVG from './placeholder.svg'
 
 export default function placeholderImage(width = 512, height = 512) {
   // We need to base64 encode this because otherwise FF will add extra escape chars
-  // eslint-disable-next-line no-undef
   const dataUri = Buffer.from(
     IconSVG.replace(/{{w}}/g, width).replace(/{{h}}/g, height).trim()
   ).toString('base64')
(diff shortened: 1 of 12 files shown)


joyenjoyer (changes requested):
I ran the lint command locally. It is a small project but the running time was very long, almost a minute. It found a surprisingly high number of issues:

✖ 370340 problems (286751 errors, 83589 warnings)
  271514 errors and 0 warnings potentially fixable with the `--fix` option.

It goes into every folder, even into build artifacts and next js runtime built files, I would ignore those, please look into it. Cypress command was ok.

packages/__docs__/buildScripts/samplemedia/placeholder-image.js
```diff
@@ -26,7 +26,6 @@ import IconSVG from './placeholder.svg'
 
 export default function placeholderImage(width = 512, height = 512) {
   // We need to base64 encode this because otherwise FF will add extra escape chars
-  // eslint-disable-next-line no-undef
```
matyasf, line 29: There were some random files that havent been linted for a while and our linter config changes, so they were showing errors. This is fixed now

matyasf:
> I ran the lint command locally. It is a small project but the running time was very long, almost a minute. It found a surprisingly high number of issues

I've fixed this, the `.next` folder was not ignored

joyenjoyer (approved):
looks ok, eslint runs well

balzss (approved):

PR #2540, opened 2026-05-05
chore(ui-utils): remove unused hash utility

HerrTopi:
## Summary
- `hash()` exported from `@instructure/ui-utils` has no internal callers — the only importer in the repo is its own test
- Remove `packages/ui-utils/src/hash.ts`, its test, the re-export from `src/index.ts`, and the `json-stable-stringify` dependency that existed solely to back it

## Test plan
- [ ] `pnpm install` resolves cleanly
- [ ] `pnpm run build:types` succeeds
- [ ] `pnpm run test:vitest ui-utils` passes
- [ ] Full `pnpm run test:vitest` passes


packages/ui-utils/src/index.ts
```diff
@@ -25,7 +25,6 @@ export { capitalizeFirstLetter } from './capitalizeFirstLetter'
 export { cloneArray } from './cloneArray'
 export { createChainedFunction } from './createChainedFunction'
 export { deepEqual } from './deepEqual'
-export { hash } from './hash'
```
matyasf, line 28: This might be used by others, please do not remove it!

matyasf (changes requested):

PR #2542, opened 2026-05-06
fix(ui-form-field): disable child inputs when FormFieldGroup is disabled

ToMESSKa:
INSTUI-4841

**ISSUE:**

- FormFieldGroup's disabled prop renders disabled styles on the group container but doesn't actually disable the child inputs — they remain interactive

**TEST PLAN:**

- test the examples on the FormFieldGroup page by adding them them disabled prop
- every child should be visually and functionally disabled (no focus or interaction is avaliable)

### DIFF
--- a/packages/ui-form-field/src/FormFieldGroup/v2/index.tsx
+++ b/packages/ui-form-field/src/FormFieldGroup/v2/index.tsx
@@ -22,10 +22,20 @@
  * SOFTWARE.
  */
 
-import { Component, Children, ReactElement, AriaAttributes } from 'react'
+import {
+  Component,
+  Children,
+  ReactElement,
+  AriaAttributes,
+  isValidElement
+} from 'react'
 
 import { Grid } from '@instructure/ui-grid/latest'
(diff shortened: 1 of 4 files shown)


packages/ui-form-field/src/FormFieldGroup/v2/index.tsx
```diff
@@ -86,20 +96,22 @@ class FormFieldGroup extends Component<FormFieldGroupProps> {
   }
 
   renderColumns() {
+    const { disabled } = this.props
```
ToMESSKa, line 99: Currently this fix:
- if a child has an explicit disabled={false} prop, it always gets overridden and the group always wins
- it is applied to every first-level child and not just InstUI components, so a `<div>some helper text</div> ` becomes `<div disabled>some helper text</div>` too
- also I did not marked this as breaking change because looks more like a bug fix

If one of this is not supposed to be like that, let me know

matyasf (approved):
looks good!

HerrTopi (approved):
The description isn't disabled. Please check with design if it should be or not. If not, consider it approved if yes, please re-ask for my review

PR #2543, opened 2026-05-06
refactor(many): extract raw svgs for cross-package use

HouseOfHawks:
This PR:
- Extracts raw SVGs into a separate `ui-icons-svg` package, free of framework dependencies, for use in other icon-generating packages,
- adds a new bucket in `ui-icons-svg/svg/Parchment` for all `Graphics`, `Icons` and `Vectors` used in Parchment, Digitary, Scribbles and Quottly applications,
- maintains backwards compatibility for Lucide/React icons in `ui-icons` while making new Parchment icons available in this React package (both legacy and beta), and
- updates the `__docs__` app so that consumers can now filter icons by organization type and asset type when viewing Parchment assets (demo below)

[Tech-Spec Outlining the Justification For This Work](https://docs.google.com/document/d/1DgZqjNhEA-shbT_3ScNd5k31sDxiOoUHadYt5QOW8hw)

Demo:

https://github.com/user-attachments/assets/14e2e073-e5f2-4bf6-885b-1e897865e1ca

Next iterations (ongoing work):

Currently there is significant asset overlap. Design teams will need to work through both asset libraries and unify that overlap (ie, InstUI's `x.svg` and Parchment's `close.svg`). Parchment application developers will need access to this full set of assets initially as teams begin implementing InstUI, but as re-design work becomes available those teams will be pointed to the InstUI icons where applicable. In other cases where assets are Parchment-specific or there is no InstUI counterpart, those Parchment assets will remain.


matyasf:
Can you please add a postfix to the new icons? e.g. `AccessibilityParchment`

packages/__docs__/src/Icons/IconsGallery.tsx
```diff
@@ -22,7 +22,14 @@
  * SOFTWARE.
  */
 
-import { useState, memo, useCallback, useMemo, useRef } from 'react'
```
matyasf, line 25: As we have discussed, isnt this a bad place for Parchment icons? Parchment devs will look for InstUI components in Storybook, and not visit this site? Also it might be confusing to them, why are we here showing only React icons, why not Angular/WebComponents/Svelte/...?

And its confusing for Instructure users too, why show here icons, that they will never use? I would not place Parchment icons into the Instructure-side InstUI documentation

packages/ui-icons/src/index.ts
```diff
@@ -32,9 +32,17 @@ export * from './generated/lucide'
 // Custom/brand icons
 export * from './generated/custom'
 
+// Parchment categories: themable icons, multi-color graphics, multi-color vectors
+export * from './generated/parchment/icons'
+export * from './generated/parchment/graphics'
+export * from './generated/parchment/vectors'
```
matyasf, line 38: Please suffix all icons, so its easily distinguishable on the code side what kind of icons are these e.g. `AddDocumentLineParchment`

matyasf (changes requested):
see my comments

packages/__docs__/src/Icons/IconsGallery.tsx
HouseOfHawks: Instructure devs will be coming here to look for icons in their Figma designs, which may include Parchment vectors/graphics.

We will point Parchment devs to the upcoming Storybook, which will show the new Instructure icons (new to them) as well as the pre-existing vectors and graphics made available from Parchment - and temporarily have some icon overlap until we dedupe/prune the list to one that is unified.

Remember, the vectors/graphics are product branded and likely to be used in Instructure apps as we integrate our products more - something we are already doing with two Instructure teams.

Since this page is only for Instructure devs, we are only showing React icons. If you recall, the expectation is not to build Angular/WebComponents/Svelte for the Instructure stakeholders as that's not a technical requirement there. That will be handled for Parchment stakeholders only and it will live in the upcoming Storybook page where we will direct Parchment stakeholders.

packages/ui-icons/src/index.ts
HouseOfHawks: On vectors/graphics, too? We had discussed adding the suffix on the icons, since there's overlap, but there's no overlap in vectors/graphics.

We can make this adjustment - I only hesitate because this makes some of these names very long and I'm just being empathetic to the dev experience.

matyasf: hmm I understand. In this case I would still add some prefix/suffix to these too, now the `ui-icons` package exports 2000+ graphics, which is super hard to navigate:

 - `*InstUIIcon` -- new InstUI icons (1900+)
 - `*ParchIcon` -- Parchment icons
 - `[no logic]` -- Parchment graphics (100+)
 - `[no logic]` -- Parchment vectors (50+)
 - `Icon*` -- Legacy, deprecated InstUI icons (must have this name for backwards compat, will be deleted in the future)

I think a prefix or a suffix would help all devs to navigate this huge number of exports.

packages/__docs__/src/Icons/IconsGallery.tsx
matyasf: hmm OK, I guess its also a better DX because we dont hide some icons, that are accessible via imports.

packages/ui-icons/src/index.ts
HouseOfHawks: Sounds good!

HouseOfHawks: Thinking through the DX - maybe pivoting on the plan by adding a _prefix_ is a better experience; and also safer via `import`. When an app dev is intentionally looking for a Parchment Icon and is typing the import, starting with `Parch...`, it's a very clear direction to Parchment icons/vectors/graphics and most (if not all) InstUIIcons get filtered - save potential edge cases. In the other direction as an Instructure dev is importing the `<AccessibilityInstUIIcon>`and has typed `Acc...`, all Parchment icons/vectors/graphics are filtered in the IDE auto-pop.

HouseOfHawks: Looks like I need the semantics deep merge fix - rebased on pushed anyways, but build is expected to fail until we have [#2561](https://github.com/instructure/instructure-ui/pull/2561)

HouseOfHawks: Odd that the build didn't fail. 🤷

PR #2544, opened 2026-05-06
feat(ui-range-input): remove deprecated thumb variant

ToMESSKa:
BREAKING CHANGE: `thumbVariant` prop removed from RangeInput

INSTUI-5016

**ISSUE:**
- deprecated thumb variant should be removed from RangeInput

**TEST PLAN:**
- check the example in the documentation it should work and have no errors


balzss (approved):
looks good, works great

matyasf (approved):

PR #2552, opened 2026-05-12
feat(ui-position): add available space to position

joyenjoyer:
Calculates and exposes popover's available room (gap between the trigger and the resolved constraint edge) as `--ui-position-available-height` / `--ui-position-available-width` CSS variables on the content node so consumers can cap their height/width with `max-height: var(--ui-position-available-height)`. This is needed to make Popover and all other components that use Popover scrollable.

**Test:** go to the Popover docs page and observe the DOM of a Popover during scrolling, value of ui-position-available-height and --ui-position-available-width CSS vars should change.

### DIFF
--- a/packages/ui-position/src/Position/__tests__/Position.test.tsx
+++ b/packages/ui-position/src/Position/__tests__/Position.test.tsx
@@ -583,4 +583,25 @@ describe('<Position />', () => {
       expect(style.display).toEqual('block')
     })
   })
+
+  it('writes --ui-position-available-{height,width} on the content node', async () => {
+    render(
+      <Position
+        constrain="window"
+        placement="bottom"
+        renderTarget={<button data-testid="target">Target</button>}
+      >
+        <div data-testid="content">Content</div>
+      </Position>
+    )
+    const content = screen.getByTestId('content')
+    await waitFor(() => {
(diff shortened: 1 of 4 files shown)


ToMESSKa (approved):

packages/ui-position/src/Position/__tests__/Position.test.tsx
```diff
@@ -583,4 +583,25 @@ describe('<Position />', () => {
...
+      </Position>
+    )
+    const content = screen.getByTestId('content')
+    await waitFor(() => {
+      expect(
+        content.style.getPropertyValue('--ui-position-available-height')
+      ).not.toBe('')
+    })
+    expect(
+      content.style.getPropertyValue('--ui-position-available-width')
```
git-nandor: This could also go inside the waitFor to make the test more stable.

git-nandor (changes requested):

packages/ui-position/src/Position/__tests__/Position.test.tsx
joyenjoyer: It is inside waitFor now.

git-nandor (approved):

PR #2554, opened 2026-05-12
build(ui-scripts): add prerelease as security release version type: pnpm run bump 

HerrTopi:
## Summary
- Adds `prerelease` as a supported value for `ui-scripts bump` so security releases can be cut via `pnpm run bump --releaseType=prerelease`, producing versions tagged `x.y.z-SECURITY.N` via lerna's `--preid=SECURITY`.
- Skips writing versioned `exports` entries to package.json files when the release is a SECURITY prerelease, since those entries are reserved for real minor/major releases.

## Test plan
- [ ] Run `pnpm run bump --releaseType=prerelease` locally and verify packages bump to `x.y.z-SECURITY.0`
- [ ] Verify no new versioned `exports` keys are added to package.json files during the prerelease bump
- [ ] Run `pnpm run bump --releaseType=patch` (or `minor`) and confirm existing behavior is unchanged (normal bump + exports entries added)


matyasf (changes requested):
Works well, but please document this new param type in `bump.js` -> `releaseType` and in the handler. Maybe this is worth adding to the root scripts of `package.json`?

PR #2555, opened 2026-05-12
feat(ui-scripts): add publish-private command for private-registry releases

balzss:
## Summary

Adds `pnpm run publish-private` for publishing all non-private packages to an npm-compatible private registry. Used to ship pre-release builds of security fixes to selected consumers ahead of the public release.

Reads `INSTUI_PRIVATE_REGISTRY` and `INSTUI_PRIVATE_REGISTRY_TOKEN` from the environment. The existing public release path (`pnpm run release`, OIDC, GitHub workflows) is untouched.

## Why a separate command

`pnpm run release` is invoked on master push and OIDC-publishes to npmjs. Overloading it would mean a mis-set env var could leak embargoed code to the public registry. A distinct command isolates the private path and its guard rails.

## Safety

- Refuses any host matching `*.npmjs.org`.
- Exits 1 if either env var is missing.
- Operator must type the registry hostname to confirm (skippable with `--yes`).
- Writes a temp `.npmrc` for the publish subprocesses only — user/project `.npmrc` untouched, cleaned up in `finally`.
- Publishes under dist-tag `security` so `pnpm install` won't resolve a private build by accident.
- Skips packages whose version already exists on the target registry.
- 500ms delay between publishes to stay under registry rate limits.

## Usage

```sh
export INSTUI_PRIVATE_REGISTRY=https://<host>/<path>/
export INSTUI_PRIVATE_REGISTRY_TOKEN=<token>
pnpm run publish-private
```

## What this PR does NOT do

- No version bumping — handled separately.
- No changes to the public release flow, OIDC auth, GitHub workflows, or `publish.js`.
- No documentation update yet — coming once the surrounding bump-script work lands and the end-to-end flow is settled.

## Test Plan

- [x] Smoke-tested against a real private registry — auth, tag, and versions all land correctly.
- [ ] `pnpm run publish-private --help` shows the `--yes` flag.
- [ ] Missing env vars and npmjs hostname both exit 1.

Draft until the version-bump PR lands and the end-to-end flow can be walked through.

### DIFF
--- a/packages/ui-scripts/lib/commands/publish-private.js
+++ b/packages/ui-scripts/lib/commands/publish-private.js
@@ -1,0 +1,199 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015 - present Instructure, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+import os from 'node:os'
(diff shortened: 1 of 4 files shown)


packages/ui-scripts/lib/commands/publish-private.js
```diff
@@ -0,0 +1,194 @@
...
+    [
+      'publish',
+      pkg.location,
+      '--tag',
+      PRIVATE_TAG,
+      '--no-git-checks'
+      // No --provenance: only npmjs supports it.
+      // No --registry flag: the temp .npmrc handles registry + auth.
+    ],
+    childEnv
```
matyasf, line 197: I would include a small delay between calling this. We ran into very inconvenient errors when calling this 80+ times without delay, the command started failing after a while (npmjs's DDOS protection?).

I suggest to add a small delay here too (we were using 500ms)

packages/ui-scripts/lib/commands/publish-private.js
```diff
@@ -0,0 +1,194 @@
...
+  await runCommandAsync(
+    'pnpm',
+    [
+      'publish',
+      pkg.location,
+      '--tag',
+      PRIVATE_TAG,
+      '--no-git-checks'
+      // No --provenance: only npmjs supports it.
+      // No --registry flag: the temp .npmrc handles registry + auth.
```
matyasf: I dont see this flag in the docs https://pnpm.io/cli/publish

matyasf (changes requested):

packages/ui-scripts/lib/commands/publish-private.js
balzss: good call, adding it

packages/ui-scripts/lib/commands/publish-private.js
balzss: it's there 
image

matyasf: I'm talking about `--registry`

matyasf (approved):

HerrTopi (approved):
Seems good to me, giving thumbs up with cautious optimism, hope everything will work properly

PR #2557, opened 2026-05-12
refactor: unify deploy base path handling

joyenjoyer:
This change is needed to be able to host PR previews on Github pages of `instructure-design-tokens` repo.

**Unify docs-app deploy base path handling**

  Make `PUBLIC_PATH` (env var → webpack `output.publicPath`) the single source of truth for "where this build lives" and thread it through every layer:

  - **`webpack.config.mjs`** — resolve `PUBLIC_PATH` once and pass it to both `output.publicPath` and `HtmlWebpackPlugin`'s `templateParameters`.
  - **`index.html`** — `<base href>` and the inline 404 SPA-redirect script both read the template parameter; the redirect is now generic instead of pattern-matching `pr-preview`/`latest`.
  - **`navigationUtils.ts`** — new `getDeployBase()` reads `__webpack_public_path__` at runtime. `parseCurrentUrl` strips it before parsing the in-app route; `getAssetBasePath` returns it directly. Renamed
  `prPrefix` → `basePrefix` and de-duplicated a double `parseCurrentUrl()` call in `buildUrl`.

  No behavior change for current `instructure.design` deploys. Unlocks hosting the docs-app at any base path — other repos (e.g. design-tokens previews) or arbitrary sub-hosts — by setting `PUBLIC_PATH` at
  build time.

**Test plan**: try if everything works well on dev server locally, open docs page also check the pr preview here in the PR.

### DIFF
--- a/packages/__docs__/src/index.html
+++ b/packages/__docs__/src/index.html
@@ -1,20 +1,20 @@
 <!DOCTYPE html>
 <html lang="en" dir="ltr">
   <head>
-    <base href="/" />
-    <!-- GitHub Pages SPA redirect for subdirectory deployments.
-         When serving this page as 404.html, redirect deep links under
-         /pr-preview/pr-N/ or /latest/ back to the subdirectory root with
-         the route encoded as a query parameter. No-op for root paths. -->
+    <base href="<%= PUBLIC_PATH %>" />
+    <!-- When GH Pages serves this as 404.html, redirect the deep link to the
+         deploy base -->
     <script>
       ;(function () {
+        var base = '<%= PUBLIC_PATH %>'
         var l = window.location
-        var m = l.pathname.match(/^(\/pr-preview\/pr-\d+|\/latest)(\/.*)$/)
-        if (m && m[2] !== '/') {
-          l.replace(
-            m[1] + '/?__spa_route=' + encodeURIComponent(m[2] + l.hash)
-          )
-        }
+        if (l.pathname === base) return
+        if (l.pathname + '/' === base) return
+        if (l.pathname.indexOf(base) !== 0) return
+        var rest = l.pathname.slice(base.length)
+        l.replace(
+          base + '?__spa_route=' + encodeURIComponent('/' + rest + l.hash)
+        )
       })()
     </script>
     <meta charset="utf-8" />
--- a/packages/__docs__/src/index.html
+++ b/packages/__docs__/src/index.html
@@ -6,14 +6,15 @@
          deploy base -->
     <script>
       ;(function () {
-        var base = '<%= PUBLIC_PATH %>'
-        var l = window.location
-        if (l.pathname === base) return
-        if (l.pathname + '/' === base) return
-        if (l.pathname.indexOf(base) !== 0) return
-        var rest = l.pathname.slice(base.length)
-        l.replace(
-          base + '?__spa_route=' + encodeURIComponent('/' + rest + l.hash)
+        const base = '<%= PUBLIC_PATH %>'
+        const { pathname, search, hash } = window.location
+        if (pathname === base || pathname + '/' === base) return
+        if (!pathname.startsWith(base)) return
+        const rest = pathname.slice(base.length)
+        window.location.replace(
+          base +
+            '?__spa_route=' +
+            encodeURIComponent('/' + rest + search + hash)
         )
       })()
     </script>
--- a/packages/__docs__/src/navigationUtils.ts
+++ b/packages/__docs__/src/navigationUtils.ts
@@ -29,16 +29,14 @@ declare const __webpack_public_path__: string
 const MINOR_VERSION_REGEX = /^v\d+_\d+$/
 
 type ParsedUrl = {
-  basePrefix: string
   minorVersion: string | null
   page: string
   sectionId: string | undefined
 }
 
 /**
- * Returns '' for a root deploy, '/latest' on instructure.design,
- * '/pr-preview/pr-<n>' for PR previews. For sub-host deployments it can be
- * '/<repo>/pr-preview/pr-<n>'.
+ * Returns webpack's `output.publicPath` with the trailing slash stripped.
+ * Always begins with `/`, or is empty for a root deploy.
  */
 function getDeployBase(): string {
   if (typeof __webpack_public_path__ === 'string' && __webpack_public_path__) {
@@ -52,20 +50,17 @@ function parseCurrentUrl(): ParsedUrl {
 
   const deployBase = getDeployBase()
   let rest = pathname
-  if (deployBase && rest.startsWith(deployBase)) {
+  if (
+    deployBase &&
+    (rest === deployBase || rest.startsWith(deployBase + '/'))
+  ) {
     rest = rest.slice(deployBase.length)
   }
 
   const cleanPath = rest.replace(/^\/+|\/+$/g, '')
   const segments = cleanPath.split('/').filter(Boolean)
 
-  let appPrefix = ''
   let idx = 0
-
-  if (segments[idx] === 'latest') {
-    appPrefix = '/latest'
-    idx++
-  }
 
   // Detect minor version prefix: /v11_7
   let minorVersion: string | null = null
@@ -83,7 +78,7 @@ function parseCurrentUrl(): ParsedUrl {
     sectionId = decodeURI(hash.replace(/^#+/, ''))
   }
 
-  return { basePrefix: deployBase + appPrefix, minorVersion, page, sectionId }
+  return { minorVersion, page, sectionId }
 }
 
 type BuildUrlOptions = {
@@ -98,7 +93,7 @@ function buildUrl(targetPage: string, options?: BuildUrlOptions): string {
       ? options.minorVersion
       : parsed.minorVersion
 
-  let url = parsed.basePrefix
+  let url = getDeployBase()
 
   if (minorVersion) {
     url += `/${minorVersion}`
@@ -130,20 +125,12 @@ function navigateToVersion(version: string | null): void {
   window.dispatchEvent(new PopStateEvent('popstate'))
 }
 
-/**
- * Returns the base path prefix for fetching static assets.
- * On PR previews this is e.g. `/pr-preview/pr-2425`, otherwise empty string.
- */
-function getAssetBasePath(): string {
-  return getDeployBase()
-}
-
 export {
   parseCurrentUrl,
   buildUrl,
   navigateTo,
   navigateToVersion,
-  getAssetBasePath,
+  getDeployBase,
   MINOR_VERSION_REGEX
 }
 export type { ParsedUrl }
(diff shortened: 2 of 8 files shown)


packages/__docs__/webpack.config.mjs
```diff
@@ -32,7 +32,14 @@ const ENV = process.env.NODE_ENV || 'production'
...
 const PR_NUMBER = process.env.PR_NUMBER
-const PUBLIC_PATH = process.env.PUBLIC_PATH
+// The URL prefix this build is deployed under. Must end with a trailing
+// slash. Fed into output.publicPath (which webpack exposes at runtime as
+// __webpack_public_path__) and into the HTML template's <base href> and
+// 404 SPA-redirect script. Keeping a single source here means a new deploy
+// target only needs to set this one env var.
+const PUBLIC_PATH =
+  process.env.PUBLIC_PATH ||
+  (PR_NUMBER ? `/pr-preview/pr-${PR_NUMBER}/` : '/')
```
balzss: `PUBLIC_PATH` has an implicit contract ("path with leading and trailing slashes") that's only documented in the comment above, and the whole point of this PR is to let external callers set it. Suggest validating loudly at build start so a misconfigured CI workflow fails the build instead of silently producing broken `<base href>` and asset URLs:

```js
const PUBLIC_PATH =
  process.env.PUBLIC_PATH ||
  (PR_NUMBER ? `/pr-preview/pr-${PR_NUMBER}/` : '/')

if (!/^\/([^?#]*\/)?$/.test(PUBLIC_PATH)) {
  throw new Error(
    `PUBLIC_PATH must be a path with leading and trailing slashes (got: "${PUBLIC_PATH}"). ` +
    `Examples: "/", "/latest/", "/repo/pr-preview/pr-123/"`
  )
}
```

Also see the top-level review note about dropping the `PR_NUMBER` fallback entirely and making every deploy job set `PUBLIC_PATH` explicitly.

packages/__docs__/src/navigationUtils.ts
```diff
@@ -22,36 +22,48 @@
...
   minorVersion: string | null
   page: string
   sectionId: string | undefined
 }
 
+/**
+ * Returns '' for a root deploy, '/latest' on instructure.design,
+ * '/pr-preview/pr-<n>' for PR previews. For sub-host deployments it can be
+ * '/<repo>/pr-preview/pr-<n>'.
+ */
```
balzss, line 40: This JSDoc bakes specific deploy targets into a generic helper's docs — the same coupling the PR is trying to remove from the code. Next time a deploy is added, this comment goes stale silently.

Suggest describing behavior, not environments:

```ts
/**
 * Returns webpack's `output.publicPath` with the trailing slash stripped.
 * Always begins with `/`, or is empty for a root deploy.
 */
```

packages/__docs__/src/navigationUtils.ts
```diff
@@ -22,36 +22,48 @@
...
+  return ''
+}
+
 function parseCurrentUrl(): ParsedUrl {
   const { pathname, hash } = window.location
-  const cleanPath = pathname.replace(/^\/+|\/+$/g, '')
+
+  const deployBase = getDeployBase()
+  let rest = pathname
+  if (deployBase && rest.startsWith(deployBase)) {
```
balzss: `deployBase` has its trailing slash stripped on line 45, so `'/latestX/foo'.startsWith('/latest')` is true and `rest` becomes `'X/foo'`. Mostly theoretical (you can't be at `/latestX/...` while served by a `/latest/` build), but it's asymmetric with the redirect script in `index.html` which matches on the trailing-slash-included form. Tightening this also documents the invariant:

```ts
if (deployBase && (rest === deployBase || rest.startsWith(deployBase + '/'))) {
  rest = rest.slice(deployBase.length)
}
```

packages/__docs__/src/navigationUtils.ts
```diff
@@ -22,36 +22,48 @@
...
-    segments[1].startsWith('pr-')
-  ) {
-    prPrefix = `/${segments[0]}/${segments[1]}`
-    idx = 2
-  }
-
-  // Detect /latest/ prefix
-  if (idx === 0 && segments[idx] === 'latest') {
-    prPrefix = '/latest'
+  if (segments[idx] === 'latest') {
```
balzss: **Dead code + redundant field.** Walk through any real deploy:

- `PUBLIC_PATH=/latest/` → `deployBase='/latest'`, the prefix is already stripped before segment parsing, so this branch can never fire.
- `PUBLIC_PATH=/pr-preview/pr-N/` → same shape.
- `PUBLIC_PATH=/` → `deployBase=''`, `segments[0]` is the page, branch doesn't fire.

In every production deploy `appPrefix` is `''`, so `basePrefix` (line 86) is identical to `deployBase` — and `deployBase` is what `getAssetBasePath()` already returns. The `/latest`-as-app-route concept is a leftover from the old architecture where `/latest/` was an app-internal prefix rather than a deploy target.

If you delete this branch and the `appPrefix` accumulation in the return on line 86, then:
- `basePrefix` can be dropped from `ParsedUrl` (line 32)
- `buildUrl` reads `getDeployBase()` directly instead of `parsed.basePrefix`
- `getAssetBasePath` collapses with `getDeployBase` (see comment on line 137)

Net result: a meaningfully smaller file with no special knowledge of any one deploy path — which is what this PR is actually trying to achieve.

packages/__docs__/src/navigationUtils.ts
```diff
@@ -123,7 +135,7 @@ function navigateToVersion(version: string | null): void {
  * On PR previews this is e.g. `/pr-preview/pr-2425`, otherwise empty string.
  */
 function getAssetBasePath(): string {
-  return parseCurrentUrl().prPrefix
+  return getDeployBase()
 }
```
balzss: Two issues here.

**1. Stale JSDoc.** "Otherwise empty string" was true in the old URL-derived implementation; the new build-pinned version returns `/latest` for the latest deploy, `/<repo>/pr-preview/pr-N` for sub-host deploys, etc. The doc directly contradicts the JSDoc on `getDeployBase` two functions up (which returns the same value).

**2. The function is a one-line wrapper.** It just forwards to `getDeployBase()` with no transformation, and `getDeployBase` is module-private. Two names for the identical value, with the public name (`AssetBasePath`) now being a slight misdirection — `__webpack_public_path__` is literally the deploy base; assets just happen to live there.

Suggest: export `getDeployBase` directly, delete `getAssetBasePath`, and update the two callers in `App/index.tsx` and `versionData.ts`. One concept, one name.

packages/__docs__/src/index.html
```diff
@@ -1,20 +1,20 @@
...
-          )
-        }
+        if (l.pathname === base) return
+        if (l.pathname + '/' === base) return
+        if (l.pathname.indexOf(base) !== 0) return
+        var rest = l.pathname.slice(base.length)
+        l.replace(
+          base + '?__spa_route=' + encodeURIComponent('/' + rest + l.hash)
+        )
       })()
```
balzss, line 19: Two small things on this inline script:

**Drops the query string.** `rest` comes from `pathname` only — `l.search` is never appended. A 404 on `<base>Button?utm_source=x` redirects to `<base>?__spa_route=/Button` and silently loses `?utm_source=x`. Same problem existed in the old regex-based version, but this PR is the moment to fix it: include `l.search` in the encoded payload and have `src/index.tsx` re-apply it after `replaceState`.

**ES5 patterns for no reason.** Inline `<script>` tags aren't bundled and don't need to be ES5 in 2026. `var l = window.location` saves one keystroke at the cost of readability; `.indexOf(base) !== 0` is the old workaround for `.startsWith`. A modern rewrite is shorter and easier to read:

```js
;(function () {
  const base = '<%= PUBLIC_PATH %>'
  const { pathname, search, hash } = window.location
  // GH Pages may serve `/foo` for a request to `/foo/` directory
  if (pathname === base || pathname + '/' === base) return
  if (!pathname.startsWith(base)) return
  const rest = pathname.slice(base.length)
  window.location.replace(
    base + '?__spa_route=' + encodeURIComponent('/' + rest + search + hash)
  )
})()
```

balzss (commented):
Nice direction — making `PUBLIC_PATH` the single source of truth is the right move and unlocks the cross-repo preview use case. Most of my notes are about following the same idea a bit further: there's still some scaffolding in the code that only makes sense in the old URL-driven world.

### Architectural notes (not inline because they span the PR)

**`PUBLIC_PATH` should be required, not fallback-derived.** Right now `webpack.config.mjs` falls back to `PR_NUMBER`-based path, then `/`. Combined with the un-validated env var, a deploy job that forgets to set `PUBLIC_PATH` will silently build with the wrong base — the assets will request `/main.js` while the page lives at `/v10/`, and nothing flags it. Suggest: drop the `PR_NUMBER` fallback from the config, and update every job in `.github/workflows/deploy.yml` (including the legacy versioned one and the PR preview job) to set `PUBLIC_PATH` explicitly. Combined with the validation comment on line 42, misconfiguration becomes a build-time error instead of a runtime mystery.

**No unit tests for the rewritten parsing.** `parseCurrentUrl` / `buildUrl` / `getDeployBase` just had their input space widened from a finite set (`''`, `/latest`, `/pr-preview/pr-N`) to an arbitrary path (`/repo/pr-preview/pr-N/...`). That's exactly when tests stop being optional. `packages/__docs__` has no test setup today, so step one is wiring vitest in there — worth doing here so the new cross-repo path stays covered.

**Subtle semantic shift in `getAssetBasePath`.** Old: derived from current URL (`parseCurrentUrl().prPrefix`). New: derived from build-time `__webpack_public_path__`. The new behavior is strictly more correct (the cross-repo case has no URL marker), but it means a build can't be served from a path different from the one it was compiled for. Worth a line in the commit body so future debuggers don't get surprised.

### Suggested test plan (the PR body's test plan is empty)

- Build with no env vars → `<base href="/">`, dev server works.
- Build with `PR_NUMBER=2557` → `<base href="/pr-preview/pr-2557/">`, chunks prefixed.
- Build with `PUBLIC_PATH=/latest/` (mirrors `deploy.yml`) → unchanged from prod.
- Build with a novel `PUBLIC_PATH=/instructure-design-tokens/pr-preview/pr-42/` (the actual use case) → confirm `<base href>`, every chunk URL, every fetch in the Network tab, the version selector, and the SPA 404 round-trip.
- Deep link `<base>SomeRoute` → 404 fires the redirect → `index.tsx` restores URL → page renders.
- Deep link with hash + query: `<base>Button?utm=x#props` → check what survives the redirect (see inline note on the SPA script).
- Side-by-side smoke test against current `instructure.design/latest/` to confirm no regression for the existing prod deploy.

HerrTopi (changes requested):
Seems good.
One small thing: packages/__docs__/src/index.tsx:33-36 comment is out of date I think. I doesn't redirect to pr preview anymore, but to the applicable path.

HerrTopi (approved):

balzss (approved):

PR #2558, opened 2026-05-14
fix(ui-scripts): fixed a shallow merge bug in generatesemantics.ts

adamlobler:
`mergeSemanticSets` used a shallow spread. Every semantic JSON file wraps its contents under a single top-level "semantic" key, so when color and layout files were merged, the layout's semantic object silently overwrote the color one. The deep merge now recursively combines both instead.

### DIFF
--- a/packages/ui-scripts/lib/build/buildThemes/generateSemantics.ts
+++ b/packages/ui-scripts/lib/build/buildThemes/generateSemantics.ts
@@ -104,8 +104,27 @@ export const resolveTypeReferences = (semantics: any, key?: any): string => {
   return `${typeof value}, `
 }
 
+const deepMerge = (target: any, source: any): any => {
+  const result = { ...target }
+  for (const key of Object.keys(source)) {
+    if (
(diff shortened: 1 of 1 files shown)


packages/ui-scripts/lib/build/buildThemes/generateSemantics.ts
```diff
@@ -104,8 +104,27 @@ export const resolveTypeReferences = (semantics: any, key?: any): string => {
...
+    } else {
+      result[key] = source[key]
+    }
+  }
+  return result
+}
+
 export const mergeSemanticSets = (semanticList: any[]) =>
-  semanticList.reduce((acc, semantic) => ({ ...acc, ...semantic }), {})
+  semanticList.reduce((acc, semantic) => deepMerge(acc, semantic), {})
```
matyasf, line 127: please use `mergeDeep` from `ui-utils`

adamlobler:
@HerrTopi will take over this issue. I close this PR.

PR #2561, opened 2026-05-18
build(ui-scripts): fix changed token/theme parsing to follow the new design token structure

HerrTopi:
## Summary

Brings InstUI's token build in line with the new design-token structure and adds runtime support for color modifiers.

**Token parsing fixes (`ui-scripts`)**
- Token references in the new structure already carry the `semantic.` prefix, so `formatReference` no longer prepends `semantics.`. Param name in generated component files renamed `semantics → semantic` to match.
- Multi-file semantic token sets now merge via `mergeDeep` instead of shallow spread, so nested keys from different files no longer clobber each other. `mergeDeep` is inlined into `generateSemantics.ts` rather than imported from `@instructure/ui-utils`, because build scripts need to run before workspace packages are compiled.

**Bootstrap design-token refresh**
- `scripts/bootstrap.js` now refreshes `@instructure/instructure-design-tokens` from its GitHub source before building themes, so a clean bootstrap doesn't reuse a stale tarball.
- Scoped via `pnpm --filter @instructure/ui-scripts update …` (only `ui-scripts` declares the dep) to avoid the workspace-wide lockfile churn that `pnpm update -r` was causing.

**Color modifiers (`emotion`, `ui-scripts`)**
- New `applyColorModifiers` util in `packages/emotion` resolves Tokens Studio `darken` / `lighten` modifiers in HSL space, matching Studio's `space: "hsl"` math (each step is a fraction of the remaining distance to black/white, so it never overshoots).
- Wired into both `useStyleNew` and `withStyleNew` so any component theme entry shaped as `{ value, modify: { type, value } }` is collapsed to a final color string before reaching the component.
- `generateComponents.js` now preserves the `$extensions["studio.tokens"].modify` payload through the token build instead of dropping it, so the modifier metadata actually reaches the runtime.

**Theme type generation refactor**
- Generated per-component theme files no longer carry an explicit return type or import their type from `componentTypes/`; the function is emitted untyped and consumers rely on the aggregate `ComponentTypes` shape. Reduces cross-file coupling in the generated output.

## Test plan
- [ ] `pnpm run bootstrap` succeeds end-to-end on a clean checkout; only `ui-scripts`-related entries change in `pnpm-lock.yaml`.
- [ ] Spot-check a generated component file under `packages/ui-themes/src/themes/*/components/` — signature is `(semantic: Semantics)` and body references resolve against `semantic.*`, not `semantics.*`.
- [ ] Spot-check a generated `semantics.ts` for at least one theme — values that come from multiple token files are all present (verifies the deep-merge fix).
- [ ] Spot-check a component whose theme uses a `darken` / `lighten` modifier (e.g. a hover/active state) — at runtime, the resolved CSS shows an `hsl(…)` / `hsla(…)` color rather than a `{ value, modify }` object, and the shade matches what Tokens Studio renders for the same token.
- [ ] `pnpm run ts:check` passes.
- [ ] `pnpm run dev` boots the docs app and components using semantic tokens (e.g. Button) render with correct theming, including any states that rely on color modifiers.


scripts/bootstrap.js
```diff
@@ -28,6 +28,17 @@ const path = require('path')
 
 const opts = { stdio: 'inherit' }
 function buildProject() {
+  console.info('Fetching design tokens...')
+  try {
+    execSync('pnpm update @instructure/instructure-design-tokens -r', opts)
```
matyasf: This has an unintended side effect of updating all peer dependencies too, this takes a long time and results in unrelated pnpm-lock.yaml changes. AI recommends to use `pnpm --filter @instructure/ui-scripts add @instructure/instructure-design-tokens@github:instructure/instructure-design-tokens`, but please double check

matyasf (changes requested):

scripts/bootstrap.js
HerrTopi: Changed to a more specific command.

packages/ui-scripts/lib/build/buildThemes/generateSemantics.ts
```diff
@@ -22,6 +22,65 @@
  * SOFTWARE.
  */
 
+//////////////////////////////////////////////////////////////////////////////////
+// START OF MERGFEDEEP
```
joyenjoyer: Fix typo

packages/ui-scripts/lib/build/buildThemes/generateSemantics.ts
```diff
@@ -22,6 +22,65 @@
...
+    !Array.isArray(item)
+  )
+}
+
+function isArray(item: unknown): boolean {
+  return Array.isArray(item)
+}
+
+//////////////////////////////////////////////////////////////////////////////////
+// END OF MERGFEDEEP
```
joyenjoyer: Fix typo

packages/emotion/src/styleUtils/applyColorModifiers.ts
```diff
@@ -0,0 +1,108 @@
...
+ * remaining distance to black/white. Plain values and unrecognized modifier
+ * types are passed through unchanged. Nested plain objects are walked
+ * recursively so modifiers anywhere in the tree get resolved.
+ *
+ * @param componentTheme - Theme map whose values can be CSS strings, `ModifyColor`
+ *   objects, or nested theme objects containing the same.
+ * @returns A new theme object mirroring the input shape, with every `ModifyColor`
+ *   collapsed to its final resolved color string.
+ */
+const applyColorModifiers = (
```
joyenjoyer, line 85: Unit test for this would be nice

packages/ui-scripts/lib/build/buildThemes/generateComponents.js
```diff
@@ -32,6 +32,12 @@ const formatComponent = (collection, key) => {
       return { ...acc, [key]: formatComponent(value, key) }
     }, {})
   }
+  if (value['$extensions']) {
```
joyenjoyer, line 39: A comment to explain what `$extensions` field is would be great

packages/ui-scripts/lib/build/buildThemes/generateComponents.js
```diff
@@ -32,6 +32,12 @@ const formatComponent = (collection, key) => {
...
+    return {
+      value: value.value,
+      modify: value['$extensions']['studio.tokens'].modify
```
joyenjoyer, line 42: Is it safe to assume that modify always exists?

packages/ui-scripts/lib/build/buildThemes/generateSemantics.ts
```diff
@@ -22,6 +22,65 @@
  * SOFTWARE.
  */
 
+//////////////////////////////////////////////////////////////////////////////////
+// START OF MERGFEDEEP
+//////////////////////////////////////////////////////////////////////////////////
+//This is needed because scripts can't have dependencies (e.g. ui-utils) because it needs to be resolved before build
+
+function mergeDeep(...args: Record<string, unknown>[]): Record<string, any> {
```
joyenjoyer, line 30: I am okay with keeping this solution but isn't there a better way to do it other than implementing our own merge deep function? Maybe using a 3rd party library?

joyenjoyer (changes requested):
Added my comments. Using both `semantic` and `semantics` for the same idea makes the code harder to read, the reader has to map the two forms and remember which one is correct in each spot, can you come up with better names?

packages/ui-scripts/lib/build/buildThemes/generateSemantics.ts
HerrTopi: This is so simple and small, I'd let this be here, let's not bloat the lib with unecessary 3rd party deps

packages/ui-scripts/lib/build/buildThemes/generateComponents.js
HerrTopi: No, but if it doesn't we want the code to fail. It needs to be fixed in the design side

packages/ui-scripts/lib/build/buildThemes/generateComponents.js
HerrTopi: Added!

packages/emotion/src/styleUtils/applyColorModifiers.ts
HerrTopi: We don't have tests for the scripts anyway, let's not add it in this pr, however a worthwile idea, let's discuss it in the next grooming

packages/ui-scripts/lib/build/buildThemes/generateSemantics.ts
HerrTopi: fixed

packages/ui-scripts/lib/build/buildThemes/generateSemantics.ts
HerrTopi: fixed

joyenjoyer (approved):

packages/emotion/src/styleUtils/applyColorModifiers.ts
matyasf: yep, lets add them in a new PR

packages/emotion/src/styleUtils/applyColorModifiers.ts
```diff
@@ -0,0 +1,108 @@
...
+  modify: {
+    type: 'lighten' | 'darken'
+    value: number
+  }
+}
+
+// Matches Tokens Studio's HSL modifier math: move L by `amount` (0–1) of the
+// remaining distance toward the endpoint, so the step shrinks as it approaches
+// black/white and never overshoots.
+const modifyLightness = (
```
matyasf, line 37: This should be in ui-color-utils in the future

matyasf (approved):
As I understand the code, we are now lying about the token type in generateComponents. Please make a ticket where we fix this

PR #2564, opened 2026-05-27
build(many): remove commonjs

joyenjoyer:
INSTUI-5036


joyenjoyer:
Discussion with @matyasf:
Adding type module into package.json may break something, consider not adding it

PR #2565, opened 2026-05-27
build: simplify and put timers into the bootstrap script

matyasf:
Also simplify and put timers into the bootstrap script.

To test:
- run bootstrap, check that the times are roughly accurate
- all CI should be green 
- docs app should work

### DIFF
--- a/scripts/bootstrap.js
+++ b/scripts/bootstrap.js
@@ -23,71 +23,75 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-const { execSync, fork } = require('child_process')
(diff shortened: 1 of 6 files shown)


scripts/bootstrap.js
```diff
@@ -23,71 +23,44 @@
...
 
 const opts = { stdio: 'inherit' }
-function buildProject() {
-  console.info('Fetching design tokens...')
-  try {
-    execSync(
-      'pnpm --filter @instructure/ui-scripts update @instructure/instructure-design-tokens',
-      opts
-    )
-  } catch (error) {
```
matyasf, line 37: these try..catch blocks were not needed, this script exits already if there is an error

ToMESSKa (approved):

joyenjoyer (approved):

PR #2567, opened 2026-05-27
Fullscreen modal margin

adamlobler:
**Modal fixes:**
- The fullscreen Modal variant previously covered the entire viewport, making it hard to tell it was a modal. A new `fullScreenMargin` token adds separation from the viewport edge to fix that a11y issue on bigger screen sizes.
- `borderRadius` from the component theme wasn't applying because Dialog's hardcoded inline `borderRadius: inherit` always overrode the Emotion CSS class. Dialog now accepts a style prop so consumers can override that default.

### DIFF
--- a/packages/ui-dialog/src/Dialog/props.ts
+++ b/packages/ui-dialog/src/Dialog/props.ts
@@ -42,6 +42,11 @@ type DialogOwnProps = {
   as?: AsElementType
   display?: 'auto' | 'block' | 'inline-block'
   /**
+   * Inline styles applied to the root element. Merged with the default
+   * `borderRadius: inherit` so consumers can override it.
+   */
+  style?: React.CSSProperties
+  /**
    * The aria-label to read for screen reader. When specified, it will automatically set role="dialog".
    */
   label?: string
@@ -77,7 +82,8 @@ const allowedProps: AllowedPropKeys = [
   'shouldCloseOnDocumentClick',
   'shouldCloseOnEscape',
   'shouldFocusOnOpen',
-  'elementRef'
+  'elementRef',
+  'style'
 ]
 
 export type { DialogProps }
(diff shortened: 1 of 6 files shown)


hajnaldo (approved):

packages/ui-dialog/src/Dialog/props.ts
```diff
@@ -41,6 +41,11 @@ type DialogOwnProps = {
    */
   as?: AsElementType
   display?: 'auto' | 'block' | 'inline-block'
+  /**
+   * Inline styles applied to the root element. Merged with the default
+   * `borderRadius: inherit` so consumers can override it.
+   */
+  style?: React.CSSProperties
```
matyasf, line 48: We do not allow to set users CSS for our components. It opens up way too many possibilities leading to inconsistent design and bugs for users during internal styles changes.

We could pass have a `borderRadius` prop if needed (that defaults to `inherit`)

packages/ui-modal/src/Modal/v2/index.tsx
```diff
@@ -99,7 +100,7 @@ class Modal extends Component<ModalProps, ModalState> {
 
   componentDidMount() {
     this.props.makeStyles?.()
-    window.addEventListener('resize', this.updateHeight)
+    window.addEventListener('resize', this.updateSize)
```
matyasf, line 103: We might need some conditionals here, `window` does not exists in SSR environments an in Jest unit tests

packages/ui-modal/src/Modal/v2/index.tsx
```diff
@@ -213,13 +217,17 @@ class Modal extends Component<ModalProps, ModalState> {
...
       liveRegion,
-      size,
       constrain,
       as,
+      size,
       styles
     } = this.props
 
     const isFullScreen = size === 'fullscreen'
+    const isSmallScreen = this.state.windowWidth < 480 // 30em at 16px base
```
matyasf, line 227: A bit off, but in the previous styling enginer breakpoints were defined at theme level. This might be better for us too, so we dont have components changing visuals lots of times when resizing a window.

matyasf (changes requested):
Please write a "How to test" part in the PR description too

matyasf:
We will fix these in a separate PR

PR #2568, opened 2026-05-27
docs: better multiversion support for docs v2

git-nandor:
INSTUI-5031

## Summary

- Moves the component version (minor version) selector from the side nav to the component pages, displayed alongside the Theme selector
- The Component version and Theme selectors are now only rendered on Components pages
- The `v11.6` option in the Component version Select is labeled `v11.6 (legacy theming)`
- Restores the full library version display in the side nav header
- Legacy Theme Overrides page always redirects to v11.6 
- New Theme Overrides page always redirects to v11.7
- Adds an info alert on all non-component pages explaining that the content reflects the latest versioning
- Modify New theme overrides page "13. Independent overrides for child parts of compound components" section
- Adds a new "Component versioning" guide page

### DIFF
--- a/docs/guides/component-versioning.md
+++ b/docs/guides/component-versioning.md
@@ -1,0 +1,95 @@
+---
+title: Component versioning
+category: Guides
+order: 2
+---
+
+## Why components are versioned
+
+When InstUI needs to make a breaking change to a component (renamed props, changed behaviour, etc.), the old version is **kept alongside the new one** instead of being replaced. Upgrading the library no longer forces you to immediately rewrite every component usage — you can migrate to new versions on your own schedule.
+
+New component versions appear with InstUI **minor** version bumps (e.g. `11.7` → `11.8`). Patch releases never include breaking changes.
+
+> The docs UI "Component version" selector labels library `v11_6` as **`v1 (legacy)`** and library `v11_7` as **`v2`**. This page uses the `v11_X` form because it matches the actual NPM import paths.
+
+## Import paths
+
+Every InstUI component package supports three import styles:
+
+### Default — `@instructure/ui-<name>`
+
+Always points to the **oldest** still-supported component version. Upgrading the library without changing your imports will keep your code working without surprises.
+
+```js
+---
+type: code
+---
+import { Alert } from '@instructure/ui-alerts'
+```
+
+### Pinned — `@instructure/ui-<name>/v11_X`
+
+Locks the import to a specific InstUI minor version of the component. When you are ready to adopt a breaking change, update the path to the next pinned version.
+
+```js
+---
+type: code
+---
+import { Alert } from '@instructure/ui-alerts/v11_7'
+```
+
+### Latest — `@instructure/ui-<name>/latest`
+
+Always points to the newest component version. This may bring breaking changes when you upgrade InstUI itself.
+
+```js
+---
+type: code
+---
+import { Alert } from '@instructure/ui-alerts/latest'
+```
+
+### Per-package or umbrella package
+
+InstUI also ships an umbrella package, `@instructure/ui`, which re-exports every component from the individual `@instructure/ui-*` packages. Two equivalent import styles work — both resolve to the same component:
+
+```js
+---
+type: code
+---
+// per-package import
+import { Alert } from '@instructure/ui-alerts/v11_7'
+
+// umbrella package import
+import { Alert } from '@instructure/ui/v11_7'
+```
+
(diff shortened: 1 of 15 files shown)


matyasf (changes requested):
- The "If you are viewing the v11.7 version, [switch to v11.6](https://instructure.design/pr-preview/pr-2568/v11_6/legacy-theme-overrides) to see the examples working correctly." text in the legacy/new theme overrides is not needed anymore since the switch is automatic.
- The message "This page always displays the latest version of the documentation" is not clear what it means. I would rephrase it to "This page is made for the latest version (v${major.minor})) of InstUI). Also the linked page https://instructure.design/pr-preview/pr-2568/v11_7/multi-version-system is way too detailed for our users, can you make a new page under Guides that explains the multi version stuff for users? We just need that components are now versioned, the import paths, and what versions work with which theming engine.
- I would not include the Component version selector for non-component pages where it does nothing
- +1: How about displaying components with the "Light" theme when switching to v11.7, since this is now the recommended theme?

docs/guides/component-versioning.md
```diff
@@ -0,0 +1,93 @@
+---
+title: Component versioning
+category: Guides
+order: 5
```
matyasf: I think this is an important document, it should be #2, right after the "Getting started" menu point. but its not that important

matyasf (approved):
Nice work, I like these changes a lot!

ToMESSKa (changes requested):
Image

Here the legacy examples are shown, but the Alert says `made for the latest version (v11.7) of InstUI`, it can be a bit confusing?

docs/guides/component-versioning.md
```diff
@@ -0,0 +1,95 @@
...
+Each version of the library is only compatible with the themes built for its theming engine:
+
+- **`v1`** (library `v11_6` and earlier) supports the legacy themes:
+
+  - `canvas`
+  - `canvasHighContrast`
+
+- **`v2`** (library `v11_7` and newer) supports themes built on the new token system. The legacy Canvas looks are preserved (re-implemented on the new engine) and two brand-new themes are added:
+
+  - `legacyCanvas` — same visual style as the old `canvas`, but on the new engine
```
balzss: This lists the v2 legacy themes as `legacyCanvas` / `legacyCanvasHighContrast`, but the new-theme-overrides page (v2) uses `theme={canvas}` throughout, and the docs Theme selector labels the v2 theme as `canvas (legacy)`. Are the real v2 theme keys actually `canvas` / `canvasHighContrast` (shown as "(legacy)") rather than `legacyCanvas` / `legacyCanvasHighContrast`? If so the names here are misleading.

docs/theming/new-theme-overrides.md
```diff
@@ -527,9 +518,52 @@ type: example
...
+          <Table.ColHeader id="h-row">TableColHeader — purple</Table.ColHeader>
+          <Table.ColHeader id="h-col">TableColHeader — purple</Table.ColHeader>
+        </Table.Row>
+      </Table.Head>
+      <Table.Body>
+        <Table.Row>
+          <Table.RowHeader>TableRowHeader — deeppink</Table.RowHeader>
+          <Table.Cell>TableCell — unchanged</Table.Cell>
+        </Table.Row>
+                <Table.Row>
```
balzss: Minor: this `<Table.Row>` is over-indented (extra leading spaces), and both `Table.ColHeader`s above share the same label and a copy-pasted `id` pattern (`h-row` / `h-col`). Quick tidy since it renders as an example.

packages/__docs__/src/App/index.tsx
```diff
@@ -741,9 +824,20 @@ class App extends Component<AppProps, AppState> {
...
+    const isGuidePage = currentData.extension === '.md'
     const documentContent = (
       <View as="div" padding="x-large none none">
         {this.renderThemeSelect()}
+        {isGuidePage &&
+          (currentData.id as string) !== 'legacy-theme-overrides' && (
+            <Alert variant="info" margin="xx-large 0">
+              This page is made for the latest version (
+              {this.getLatestMinorVersion()?.replace('_', '.')}) of InstUI.{' '}
+              <Link href="component-versioning">
```
balzss: This (and the same `<Link href="component-versioning">` in `Document/index.tsx`) uses a bare relative href. Does it resolve correctly from a versioned URL like `/v11_7/Alert` — navigating within the SPA rather than triggering a full reload — given the deploy-base + `v11_X` prefix routing? Worth a quick check.

git-nandor: Thanks, fixed

docs/guides/component-versioning.md
git-nandor: Modified

matyasf (commented):
nice work, the whole documentation is much better understandable now!

balzss (approved):

ToMESSKa (approved):

PR #2574, opened 2026-06-02
fix(ui-table): fix caption for table to be responsive to sorting

HerrTopi:
changes the Table caption prop from an auto-appended hardcoded English "Sorted by..." string to a consumer-supplied function (sortByHeader, sortDirection) => string, so the caption reflects sort state in a fully localizable way. In v1 this is non-breaking (caption accepts a ReactNode or a function), while v2 now requires caption to be a function (breaking change), with tests, docs, and the regression page updated accordingly.

TESTS_PLAN
- With a screen reader (NVDA/VoiceOver), sort a column and confirm the caption announcement reflects the new sort state and uses the consumer-provided (localizable) text, not a hardcoded English suffix.
- v1 only: pass a plain string caption → confirm it renders unchanged and no sort announcement fires (expected behavior); pass a function → confirm announcements fire on sort change.
- Confirm sort announcement only fires when the sort actually changes (no duplicate/repeat announcements).

INSTUI-5047

### DIFF
--- a/packages/ui-table/src/Table/v2/props.ts
+++ b/packages/ui-table/src/Table/v2/props.ts
@@ -32,12 +32,20 @@ import type { NewComponentTypes } from '@instructure/ui-themes'
 
 type RowChild = React.ReactElement<{ children: React.ReactElement }>
 
+type TableCaption = (
+  sortByHeader: string,
+  sortDirection: 'none' | 'ascending' | 'descending'
+) => string
+
 type TableOwnProps = {
   /**
-   * Provide a screen reader friendly description. Anything passed to this
-   * prop will be wrapped by `<ScreenReaderContent>` when it is rendered.
+   * Provide a screen reader friendly description. The returned string is
+   * wrapped by `<ScreenReaderContent>` when it is rendered.
+   *
+   * This is a function so consumers can build a localized caption that also
+   * reflects the current sort state. See {@link TableCaption}.
    */
-  caption: React.ReactNode
+  caption: TableCaption
   /**
    * Valid values are `0`, `none`, `auto`, `xxx-small`, `xx-small`, `x-small`,
    * `small`, `medium`, `large`, `x-large`, `xx-large`. Apply these values via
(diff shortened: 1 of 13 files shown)


ToMESSKa (changes requested):
- In v11.6 examples, there is no announcement which column the table is sorted by (see below)
- also there is no announcement when I change the sorting in the v11.6 examples
image

Can you please take a look at these?

packages/ui-table/src/Table/v2/props.ts
```diff
@@ -32,12 +32,20 @@ import type { NewComponentTypes } from '@instructure/ui-themes'
...
-   * Provide a screen reader friendly description. Anything passed to this
-   * prop will be wrapped by `<ScreenReaderContent>` when it is rendered.
+   * Provide a screen reader friendly description. The returned string is
+   * wrapped by `<ScreenReaderContent>` when it is rendered.
+   *
+   * This is a function so consumers can build a localized caption that also
+   * reflects the current sort state. See {@link TableCaption}.
    */
-  caption: React.ReactNode
+  caption: TableCaption
```
matyasf, line 48: Can you put this change also in the upgrade guide for multi version (make sure to rebase)

matyasf (changes requested):

packages/ui-table/src/Table/v2/props.ts
HerrTopi: Nothing to add, check the pr again pls

ToMESSKa (changes requested):
> * In v11.6 examples, there is no announcement which column the table is sorted by (see below)
> * also there is no announcement when I change the sorting in the v11.6 examples
> 
> image
> Can you please take a look at these?

@HerrTopi The title of the table (Top rated movies) is not announced in the v11.6 example.
Can you please fix that too?

image

matyasf (approved):

ToMESSKa (approved):

PR #2584, opened 2026-06-09
fix(ui-modal): skip Modal.Body tab stop when it has focusable children

ToMESSKa:
INSTUI-5046

**ISSUE:**
- Modal.Body gets a redundant tabIndex={0} when it's scrollable but already contains focusable elements, creating a redundant tab stop. It should only get a tabIndex={0}  when it has no interactive elements.

**TEST PLAN:**
- Copy the code below into the documentation.
- Open the modal using only the keyboard.
   - Navigate to the modal body using the Tab key.
   - The modal body should receive focus.
   - Once focused, the body should be scrollable using the Down Arrow and Up Arrow keys.
- Keep the modal open.
   - Navigate to the "Scrollable body" toggle and turn it off.
   - While the modal remains open, navigate back to the modal body.
   - This time, the modal body should not receive focus.
- Keep the modal open.
   - Turn the "Scrollable body" toggle back on.
   - Verify that the modal body can be focused again.
   - While the modal is open, turn on the "Focusable element in body" toggle.
   - Navigate to the modal body again — this time, the whole body should not be focused. Instead, focus should jump only to the button inside the body.
- Turn the "Focusable element in body" toggle back off.
   - The modal body should once again be focusable.

- Check that Modal.Body gets tabIndex={0} when it contains no focusable elements.
- Check that Modal.Body does NOT get tabIndex={0} when it contains at least one focusable element.
- should work in v1 and v2 too

```
const longText = lorem.paragraphs(100)
const shortText = lorem.paragraphs(1)

const Example = () => {
  const [scrollable, setScrollable] = useState(true)
  const [hasFocusable, setHasFocusable] = useState(false)

  const handleButtonClick = () => {
    setOpen((state) => !state)
  }

  const renderCloseButton = () => {
    return (
      <CloseButton
        placement="end"
        offset="small"
        onClick={handleButtonClick}
        screenReaderLabel="Close"
      />
    )
  }

  return (
    <div style={{ padding: '0 0 11rem 0', margin: '0 auto' }}>
      <Button onClick={handleButtonClick}>
        {open ? 'Close' : 'Open'} the Modal
      </Button>
      <Modal
        open={open}
        onDismiss={() => {
          setOpen(false)
        }}
        size="medium"
        label="Hello World"
        shouldCloseOnDocumentClick
      >
        <Modal.Header>
          {renderCloseButton()}
        </Modal.Header>
        <Modal.Body>
          {hasFocusable && (
            <View as="div" margin="0 0 small 0">
              <Button>A focusable element inside the body</Button>
            </View>
          )}
          <Text lineHeight="double">{scrollable ? longText : shortText}</Text>
        </Modal.Body>
        <Modal.Footer>
          <View as="div" margin="0 auto 0 0">
            <View as="div" display="block" margin="0 0 small 0">
              <Checkbox
                label="Scrollable body"
                variant="toggle"
                size="small"
                checked={scrollable}
                onChange={() => setScrollable((v) => !v)}
              />
            </View>
            <Checkbox
              label="Focusable element in body"
              variant="toggle"
              size="small"
              checked={hasFocusable}
              onChange={() => setHasFocusable((v) => !v)}
            />
          </View>
          <Button onClick={handleButtonClick}>Close</Button>
        </Modal.Footer>
      </Modal>
    </div>
  )
}

render(<Example />)
```

### DIFF
--- a/packages/ui-modal/src/Modal/v1/ModalBody/index.tsx
+++ b/packages/ui-modal/src/Modal/v1/ModalBody/index.tsx
@@ -26,7 +26,7 @@ import { Component } from 'react'
 
 import { View } from '@instructure/ui-view/v11_6'
 import { omitProps } from '@instructure/ui-react-utils'
-import { getCSSStyleDeclaration } from '@instructure/ui-dom-utils'
+import { getCSSStyleDeclaration, findTabbable } from '@instructure/ui-dom-utils'
 
 import { withStyle } from '@instructure/emotion'
 import generateStyle from './styles.js'
@@ -56,6 +56,8 @@ class ModalBody extends Component<ModalBodyProps> {
   }
 
   ref: UIElement | null = null
+  private resizeObserver?: ResizeObserver
+  private mutationObserver?: MutationObserver
 
   handleRef = (el: UIElement | null) => {
     const { elementRef } = this.props
@@ -87,10 +89,37 @@ class ModalBody extends Component<ModalBodyProps> {
     if (isFirefox) {
       this.setState({ isFirefox })
     }
+
+    const finalRef = this.getFinalRef(this.ref)
+    if (finalRef && typeof ResizeObserver !== 'undefined') {
+      this.resizeObserver = new ResizeObserver(() => this.forceUpdate())
+      this.resizeObserver.observe(finalRef)
+      this.mutationObserver = new MutationObserver(() => this.forceUpdate())
+      this.mutationObserver.observe(finalRef, {
+        childList: true,
+        subtree: true,
+        attributes: true,
+        attributeFilter: [
+          'disabled',
+          'tabindex',
+          'hidden',
+          'href',
+          'contenteditable',
+          'type',
+          'class',
+          'style'
+        ]
+      })
+    }
   }
 
   componentDidUpdate() {
     this.props.makeStyles?.()
+  }
+
+  componentWillUnmount() {
+    this.resizeObserver?.disconnect()
+    this.mutationObserver?.disconnect()
   }
 
   getFinalRef(el: UIElement): Element | undefined {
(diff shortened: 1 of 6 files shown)


packages/ui-modal/src/Modal/v1/ModalBody/index.tsx
```diff
@@ -86,12 +88,39 @@ class ModalBody extends Component<ModalBodyProps> {
...
+        attributeFilter: [
+          'disabled',
+          'tabindex',
+          'hidden',
+          'href',
+          'contenteditable',
+          'type',
+          'class',
+          'style'
+        ]
```
ToMESSKa, line 111: based on the set of attributes that `findTabbable` reads

packages/ui-modal/src/Modal/v1/ModalBody/index.tsx
```diff
@@ -86,12 +88,39 @@ class ModalBody extends Component<ModalBodyProps> {
     if (isFirefox) {
       this.setState({ isFirefox })
     }
+
```
ToMESSKa, line 92: This was needed to handle the the case when the modal body content changes  dinamically (a scrollbar or a focusable element appears or disappears)

packages/ui-modal/src/Modal/v1/ModalBody/index.tsx
matyasf: Can you put this as a comment to `findTabbble`? This is important because we might replace findTabbable in the future, it has several flaws

matyasf (changes requested):
looks good, just a small comment. In the future we should look into replacing this with some native code (e.g. https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog ) because there are constantly issues with this code

joyenjoyer (approved):
great work!

packages/ui-modal/src/Modal/v1/ModalBody/index.tsx
ToMESSKa: added

matyasf (approved):

PR #2587, opened 2026-06-10
feat(many): support current spacing tokens in the margin prop for v2 components

balzss:
## What & why

The **current (era-3) spacing tokens** — `general.*`, `gap.*`, `padding.card.*` (dot-path notation, e.g. `margin="general.spaceMd"`) — already resolved at runtime in View v2, but the `Spacing` prop type and the docs only advertised the two older token eras. So the new tokens had no autocomplete, no type hints, and weren't documented.

This PR makes them first-class on the `margin` prop of v2 components. **Non-breaking** — the older tokens still resolve at runtime and are only marked `@deprecated`.

## What changed

| Area | Change |
| --- | --- |
| **Types** (`emotion`) | Add `CurrentSpacingValues` (era-3 dot-paths) to the `Spacing` union; mark era-1 / era-2 unions `@deprecated`. |
| **v2 components** | Sync the `margin` JSDoc across all v2 `props.ts` files — link to the spacing guide + current-token example. |
| **v2 READMEs** | Convert legacy spacing keywords in `margin` examples to `general.*` tokens (value-preserving mapping). |
| **Docs guide** | Re-add the *Layout Spacing* page (removed on master during the docs rework) as a **v11.7+ version-gated page**, following the `theme-overrides` version-banner pattern. |
| **Tests** | era-3 dot-path resolution in `calcSpacingFromShorthand`; View v2 integration test (`general.spaceMd` → `0.75rem`); regression-test page examples. |

## Reviewer notes

- **Legacy → era-3 example mapping is value-preserving** where an equivalent exists (`small`→`general.spaceMd`, `medium`→`general.spaceXl`, …). `large`/`x-large` have no `general` equivalent (scale tops out at 2rem) and compress to `general.space2xl` — the only intentional visual change, limited to doc examples.
- **Live doc examples use `margin`/`padding` only.** These resolve era-3 dot-paths via `calcSpacingFromShorthand`. The `Flex` `gap` prop uses a separate flat-key resolver that doesn't understand dot-paths, so it's intentionally not shown with them.
- **Scope:** v2 components only; the dot-path resolver / token-merge logic is unchanged.

## Testing

- `pnpm run test:vitest emotion` → 78 passed
- `pnpm run test:vitest ui-view` → 16 passed
- Visual regression (Cypress/Chromatic) not run; doc/regression page examples added for it.

## Follow-ups (not in this PR)

- `padding` prop examples still use legacy tokens (kept out of scope; easy identical sweep).
- Misspelled tokens compile (type ends in `(string & {})`) and only warn at runtime — a louder dev-mode warning could be added later.

### DIFF
--- a/docs/guides/layout-spacing.md
+++ b/docs/guides/layout-spacing.md
@@ -7,18 +7,19 @@ relevantForAI: true
 
 ## Layout Spacing
 
-```js
----
-type: embed
----
-<Alert variant="warning" margin="0 0 medium">
-  The spacing tokens documented on this page (such as <code>general.spaceMd</code>) require <strong>v11.7+</strong> components and are applied through the <code>margin</code> and <code>padding</code> props. If you are viewing the v11.6 version, <Link href={window.location.pathname.match(/v\d+_\d+/) ? window.location.pathname.replace(/v\d+_\d+/, 'v11_7') : `/v11_7${window.location.pathname}`}>switch to v11.7</Link> to see the examples working correctly.
-</Alert>
-```
+The spacing tokens documented on this page (such as `general.spaceMd`) require **v11.7+** components and are applied through the `margin` and `padding` props.
 
 Our design system provides a set of spacing tokens for consistent layouts and components. The current tokens are organized into a general t-shirt scale plus a handful of semantic tokens. Some tokens share a value but carry different meaning — prefer the semantically correct token for the context (e.g. use `gap.buttons` for spacing between buttons).
 
 The `margin` and `padding` props on InstUI components accept these tokens via **dot-path notation** (for example `margin="general.spaceMd"` or `padding="padding.card.lg"`), and support the familiar CSS-like 1–4 value shorthand.
+
+### Where these tokens come from
+
+A prop like `margin="general.spaceMd"` is resolved at runtime against the active theme's `sharedTokens.spacing` object. `sharedTokens` doesn't hold spacing values of its own — it reshapes the theme's **semantic** tokens into the flat, prop-facing structure shown below. The full pipeline is:
+
+`primitives` (raw `rem` values) → `semantics.spacing` (named tokens per theme) → `sharedTokens.spacing` (the dot-paths this page documents) → the `margin` / `padding` prop.
+
+Because the values originate in each theme's semantic tokens, they are **theme-dependent** in principle — a theme can resolve `general.spaceMd` to a different value. In practice all themes InstUI currently ships resolve these tokens to the same values (those listed in the tables below).
 
 ## Tokens
 
--- a/docs/guides/layout-spacing.md
+++ b/docs/guides/layout-spacing.md
@@ -1,0 +1,95 @@
+---
+title: Layout Spacing
+category: Guides
+order: 7
+relevantForAI: true
+---
+
+## Layout Spacing
+
+```js
+---
+type: embed
+---
+<Alert variant="warning" margin="0 0 medium">
+  The spacing tokens documented on this page (such as <code>general.spaceMd</code>) require <strong>v11.7+</strong> components and are applied through the <code>margin</code> and <code>padding</code> props. If you are viewing the v11.6 version, <Link href={window.location.pathname.match(/v\d+_\d+/) ? window.location.pathname.replace(/v\d+_\d+/, 'v11_7') : `/v11_7${window.location.pathname}`}>switch to v11.7</Link> to see the examples working correctly.
+</Alert>
+```
+
+Our design system provides a set of spacing tokens for consistent layouts and components. The current tokens are organized into a general t-shirt scale plus a handful of semantic tokens. Some tokens share a value but carry different meaning — prefer the semantically correct token for the context (e.g. use `gap.buttons` for spacing between buttons).
+
+The `margin` and `padding` props on InstUI components accept these tokens via **dot-path notation** (for example `margin="general.spaceMd"` or `padding="padding.card.lg"`), and support the familiar CSS-like 1–4 value shorthand.
+
+## Tokens
+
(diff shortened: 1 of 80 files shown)


packages/ui-alerts/src/Alert/v2/props.ts
```diff
@@ -79,9 +79,9 @@ type AlertOwnProps = {
    */
   timeout?: number
   /**
-   * Valid values are `0`, `none`, `auto`, `xxx-small`, `xx-small`, `x-small`,
-   * `small`, `medium`, `large`, `x-large`, `xx-large`. Apply these values via
-   * familiar CSS-like shorthand. For example: `margin="small auto large"`.
+   * Valid values are `0`, `none`, `auto`, and Spacing token values,
+   * see https://instructure.design/layout-spacing. Apply these values via
+   * familiar CSS-like shorthand. For example, `margin="general.spaceMd auto"`.
```
git-nandor, line 84: In v2 Avatar, Flex, FormFieldLayout, Link still recommend the legacy margin="small auto large"

ToMESSKa (changes requested):
I haven't checked all, but some of the v2 examples still use older spacing values in the margin prop (e.g. Alert, Badge), those should be updated too.

git-nandor (changes requested):
See my comment

git-nandor (approved):

ToMESSKa (approved):

docs/guides/layout-spacing.md
```diff
@@ -0,0 +1,95 @@
...
+---
+
+## Layout Spacing
+
+```js
+---
+type: embed
+---
+<Alert variant="warning" margin="0 0 medium">
+  The spacing tokens documented on this page (such as <code>general.spaceMd</code>) require <strong>v11.7+</strong> components and are applied through the <code>margin</code> and <code>padding</code> props. If you are viewing the v11.6 version, <Link href={window.location.pathname.match(/v\d+_\d+/) ? window.location.pathname.replace(/v\d+_\d+/, 'v11_7') : `/v11_7${window.location.pathname}`}>switch to v11.7</Link> to see the examples working correctly.
```
matyasf: > If you are viewing the v11.6 version, [switch to v11.7](https://instructure.design/pr-preview/pr-2587/v11_7/layout-spacing) to see the examples working correctly.

no need for this text, it always shows 11.7

docs/guides/layout-spacing.md
```diff
@@ -0,0 +1,95 @@
...
+</Alert>
+```
+
+Our design system provides a set of spacing tokens for consistent layouts and components. The current tokens are organized into a general t-shirt scale plus a handful of semantic tokens. Some tokens share a value but carry different meaning — prefer the semantically correct token for the context (e.g. use `gap.buttons` for spacing between buttons).
+
+The `margin` and `padding` props on InstUI components accept these tokens via **dot-path notation** (for example `margin="general.spaceMd"` or `padding="padding.card.lg"`), and support the familiar CSS-like 1–4 value shorthand.
+
+## Tokens
+
+### General scale
```
matyasf, line 26: I would also add that the actual values are theme-dependent

docs/guides/layout-spacing.md
```diff
@@ -0,0 +1,95 @@
...
+| gap.cards.nestedContainers.sm | 0.5rem  | 8px             |
+| gap.cards.nestedContainers.md | 0.75rem | 12px            |
+| gap.cards.nestedContainers.lg | 1rem    | 16px            |
+| gap.inputs.horizontal         | 0.75rem | 12px            |
+| gap.inputs.vertical           | 1rem    | 16px            |
+| padding.card.sm               | 0.5rem  | 8px             |
+| padding.card.md               | 0.75rem | 12px            |
+| padding.card.lg               | 1rem    | 16px            |
+
+## Applying spacing
```
matyasf, line 57: I would also mention which part of our themes contain these tokens (is it in sharedTokens?)

matyasf (changes requested):
see my comments. Also looking at the token code, its really not clear for me where the tokens are coming from. Is it in `SharedTokens` or in `Semantics`? I guess its in `SharedTokens`, otherwise we could have different values for each theme. But then the token values from semaintics need to be deleted? e.g. `ui-themes/src/themes/newThemeTokens/dark/semantics.ts` vs `ui-themes/src/themes/newThemeTokens/dark/sharedTokens.ts`

balzss:
@matyasf thanks — the "where do the tokens come from" question was a real doc gap, so I've added a **"Where these tokens come from"** section to the Layout Spacing guide that spells out the pipeline.

To answer directly: the `margin`/`padding` props resolve against **`sharedTokens.spacing`** (see `ui-view/src/View/v2/styles.ts` — it merges `sharedTokens.spacing` with `sharedTokens.legacy.spacing`). So your guess about `SharedTokens` is right about where the *prop* reads from.

But **nothing in `semantics.ts` should be deleted** — there's no duplication. `sharedTokens.ts` is a function `(semantic: Semantics) => SharedTokens` that doesn't hold values of its own; it *references* `semantic.spacing.*` and just reshapes it into the flat dot-path structure the prop API exposes (`spacing.general.*`, `spacing.gap.*`, `spacing.padding.card.*`). The layering is: `primitives` → `semantics` (source of truth, per theme) → `sharedTokens` (prop-facing view) → prop. Deleting the semantic spacing values would break `sharedTokens`, which reads from them.

On theme-dependence: because the values originate in each theme's semantic tokens, they *can* differ per theme — but all themes we currently ship resolve them to the same values (the ones in the tables). I've noted that in the guide too.

Also dropped the "switch to v11.7" banner text per your other comment.

matyasf (approved):

PR #2588, opened 2026-06-11
refactor(many): upgrade to pnpm 11

matyasf:
also do not allow some dependencies to run scripts for better security. Some dependencies needed to be added because pnpm now has a more strict peer dependency resolution.

pnpm 11 has lots of good new features:
- it disables dependencies running scripts by default, we are manually allowing them in `pnpm-workspace.yml`. Please only add here new entries if you checked that its needed and safe (Claude can do this)
- by default it will only install dependencies that are more than 1 day old. This gives some time npmjs.org to detect and take down malicious packages
- it seems to have a stricter package resolution, not allowing bleeding trough packages

INSTUI-5064

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,16 +50,11 @@
     "ts:check": "pnpm -r --stream ts:check"
   },
   "license": "MIT",
-  "pnpm": {
-    "overrides": {
-      "react": "18.3.1",
-      "react-dom": "18.3.1",
-      "@types/react": "18.3.26",
-      "git-raw-commits>dargs": "7.0.0"
-    }
-  },
   "devDependencies": {
     "@babel/cli": "^7.27.2",
+    "babel-loader": "^10.1.1",
+    "@babel/runtime": "^7.29.2",
+    "moment-timezone": "^0.6.2",
     "@commitlint/cli": "^20.5.3",
     "@commitlint/config-conventional": "^20.5.3",
     "@emotion/cache": "^11.14.0",
@@ -140,5 +135,5 @@
   "browserslist": [
     "extends @instructure/browserslist-config-instui"
   ],
-  "packageManager": "pnpm@10.23.0+sha512.21c4e5698002ade97e4efe8b8b4a89a8de3c85a37919f957e7a0f30f38fbc5bbdd05980ffe29179b2fb6e6e691242e098d945d1601772cad0fef5fb6411e2a4b"
+  "packageManager": "pnpm@11.5.3+sha512.7ac1c919341c213a34dc0d02afb7143c5c26ac26ee8c4782deea821b8ac64d2134a081fd8941dae6e29bbb48f58dfc2b7fbceeccc07cb2f09d219d342a4969ed"
 }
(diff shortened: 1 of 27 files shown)


packages/__docs__/package.json
```diff
@@ -114,14 +113,18 @@
...
     "lorem-ipsum": "^3.0.0",
     "marked-react": "^4.0.0",
     "moment": "^2.30.1",
     "react": "18.3.1",
     "react-dom": "18.3.1",
     "uuid": "^14.0.0",
-    "webpack-merge": "^6.0.1"
+    "webpack-merge": "^6.0.1",
+    "hoist-non-react-statics": "^3.3.2",
+    "buffer": "^6.0.3"
```
matyasf, line 127: I needed to add lots of dependencies because in earlier versions pnpm read these if they were in the root node_modules folder, now it needs to be added explicitly. This also uncovered some bugs, e.g. babel runtime not added everywhere

package.json
```diff
@@ -50,16 +50,11 @@
...
   },
   "license": "MIT",
-  "pnpm": {
-    "overrides": {
-      "react": "18.3.1",
-      "react-dom": "18.3.1",
-      "@types/react": "18.3.26",
-      "git-raw-commits>dargs": "7.0.0"
-    }
-  },
```
matyasf, line 60: This is now in `[pnpm-workspace.yaml`

balzss (approved):
nice work, works great. pnpm automatically switched to v11 when i run install

packages/ui-scripts/package.json
```diff
@@ -37,7 +37,9 @@
     "style-dictionary": "4.4.0",
     "webpack-cli": "^7.0.2",
     "webpack-dev-server": "^5.2.3",
-    "yargs": "^17.7.2"
+    "yargs": "^17.7.2",
+    "lucide-react": "1.7.0",
```
joyenjoyer, line 41: Why is lucide-react added here?

joyenjoyer (changes requested):

packages/ui-scripts/package.json
balzss: i didn't check so i'm not sure but one guess is because of the icon scripts located in ui-scripts import something from the lucide package

joyenjoyer:
Please take a look into this issue, upgrade might break something: https://github.com/pnpm/pnpm/issues/11513

@matyasf

packages/ui-scripts/package.json
matyasf: yep, because e.g. `generate-lucide-index` imports it

matyasf:
> Please take a look into this issue, upgrade might break something: [pnpm/pnpm#11513](https://github.com/pnpm/pnpm/issues/11513)
> 
> @matyasf

Good catch, we're using `action-setup` v4, and pnpm 11 support was added in version 6 https://github.com/pnpm/action-setup/releases

matyasf:
@joyenjoyer I've updated all our GH actions to the latest version

joyenjoyer (approved):
great work

PR #2595, opened 2026-06-17
fix(ui-dialog): allow consumers to override inherited border-radius

HerrTopi:
## Summary
- Replace Dialog's forced inline `border-radius: inherit` with an `@layer`-wrapped emotion rule so unlayered consumer styles (e.g. Modal) can override it while the inherit fallback still applies when none is set.

## Test Plan
- Open a Modal and confirm it renders with its own `border-radius` (not the wrapping container's).
- Verify Menu / DateInput rendered inside a rounded `View` still inherit the View's corner radius.

Fixes INSTUI-5068

### DIFF
--- a/packages/ui-dialog/src/Dialog/index.tsx
+++ b/packages/ui-dialog/src/Dialog/index.tsx
@@ -23,6 +23,7 @@
  */
 
 import { Component } from 'react'
+import { css } from '@instructure/emotion'
 
 import { omitProps, getElementType } from '@instructure/ui-react-utils'
 import { findDOMNode, requestAnimationFrame } from '@instructure/ui-dom-utils'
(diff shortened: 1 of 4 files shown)


packages/ui-dialog/src/Dialog/index.tsx
```diff
@@ -173,7 +174,11 @@ class Dialog extends Component<DialogProps> {
           role === 'dialog' && this.props.shouldContainFocus ? true : undefined
         }
         className={this.props.className} // TODO in V2 remove className, there is no usage of it.
-        style={{ borderRadius: 'inherit' }} // ensure the dialog inherits border radius from View
+        css={css`
```
matyasf: so as I see the goal is here to decrease specificity. There seems to be no need for `@layer`, just using
```
css={css`{border-radius: inherit;}`}
```
 seems to merge the CSS for me

packages/ui-dialog/src/Dialog/index.tsx
```diff
@@ -23,6 +23,7 @@
  */
 
 import { Component } from 'react'
+import { css } from '@emotion/react'
```
matyasf: this needs to be added to the dependencies

matyasf (changes requested):

packages/ui-dialog/src/Dialog/index.tsx
HerrTopi: No need for the import

packages/ui-dialog/src/Dialog/index.tsx
HerrTopi: Good catch, it seems not even the css`{ part needed. Emotion resolves nested rules backwards imho.

matyasf (approved):

git-nandor (approved):

PR #2597, opened 2026-06-19
Add missing type exports to emotion and the ui packages; remove not needed dependencies

matyasf:
The added exports are used by public props, Canvas codebase exposed, that they should be exported

To test: just check that the build and the docs app is OK


packages/ui/src/v11_6.ts
```diff
@@ -460,6 +469,10 @@ export type {
   TopNavBarActionItemsOwnProps,
   TopNavBarBrandProps,
   TopNavBarBrandOwnProps,
+  /**
+   * This is an alias for the `ItemChild` type in `TopNavBarItem`
+   */
+  ItemChild as TopNavBarItemChild,
```
matyasf, line 475: We should be more careful with type/component names in the future, so its clear to see which component it belongs to

joyenjoyer (approved):

HerrTopi (approved):

PR #2599, opened 2026-06-19
INSTUI-5072 Restore Modal open/close transition animation

matyasf:
## Summary
- Move `ModalContext.Provider` outside `Transition` (v1 + v2) so the dialog/Mask is `Transition`'s direct child again and the fade lifecycle runs.
- v2: pass `open` to the `Dialog` unconditionally (was gated on `!transitioning`) so the window renders during the transition and fades together with the Mask.
- v1 + v2: release focus from the `Dialog` at the start of the exit transition so focus isn't retained inside the `aria-hidden` fading subtree (fixes the retained-focus console warning).
- docs: fix the `Mask` z-index theme-override example in the v2 README.

## Test Plan
- On the Modal docs page, open and close the first example: both the Mask and the window should fade in and out together (~300ms), not appear/disappear instantly.
- Close the Modal via the Close button (or Esc / click-away) with focus inside it and confirm no `aria-hidden`/retained-focus warning in the console, and that focus returns to the trigger.
- Spot-check with a screen reader that focus moves into the Modal on open and back to the trigger on close.

Fixes INSTUI-5072

### DIFF
--- a/packages/ui-modal/src/Modal/v2/README.md
+++ b/packages/ui-modal/src/Modal/v2/README.md
@@ -618,16 +618,14 @@ You can do this with the `insertAt` prop or a theme override:
   type: code
 ---
 <InstUISettingsProvider
-      theme={{
-        themeOverrides: {
-            componentOverrides: {
-              Mask: {
-                zIndex: 555,
-              }
-            }
-          }
-      }}
-    >
+  themeOverride={{
+    components: {
+      Mask: {
+        zIndex: 555
+      }
+    }
+  }}
+>
   <Modal />
 </InstUISettingsProvider>
 ```
--- a/packages/ui-modal/src/Modal/v1/index.tsx
+++ b/packages/ui-modal/src/Modal/v1/index.tsx
@@ -107,6 +107,14 @@ class Modal extends Component<ModalProps, ModalState> {
   componentDidUpdate(prevProps: ModalProps) {
     if (this.props.open !== prevProps.open) {
       this.setState({ transitioning: true, open: !!this.props.open })
+      // When closing, release focus from the Dialog.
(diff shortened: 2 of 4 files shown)


packages/ui-modal/src/Modal/v2/README.md
```diff
@@ -618,16 +618,12 @@ You can do this with the `insertAt` prop or a theme override:
...
-            }
-          }
-      }}
-    >
+  themeOverride={{
+    Mask: {
+      zIndex: 555
+    }
+  }}
+>
```
matyasf, line 628: Note: This token does not exist yet in `v2`, I've asked Adam to create it

packages/ui-modal/src/Modal/v2/README.md
```diff
@@ -618,16 +618,12 @@ You can do this with the `insertAt` prop or a theme override:
...
-        themeOverrides: {
-            componentOverrides: {
-              Mask: {
-                zIndex: 555,
-              }
-            }
-          }
-      }}
-    >
+  themeOverride={{
```
balzss, line 621: i don't think this is the right syntax. shouldn't it be:
```tsx
<InstUISettingsProvider
  themeOverride={{
    components: {
      Mask: {
        zIndex: 555
      }
    }
  }}
>
  <Modal />
</InstUISettingsProvider>
```

balzss (commented):
looks good besides the themeoverride example in the readme

packages/ui-modal/src/Modal/v2/README.md
matyasf: fixed it

balzss (approved):
looks good now

joyenjoyer (approved):

PR #2601, opened 2026-06-23
build: run the Babel and TypeScript builds in parallel during bootstrap

joyenjoyer:
## Summary
- Run the Babel package build and the `tsc -b` declaration build concurrently during `bootstrap` (`concurrently --kill-others-on-fail`) -> this makes the build run 20-30% faster!!!

## Test Plan
- Run `pnpm run bootstrap` from a clean tree and confirm it completes

Fixes INSTUI-5075

### DIFF
--- a/scripts/bootstrap.js
+++ b/scripts/bootstrap.js
@@ -66,16 +66,21 @@ execSync(
   opts
 )
 
-mark('Building packages with Babel')
-execSync('pnpm run build', opts)
+// The Babel build and the TypeScript declaration build
+// (tsc -b, src -> .d.ts) are mutually independent, so we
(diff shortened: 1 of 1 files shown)


balzss (approved):
tested it multiple times both on this branch and master for comparison and based on the results it saves us around 30s right now and could save even more migrating away from babel and tsc

my results: the non-parallel version was about 50s (babel) + 45s (tsc) = 90s together, this version is usually around 60s for the combined time (i guess the extra 10s is because the cpu is more under load)

question: this version added an extra 15s for fetching design tokens which the master branch did not do, i guess rebasing will solve that?

HerrTopi (approved):

PR #2602, opened 2026-06-23
Breakpoints are set from sharedTokens

matyasf:
- FormFieldLayout and Grid now uses design tokens from sharedTokens to set breakpoint values.
- Document app has fewer CSS breakpoints and their value are the same as the breakpoints in SharedTokens.
- update button example in regression test with more valuers

To test:
- check the docs app at different sizes and on mobile that it looks OK.
- check FormFieldLayout and Grid docs pages that they look OK.
- check the new Button examples in the regression test

INSTUI-5074

### DIFF
--- a/docs/upgrading/upgrade-guide.md
+++ b/docs/upgrading/upgrade-guide.md
@@ -695,7 +695,8 @@ type: embed
   removed={[
     {name:"spacing",note:""},
     {name:"inlinePadding",note:""},
-    {name:"asteriskColor",note:""}
+    {name:"asteriskColor",note:""},
+    {name:"stackedOrInlineBreakpoint",note:"uses now sharedTokens.breakpoints.md"}
   ]}
   changed={[
     {oldName:"color",newName:"textColor",note:""}
@@ -746,10 +747,10 @@ type: embed
 type: embed
 ---
 <V12ChangelogTable
-  changed={[
-    {oldName:"mediumMin",newName:"mediumMin",note:"now only provides the value, not the full media query (e.g., '768px' instead of 'min-width: 768px'). The component now adds the `min-width:` prefix automatically in the generated styles."},
-    {oldName:"largeMin",newName:"largeMin",note:"now only provides the value, not the full media query (e.g., '1024px' instead of 'min-width: 1024px'). The component now adds the `min-width:` prefix automatically in the generated styles."},
-    {oldName:"xLargeMin",newName:"xLargeMin",note:"now only provides the value, not the full media query (e.g., '1440px' instead of 'min-width: 1440px'). The component now adds the `min-width:` prefix automatically in the generated styles."}
+  removed={[
+    {name:"mediumMin",note:"value is read from sharedTokens.breakpoints.md"},
+    {name:"largeMin",note:"value is read from sharedTokens.breakpoints.lg. Its value is now 64em instead of 62em."},
+    {name:"xLargeMin",note:"value is read from sharedTokens.breakpoints.xl. Its value is now 80em instead of 75em."}
   ]}
 />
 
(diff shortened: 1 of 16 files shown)


packages/__docs__/src/App/index.tsx
```diff
@@ -97,16 +96,12 @@ import { AppContext } from '../appContext'
...
   static displayName = 'App'
-  static allowedProps = allowedProps
   static contextType = AppContext
   declare context: React.ContextType<typeof AppContext>
 
   private navRef = createRef<HTMLElement>()
   private navFocusRegion: FocusRegion | null = null
 
-  static defaultProps = {
-    trayWidth: 300
```
matyasf, line 108: For some reason from ancient times, the App component had this prop. It's now hardcoded

adamlobler (approved):

HerrTopi (approved):

docs/upgrading/upgrade-guide.md
```diff
@@ -696,6 +696,7 @@ type: embed
     {name:"spacing",note:""},
     {name:"inlinePadding",note:""},
     {name:"asteriskColor",note:""}
```
git-nandor: A missing trailing comma breaks the rendering.

Image

docs/upgrading/upgrade-guide.md
```diff
@@ -746,10 +747,10 @@ type: embed
...
 ---
 <V12ChangelogTable
-  changed={[
-    {oldName:"mediumMin",newName:"mediumMin",note:"now only provides the value, not the full media query (e.g., '768px' instead of 'min-width: 768px'). The component now adds the `min-width:` prefix automatically in the generated styles."},
-    {oldName:"largeMin",newName:"largeMin",note:"now only provides the value, not the full media query (e.g., '1024px' instead of 'min-width: 1024px'). The component now adds the `min-width:` prefix automatically in the generated styles."},
-    {oldName:"xLargeMin",newName:"xLargeMin",note:"now only provides the value, not the full media query (e.g., '1440px' instead of 'min-width: 1440px'). The component now adds the `min-width:` prefix automatically in the generated styles."}
+  removed={[
+    {name:"mediumMin",note:"value is read from sharedTokens.breakpoints.md"},
+    {name:"largeMin",note:"value is read from sharedTokens.breakpoints.lg"},
+    {name:"xLargeMin",note:"value is read from sharedTokens.breakpoints.xl"}
```
git-nandor: These largeMin/xLargeMin notes only say the value now comes from sharedTokens, but the actual breakpoints also shift (large 62em → 64em, xLarge 75em → 80em) for Grid, Grid.Col and Grid.Row — might be worth mention the value change.

packages/__docs__/src/App/index.tsx
```diff
@@ -122,10 +117,8 @@ class App extends Component<AppProps, AppState> {
     // determine what page we're loading
     const [page] = this.getPathInfo()
 
-    // if there's room for the tray + 700px, load with the tray open (unless it's the homepage)
-    const smallerScreen = window.matchMedia(
-      `(max-width: ${props.trayWidth + 700}px)`
-    ).matches
+    // if there's room for the tray + 724px, load with the tray open (unless it's the homepage)
+    const smallerScreen = window.matchMedia(`(max-width: 1024px)`).matches
```
git-nandor, line 118: Since this is now a hardcoded 1024px rather than a calculation based on trayWidth, the "tray + 724px" comment is a bit confusing

git-nandor (changes requested):

docs/upgrading/upgrade-guide.md
matyasf: good catch, fixed

docs/upgrading/upgrade-guide.md
matyasf: added it

packages/__docs__/src/App/index.tsx
matyasf: Actually this was a weird logic. Now we only hide the side navigation if the user is on smaller screens (we always show it on the homepage, I have no idea why this way not the case before)

git-nandor (approved):

PR #2604, opened 2026-06-23
feat(ui-view,emotion): add shared design token values to View v2 style props

HerrTopi:
## Description

Extends **View v2** (and the shared `ThemeablePropValues` enums) so the style props accept the new shared design tokens in addition to the existing legacy values. All legacy values keep working, so this is purely additive for consumers.

New values now accepted by View v2:

- **`shadow`** — `elevation1`, `elevation2`, `elevation3`, `elevation4` (alongside legacy `resting` / `above` / `topmost` / `none`).
- **`borderWidth`** — `sm`, `md`, `lg` (alongside legacy `small` / `medium` / `large`).
- **`borderRadius`** — the shared radius scale `xs`, `sm`, `md`, `lg`, `xl`, `xxl`, `full` plus the card radii `card.sm` / `card.md` / `card.lg` and `card.nestedContainer.sm` / `.md` / `.lg` (alongside legacy `small` / `medium` / `large`, `circle`, `pill`, and raw CSS lengths).
- **`borderColor`** — the shared stroke tokens `strongColor`, `visualSeparator` and the accent palette (`accentAsh`, `accentAurora`, `accentBlue`, `accentGreen`, `accentGrey`, `accentHoney`, `accentOrange`, `accentPlum`, `accentRed`, `accentSea`, `accentSky`, `accentStone`, `accentViolet`) — in addition to the legacy contextual colors and any valid CSS color string.

Supporting changes:

- `packages/emotion` — added the new `elevation1`–`elevation4` shadow values and `sm` / `md` / `lg` border-width values to `ThemeablePropValues`.
- **v1 components** (`View` v1, `ContextView`, `Popover`, `Expandable`) — the `shadow` prop type is inlined to the legacy union (`'resting' | 'above' | 'topmost' | 'none'`) so the v1 API stays unchanged and does not surface the new elevation values.
- `View/v2/styles.ts` — maps each new value to the corresponding `sharedTokens` entry.
- Updated the View v2 README with examples and prop documentation for the new values.

No breaking changes — existing prop values are preserved.

## Test plan

- [x] `pnpm run test:vitest ui-view` passes.
- [x] In the docs app (`pnpm run dev`), open the **View** page and verify the shadow, borderWidth, borderRadius, and borderColor examples render, including the newly added `elevation4`, `card.md`, `lg` radius, and accent color examples.
- [x] Legacy values (`shadow="resting"`, `borderWidth="small"`, `borderRadius="medium"`, `borderColor="primary"`) still render identically to before.
- [ ] New values apply the expected shared-token styles (elevation shadows, stroke widths, radius scale, accent stroke colors).
- [ ] v1 `View` / `ContextView` / `Popover` / `Expandable` still type-check and their `shadow` prop only accepts the legacy values.
- [ ] TypeScript build is clean (`pnpm run ts:check` / references check).

INSTUI-4945

### DIFF
--- a/packages/emotion/src/index.ts
+++ b/packages/emotion/src/index.ts
@@ -61,8 +61,6 @@ export type {
   Shadow,
   Stacking,
   Background,
-  BorderRadiiValues,
-  BorderRadii,
   BorderWidthValues,
   BorderWidth
 } from './styleUtils'
--- a/packages/emotion/src/styleUtils/ThemeablePropValues.ts
+++ b/packages/emotion/src/styleUtils/ThemeablePropValues.ts
@@ -29,7 +29,11 @@ const ThemeablePropValues = {
     resting: 'resting',
     above: 'above',
     topmost: 'topmost',
-    none: 'none'
+    none: 'none',
+    elevation1: 'elevation1',
+    elevation2: 'elevation2',
+    elevation3: 'elevation3',
+    elevation4: 'elevation4'
   },
 
   STACKING_TYPES: {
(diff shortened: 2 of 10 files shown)


balzss (approved):
the changes looks good and work nicely, hence the approve but there are 2 minor things that i think would be a nice addition:
1. add some examples with the new values to the visual regression test, maybe in collaboration with design so they can tell us what they'd like to see there
2. please remove the lorem ipsum generation from the view examples, they make it really annoying to test examples

matyasf (changes requested):
- **v1 View and PopOver's `borderWidth` type allows the new values when it should not.** 
- in the View's `borderRadius` example only show non-deprecated values.
- in the View's `borderColor` example do not show the legacy values
- Please note with a comment in `ThemeablePropValues` which values became deprecated
- `type BorderRadii` is broken, I'd just make it a `string`. Its also used for V1 (but I would just delete it, or make a proper type -- which is tricky because we allow CSS length values too)


I agree with Balazs, that some explanation to these values and closer to real-life examples would be nice. (this can be a future ticket)

matyasf (approved):

PR #2606, opened 2026-06-24
feat(ui-text-input,ui-select): add renderBeforeInputElementGap prop for even wrapped-tag padding in multiple Select

ToMESSKa:
INSTUI-5094 (includes regression-test font-determinism fix)

INSTUI-5079

ISSUE:
- multiple select with multiple rows of Tags, the bottom and top margins aren't consistent
image

-TEST PLAN:
- check the code below if the examples have consistent top and bottom padding in the legacy and new themes
- check the disabled state in the code examples too, the tag spacing should work as well
- Tags in Select/TextInput should work the same when the new prop is not set and the current change should cause a breaking change

```
const MultiSelectSizeExample = ({ size, options }) => {
    const [inputValue, setInputValue] = useState('')
    const [isShowingOptions, setIsShowingOptions] = useState(false)
    const [highlightedOptionId, setHighlightedOptionId] = useState(null)
    const [selectedOptionId, setSelectedOptionId] = useState(
      // the large version starts with one fewer (5) selected state
      size === 'large'
        ? ['opt1', 'opt2', 'opt3', 'opt4', 'opt5']
        :  ['opt1', 'opt2', 'opt4', 'opt5', 'opt6', 'opt7']
    )
    const inputRef = useRef()

    const getOptionById = (id) => options.find((o) => o.id === id)

    const availableOptions = () =>
      options.filter((o) => !selectedOptionId.includes(o.id))

    const focusInput = () => {
      if (inputRef.current) {
        inputRef.current.blur()
        inputRef.current.focus()
      }
    }

    const handleSelectOption = (event, { id }) => {
      if (!getOptionById(id)) return
      focusInput()
      setSelectedOptionId([...selectedOptionId, id])
      setHighlightedOptionId(null)
      setInputValue('')
      setIsShowingOptions(false)
    }

    const handleKeyDown = (event) => {
      // remove last selected option on backspace, if input has no entered text
      if (event.keyCode === 8 && inputValue === '' && selectedOptionId.length > 0) {
        setHighlightedOptionId(null)
        setSelectedOptionId(selectedOptionId.slice(0, -1))
      }
    }

    const dismissTag = (e, tag) => {
      // prevent closing of list
      e.stopPropagation()
      e.preventDefault()
      setSelectedOptionId(selectedOptionId.filter((id) => id !== tag))
      setHighlightedOptionId(null)
      inputRef.current.focus()
    }

    const renderTags = () =>
      selectedOptionId.map((id) => (
        <Tag
          dismissible
          key={id}
          size={size}
          text={
            <AccessibleContent alt={`Remove ${getOptionById(id).label}`}>
              {getOptionById(id).label}
            </AccessibleContent>
          }
          onClick={(e) => dismissTag(e, id)}
        />
      ))

    return (
      <Select
        htmlSize={2}
        renderLabel={`Multiple Select (${size})`}
        size={size}
        width="30rem"
        renderBeforeInputElementGap="even"
        assistiveText="Type or use arrow keys to navigate options. Multiple selections allowed."
        inputValue={inputValue}
        isShowingOptions={isShowingOptions}
        inputRef={(el) => {
          inputRef.current = el
        }}
        onInputChange={(e) => setInputValue(e.target.value)}
        onRequestShowOptions={() => setIsShowingOptions(true)}
        onRequestHideOptions={() => setIsShowingOptions(false)}
        onRequestHighlightOption={(e, { id }) => setHighlightedOptionId(id)}
        onRequestSelectOption={handleSelectOption}
        onKeyDown={handleKeyDown}
        renderBeforeInput={selectedOptionId.length > 0 ? renderTags() : null}
      >
        {availableOptions().length > 0 ? (
          availableOptions().map((option) => (
            <Select.Option
              id={option.id}
              key={option.id}
              isHighlighted={option.id === highlightedOptionId}
            >
              {option.label}
            </Select.Option>
          ))
        ) : (
          <Select.Option id="empty-option" key="empty-option">
            ---
          </Select.Option>
        )}
      </Select>
    )
  }

  const states = [
    { id: 'opt1', label: 'Alabama' },
    { id: 'opt2', label: 'Alaska' },
    { id: 'opt3', label: 'Arizona' },
    { id: 'opt4', label: 'Arkansas' },
    { id: 'opt5', label: 'California' },
    { id: 'opt6', label: 'Colorado' },
    { id: 'opt7', label: 'Connecticut' },
    { id: 'opt8', label: 'Delaware' },
    { id: 'opt9', label: 'Florida' },
    { id: 'opt10', label: 'Georgia' }
  ]

  render(
    <View as="div">
      <MultiSelectSizeExample size="small" options={states} />
      <View as="div" margin="medium 0 0 0">
        <MultiSelectSizeExample size="medium" options={states} />
      </View>
      <View as="div" margin="medium 0 0 0">
        <MultiSelectSizeExample size="large" options={states} />
      </View>
    </View>
  )
```


packages/ui-select/src/Select/v2/README.md
```diff
@@ -546,6 +543,7 @@ type: example
       <div>
         <Select
           renderLabel="Multiple Select"
+          contentSpacing="even"
```
git-nandor: When contentSpacing="even" is set, the even tag spacing only works while the input is enabled, because beforeElement is display: contents only in the enabled state — in the disabled state it becomes a plain block, so the tags are no longer individual flex items.



Image

git-nandor (changes requested):

packages/ui-select/src/Select/v2/README.md
ToMESSKa: @git-nandor Nice catch, I fixed it, can you please check again?

git-nandor (approved):

packages/ui-text-input/src/TextInput/v2/props.ts
```diff
@@ -121,6 +121,13 @@ type TextInputOwnProps = {
    */
   shouldNotWrap?: boolean
 
+  /**
+   * How wrapped `renderBeforeInput` content (e.g. tags) is spaced.
+   * - `'default'`: wrapped rows are not evenly padded.
+   * - `'even'`: wrapped rows get even padding.
+   */
+  contentSpacing?: 'default' | 'even'
```
matyasf: An example here in TextInput would be nice

packages/ui-select/src/Select/v2/props.ts
```diff
@@ -248,6 +248,13 @@ type PropsFromTextInput = {
    */
   shouldNotWrap?: boolean
 
+  /**
+   How wrapped `renderBeforeInput` content (e.g. tags) is spaced.
+   * - `'default'`: wrapped rows are not evenly padded.
+   * - `'even'`: wrapped rows get even padding.
```
matyasf: I'd rephrase:

Whether to apply a gap between `renderBeforeInput` (which is a CSS Flex container) elements.
- 'default' (or no value): no gap is applied between the items in RenderBeforeInput .
- 'even': Wrapped rows are evenly spaced by adding a vertical and horizontal gap.

I'd also rename this to `renderBeforeInputElementGap`

matyasf (changes requested):
Also change the example in TextInput to use this prop, change the example not to use `margin="xxx-small xxx-small xxx-small none"`

packages/ui-text-input/src/TextInput/v2/props.ts
ToMESSKa: I updated the tag example in TextInput (the one under Prepending and appending content) using the new prop.

packages/ui-select/src/Select/v2/props.ts
ToMESSKa: I renamed the prop and updated the prop descriptions in TextInput and Select

matyasf (approved):

packages/ui-select/src/Select/v2/README.md
```diff
@@ -531,6 +529,7 @@ type: example
           onRequestSelectOption={handleSelectOption}
           onKeyDown={handleKeyDown}
           renderBeforeInput={selectedOptionId.length > 0 ? renderTags() : null}
+          htmlSize={2}
```
ToMESSKa, line 532: Dana noticed that sometimes the cursor jumps to a new line even when there is plenty of space remaining after the last tag on the previous line (e.g. in the picture). By default, the `<input>` element requires about 20 characters of space before wrapping, which is why it often forces a new line. To fix this, I reduced the amount of space needed by using this prop in this example.

Image

PR #2612, opened 2026-06-26
fix(ui-drilldown): disabled options no longer show hover/focus styles

ToMESSKa:
Closes: INSTUI-5077

**ISSUE:**
- disabled Drilldown options currently look like enabled options when hovered with the mouse or focused with the keyboard, they should not recieve hover of focus styles

**TEST PLAN:**
- in Drilldown v1 and v2 check the first two examples under the 'Options' header with a mouse and keyboard 
- the disabled options should not get styling when they get hovered or keyboard focused


balzss (approved):
works great, nice job

matyasf (approved):

PR #2616, opened 2026-07-01
fix(ui-select): fix screen reader labels on multiple select

git-nandor:
INSTUI-5076

## Summary

When a Select renders selected options as dismissible Tag pills their "Remove …" labels always announced.

## Changes
The multiple-select example sets an explicit aria-label on the Select, which overrides the wrapping-<label> name computation, so the pills are excluded from the combobox name — while each pill keeps its own "Remove …" name, so it is still announced when focused directly.

## Test plan
Uses the "Highlighting and selecting options" multiple-select example on the Select v2 docs page.
- Tab to the Select → announces "Multiple Select, Alaska Selected, …", not announces "Remove …".
- Navigate onto a pill → announces "Remove Alaska"

### DIFF
--- a/packages/ui-select/src/Select/v2/index.tsx
+++ b/packages/ui-select/src/Select/v2/index.tsx
@@ -30,8 +30,7 @@ import {
   matchComponentTypes,
   omitProps,
   getInteraction,
-  withDeterministicId,
-  callRenderProp
+  withDeterministicId
 } from '@instructure/ui-react-utils'
 import {
   getBoundingClientRect,
@@ -184,9 +183,6 @@ class Select extends Component<SelectProps> {
   ref: HTMLSpanElement | null = null
   _input: HTMLInputElement | null = null
   private _defaultId = this.props.deterministicId!()
-  // Dedicated id for a hidden label element used as the combobox's accessible
-  // name when content is rendered before the input (e.g. multiple-select pills).
-  private _labelId = this.props.deterministicId!()
   private _inputContainer: HTMLSpanElement | null = null
   private _listView: Element | null = null
   // temporarily stores actionable options
@@ -753,17 +749,6 @@ class Select extends Component<SelectProps> {
       overrideProps.autoComplete = passthroughProps['autoComplete']
     }
 
-    // When content is rendered before the input via `renderBeforeInput` (e.g.
-    // the dismissible Tag "pills" of a multiple select), that content lives
-    // inside the wrapping <label> and would otherwise be pulled into the
-    // combobox's accessible name (announced as e.g.
-    // "Multiple Select Remove Alaska Remove American Samoa ..."). Point the
-    // input's name at a dedicated hidden label element holding only the field
-    // label, so the pills are only read when focused directly.
-    if (renderBeforeInput != null) {
-      overrideProps['aria-labelledby'] = this._labelId
-    }
-
     const inputProps: Partial<TextInputProps> = {
       id: this.id,
       renderLabel,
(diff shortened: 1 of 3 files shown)


ToMESSKa (changes requested):
@git-nandor I really like the approach.
But I'm thinking maybe there are some use cases when there is genuine useful information in the renderBeforeInput that now gets ignored?
Like try this example before and after your fix, the +1 country code is ignored.
```
const CountryCodeExample = () => {
    const options = [
      { id: 'us', label: 'United States', dial: '+1' },
      { id: 'hu', label: 'Hungary', dial: '+36' },
      { id: 'jp', label: 'Japan', dial: '+81' },
      { id: 'br', label: 'Brazil', dial: '+55' }
    ]

    const [isShowingOptions, setIsShowingOptions] = useState(false)
    const [highlightedOptionId, setHighlightedOptionId] = useState(null)
    const [selectedOptionId, setSelectedOptionId] = useState('us')
    const inputRef = useRef()

    const getOptionById = (id) => options.find((o) => o.id === id)
    const selected = getOptionById(selectedOptionId)

    const focusInput = () => {
      if (inputRef.current) {
        inputRef.current.blur()
        inputRef.current.focus()
      }
    }

    return (
      <div>
        <Select
          renderLabel="Country code"
          assistiveText="Use arrow keys to navigate options."
          inputValue={selected.label}
          isShowingOptions={isShowingOptions}
          inputRef={(el) => {
            inputRef.current = el
          }}
          onRequestShowOptions={() => setIsShowingOptions(true)}
          onRequestHideOptions={() => {
            setIsShowingOptions(false)
            setHighlightedOptionId(null)
          }}
          onRequestHighlightOption={(e, { id }) => setHighlightedOptionId(id)}
          onRequestSelectOption={(e, { id }) => {
            focusInput()
            setSelectedOptionId(id)
            setIsShowingOptions(false)
          }}
          // MEANINGFUL prefix — the only place the selected dial code is shown.
          renderBeforeInput={
            <span aria-hidden={false} style={{ whiteSpace: 'nowrap' }}>
              {selected.dial}
            </span>
          }
        >
          {options.map((option) => (
            <Select.Option
              id={option.id}
              key={option.id}
              isHighlighted={option.id === highlightedOptionId}
              isSelected={option.id === selectedOptionId}
            >
              {`${option.label} (${option.dial})`}
            </Select.Option>
          ))}
        </Select>
      </div>
    )
  }

  render(<CountryCodeExample />)
```

git-nandor:
Good catch — you're right. The fix was too broad: it stripped all renderBeforeInput from the accessible name. I've reverted the automatic name override in Select so meaningful prefixes stay in the name. Updated the docs example and tests accordingly.

ToMESSKa (approved):

packages/ui-select/src/Select/v2/index.tsx
```diff
@@ -111,18 +111,18 @@ const MemoedOption = memo(
...
+        nextProps.selectOption.props.isDisabled &&
       prevProps.selectOption.props.children ===
-      nextProps.selectOption.props.children &&
+        nextProps.selectOption.props.children &&
       prevProps.selectOption.props.id === nextProps.selectOption.props.id &&
       prevProps.selectOption.props.renderBeforeLabel ===
-      nextProps.selectOption.props.renderBeforeLabel &&
+        nextProps.selectOption.props.renderBeforeLabel &&
       prevProps.selectOption.props.renderAfterLabel ===
-      nextProps.selectOption.props.renderAfterLabel &&
```
matyasf, line 125: FYI your formatter is OK, it was broken here: https://github.com/instructure/instructure-ui/commit/cda547ac0af8664831097598be23dc0eb989602a#diff-86c34ad67bd876370225bf0df1e248d809d3ac768f41267d08d576d0ef2d0ec1

matyasf (approved):

PR #2617, opened 2026-07-01
fix(ui-buttons): fix ai-secondary button size diff

git-nandor:
INSTUI-5061

### Summary
An ai-secondary button was 2px taller (and, for icon-only buttons, 2px wider) than the same button in any other color.

ai-secondary removes the real border and renders its gradient "border" as a ::before ring whose space is reserved by padding. That padding is additive to the content's full minHeight, adding 2px to the outer size.

The fix shrinks the ai-secondary content box by 2 × borderWidth (height, plus width for icon-only buttons), so the reserved padding is compensated and the outer size is identical regardless of color.

### Update (review feedback)
- Condensed sizes also covered. 
- Refactor. Collapsed the duplicated calc / size list / height tokens into a single `aiSecondaryContentSize` const.
- Regression coverage. The button page now renders every size's ai-secondary button next to a default one so the screenshot catches any size drift, condensed included.

### Test:
Verify that the buttons have the same size.

````
{[
  'small',
  'medium',
  'large',
  'condensedSmall',
  'condensedMedium'
].map((size) => (
  <div
    key={size}
  >
    <h3>{size}</h3>

    {/* text: ai-secondary vs default */}
    <Button size={size} color="ai-secondary">{size}</Button>
    <Button size={size} color="secondary">{size}</Button>

    {/* icon-only: ai-secondary vs default (width parity) */}
    <IconButton size={size} color="ai-secondary" screenReaderLabel="Add">
      <IconAddLine />
    </IconButton>
    <IconButton size={size} color="secondary" screenReaderLabel="Add">
      <IconAddLine />
    </IconButton>

    {/* icon-only circle ai-secondary */}
    <IconButton size={size} shape="circle" color="ai-secondary" screenReaderLabel="Add">
      <IconAddLine />
    </IconButton>
    <IconButton size={size} shape="circle" color="secondary" screenReaderLabel="Add">
      <IconAddLine />
    </IconButton>
  </div>
))}
````

### DIFF
--- a/packages/ui-buttons/src/BaseButton/v2/styles.ts
+++ b/packages/ui-buttons/src/BaseButton/v2/styles.ts
@@ -55,6 +55,26 @@ const generateStyle = (
     condensedMedium: componentTheme.gapButtonContentSm
   }
 
+  // Outer height token per size, keyed like `sizeVariants`. Used to compensate
+  // the ai-secondary gradient-border padding.
+  const heightForSize = {
+    small: componentTheme.smallHeight,
+    medium: componentTheme.mediumHeight,
+    large: componentTheme.largeHeight,
+    condensedSmall: componentTheme.heightXxs,
+    condensedMedium: componentTheme.heightXs
+  }
+
+  // ai-secondary draws its gradient "border" as a ::before ring whose space is
+  // reserved by padding on the button wrapper (see the `&&&&&&&&&&` block below).
+  // That padding is additive to the content box, so an ai-secondary button ends
+  // up 2 × borderWidth taller — and, when icon-only, wider — than the same button
+  // in any other color. Shrink the content box by the same amount to compensate.
+  const aiSecondaryContentSize =
+    color === 'ai-secondary'
+      ? `calc(${heightForSize[size!]} - ${componentTheme.borderWidth} * 2)`
+      : undefined
+
   const shapeVariants = {
     circle: { borderRadius: componentTheme.borderRadiusFull },
     rectangle: {}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by balzss** [comment]
Nice fix, root cause is spot on and I checked that small/medium/large are pixel-perfect now (both icon and text buttons).

**Condensed sizes are still off.** The fix only kicks in for `small`/`medium`/`large`, but the wrapper padding at line 477 gets added for every size — so `condensedSmall` and `condensedMedium` ai-secondary buttons are still 2px taller and wider. Confirmed by measuring. Since condensed is a real combo we need to cover it too — heads up, condensed uses `height` instead of `minHeight`, so it needs its own little branch subtracting `2 × borderWidth` there.

**Add a case to the regression test suite** — render the ai-secondary buttons next to the default ones (all sizes) on the button page so the screenshot catches any size drift.

**Small cleanup on the code itself.** Right now the calc is written twice (minHeight + width), the size list is repeated (once as the `heightForAiSize` keys, once in the `size === ...` check), and `heightForAiSize` duplicates height tokens we already have. You can collapse all three — pull the value into one const and index the theme directly:

```js
const aiContentSize =
  color === 'ai-secondary' && componentTheme[`${size}Height`]
    ? `calc(${componentTheme[`${size}Height`]} - ${componentTheme.borderWidth} * 2)`
    : undefined
```

then use `aiContentSize` for both `minHeight` and the icon-only `width`. No map, no repeated calc, no repeated size list. (You'll still want the separate condensed branch since those don't have a `${size}Height` token.)

Heads up: this only works because it's spread last and overrides the `minHeight` from `sizeVariants` — that's the existing pattern here, just good to know.

balzss (commented):
see my (and claude's 😺 ) comment

git-nandor:
Thanks, the fix and the cleanup is done.

balzss (approved):
nice work

joyenjoyer (approved):
looks good

PR #2620, opened 2026-07-02
fix(ui-source-code-editor): apply current theme to v2 search panel

ToMESSKa:
INSTUI-5081

**ISSUE:**
- the current theme is never applied to SourceCodeEditor's search panel

**TEST PLAN:**
- check out the example under the Search header in SourceCodeEditor
- open the search panel in each theme using `cmd/ctrl+f `
- the current theme should be applied to the search panel and the dark theme should work too
- pressing `esc` should close the search panel

### DIFF
--- a/packages/ui-source-code-editor/src/SourceCodeEditor/v2/index.tsx
+++ b/packages/ui-source-code-editor/src/SourceCodeEditor/v2/index.tsx
@@ -23,6 +23,7 @@
  */
 
 import { Component } from 'react'
+import { createPortal } from 'react-dom'
 import { deepEqual as isEqual } from '@instructure/ui-utils'
 
 import { EditorSelection, EditorState, StateEffect } from '@codemirror/state'
@@ -89,7 +90,7 @@ import { textDirectionContextConsumer } from '@instructure/ui-i18n'
 
 import { withStyleNew } from '@instructure/emotion'
 
-import customSearch from './SearchPanel.js'
+import customSearch, { SearchPanel } from './SearchPanel.js'
 
 import generateStyle from './styles.js'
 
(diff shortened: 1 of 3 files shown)


packages/ui-source-code-editor/src/SourceCodeEditor/v2/index.tsx
```diff
@@ -687,6 +727,16 @@ class SourceCodeEditor extends Component<SourceCodeEditorProps> {
             css={styles?.codeEditorContainer}
           />
         </label>
+        {this.state.searchPanels.map(({ id, dom, view }) =>
+          createPortal(
+            <SearchPanel
+              view={view}
+              searchConfig={this.props.searchConfig!}
```
matyasf: This might be `undefined` if the prop's value changes runtime. Can you add something that only renders this if `searchConfig` is not null?

packages/ui-source-code-editor/src/SourceCodeEditor/v2/index.tsx
matyasf, line 370: add here `searchConfig` too

matyasf (changes requested):
- `SearchPanel` shows in the docs app, this is an internal component that should not show
- see my inline comments also

packages/ui-source-code-editor/src/SourceCodeEditor/v2/index.tsx
ToMESSKa: Added, thanks

packages/ui-source-code-editor/src/SourceCodeEditor/v2/index.tsx
ToMESSKa: Added

ToMESSKa:
> * `SearchPanel` shows in the docs app, this is an internal component that should not show
> * see my inline comments also

SearchPanel does not show in the docs app now, please check

matyasf (approved):

git-nandor (approved):

PR #2623, opened 2026-07-03
chore: update dependencies

matyasf:
Update most dependencies to their latest versions.

- Bump `vitest` from ^3.2.2 to ^4.1.9 across all packages. Adjust `vitest.setup.ts` and `tsconfig` files for the new major versions.
- Drop the deprecated `@lerna/project` dependency and read the project version
directly by walking up to the nearest `lerna.json`.
- Drop the deprecated `@lerna/package` dependency and implement it directly
- Delete `pkg-utils` and move its code to `ui-scripts/lib/utils/pkg-utils`
- Convert the root scripts/*.js to ESM (.mjs) so they can consume the ESM-only package, and update these

I suggest to look at the PR per commit, its divided into distinct parts

to test:
- smoke test docs app


packages/ui-dom-utils/src/__tests__/getScrollParents.test.tsx
```diff
@@ -27,7 +27,8 @@ import '@testing-library/jest-dom'
 import { getOffsetParents } from '../getOffsetParents.js'
 
 describe('getOffsetParents', () => {
-  // At rendering, we receive the node_wrapper and the body as extra offsetParents
+  // getOffsetParents returns only transformed or non-statically-positioned
+  // ancestors, plus the document body (always included as the last entry).
```
matyasf, line 31: This change is here because Jest was updated

packages/ui-scripts/lib/utils/npm.ts
```diff
@@ -29,32 +29,37 @@ import {
...
   info
 } from '@instructure/command-utils'
+import path from 'node:path'
+import fs from 'fs'
 
-// @ts-expect-error no types published for @lerna/project
-import { Project } from '@lerna/project'
-const syncRootPackageVersion = async (useProjectVersion: boolean) => {
-  const project = new Project(process.cwd())
+const syncRootPackageVersion = async () => {
```
matyasf, line 35: I've removed the deprecated `lerna/project` package and added code that has the same behaviour

packages/__docs__/package.json
```diff
@@ -113,15 +113,15 @@
     "@instructure/ui-utils": "workspace:*",
     "@instructure/ui-view": "workspace:*",
     "babel-loader": "^10.1.1",
-    "@babel/runtime": "^7.29.2",
+    "@babel/runtime": "^7.29.7",
     "@babel/standalone": "^7.29.3",
-    "codesandbox": "^2.2.3",
+    "codesandbox-import-utils": "^2.2.3",
```
matyasf, line 118: This package has wayyy less transitive dependencies

packages/ui-scripts/lib/utils/pkg-utils/package-utils.ts
```diff
@@ -0,0 +1,126 @@
...
+ * SOFTWARE.
+ */
+
+import fs from 'fs'
+import path from 'path'
+import { readPackageUpSync } from 'read-package-up'
+import type { PackageJson, NormalizeOptions } from 'read-package-up'
+import childProcess from 'child_process'
+
+export class Package {
```
matyasf, line 31: This is copied from Lerna's source code, they are no longer exporting this class

scripts/eslint/no-relative-imports-plugin.mjs
```diff
@@ -143,6 +147,16 @@ Object.assign(plugin.configs, {
...
+    settings: {
+      // The rule resolves import paths via eslint-module-utils. Without
+      // TypeScript extensions the default node resolver can't resolve .ts/.tsx
+      // files, resolve.default() returns undefined and the rule silently bails.
+      "import/resolver": {
+        node: {
+          extensions: [".js", ".jsx", ".ts", ".tsx", ".json"]
+        }
+      }
+    },
```
matyasf, line 159: While doing some refactoring here to remove `pkg-utils` it turned out that this plugin was not working at all :/ 

I really dont like `ESLint` its slow and full of ways where it fails silently

cypress/component/Tooltip.cy.tsx
```diff
@@ -158,7 +158,7 @@ describe('<Tooltip/>', () => {
     })
   })
 
-  it('should close when Esc key is pressed', async () => {
```
matyasf, line 161: This test was randomly failing because its async and does not have an await. 

This seems to be a systematic issue with Cypress tests, I'll fix them in a later PR

git-nandor (approved):

HerrTopi (approved):

PR #2625, opened 2026-07-06
fix(many): keep form field messages out of the control's accessible name

git-nandor:
INSTUI-4724

## Summary

Form control messages render inside the wrapping <label>, so screen readers were reading them as part of the field's accessible name (e.g. "Password Password must be at least 6 characters") instead of as a description.

Each single form control now points its accessible name at the label text only (aria-labelledby) and references the messages as its description (aria-describedby), so a screen reader announces the label first, then the messages. The wrapping <label> is unchanged, so click-on-label focus is preserved.

## Test plan

Tab through the given exaple and verify in DevTools Accessibility tab that the Name and Description values are correct. With a screen reader enabled, confirm that the label is announced as the field name, followed by a brief pause before the description is announced.

image

````
const messages = [{ type: 'error', text: 'This field has an error' }]

render(
  <View as="div" display="block" width="24rem">
    <Text as="p">
      Tab through and verify in DevTools Accessibility tab that the Name and Description values are correct. With a screen reader enabled, confirm that the label is announced as the field name, followed by a brief pause before the description is announced.
    </Text>

    <View as="div" margin="medium 0 0">
      <TextInput renderLabel="Text Input" messages={messages} />
    </View>
    <View as="div" margin="medium 0 0">
      <NumberInput renderLabel="Number Input" messages={messages} />
    </View>
    <View as="div" margin="medium 0 0">
      <TextArea label="Text Area" messages={messages} />
    </View>
    <View as="div" margin="medium 0 0">
      <Checkbox label="Checkbox (simple)" value="c1" messages={messages} />
    </View>
    <View as="div" margin="medium 0 0">
      <Checkbox
        label="Checkbox (toggle)"
        variant="toggle"
        value="c2"
        messages={messages}
      />
    </View>
    <View as="div" margin="medium 0 0">
      <RangeInput
        label="Range Input"
        name="range"
        min={0}
        max={100}
        defaultValue={50}
        messages={messages}
      />
    </View>
  </View>
)
````

### DIFF
--- a/packages/ui-checkbox/src/Checkbox/v2/index.tsx
+++ b/packages/ui-checkbox/src/Checkbox/v2/index.tsx
@@ -84,8 +84,12 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
     }
 
     this._defaultId = props.deterministicId!()
+    this._messagesId = props.deterministicId!('Checkbox-messages')
+    this._labelId = props.deterministicId!('Checkbox-label')
   }
   private readonly _defaultId: string
+  private readonly _messagesId: string
+  private readonly _labelId: string
   private _input: HTMLInputElement | null = null
 
   ref: Element | null = null
@@ -186,6 +190,10 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
     )
   }
 
+  get hasMessages() {
+    return !!this.props.messages && this.props.messages.length > 0
+  }
+
   get invalid() {
     return !!this.props.messages?.find(
       (m) => m.type === 'newError' || m.type === 'error'
@@ -217,6 +225,24 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
       `[Checkbox] The \`simple\` variant does not support the \`labelPlacement\` property.  Use the \`toggle\` variant instead.`
     )
 
+    // The label text gets its own id so that, when there are messages, the
+    // input's accessible name can point at the label only (via
+    // `aria-labelledby`) instead of also including the message text.
+    const labelContent = (
+      <span id={this._labelId}>
+        {label}
+        {isRequired && label && (
+          <span
+            css={this.invalid ? styles?.requiredInvalid : {}}
+            aria-hidden={true}
+          >
+            {' '}
+            *
+          </span>
+        )}
+      </span>
+    )
+
     if (variant === 'toggle') {
       return (
         <ToggleFacade
@@ -232,16 +258,7 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
           themeOverride={themeOverride}
           invalid={this.invalid}
         >
-          {label}
-          {isRequired && label && (
-            <span
-              css={this.invalid ? styles?.requiredInvalid : {}}
-              aria-hidden={true}
-            >
-              {' '}
-              *
-            </span>
-          )}
+          {labelContent}
         </ToggleFacade>
       )
     } else {
@@ -257,16 +274,7 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
           themeOverride={themeOverride}
           invalid={this.invalid}
         >
-          {label}
-          {isRequired && label && (
-            <span
-              css={this.invalid ? styles?.requiredInvalid : {}}
-              aria-hidden={true}
-            >
-              {' '}
-              *
-            </span>
-          )}
+          {labelContent}
         </CheckboxFacade>
       )
     }
@@ -286,7 +294,7 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
             : styles?.indentedError)
         }
       >
-        <FormFieldMessages messages={messages} />
+        <FormFieldMessages id={this._messagesId} messages={messages} />
       </View>
     ) : null
   }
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: packages/ui-text-input/src/TextInput/v2/index.tsx [hunk 1]
```diff
@@ -231,6 +233,20 @@ class TextInput extends Component<TextInputProps> {
     if (props['aria-describedby']) {
       descriptionIds = `${props['aria-describedby']}`
     }
+    // When there are messages, associate them with the input as its description
+    // (they are rendered by FormField with `id={this._messagesId}`) and point
+    // the accessible name at the label text only via `aria-labelledby`. This
+    // keeps the messages — which live inside the wrapping <label> — out of the
+    // control's name while preserving the click-on-label focus behavior.
+    // Any consumer-provided `aria-labelledby` (e.g. Select's pill handling)
+    // takes precedence.
```
ToMESSKa: I think the Select's pill handling doesn't use `aria-labelledby` anymore see https://github.com/instructure/instructure-ui/pull/2616#issuecomment-4896977216 so this is not a relevant example

ToMESSKa (changes requested):

packages/ui-text-input/src/TextInput/v2/index.tsx
git-nandor: Thanks, modified

ToMESSKa (changes requested):
A vitest test is failing, can you please check?

ToMESSKa (approved):

packages/ui-checkbox/src/Checkbox/v2/index.tsx
```diff
@@ -337,6 +345,24 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
...
+            // Keep messages in the description so the accessible name contains only the label.
+            aria-labelledby={
+              this.hasMessages
+                ? ((props as Record<string, unknown>)[
+                    'aria-labelledby'
+                  ] as string) || this._labelId
+                : ((props as Record<string, unknown>)['aria-labelledby'] as
+                    | string
+                    | undefined)
+            }
```
matyasf, line 352: These complex type casts are not nice and are not needed

```
        aria-labelledby={
          this.hasMessages
            ? props['aria-labelledby'] || this._labelId
            : props['aria-labelledby']
```
works too

packages/ui-checkbox/src/Checkbox/v2/index.tsx
```diff
@@ -337,6 +345,24 @@ class Checkbox extends Component<CheckboxProps, CheckboxState> {
...
+            aria-describedby={
+              [
+                (props as Record<string, unknown>)['aria-describedby'],
```
matyasf: cast is not needed

packages/ui-text-area/src/TextArea/v2/index.tsx
```diff
@@ -393,6 +401,22 @@ const TextArea = forwardRef<TextAreaElement, TextAreaProps>((props, ref) => {
...
+          .join(' ') || undefined
+      }
+      aria-labelledby={
+        hasMessages
+          ? ((rest as Record<string, unknown>)['aria-labelledby'] as string) ||
+            labelId
+          : ((rest as Record<string, unknown>)['aria-labelledby'] as
+              | string
+              | undefined)
+      }
```
matyasf, line 410: casts are not needed, e.g. 

```
hasMessages
          ? rest['aria-labelledby'] || labelId
          : rest['aria-labelledby']
```

packages/ui-text-input/src/TextInput/v2/index.tsx
```diff
@@ -231,6 +233,19 @@ class TextInput extends Component<TextInputProps> {
     if (props['aria-describedby']) {
       descriptionIds = `${props['aria-describedby']}`
     }
+    // When there are messages, associate them with the input as its description
+    // (they are rendered by FormField with `id={this._messagesId}`) and point
+    // the accessible name at the label text only via `aria-labelledby`. This
+    // keeps the messages — which live inside the wrapping <label> — out of the
+    // control's name while preserving the click-on-label focus behavior.
```
matyasf: why does this needs to be prevented?

matyasf (changes requested):
- dateTimeInput is wrong, see the example in the /forms page: aria-describedby and aria-labelledby reference nothing.
- While it works, I generally dislike this component structure. All this mess is needed, because we wrap everything in a <label>, the nice way would be not to do so. I think its OK for now, but we need to completely revamp this in the future. (also support SSR better, now the layout falls apart when loading)
- see my other comments

packages/ui-text-input/src/TextInput/v2/index.tsx
git-nandor: FormField renders this control and its `messages` inside a single wrapping
  <label>, so by default the messages' text becomes part of the control's
  accessible name

packages/ui-text-input/src/TextInput/v2/index.tsx
```diff
@@ -160,7 +160,12 @@ class TextInput extends Component<TextInputProps> {
   }
 
   get hasMessages() {
-    return !!this.props.messages && this.props.messages.length > 0
+    // FormField only renders messages that have text, so only treat the field
+    // as having messages then. Otherwise the input's `aria-describedby` would
+    // reference a messages element that was never rendered — e.g. DateTimeInput
+    // passes an empty-text error message to its sub-inputs just to force the
+    // invalid styling.
```
matyasf: Please make a bit more brief comments. Claude can be very very verbose and this just makes the code hard to read. For example this comment is totally unnecessary IMO, the code is pretty self-explanatory

packages/ui-text-input/src/TextInput/v2/index.tsx
```diff
@@ -231,6 +238,22 @@ class TextInput extends Component<TextInputProps> {
...
     }
+    // FormField renders this control and its `messages` inside a single wrapping
+    // <label>, so by default the messages' text becomes part of the control's
+    // accessible *name* (e.g. "Password Password must be at least 6 characters"),
+    // which is confusing and repeats on every announcement. Messages should be
+    // the field's *description* instead. So when there are messages, reference
+    // them via `aria-describedby` and pin the name to the label text only via
+    // `aria-labelledby`. Keeping the messages inside the <label> preserves the
+    // native click-on-label focus behavior. A consumer-provided `aria-labelledby`
+    // takes precedence.
```
matyasf: Again, way too verbose. I'd just write:

"Use the contents of the messages field as ARIA label; user supplied value takes precedence"

matyasf (changes requested):
looks good, just a small nitpick

packages/ui-text-input/src/TextInput/v2/index.tsx
git-nandor: I shortened it.

packages/ui-text-input/src/TextInput/v2/index.tsx
git-nandor: Ok, i removed this one

matyasf (approved):

PR #2629, opened 2026-07-08
Fix silently false-passing Cypress component tests

matyasf:
## Summary

Every test in `cypress/component` was written as `it('…', async () => { cy… })` but never awaited. Cypress commands are queued `Chainables`, not `Promise`s, so the resolved async callback let `Mocha` mark tests passed before the queued assertions ran — silent false-passes.

This PR removes the erroneous async, sets retries: 0, and fixes every failure that removal exposed.

**Changes**                                                                                                                                                                                                                                     
   
- Speed up the whole suite by using `ESBuild` and caching build output and by removing a `wait` from the `DateInput` test
- Deleted some tests, that were testing unnecessary things/impossible scenarios.   
- Removed unnecessary async from all cypress/component test callbacks; set retries: 0.                                                                                                                                                      
- Fixed failures uncovered across 52 spec files, mainly:                                                                                                                                                                                    
  - Outdated values referring to v1 components — updated tests for v2 code.                                                          
  - Measurement/timing races — measure after layout settles.                                                                                
  - Cross-test pollution — move the real-events cursor to 0,0 between tests / reset leaked state between tests.                                                                           
  - Wrong-element assertions — target the element that actually carries the behavior.
- Skipped `TruncateText` tests because they fail due to a real bug: lineHeight token type changed, a separate PR. With the bugfix these will be enabled again.

To test:

**Note: For easier review do not view the first commit, it just removed the `asycn`-s**
Just check the code changes, make sure we are testing correct behaviour not for bugs.

## Jira
INSTUI-5091


cypress/component/DateInput.cy.tsx
```diff
@@ -399,10 +404,7 @@ describe('<DateInput/>', () => {
     cy.get('thead th')
       .eq(2)
       .within(() => {
-        cy.get('.plugin-cache-1sr5vj2-screenReaderContent').should(
```
matyasf, line 402: another common error -- hardcoded generated CSS class name

cypress/component/Drilldown.cy.tsx
```diff
@@ -43,6 +43,10 @@ const renderOptions = (page: string) => {
 }
 
 describe('<Drilldown/>', () => {
+  beforeEach(() => {
+    cy.get('body').realMouseMove(0, 0, { position: 'bottomRight' })
+  })
```
matyasf, line 48: a very common pattern, tests were failing because the cursor was kept in the same position, causing things not to trigger

cypress/component/ColorMixer.cy.tsx
```diff
@@ -109,52 +109,7 @@ describe('<ColorMixer/>', () => {
     cy.wrap(onChange).should('have.been.calledTwice')
   })
 
-  it('should change hue value when click at the outside of the slider to the left', async () => {
```
matyasf, line 112: This should not happen

cypress/component/ColorMixer.cy.tsx
```diff
@@ -422,54 +377,7 @@ describe('<ColorMixer/>', () => {
     cy.wrap(onChange).should('have.been.calledTwice')
   })
 
-  it('should change alpha value when click at the outside of the bar to the left', async () => {
```
matyasf, line 425: This should not happen

cypress/component/Tabs.cy.tsx
```diff
@@ -24,54 +24,12 @@
 
 import React from 'react'
 import { Tabs } from '@instructure/ui/latest'
-import { expect } from 'chai'
 
 import '../support/component'
 import 'cypress-real-events'
 
 describe('<Tabs/>', () => {
-  it('should preserve Tab.Panel keys', async () => {
```
matyasf, line 33: The keys are not preserved even in InstUI v10 because Tabs are cloned

cypress/component/DateInput.cy.tsx
```diff
@@ -951,57 +950,7 @@ describe('<DateInput/>', () => {
     cy.get('@yearPicker').should('have.value', '2021')
   })
 
-  it('should display -- sign in yearPicker if no date value or date is out of range', async () => {
```
matyasf, line 954: this test broke, v2 components display placeholders differently, I'll investigate further

cypress/component/DateInput.cy.tsx
```diff
@@ -618,32 +617,34 @@ describe('<DateInput/>', () => {
 
         cy.get('table').should('be.visible')
 
-        cy.contains('button', dayForSelect)
-          .should('be.enabled')
-          .click()
-          .wait(500)
```
matyasf, line 624: Removing this is a huge speedup since this test runs 83 times

packages/ui-billboard/src/Billboard/__tests__/Billboard.test.tsx
```diff
@@ -147,13 +147,13 @@ describe('<Billboard />', () => {
...
 
-    it('should not be clickable', () => {
+    it('should not be clickable', async () => {
       const onClick = vi.fn()
 
       render(<Billboard onClick={onClick} readOnly />)
       const button = screen.getByRole('button')
 
-      userEvent.click(button)
+      await userEvent.click(button)
```
matyasf, line 156: an unrelated fix for a bug that causes this test to be flaky. Note that this is again a pattern, that we need to fix in the whole suite

git-nandor (approved):

cypress/component/Emotion-useStyle.cy.tsx
```diff
@@ -58,6 +58,7 @@ const grey1111 = 'rgb(0, 128, 0)'
 const green4570 = 'rgb(10, 10, 10)'
 const blue4570 = 'rgb(255, 255, 0)'
 const exampleTheme: Theme = {
+  // TODO update to the new theme syntax
```
joyenjoyer, line 61: Create a jira ticket for this so it won't be forgotten

packages/ui-billboard/src/Billboard/__tests__/Billboard.test.tsx
joyenjoyer: What is the whole suite fix for this bug?

joyenjoyer (approved):

packages/ui-billboard/src/Billboard/__tests__/Billboard.test.tsx
matyasf: just for this bug. We are handling the fix in a different ticket

PR #2630, opened 2026-07-08
docs(ui-select): remove dead screen reader announcement code from Sel…

ToMESSKa:
…ect examples
INSTUI-5024

**ISSUE:**
- Réka noticed some screenreader announcements in the Select examples are not in use so they should be removed as the dropdown’s state-change and the number of options are announced without these already.  As part of the a11y audit, https://github.com/instructure/instructure-ui/pull/1991#discussion_r2123515917 Select examples were reworked and these announcements were dropped but some of this code was left behind.

**TEST PLAN:**
- make sure the deleted code was in fact unused
- check the Autocomplete example under Providing autocomplete behavior header with a screenreader
- when pressing 'a', the number of matches (3) should be announced and pressing 'a' again, the number of matches should be announced again (1)
- for the remaining examples, the number of options/option groups should be announced when the user presses Space in the input field.

### DIFF
--- a/packages/ui-select/src/Select/v1/README.md
+++ b/packages/ui-select/src/Select/v1/README.md
@@ -26,7 +26,6 @@ type: example
     const [isShowingOptions, setIsShowingOptions] = useState(false)
     const [highlightedOptionId, setHighlightedOptionId] = useState(null)
     const [selectedOptionId, setSelectedOptionId] = useState(options[0].id)
-    const [announcement, setAnnouncement] = useState(null)
     const inputRef = useRef()
 
     const focusInput = () => {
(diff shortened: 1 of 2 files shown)


packages/ui-select/src/Select/v1/README.md
```diff
@@ -335,6 +304,13 @@ type: example
             </Select.Option>
           )}
         </Select>
+        <Alert
```
ToMESSKa, line 307: After checking every example, this was the only case the number of options wasn't announced properly (see the case in the test plan) so I added this.

HerrTopi (approved):

git-nandor (approved):

PR #2634, opened 2026-07-08
fix(ui-options,ui-select): stop VoiceOver re-announcing Select option

git-nandor:
Bug: Arrowing through a SimpleSelect (single-select listbox) with VoiceOver announces each option, then appends "not selected" a beat later and restarts the announcement.

## Summary
- Express the `Options.Item` highlight variant through a stable `data-variant` attribute + static `&[data-variant="…"]` CSS instead of a variant-dependent Emotion class. The option's `<li role="none">` no longer swaps its class in a deferred second commit (via `withStyleNew`'s `makeStyles`→`setStyles`) one tick after `aria-activedescendant` moves — that ancestor mutation was what made VoiceOver restart the announcement and append "not selected".

## Test Plan
- Chrome + VoiceOver on macOS: open a v11.7 `Select`/`SimpleSelect`, open the listbox, and arrow up/down through the options. Each option should be announced once, cleanly (label + state + position), with no interrupt/restart 
````
const Example = ({ options }) => {
  const [value, setValue] = useState('all')
  
  return (
    <div style={{ padding: 40, maxWidth: 320 }}>
      <SimpleSelect
        renderLabel="Filter"
        value={value}
        onChange={(_e, { value }) => setValue(value)}
      >
        {options.map((o) => (
          <SimpleSelect.Option key={o.value} id={o.value} value={o.value}>
            {o.label}
          </SimpleSelect.Option>
        ))}
      </SimpleSelect>
    </div>
  )
}
render(
  <Example
    options={[
  { label: 'All notes', value: 'all' },
  { label: 'Important', value: 'important' },
  { label: 'Unclear', value: 'unclear' },
]}
  />
)
````

Fixes INSTUI-5082

### DIFF
--- a/packages/ui-options/src/Options/__tests__/Item.test.tsx
+++ b/packages/ui-options/src/Options/__tests__/Item.test.tsx
@@ -208,4 +208,48 @@ describe('<Item />', () => {
     expect(link).toHaveAttribute('href', '/helloWorld')
     expect(link?.tagName).toBe('A')
   })
+
+  // VoiceOver double-announces options because the highlight used to
+  // be a variant-dependent Emotion class. The variant is now expressed
+  // through a stable `data-variant` attribute + static CSS, so the class must NOT
+  // change when the variant does.
+  describe('variant applied via a stable data-variant attribute', () => {
+    it('exposes the variant through a data-variant attribute on the outer element', async () => {
+      const { container } = render(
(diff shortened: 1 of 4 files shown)


ToMESSKa (changes requested):
Unfortunately I couldn't reproduce the issue on the the master, can you please give more information how to do that?

PR #2635, opened 2026-07-13
fix(ui-truncate-text): remove lineHeight theme input, default v2 truncation to 1.2

HerrTopi:
## Summary
- Remove `lineHeight` as a theme/style input for v2 `TruncateText`: it's dropped from the `TruncateTextStyle` `ComponentStyle` union (`props.ts`), no longer emitted by `styles.ts`, and no longer threaded from `index.tsx` into the truncator options.
- Replace it with a module-level `DEFAULT_LINE_HEIGHT = 1.2` constant in `utils/truncate.ts`. When the inherited computed `line-height` resolves to `normal`, truncation height is calculated as `1.2 * fontSize`; a concrete `line-height` is still parsed from the computed style and used as-is.

## Test Plan
- Render a v2 `TruncateText` in a container whose `line-height` is `normal` and confirm truncation lands on the right line (now driven by the 1.2 default).
- Render one with a concrete numeric/px `line-height` and confirm truncation is unchanged (no regression).

### DIFF
--- a/packages/ui-truncate-text/src/TruncateText/v2/index.tsx
+++ b/packages/ui-truncate-text/src/TruncateText/v2/index.tsx
@@ -209,8 +209,7 @@ class TruncateText extends Component<TruncateTextProps, TruncateTextState> {
     if (canUseDOM) {
       const result = truncate(this._stage!, {
         ...this.props,
-        parent: this.ref ? this.ref : undefined,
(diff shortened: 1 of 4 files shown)


matyasf (approved):
Looks good! Lets make sure that design removed the unused token

git-nandor (approved):

PR #2636, opened 2026-07-14
fix(ui-instructure): use CondensedButton for AI component triggers

joyenjoyer:
Link is changed to CondensedButton to comply with WCAG rules.

Test:

- inspect AI components and observe if tags that were previously anchors are buttons now


joyenjoyer:
@adamlobler I add you as a reviewer too, please check if CondensedButton is a good choice from a design perspective.

HerrTopi (approved):
consider the ai's suggestion about the aria dialog. Otherwise, looks good

git-nandor (approved):

PR #2637, opened 2026-07-14
feat(ui-codemods): update tokens and move theme.componentOverrides to…

ToMESSKa:
… themeOverride.components

INSTUI-4840

**1. Renames and removes theme tokens.**
- Renames or drops theme variables that changed in the new theming (e.g. background → backgroundColor, color → textColor).
- Finds them in themeOverride props and in InstUISettingsProvider overrides and rewrites them to the new names.
- Removes tokens that no longer exist.
- Leaves a warning wherever it can't safely decide.

**2. Updates how overrides are set on InstUISettingsProvider.**
- Relocates overrides from the old theme={{ componentOverrides: { Checkbox: {...} } }} to the new themeOverride={{ components: { Checkbox: {...} } }}.
- Always migrates, since it can't tell which version the components inside the InstUISettingsProvider are.
- Warns you to bump anything still on v11.6-or-earlier to v11.7/latest so its overrides keep working.
- Runs in the same pass as 1., so the token rewrites and this relocation boht happen together

**3. Revises the token list in the upgrade guide.**
- Double-checks every existing entry against the actual styles.ts files of the v1 and v2 components.
- Corrects entries that were wrong (bad renames, tokens listed that don't actually change) so the guide matches what the codemod does.

**4. Ships with a full test suite.**
- Adds ~60 before/after fixtures covering both token renames and override relocation, including the tricky cases (mixed v1/v2 imports, quoted keys, spreads, unmergeable conflicting overrides, providers that end up empty).
- Pairs every transform fixture with a warning fixture so the exact warning text is locked down.

TEST PLAN:

**1. Confirm the tests actually catch mistakes**
- Edit a fixture output (e.g. test1.output.tsx) — change one expected token, e.g. `backgroundColor` → `bogusColor`.
- Run `pnpm run test:vitest ui-codemods` → that fixture fails.

**2. Come up with new edge cases the fixtures don't cover**
- Think of tricky cases especially cases where it might transform something it shouldn't or cases it should


packages/ui-codemods/lib/__node_tests__/__testfixtures__/multiVersionThemeVariablesCodemod/test38.input.tsx
```diff
@@ -0,0 +1,21 @@
+// @ts-nocheck
```
ToMESSKa, line 1: From the codemod's point of view Checkbox is tricky because the same old token maps to different new tokens depending on the variant (simple vs toggle), so it can't do a straight one-to-one rename — and when the variant can't be determined statically it has to leave those tokens alone and warn. On top of that it has separate facade keys (Checkbox/CheckboxFacade/ToggleFacade) that each keep their own tokens, so they can't be merged.

docs/upgrading/upgrade-guide.md
```diff
@@ -146,6 +146,37 @@ type: embed
...
+  removed={[
+    {name:"boxShadowColor",note:"replaced by a single boxShadow object token"},
+    {name:"boxShadowBlur",note:"replaced by a single boxShadow object token"},
+    {name:"color",note:"color system reworked to accent1-6; use the <color>BackgroundColor/<color>TextColor tokens"},
+    {name:"colorShamrock",note:"color system reworked to accent1-6; use the <color>BackgroundColor/<color>TextColor tokens"},
+    {name:"colorBarney",note:"color system reworked to accent1-6; use the <color>BackgroundColor/<color>TextColor tokens"},
+    {name:"colorCrimson",note:"color system reworked to accent1-6; use the <color>BackgroundColor/<color>TextColor tokens"},
+    {name:"colorFire",note:"color system reworked to accent1-6; use the <color>BackgroundColor/<color>TextColor tokens"},
+    {name:"colorLicorice",note:"color system reworked to accent1-6; use the <color>BackgroundColor/<color>TextColor tokens"},
+    {name:"colorAsh",note:"color system reworked to accent1-6; use the <color>BackgroundColor/<color>TextColor tokens"},
```
matyasf: This is very hard to read/interpret. These should go to the 'changed' part, e.g. color -> blueBackgroundColor, colorAsh -> greyBackgroundColor, colorBarney -> blueBackgroundColor, colorCrimson -> redBackgroundColor,..

docs/upgrading/upgrade-guide.md
```diff
@@ -146,6 +146,37 @@ type: embed
...
+    {name:"aiFontColor",note:"the AI variant text now uses textOnColor"},
```
matyasf: Then its a change, not a removal

docs/upgrading/upgrade-guide.md
```diff
@@ -146,6 +146,37 @@ type: embed
...
+    {name:"fontFamily",note:""}
```
matyasf: This variable still exists in 11.7

docs/upgrading/upgrade-guide.md
```diff
@@ -146,6 +146,37 @@ type: embed
...
+  ]}
+  changed={[
+    {oldName:"background",newName:"backgroundColor",note:""},
+    {oldName:"borderWidthSmall",newName:"borderWidthSm",note:""},
+    {oldName:"borderWidthMedium",newName:"borderWidthMd",note:""}
+  ]}
+/>
```
matyasf, line 175: Also as I see there were some TextColor tokens added? Like `ashTextColor`, `blueTextColor`,.. 
and the `fontSizeXs` are new too... We should check/regenerate all the tables here based on the token changes Google doc I think

matyasf (changes requested):

docs/upgrading/upgrade-guide.md
ToMESSKa: These are unfortunately not one-to-one renames, so a remove + warn is the best solution in my opinion
eg.: colorCrimson -> redBackgroundColor ?

v1 (`v1/styles.ts`)
- non-inverse: text = `colorCrimson`, background = white
- inverse: text = white, background = `colorCrimson`

v2 (`v2/styles.ts`)
- non-inverse: text = `redTextColor`, background = white
- inverse: text = `textOnColor`, background = `redBackgroundColor`

docs/upgrading/upgrade-guide.md
ToMESSKa: It's not a one-on-one rename either 
v1 (v1/styles.ts)
- ai variant: text = aiFontColor, background = AI gradient
- aiFontColor is used only for the ai variant's text/icon

v2 (v2/styles.ts)
- ai variant: text = textOnColor, background = AI gradient
- but textOnColor is also the inverse-mode text for every accent color: color: hasInverseColor ? textOnColor : colorVariants[color].text

I updated the warnings tho

docs/upgrading/upgrade-guide.md
ToMESSKa: No, it is in Avatar/v1/styles.ts but not in Avatar/v2/styles.ts

docs/upgrading/upgrade-guide.md
ToMESSKa: The Google docs is a really unreliable source unfortunately what why I checked the styles.ts v1 vs v2 files instead and used as a source of true.

matyasf (approved):
nice work! The test caught a bug too (INSTUI-5103)

packages/ui-codemods/lib/multiVersionThemeVariablesCodemod/migrateComponentOverridesToThemeOverride.ts
```diff
@@ -0,0 +1,459 @@
...
+              componentsValue as Parameters<typeof j.spreadElement>[0]
+            )
+          )
+        }
+      } else {
+        toObject.properties.push(
+          j.objectProperty(j.identifier('components'), componentsValue)
+        )
+      }
+    } else if (hasOverrides) {
```
balzss, line 446: **Blocking — emits a duplicate `themeOverride`.** This `else if (hasOverrides)` is reached when an existing `themeOverride` prop is present but *not* an object literal (a variable, call, ternary…), because the branch above only matches `ObjectExpression`. So we splice in a second `themeOverride` alongside the existing one → invalid JSX; React keeps the last, silently dropping the migrated components, with no warning.

Confirmed by running the codemod on `theme={{ componentOverrides: {...} }}` + `themeOverride={someVar}` → output has two `themeOverride=` attributes.

Suggest: detect the existing non-object `themeOverride` here and warn + skip (consistent with the other opaque-value cases), and add a fixture for this shape.

packages/ui-codemods/lib/multiVersionThemeVariablesCodemod/migrateComponentOverridesToThemeOverride.ts
```diff
@@ -0,0 +1,459 @@
...
+      componentsValue.properties.length > 0
+
+    // Overrides were relocated to the new channel; warn that any v1 (v11.6 or
+    // earlier) component in this InstUISettingsProvider will no longer pick them up.
+    if (hasOverrides) {
+      printWarning(filePath, coProp.loc?.start.line, V1_OVERRIDE_WARNING)
+    }
+
+    // Case 5: element already has a `themeOverride` prop -> merge into it.
+    const existingThemeOverride = attributes.find(
```
balzss: (Same bug, root cause) `existingThemeOverride` is found here, but the branch condition below only handles the object-literal case. Any non-object existing value falls through to the "build fresh attribute" path. Whatever guard you add, base it on this variable being defined.

packages/ui-codemods/lib/multiVersionThemeVariablesCodemod/themeVariableMappings/index.ts
```diff
@@ -0,0 +1,1255 @@
...
+      },
+      color: {
+        warning:
+          "Avatar's color system was reworked to `accent1`-`accent6`; use `blueBackgroundColor`/`blueTextColor` instead of `color`."
+      },
+      colorShamrock: {
+        warning:
+          "Avatar's color system was reworked to `accent1`-`accent6`; use `greenBackgroundColor`/`greenTextColor` instead of `colorShamrock`."
+      },
+      colorBarney: {
```
balzss, line 133: `colorBarney` maps to `blueBackgroundColor/blueTextColor`, same as `color` two entries up — but Barney is purple in the palette, so this looks like a copy-paste slip. Whole table should get a domain cross-check against the v1/v2 `theme.ts` diffs, but flagging this one specifically.

packages/ui-codemods/lib/__node_tests__/runTest.ts
```diff
@@ -36,11 +36,14 @@ import { expect, vi } from 'vitest'
  * - Fixtures must be named `XY.input.[ts|tsx|js]` and `XY.output.[ts|tsx|js]`
  * @param codemod The codemod to run
  */
-export function runTest(codemod: Transform) {
-  const entries = fs.readdirSync(
-    `${__dirname}/__testfixtures__/${codemod.name}`,
-    { withFileTypes: true }
-  )
+export function runTest(codemod: Transform, subDir?: string) {
```
balzss: The new `subDir` param isn't passed by any caller (`multiVersionThemeVariablesCodemod` fixtures resolve via `codemod.name` alone). Remove it, or wire it up if the intent was to nest the fixtures.

packages/ui-codemods/lib/multiVersionThemeVariablesCodemod/migrateComponentOverridesToThemeOverride.ts
```diff
@@ -0,0 +1,459 @@
...
+      seen.set(prop.key.name, prop)
+    }
+    merged.push(prop)
+  })
+  // eslint-disable-next-line no-param-reassign
+  components.properties = merged
+}
+
+/**
+ * Converts dotted child-component keys to the dot-stripped `componentId` form
```
balzss: This docblock ("Converts dotted child-component keys…") sits above `const SPREAD_WARNING`, but it actually describes `normalizeComponentKeys` further down. Move it onto that function.

packages/ui-codemods/lib/multiVersionThemeVariablesCodemod/migrateComponentOverridesToThemeOverride.ts
ToMESSKa: I fixed this and added a text fixture for this scenario

packages/ui-codemods/lib/multiVersionThemeVariablesCodemod/migrateComponentOverridesToThemeOverride.ts
ToMESSKa: Fixed

packages/ui-codemods/lib/multiVersionThemeVariablesCodemod/themeVariableMappings/index.ts
ToMESSKa: Barney maps to blueBackgroundColor/blueTextColor because it in v1 colorBarney resolves to blue so users of color="barney" already rendered blue. Anc color is also maps ti blue.

packages/ui-codemods/lib/__node_tests__/runTest.ts
ToMESSKa: Removed

packages/ui-codemods/lib/multiVersionThemeVariablesCodemod/migrateComponentOverridesToThemeOverride.ts
ToMESSKa: I moved this

packages/ui-codemods/lib/multiVersionThemeVariablesCodemod/themeVariableMappings/index.ts
balzss: embarrassing claude slip up :)

balzss (approved):

PR #2638, opened 2026-07-14
POC: run tests in real-browser vitest mode

matyasf:
Add real-browser Vitest (Playwright/chromium) test POC, that replaces both Cypress component tests AND Vitest unit tests.

## Summary
- Port Alert (`ui-alerts`) and Dialog (`ui-dialog`) and some other suites to `vitest/browser`
- add `test:browser` scripts and new vitest configs.

## Test Plan
- `pnpm run test:browser` runs the browser tests in the console
- `pnpm run test:browser-ui` runs the browser tests in a browser
- Check the code diff, it should do the same thing. Also check how you like the new syntax, any pros or cons
- Read about it on https://vitest.dev/guide/browser/ or watch a short Youtube video

Future additions:
- Vitest can automatically screenshot test failures, we could add it to the CI somehow
- Vitest browser has a Visual regression test add-on, we could use it.

Fixes INSTUI-5098


packages/ui-dialog/src/Dialog/__tests__/Dialog.test.tsx
```diff
@@ -91,104 +94,72 @@ const DialogExample = forwardRef((props: DialogProps, ref: React.Ref<any>) => {
 })
 DialogExample.displayName = 'DialogExample'
 
-const NestedDialogExample = (props: DialogProps) => {
```
matyasf, line 94: This was not used by the test at all

vitest.config.mts
```diff
@@ -62,6 +147,44 @@ export default defineConfig({
...
+            'packages/ui-alerts/**/__tests__/**/*.test.tsx',
+            'packages/ui-dialog/**/__tests__/**/*.test.tsx'
+          ],
+          setupFiles: './vitest.setup.browser.ts',
+          name: { label: 'browser', color: 'green' },
+          browser: {
+            enabled: true,
+            provider: playwright(),
+            headless: true,
+            instances: [{ browser: 'chromium' }],
```
matyasf, line 173: we can add here other browsers easily to run our test suite on e.g. Firefox (but CI support might be tricky)

vitest.config.mts
```diff
@@ -24,8 +24,85 @@
 
 /// <reference types="vitest" />
 
-import { defineConfig } from 'vitest/config'
+import { defineConfig, configDefaults } from 'vitest/config'
```
matyasf, line 27: I havent checked this config thoroughly, we should do it, if this is the way forward

.github/workflows/pr-validation.yml
```diff
@@ -37,6 +37,24 @@ jobs:
...
+      - uses: actions/checkout@v6
+      - uses: pnpm/action-setup@v6
+      - uses: actions/setup-node@v6
+        with:
+          node-version: '24'
+          cache: 'pnpm'
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+      - name: Install Playwright browser
+        run: pnpm exec playwright install --with-deps chromium
```
balzss, line 53: why isnt it installed from dev deps?

vitest.config.mts
```diff
@@ -24,8 +24,85 @@
...
+          })
+        }
+      }
+    }
+    // Deep source imports: `@instructure/ui-x/src/...` -> package src dir
+    aliases.push({
+      find: `${pkgName}/src`,
+      replacement: path.join(pkgPath, 'src')
+    })
+    // Fallback bare-name -> package dir (for packages without an exports map)
```
balzss: is this an existing scenario or just AI being cautious?

vitest.config.mts
```diff
@@ -49,6 +126,14 @@ export default defineConfig({
           // Allows using APIs like Jest without importing them
           globals: true,
```
balzss, line 115: not part of the changeset, but maybe we should consider changing this to false?

vitest.setup.browser.ts
```diff
@@ -0,0 +1,34 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+// Setup for the `browser` project (real-browser tests via vitest-browser).
+import '@testing-library/jest-dom/vitest'
+
+// @instructure/console only outputs logs if process.env.NODE_ENV !== 'production'
+// these messages are sometimes used in test assertions
+if (typeof globalThis.process === 'undefined') {
```
balzss, line 30: this file is a bit hard to parse. is it only needed because of the `console` package?

matyasf: I dont know yet, it could be that other errors surface when we dont mock it

vitest.config.mts
joyenjoyer: Good idea; I would add Safari if it is possible, it has many tricky cases when it comes to a11y

.github/workflows/pr-validation.yml
joyenjoyer: I also had to run this locally, `pnpm install` is not enough

joyenjoyer:
I like the browser-based runtime. There isn't a huge difference in the code syntax, which is also a great thing. I can't see any major drawbacks yet. I'd replace the component tests and Vitest tests with this if the test run is significantly faster.

joyenjoyer (approved):

.github/workflows/pr-validation.yml
matyasf: there's no way to express it as a dependency.
`playwright` ships the JS driver only; the actual Chromium build (~150 MB) lives in `~/.cache/ms-playwright`, outside `node_modules`,  and is fetched by `playwright install`.  I've added this to the package.json scripts.
The `with-deps` flag installs system deps too, these are present on PCs with GUIs and would just slow down test runs

vitest.config.mts
matyasf: AI being too cautinous, I've removed it

vitest.config.mts
matyasf: yep, but in a different PR

vitest.setup.browser.ts
matyasf: as I see yes. We're using the `NODE_ENV` in 3 types of places:
1. to add console warnings/errors in non-prod builds.  
2. to have different `displayName` for stuff that is wrapped in decorators `withStyle` or `WithDeterministicId`. 
3. to generate a less readable, but shorter `uid` in prod.

As I see some tests are checking for console errors, this is why this mock is needed. 
IMO in the future we should remove all `NODE_ENV` checks.

balzss (approved):
this is good as is for preparing the complete migration. please squash the commit before rebasing to master

HerrTopi (approved):

PR #2640, opened 2026-07-15
build: generate AI docs from latest version

ToMESSKa:
INSTUI-5092

**ISSUE:**
- the AI documentation the build generates (llms.txt + the per-component markdowns/*.md) was being built using the default export, which still resolves to the older stable v11_6, so the docs described outdated props/examples instead of the latest v11_7 / v2 components

**TEST PLAN:**
- build the docs from the repo using `pnpm run dev`
- check the a few component's markdown files in `packages/__docs__/__build__/markdowns` and confirm they reflect the v2 version of the component e.g: in Link.md `isWithinText` is removed (compared to v1) or in Avatar.md the first example of Avatar uses accent colors (compared to v1)
- check a versioned component's file in `packages/__docs__/__build__/markdowns`, it should show the versioned import like `import { Alert } from '@instructure/ui-alerts/v11_7'` under "Import the component" header:
- check a non-versioned component's file in `packages/__docs__/__build__/markdown`s, it should show a plain import like`import { Dialog } from '@instructure/ui-dialog'` under the "Import the component" header:
- check `packages/__docs__/__build__/llms.txt` the file should include  `"All component documentation below always reflects the latest InstUI version noted above.."` warning and "Component versioning" should be there under "Guides"

### DIFF
--- a/packages/__docs__/buildScripts/build-docs.mts
+++ b/packages/__docs__/buildScripts/build-docs.mts
@@ -226,21 +226,38 @@ async function buildDocs() {
       JSON.stringify(docsVersionsManifest)
     )
 
-    // Generate AI accessible documentation from default version
-    const defaultVersionDocsDir = buildDir + 'docs/' + defaultVersion + '/'
-    generateAIAccessibleMarkdowns(defaultVersionDocsDir, buildDir + 'markdowns/')
+    // Generate AI accessible documentation from the latest library version.
+    const latestVersion =
+      versionMap.libraryVersions[versionMap.libraryVersions.length - 1] ||
+      defaultVersion
+    const latestVersionDocsDir = buildDir + 'docs/' + latestVersion + '/'
+    const latestVersionSourcesData =
+      latestVersionDocsDir + 'markdown-and-sources-data.json'
 
-    generateAIAccessibleLlmsFile(
-      buildDir + 'markdown-and-sources-data.json',
-      {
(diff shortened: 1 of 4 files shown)


packages/__docs__/buildScripts/build-docs.mts
```diff
@@ -226,22 +226,39 @@ async function buildDocs() {
...
-      {
-        outputFilePath: path.join(buildDir, 'llms.txt'),
-        baseUrl: 'https://instructure.design/markdowns/',
-        summariesFilePath: path.join(
-          __dirname,
-          '../buildScripts/ai-accessible-documentation/summaries-for-llms-file.json'
-        )
-      }
+    generateAIAccessibleMarkdowns(
+      latestVersionDocsDir,
```
git-nandor, line 247: generateAIAccessibleMarkdowns is called without await, so any error would bypass the buildDocs error wrapper as an unhandled rejection.

git-nandor (changes requested):

git-nandor:
😄 I think there is a bit more 'snow' in this test plan than we actually need 🌨️ ❄️

packages/__docs__/buildScripts/build-docs.mts
ToMESSKa: Thanks, I fixed this

git-nandor (approved):

joyenjoyer (approved):

PR #2641, opened 2026-07-15
chore: replace ESLint with oxlint

joyenjoyer:
## Summary
- Add `.oxlintrc.json` (root + `regression-test`) porting the existing ESLint ruleset to oxlint's native plugins, its alpha `jsPlugins` bridge for 5 `eslint-plugin-react` rules oxlint lacks natively, and two custom rules (`no-relative-imports`, `notice`) reimplemented for oxlint's JS-plugin API
- Wire oxlint into `lint-staged`, `ui-scripts lint`, and `regression-test`'s lint script; remove ESLint's config files, its superseded custom plugin, and 11 now-unused dependencies. `eslint-plugin-react` and `read-package-up` are kept, since both are still runtime deps of oxlint's `jsPlugins` bridge and custom rules.
- Fix 2 `eslint-disable` comments whose rule names didn't match oxlint's actual rule IDs and had silently stopped suppressing anything; delete 6 more that are genuinely dead under oxlint's `role-has-required-aria-props` implementation
- Relocate `ui-scripts`' `lint.ts` out of a legacy `lib/test/` folder into `lib/commands/`, alongside the other command modules

## Test Plan
- No manual verification needed; covered by existing tests. `tsc -b`, `vitest`, and a full-repo oxlint dry run all confirmed clean (only 5 pre-existing, unrelated `no-undef` errors remain in `ui-icons`'s `.cjs` configs, unchanged from before this PR).

Fixes INSTUI-5035


balzss (commented):
the changeset look ok and linting works fine for me, however this breaks my editor integration which showed realtime feedback about lint errors. i can work around it and find a plugin that fills this gap but it worths checking if other editors like webstorm and vscode support it or not

scripts/oxlint/notice-plugin.mjs
```diff
@@ -0,0 +1,123 @@
...
+
+const MUST_MATCH = 'The MIT License'
+const CHARS = 1000
+
+/**
+ * Pure rule logic, independent of oxlint's context/AST shape, so it can be unit-tested directly.
+ */
+export function hasNoticeHeader(sourceText) {
+  const leading = String(sourceText).replace(/\r\n/g, '\n').slice(0, CHARS)
+  return new RegExp(MUST_MATCH).test(leading)
```
matyasf: Why just match this short text? All text must be included

package.json
```diff
@@ -85,33 +83,23 @@
...
     "esbuild": "^0.28.1",
     "esbuild-loader": "^4.5.0",
-    "eslint": "^9.39.4",
-    "eslint-config-prettier": "^10.1.8",
-    "eslint-import-resolver-node": "^0.4.0",
-    "eslint-module-utils": "2.13.0",
-    "eslint-plugin-compat": "^7.0.2",
-    "eslint-plugin-jsx-a11y": "^6.10.2",
-    "eslint-plugin-notice": "^1.0.0",
     "eslint-plugin-react": "^7.37.5",
```
matyasf, line 93: I think this one can be deleted too

.oxlintrc.json
```diff
@@ -0,0 +1,546 @@
...
+      "plugins": ["vitest"],
+      "rules": {
+        "vitest/valid-expect": "off"
+      }
+    },
+    {
+      "files": ["**/*"],
+      "excludeFiles": ["**/__tests__/**"],
+      "plugins": ["jsx-a11y"],
+      "rules": {
```
matyasf, line 373: Do we really need such huge config for each plugin? Why not use some defaults?

.oxlintrc.json
```diff
@@ -0,0 +1,546 @@
...
+        ]
+      }
+    },
+    {
+      "files": ["packages/ui-truncate-text/src/TruncateText/__tests__/**"],
+      "plugins": ["vitest"],
+      "rules": {
+        "vitest/valid-expect": "off"
+      }
+    },
```
matyasf: Why is there a rule for just this one?

.oxlintrc.json
```diff
@@ -0,0 +1,546 @@
...
+    ],
+    "prefer-rest-params": "error",
+    "prefer-spread": "error",
+    "@typescript-eslint/ban-ts-comment": [
+      "error",
+      {
+        "ts-ignore": "allow-with-description"
+      }
+    ],
+    "@typescript-eslint/no-array-constructor": "error",
```
matyasf, line 214: Cant we just use defaults for these? For most cases it would be much cleaner IMO

matyasf (changes requested):
While the commands work, I get no linting in WebStorm, it does not show me anything thats an error.
Pease add it to `contributing-getting-started`/Linters part on how to use it

package.json
joyenjoyer: There are a few ESLint rules that don't have native implementations in Oxlint yet:
  - react-js/prop-types                                                                                                                                                                                                                  
  - react-js/forbid-foreign-prop-types                                                                                                                                                                                                   
  - react-js/require-default-props                                                                                                                                                                                                       
  - react-js/no-typos                                                                                                                                                                                                                    
  - react-js/no-deprecated)
  
 Therefore, ESLint needs to be added as a dependency. Oxlint can use ESLint rules through its jsPlugins bridge. To answer your other comments here @matyasf, we can decide removing these and other specific rules too and just trust what Oxlint checks by default, I wanted to cover as many rules as we did with ESLint to not lose coverage.

scripts/oxlint/notice-plugin.mjs
joyenjoyer: Eslint-plugin-notice also scans the first 1000 characters by default that is why I implemented this way.

matyasf: But here here are comparing this the MUST_MATCH text, not 1000 characters

joyenjoyer:
@HerrTopi I need you back here :D, you are the main VSCode user in the team, please try the new Oxlint rules with this
[plugin](https://marketplace.visualstudio.com/items?itemName=oxc.oxc-vscode)

HerrTopi (approved):
It seems to be working for vscodium

scripts/oxlint/notice-plugin.mjs
joyenjoyer: Fixed, full codebase linting became slower though.

.oxlintrc.json
joyenjoyer: This was added to bypass this oxlint rule flagging an old Chai style assertion, I updated these assertions in TruncateText test files and removed this rule bypass from the config file.

.oxlintrc.json
joyenjoyer: I am going to revisit all the rules in a separate PR as we discussed. Please accept this PR.

.oxlintrc.json
joyenjoyer: I am going to revisit all the rules in a separate PR as we discussed. Please accept this PR.

matyasf: good idea

matyasf (approved):

PR #2643, opened 2026-07-17
feat(ui-text-input,ui-select): add forcePlaceholder prop to TextInput

HerrTopi:
## Summary
- Add optional `forcePlaceholder` prop (default `false`) to TextInput v2 that renders the placeholder when empty regardless of interaction type
- Pass `forcePlaceholder` to Select v2's trigger TextInput so a Select shows its placeholder outside the enabled interaction state

## Test Plan
- Render a TextInput in `readonly`/`disabled` interaction with `forcePlaceholder` and confirm the placeholder text shows while empty, and disappears once a value is entered
- Verify a Select displays its placeholder when not in the enabled interaction state
- RTL + a11y spot-check on both components

Fixes INSTUI-5104

### DIFF
--- a/packages/ui-select/src/Select/v2/index.tsx
+++ b/packages/ui-select/src/Select/v2/index.tsx
@@ -111,18 +111,18 @@ const MemoedOption = memo(
   (prevProps, nextProps) => {
     return (
       prevProps.selectOption.props.isHighlighted ===
-      nextProps.selectOption.props.isHighlighted &&
+        nextProps.selectOption.props.isHighlighted &&
       prevProps.selectOption.props.isSelected ===
-      nextProps.selectOption.props.isSelected &&
+        nextProps.selectOption.props.isSelected &&
       prevProps.selectOption.props.isDisabled ===
-      nextProps.selectOption.props.isDisabled &&
+        nextProps.selectOption.props.isDisabled &&
       prevProps.selectOption.props.children ===
-      nextProps.selectOption.props.children &&
+        nextProps.selectOption.props.children &&
       prevProps.selectOption.props.id === nextProps.selectOption.props.id &&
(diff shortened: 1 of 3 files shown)


packages/ui-select/src/Select/v2/index.tsx
```diff
@@ -111,18 +111,18 @@ const MemoedOption = memo(
...
+        nextProps.selectOption.props.isDisabled &&
       prevProps.selectOption.props.children ===
-      nextProps.selectOption.props.children &&
+        nextProps.selectOption.props.children &&
       prevProps.selectOption.props.id === nextProps.selectOption.props.id &&
       prevProps.selectOption.props.renderBeforeLabel ===
-      nextProps.selectOption.props.renderBeforeLabel &&
+        nextProps.selectOption.props.renderBeforeLabel &&
       prevProps.selectOption.props.renderAfterLabel ===
-      nextProps.selectOption.props.renderAfterLabel &&
```
matyasf, line 125: The formatting was broken in this commit: https://github.com/instructure/instructure-ui/commit/cda547ac0af8664831097598be23dc0eb989602a#diff-86c34ad67bd876370225bf0df1e248d809d3ac768f41267d08d576d0ef2d0ec1

This is just a revert (and makes the whole thing a pain to read!)

matyasf (approved):

ToMESSKa (changes requested):
The forced placeholder text currently has no styling (any styling depends on it being interaction === 'enabled' in TextInput/v2/styles.ts). I think it has too low of a contrast in the dark theme, around 2.1:1.

I also see a comment in TextInput/v2/styles.ts which says "placeholder is not rendered in the readOnly and disabled state", which should be removed now.

Image

ToMESSKa (approved):

PR #2644, opened 2026-07-17
test(many): add missing await on userEvent calls in vitest

ToMESSKa:
INSTUI-5100

**ISSUE:**

- 63 userEvent.* calls in vitest tests were missing await. Without "wait," the test fires the click and immediately runs its check — before the click has actually done anything. So the test can say "passed" even when the feature is broken. 

**TEST PLAN:**
- confirm no userEvent action is left un-awaited anywhere

### DIFF
--- a/packages/ui-checkbox/src/CheckboxGroup/__tests__/CheckboxGroup.test.tsx
+++ b/packages/ui-checkbox/src/CheckboxGroup/__tests__/CheckboxGroup.test.tsx
@@ -138,10 +138,10 @@ describe('<CheckboxGroup />', () => {
     expect(checkboxes[0]).not.toBeChecked()
     expect(checkboxes[1]).not.toBeChecked()
 
-    userEvent.click(checkboxes[0])
+    await userEvent.click(checkboxes[0])
 
     await waitFor(() => {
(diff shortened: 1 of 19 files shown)


packages/ui-checkbox/src/CheckboxGroup/__tests__/CheckboxGroup.test.tsx
```diff
@@ -138,10 +138,10 @@ describe('<CheckboxGroup />', () => {
     expect(checkboxes[0]).not.toBeChecked()
     expect(checkboxes[1]).not.toBeChecked()
 
-    userEvent.click(checkboxes[0])
+    await userEvent.click(checkboxes[0])
 
     await waitFor(() => {
-      expect(onChange).not.toHaveBeenCalled()
+      expect(onChange).toHaveBeenCalledWith(['tester', 'test-value-1'])
```
ToMESSKa, line 144: A controlled CheckboxGroup is supposed to fire onChange on every click (that's how it reports the user's intent to the parent) — it just never changes its own checked state.

git-nandor (approved):

matyasf (approved):

PR #2654, opened 2026-07-23
feat(ui-tag): remove the inline variant from Tag v2

ToMESSKa:
BREAKING CHANGE: the inline variant and the variant prop have been removed from Tag v2.

Closes: INSTUI-5127

**ISSUE:**
- the inline variant of tag is no longer supported so it needs to get removed from v2

**TEST PLAN:**
- check the examples on the Tag documentation page, they should work as usual

### DIFF
--- a/packages/ui-tag/src/Tag/v2/styles.ts
+++ b/packages/ui-tag/src/Tag/v2/styles.ts
@@ -41,7 +41,7 @@ const generateStyle = (
   props: TagProps,
   sharedTokens: SharedTokens
 ): TagStyle => {
-  const { variant, size, dismissible, onClick, disabled } = props
+  const { size, dismissible, onClick, disabled } = props
 
   const isButton = !!onClick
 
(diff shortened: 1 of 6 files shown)


packages/ui-tag/src/Tag/v2/styles.ts
```diff
@@ -118,66 +118,6 @@ const generateStyle = (
...
-    inline: {
-      tag: {
-        backgroundColor: componentTheme.inlineBackground,
-        border: `${componentTheme.inlineBorderWidth} ${componentTheme.inlineBorderStyle} ${componentTheme.inlineBorderColor}`,
-        borderRadius: componentTheme.inlineBorderRadius,
-        color: componentTheme.inlineColor,
-        cursor: 'text',
-        margin: '0 0.1875rem 0.1875rem',
-        ...(isButton && {
-          '&:hover': { backgroundColor: componentTheme.inlineBackgroundHover }
```
matyasf, line 147: we should ask design to remove these tokens

matyasf (approved):

joyenjoyer (approved):
looks good

PR #2657, opened 2026-07-24
feat(docs-app,ui-buttons): add auto-generated prop playground

balzss:
## Summary
- Add internal `PropEditor` docs component that reads a component's react-docgen metadata at runtime and auto-generates prop controls (enum selects, boolean toggle, text/number inputs) with a live preview and generated JSX.
- Register `PropEditor` as a docs global so it can be embedded in any README via a `type: embed` block; supports an optional `config` for per-component tuning (include/exclude, control overrides, sample children).
- Wire it into the Button v2 README as a proof of concept.

## Test Plan
- On the Button docs page, scroll to the new "Playground" section: confirm controls are auto-generated and editing them updates the preview and code live.
- Verify complex/function props (`as`, `margin`, `cursor`, `renderIcon`, `onClick`) show under "Not editable here" and are not rendered as controls.
- Switch docs theme/version and confirm the preview re-themes and metadata still loads.

Fixes INSTUI-5128

### DIFF
--- a/.claude/commands/commit.md
+++ b/.claude/commands/commit.md
@@ -26,6 +26,7 @@ Co-Authored-By: Claude <noreply@anthropic.com>
 
 ## Steps
 
+0. **Comprehension gate.** Unless the diff is trivial (pure docs/config/formatting/version bump/generated files), run the `/comprehension-check` skill first and only continue once the author has demonstrated understanding. Skip silently for trivial diffs.
 1. `git status` + `git diff` (and `git diff --staged` if anything's staged), and **check the current branch is the right place for this commit**:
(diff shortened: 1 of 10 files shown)


adamlobler (commented):
Thanks for the quick work, this is exactly what I had in mind! 🔥

I would only adjust the layout a bit because the prop selectors are very wide, while the preview is quite small. I think a layout where the component preview and the prop selectors are next to each other would work better. Right now, if a component has a lot of props, you can’t see the preview without scrolling.

Also, if it’s possible, it would be great to be able to change the theme locally within this view. It’s a small UX improvement, but it’s something we do a lot. Currently, you have to scroll back up, change the theme, wait for the page to reload, and then set the props again, which is a lot of effort for a quick check that we do a lot.

git-nandor (changes requested):
The theme selector is not in sync with the version selected on the page.

Image

git-nandor (approved):

HerrTopi (approved):

PR #2660, opened 2026-07-27
feat(ui-card): add Card component with base and nested variants

adamlobler:
## Summary
- Add `@instructure/ui-card`: a wrapper with `variant` (`base` | `nested`) and `size` (`sm` | `md` | `lg` | `auto`), built on the v11.7 token system (`useStyleNew` + generated Card tokens) and exported from `@instructure/ui/v11_7`
- Bump `instructure-design-tokens` to `v1.8.0` (the Card token set first landed in `v1.7.0`; `v1.8.0` flattens Card's token structure)
- Add `size="auto"` for `variant="base"`: measures the Card's own rendered width via a CSS container query and picks `sm`/`md`/`lg` padding and border radius itself, using the same breakpoint tokens as the explicit sizes. Not supported for `nested`, where size is a deliberate content choice rather than a reflection of available width — falls back to a normal render.
- Document Card with per-size nested examples, mixed sizes inside one base Card, do/don't composition guidelines, and a live auto-sizing example
- Add unit tests asserting concrete token values (including `size="auto"` at fixed container widths, its extra wrapper element, and its `nested` fallback), plus the `/card` visual regression page (now including `size="auto"` cases) and its axe check

## Test Plan
- Docs → Card: switch the Theme selector to a dark theme and confirm example copy stays legible (examples use `Heading`/`Text` so they inherit token-driven colours rather than a browser default)
- Docs → Card → "Mixed sizes": confirm differently-sized `nested` Cards sit side by side cleanly inside one `lg` `base` Card
- Docs → Card → "Auto sizing (prototype)": resize the example panel and confirm padding/border-radius shift at the `md`/`lg` breakpoints without a `size` prop
- Under the `canvas (legacy)` theme, compare `base` Card corner radii across `sm`/`md`/`lg` — see the token note below
- Visual regression: review the new `/card` baseline

## Notes
- `pnpm-lock.yaml` carries 4 lines of drift unrelated to Card: vitest re-resolved `es-module-lexer` `2.1.0` → `2.2.0` within its pre-existing `^2.0.0` range, which also cleared a now-stale `optional: true` flag on the 2.2.0 entry. No `package.json` range was changed and webpack still resolves 2.1.0.
- `size="auto"` renders an extra wrapper `div` around the styled Card (a CSS container query can't target the element that declares `containerType` itself, only its descendants) — every other mode keeps the existing single-`div` output.

### DIFF
--- a/packages/__docs__/package.json
+++ b/packages/__docs__/package.json
@@ -46,6 +46,7 @@
     "@instructure/ui-buttons": "workspace:*",
     "@instructure/ui-byline": "workspace:*",
     "@instructure/ui-calendar": "workspace:*",
+    "@instructure/ui-card": "workspace:*",
     "@instructure/ui-checkbox": "workspace:*",
     "@instructure/ui-color-picker": "workspace:*",
     "@instructure/ui-color-utils": "workspace:*",
--- a/packages/__docs__/tsconfig.build.json
+++ b/packages/__docs__/tsconfig.build.json
@@ -69,6 +69,9 @@
       "path": "../ui-calendar/tsconfig.build.json"
     },
     {
+      "path": "../ui-card/tsconfig.build.json"
+    },
+    {
       "path": "../ui-checkbox/tsconfig.build.json"
     },
     {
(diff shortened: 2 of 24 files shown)


adamlobler:
@copilot resolve the merge conflicts in this pull request

PR #2663, opened 2026-07-28
Migrate the whole test suite to Vitest browser mode and delete Cypress tests

matyasf:
Convert Jest unit tests and Cypress tests to Vitest browser tests because:

- Much faster now: ~15 secs in total vs ~22 secs + 5 mins.  
- Everything tests in a real browser.   
- Single test suite for everything  

**Changes in detail:**
                                                                                                 
  - All 181 test files are converted off React Testing Library onto the vitest-browser API — This was a mechanical codemod pass with no change of test intent. 
  - The 56 Cypress specs move into their matching `__tests__` files, `Truncate.cy.tsx` and `i18n.cy.tsx` land in new files (`truncate.test.tsx`, `getTextDirection.test.tsx`) since they had no unit test file to merge into. The whole Cypress component-testing setup and its docs are then deleted.            
  - The cases that needed real thought are in a separate commit  (https://github.com/instructure/instructure-ui/commit/eb53eb36bdb80e18d34b13e74f81a2be14593c8f)
  - Performance: `ui-icons` and `ui-themes` resolve to their built `es/` output, which roughly halves suite runtime (with a catch, see below)
  - `DateInput` (the largest Cypress spec) is split across two files plus its optimized a bit for speed.
  -  `vitest.setup.ts` and its jsdom shims are removed as unnecessary.        
  - One product fix: `View.omitViewProps` logged bad error messages 
 
> Note: after editing the source of `ui-icons` or `ui-themes` you now have to build that package before browser tests see the change. 
                                                                                                                                                                 
  For reviewers   

  - Run the full suite locally (`pnpm run test:vitest`) and confirm it's green and meaningfully faster than on master
  - Verify the CI config: one Vitest job instead of two, and confirm nothing that used to be covered by the Cypress job is now unrun.                                                            
  - Spot-check the "needed real thought" commit (eb53eb36bd)                                                                                                                                    
  - Sanity-check the ported Cypress specs against the deleted `cypress/component/` files                                                       
  - Test the `instui-prebundle-stamp` invalidation by hand: edit `ui-themes` source, rebuild it, and confirm the tests pick up the change; check the other changes in `vitest.config.mts`                     
  - Check the DateInput split.


tsconfig.json
```diff
@@ -1,32 +1,8 @@
 {
-  "compilerOptions": {
```
matyasf, line 2: turns out that `tsconfig.json` can be very similar to `tsconfig.build.json`, this makes the tests not emit TS errors (because the old one was using an old value for moduleResolution)

vitest.config.mts
```diff
@@ -37,6 +38,45 @@ const momentDir = path.dirname(
...
+
+
+const packageDir = (pkgName: string) =>
+  path.resolve(__dirname, 'packages', pkgName.replace('@instructure/', ''))
+
+// Packages deliberately NOT aliased to source, so Vite resolves them to their
+// built `es/` output.
+// This trades DX for speed — after editing the *source* of one
+// of these, run that package's `build` before browser tests will see the change.
+const PREBUNDLED_PACKAGES = ['@instructure/ui-icons', '@instructure/ui-themes']
```
matyasf, line 50: I am unsure, that this is a good change. The tests run in:

- ~45 secs is nothing is pre-bundled
- ~25 secs if only `ui-icons` is pre-bundled
- ~15 secs if `ui-icons` and `ui-themes` are pre-bundled
- (~11 secs if everything is pre-bundled, but in this case one needs to rebuild the component(s) to see changes)

packages/ui-view/src/View/v2/index.tsx
```diff
@@ -99,8 +99,11 @@ class View extends Component<ViewProps> {
       (process?.env?.GITHUB_PULL_REQUEST_PREVIEW === 'true' ||
         process?.env?.NODE_ENV !== 'production')
     ) {
+      const name = Component.displayName
+        ? Component.displayName
+        : Component.name
       Object.keys(pickProps(props, propsToOmit)).forEach((prop) => {
-        error(false, `[${Component.name}] prop '${prop}' is not allowed.`)
+        error(false, `[${name}] prop '${prop}' is not allowed.`)
```
matyasf, line 106: This (and the same change for `v1`) is the only code change

joyenjoyer (changes requested):
Please take a look at my comments here: https://github.com/instructure/instructure-ui/commit/eb53eb36bdb80e18d34b13e74f81a2be14593c8f#diff-a6ee97c84dbd70f44ac3ed51e0a2be0bfe0ef24529be5407eb92a3cb7e8cc518

Those are minor stuff but go through them anyways, very nice work!

HerrTopi (approved):
Only question is if you thought about the testing documentation. Does it need changing? Otherwise nice work, approved

joyenjoyer (approved):

Issue #2276, opened 2025-12-01
Alert wants `role="alert"` added to live region


buckett:
> Important note: if you are an Instructure employee please use Slack for bug reports/questions/feature requests. We can assist you much quicker and easier that way. If you are a third party user please ignore this message.

### Background Information

**Package Version(s)**:

10.23.0

**Browser**:

Chrome

**OS:**

MacOS 

**Device:**

MacBook Pro

**Component**:

Alert

**Screen Reader:**

VoiceOver

### Describe the Issue

When specifying a `liveRegion` for an `<Alert>` an error is put in the console unless the live region has `role="alert"` specified. In practice there are other ways to specify this (eg `aria-live="polite"`) and `role="alert"` implies assertive announcements which can be misleading.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Guides/Live_regions

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alert_role

### Steps To Reproduce

If you have this snippet of JSX:

```
        <div
          className="alerts-container"
          aria-live="polite"
          id="live-region"
        ></div>
        <Alert
          variant={alert.variant}
          renderCloseButtonLabel="Dismiss alert"
          margin="xx-small 0"
          liveRegion={() => document.getElementById("live-region")}
          liveRegionPoliteness="polite"
        >
          Just an alert.
        </Alert>
```

it will generate a console error.

You can see this running in:

https://codesandbox.io/p/sandbox/hungry-leftpad-pky6dg

and open the codesandbox preview console to see:

<img width="683" height="907" alt="Image" src="https://github.com/user-attachments/assets/63cac517-cb86-48e9-8bf5-b70ccc1d690f" />

### Expected Behavior

Either no error message, or documentation updated to explain why the live region needs to be assertive when defined in the DOM.

### Additional Information

**Current Workaround(s):**

Just ignore the error.

**Products Affected:**

**Are you willing to submit a PR to fix?**

- [ ] Yes, I'm willing to submit a PR

**Requested Priority:**

Low

PR #2664, opened 2026-07-29
fix(ui-alerts,console): remove Alert live-region role warning

ToMESSKa:
INSTUI-5131

**ISSUE:** 
- Alert warns that your live region must have role="alert", but role="alert" means **assertive** per WAI-ARIA, so even when you'd set liveRegionPoliteness="polite" (**non-assertive**), you got a warning asking for role="alert". It was likely a leftover before the liveRegionPoliteness prop was added. GitHub issue: https://github.com/instructure/instructure-ui/issues/2276

> Setting role="alert" is equivalent to setting aria-live="assertive" and aria-atomic="true". [...] Dynamic changes that are less urgent should use a less aggressive method, such as including aria-live="polite" or using an other live region role like status.
> https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alert_role

**TEST PLAN:**
- run it locally
- add this to `packages/__docs__/src/index.html` after line 4 (webpack 5 doesn't provide `process` in the browser so typeof process resolves to "undefined" in `packages/console/src/console.ts` and log messages are not emitted)
```
     window.process = window.process || { env: {} }
    </script>
    <script>
```
- test this example, it should not emit a warning

```
<div>
  <div className="alerts-container" id="live-region"></div>
  <Alert
    variant="info"
    renderCloseButtonLabel="Dismiss alert"
    liveRegion={() => document.getElementById("live-region")}
    liveRegionPoliteness="polite"
  >
    Just an alert.
  </Alert>
</div>
```
- if you revert the fix, the warning should reappear

### DIFF
--- a/packages/ui-alerts/src/Alert/v2/README.md
+++ b/packages/ui-alerts/src/Alert/v2/README.md
@@ -231,7 +231,7 @@ type: embed
     <Figure.Item>Use the Error alert to notify user of an error</Figure.Item>
     <Figure.Item>Use the Warning alert to notify user of a warning</Figure.Item>
     <Figure.Item>Use the Success alert to notify user of a success event or action</Figure.Item>
-    <Figure.Item>Use the `variantScreenReaderLabel` prop to indicate the alert variant to screen reader users</Figure.Item>
+    <Figure.Item>Use the <code>variantScreenReaderLabel</code> prop to indicate the alert variant to screen reader users</Figure.Item>
   </Figure>
   <Figure recommendation="no" title="Don't">
     <Figure.Item>Have alert messaging that is more than two lines long</Figure.Item>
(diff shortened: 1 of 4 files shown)


packages/ui-alerts/src/Alert/v2/README.md
```diff
@@ -250,6 +250,7 @@ type: embed
     <Figure.Item>aria-live="polite" alerts will only be announced if the user is not currently doing anything. Polite should be used in most situations involving live regions that present new info to users</Figure.Item>
     <Figure.Item>aria-live="assertive" alerts will be announced to the user as soon as possible, but not necessarily immediately. Assertive should be used if there is information that a user must know about right away, for example, a warning message in a form that does validation on the fly</Figure.Item>
     <Figure.Item>The aria-atomic=BOOLEAN is used to set whether or not the screen reader should always present the live region as a whole, even if only part of the region changes. The possible settings are: false or true. The default setting is false.</Figure.Item>
+    <Figure.Item>If the `liveRegion` element has a `role`, match it to the `liveRegionPoliteness` level: use `role="status"` for `polite` and `role="alert"` for `assertive`.</Figure.Item>
```
matyasf, line 253: hmm this is not the best description, its more complicated. Please add this:

<Figure.Item>
For the `liveRegion` DOM element, note that the [`role` prop adds implicit ARIA attributes, but these are not applied, unless you set `liveRegionPoliteness` or `isLiveRegionAtomic` to `undefined`.](https://www.w3.org/TR/wai-aria-1.2/#implictValueForRole)
</Figure.Item>

matyasf (changes requested):
see my comment

packages/ui-alerts/src/Alert/v2/README.md
ToMESSKa: @matyasf `liveRegionPoliteness `and `isLiveRegionAtomic` both have default values.
```
65    static defaultProps = {
...
72      liveRegionPoliteness: 'assertive', 
73      isLiveRegionAtomic: false,
...
76    }
```
So `aria-live` and `aria-atomic` are always explicitly set even if these props are undefined, e.g.:
```
<div>
  <div className="alerts-container" id="live-region"></div>
  <Alert
    variant="info"
    renderCloseButtonLabel="Dismiss alert"
    liveRegion={() => document.getElementById("live-region")}
  >
    Just an alert.
  </Alert>
</div>
```

matyasf: hmm, good catch with the defaults. An explicit `aria-live` overrides the role's implicit value: `<div role="alert" aria-live="polite">` is announced politely.
so a better description here would be:

<Figure.Item>Do not set the`role` prop on the liveRegion because `liveRegionPoliteness` and `isLiveRegionAtomic` overrides the set values.</Figure.Item>

@ToMESSKa no need for you to be working on this ticket, Nandi has taken it over

matyasf (changes requested):
see my comment

matyasf (approved):

HerrTopi (approved):

PR #2665, opened 2026-07-29
fix: lint:changes command to only lint changed files

joyenjoyer:
## Summary
- Replace broken `--since HEAD^` flag with `git merge-base` to properly target changed files
- Pass changed files directly to oxlint instead of routing through workspace runner
- Fixes lint:changes to skip unchanged portions of the codebase

## Test Plan
- `pnpm run lint:changes` now only lints files in the diff since merge-base

Fixes INSTUI-5090

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "test:vitest-watch": "playwright install chromium && vitest --project browser --project node",
     "test:vitest-browser-ui": "playwright install chromium && vitest --project browser --browser.headless=false --ui",
     "lint": "pnpm -r --stream lint",
-    "lint:changes": "pnpm run lint -- --since HEAD^",
+    "lint:changes": "git diff --name-only $(git merge-base HEAD origin/master)..HEAD | xargs -r oxlint --format=unix",
     "lint:fix": "pnpm -r --stream lint:fix",
     "lint:commits": "commitlint --from=HEAD^1",
     "bootstrap": "node scripts/bootstrap.mjs",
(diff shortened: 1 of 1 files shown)


balzss (changes requested):
this command breaks workflows without js/ts changes since oxlint exits with 1. ironically this very PR is a good example

package.json
```diff
@@ -28,7 +28,7 @@
     "test:browser-watch": "pnpm run playwright:install && vitest --project browser",
     "test:browser-ui": "pnpm run playwright:install && vitest --project browser --browser.headless=false --ui",
     "lint": "pnpm -r --stream lint",
-    "lint:changes": "git diff --name-only $(git merge-base HEAD origin/master)..HEAD | xargs -r oxlint --format=unix",
+    "lint:changes": "git diff --name-only $(git merge-base HEAD origin/master)..HEAD | { grep -E '\\.(js|jsx|ts|tsx|mjs|cjs)$' || true; } | xargs -r oxlint --format=unix",
```
joyenjoyer, line 31: @balzss 
- git diff: list files changed since branching from master
- grep (+ || true): keep only lintable extensions; don't fail if none match
- xargs -r oxlint: only run oxlint if files remain, so no-match no longer exits 123

balzss (changes requested):
**Issue with the current solution:** `git diff --name-only` also lists deleted files, and oxlint exits 1 (`No files found to lint`) when every path it gets no longer exists — so a PR whose only lintable change is a deletion (e.g. removing an obsolete test) will fail the lint job. The pre-commit hook won't catch it either, since lint-staged skips deleted files.

**Suggested fix:** add `--diff-filter=ACMR` to the `git diff` call so deleted paths never reach oxlint.

please note that the fix was suggested by claude and needs a double check

balzss (approved):
looks good, i'll aprove but i'll also mention 2 nitpicks:
- `docs/contributing/contributing-getting-started.md:54` mentions this command linting since the last commit but actually it lints since branching off from master
- maybe scripts like this could live in ui-scrips and only be referenced here for better readability

git-nandor (approved):

PR #2666, opened 2026-07-29
fix(ui-table): don't describe a sort state on unsorted v1 tables

drakeaharper:
✨ _Authored with AI assistance._

## Summary

A v1 `Table` given a plain string caption appends `" Sorted by undefined (undefined)"` to both its `<caption>` and its accessible name whenever **no** column is sorted:

```
aria-label="Movies Sorted by undefined (undefined)"
<caption>Movies Sorted by undefined (undefined)</caption>
```

`getCaptionText` lost its early return for the unsorted case between `11.7.4-SECURITY.3` and `11.7.4`:

```js
// 11.7.4-SECURITY.3
const caption = props.caption as string
if (!sortInfo) return caption
const sortText = ` Sorted by ${sortInfo.header} (${sortInfo.direction})`

// 11.7.4 — early return gone, so sortInfo is always interpolated
const sortText = ` Sorted by ${sortInfo?.header} (${sortInfo?.direction})`
return caption ? caption + sortText : sortText.trim()
```

This is user-facing for screen reader users, not only a test problem. It still reproduces on `11.7.5-snapshot-14`.

## Changes

- Restore the early return so an unsorted table uses its caption as given. The `caption`-as-function path is untouched, and the cast matches what SECURITY.3 did (`props.caption as string`).
- Add `packages/ui-table/src/Table/v1/__tests__/Table.test.tsx`. The existing `Table` tests import `@instructure/ui-table/latest` and only ever pass a caption *function*, so the v1 string-caption path had no coverage — which is how this shipped. The new tests import relatively so they exercise source rather than built output.

## Verification

- New tests fail on unpatched source (2 of 3, printing the literal `Movies Sorted by undefined (undefined)`) and pass with the fix. The third covers the sorted case and passes either way, guarding against over-correcting.
- `pnpm run test:vitest ui-table` — 25 passed.
- `pnpm run bootstrap` clean, including `build:types` (the first draft of this fix failed type-checking; `aria-label` needs a `string`).

## Context

Found upgrading canvas-lms off the Artifactory-only `11.7.4-SECURITY.3` onto public `11.7.4`, where it broke 10 vitest files and 6 Selenium specs that locate tables by accessible name. Reported in #instui. Canvas is carrying a `patch-package` patch for this that we'll drop once a fixed release ships.

## One question for reviewers

`componentDidUpdate` still has a commented-out condition:

```js
if (
  sortingChanged &&
  currentSortInfo &&
  // typeof this.props.caption === 'function' &&
  this._liveRegionRef
) {
```

That looks like this change was mid-flight, and the accompanying comment says sort changes should only be announced for function captions. I've left it exactly as-is because I don't know the intent — if the live-region announcement is also meant to be gated, that's a follow-up (or tell me and I'll fold it in here).

### DIFF
--- a/packages/ui-table/src/Table/v1/__tests__/Table.test.tsx
+++ b/packages/ui-table/src/Table/v1/__tests__/Table.test.tsx
@@ -1,0 +1,92 @@
+/*
+ * The MIT License (MIT).
+ *
+ * Copyright (c) 2015 - present Instructure, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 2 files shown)


balzss:
Closing in favor of #2671

PR #2667, opened 2026-07-29
fix(ui-motion): only pass elementRef to children that accept it

drakeaharper:
## Summary

`BaseTransition.renderChildren()` treats `typeof child.type === 'object'` as "this is a `withStyle`-decorated InstUI component":

```js
// `typeof type === 'object'` => forwardRef wrapper (withStyle-decorated InstUI components)
const refProps =
  typeof child.type === 'object'
    ? { elementRef: createChainedFunction(...), ref: elementOnlyRef }
    : { ref: (el) => this.handleRef(...) }
```

But that's true of **every** `forwardRef` wrapper. `@emotion/react`'s wrapper matches it and forwards unknown props straight to the DOM node, so any emotion element wrapped in a `Transition` warns:

```
Warning: React does not recognize the `elementRef` prop on a DOM element.
    at span
    at @emotion/react/dist/emotion-element…
    at BaseTransition (@instructure/ui-motion/es/Transition/BaseTransition/index.js:52)
    at Transition → DrawerTray → DrawerLayout
```

`DrawerLayout.Tray` renders an emotion element inside a `Transition`, so every tray and dialog built on it logs this. `11.7.4-SECURITY.3` used `ref` + `findDOMNode` and never passed `elementRef`, so this appeared in `11.7.4`. Still present on `11.7.5-snapshot-14`.

## Changes

- Gate the `elementRef` branch on the child actually declaring `elementRef` in `allowedProps`. InstUI components decorated with `withStyle` expose it; emotion's wrapper doesn't, so it falls back to the plain `ref` path — the behaviour that shipped before this check existed.
- Keep the chaining for children that do accept it, so a child's own `elementRef` still fires rather than being overwritten.
- Add three tests under `<Transition />` covering ref forwarding.

## Verification

- The first new test fails on unpatched source and passes with the fix. The other two are guards: one proves the child node is still found via the fallback `ref`, and one proves an `allowedProps`-declaring child still receives a chained `elementRef` — that pair would catch over-correcting by dropping the `elementRef` branch entirely.
- `pnpm run test:vitest ui-motion` — 21 passed.
- `pnpm run bootstrap` clean, including `build:types`.

## Context

Found upgrading canvas-lms off the Artifactory-only `11.7.4-SECURITY.3` onto public `11.7.4`. canvas-rce's vitest setup escalates `console.error` to a failure, so this alone failed **380 tests across 23 files**, all Tray/Dialog specs. Reported in #instui. Canvas is carrying a `patch-package` patch for this that we'll drop once a fixed release ships.

## Alternative worth considering

Gating on `allowedProps` is the narrowest fix that keeps your new chaining behaviour. If you'd rather not rely on that static, the other options are an explicit opt-in prop on the child, or reverting to the pre-11.7.4 `ref` + `findDOMNode` approach and solving the overwrite problem differently. Happy to reshape this if you prefer one of those.

### DIFF
--- a/packages/ui-motion/src/Transition/BaseTransition/index.ts
+++ b/packages/ui-motion/src/Transition/BaseTransition/index.ts
@@ -347,26 +347,39 @@ class BaseTransition extends Component<
       }
     }
 
-    // `typeof type === 'object'` => forwardRef wrapper (withStyle-decorated InstUI components)
-    const refProps =
-      typeof child.type === 'object'
-        ? {
-            // chain so the child's own elementRef still fires instead of being overwritten
-            elementRef: createChainedFunction(
-              (child.props as { elementRef?: (el: Element | null) => void })
-                ?.elementRef,
-              this.handleRef
-            ),
-            // fallback for forwardRef children that expose their node via `ref`, not elementRef
-            ref: elementOnlyRef
-          }
-        : {
-            // for host el / plain class|fn: findDOMNode is the fallback
-            ref: (el: ReactInstance | Element | null) =>
-              this.handleRef(
-                el instanceof Element ? el : (findDOMNode(el) as Element) ?? null
-              )
-          }
+    // `typeof type === 'object'` covers every forwardRef wrapper, not just
+    // withStyle-decorated InstUI components. `@emotion/react`'s wrapper matches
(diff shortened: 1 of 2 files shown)


balzss:
Closing in favor of #2672

PR #2669, opened 2026-07-30
ci: keep visual-regression baselines refreshing and surface Cypress failures

balzss:
## Summary

Repairs the self-hosted visual-regression pipeline and makes the report far more useful for triage. Two related concerns, in separate commits:

**1. Baseline pipeline fix (`ci`)**
- `visual-baselines.yml`: decouple baseline capture from the a11y/console gate — Cypress runs with `continue-on-error`, screenshots always flatten + push (`if: always()`), and a trailing step fails the job for visibility if Cypress failed. Fixes the **silent baseline freeze**: the `visual-baselines` branch had stopped updating on every merge since the dark/light theme suite landed (Cypress aborted the job before the push step).
- `visual-regression.yml`: the sticky PR comment now leads with a **"Cypress assertions (a11y + console errors)" ✅/❌** line, so failures are visible even when the visual diff is clean.
- Removed the leftover `TEMP inject diff-demo fixture` step and `regression-test/cypress/diff-demo/`.

**2. A11y in the visual-diff report (`feat`)**
- The suite already ran axe per page/theme but discarded the results. A new `recordA11y` Cypress task now persists them to `cypress/a11y.json`, keyed by `<slug>-<theme>` (mirrors `meta.json`).
- `ui-scripts visual-diff` gains a `--a11y` flag that renders a per-page **`⚠ N a11y` badge**, an **`⚠ A11y` filter chip**, a header count, and a collapsible **violation list** (rule linked to its axe `helpUrl`, impact, offending selector, contrast summary). `a11yPages`/`a11yViolations` added to `summary.json`. Additive — without `--a11y` the report is unchanged.
- Wired `--a11y cypress/a11y.json` into the diff step.

## Test Plan
- Unit: `visual-diff` suite (34 tests, incl. new `esc`/`a11yFor`).
- Verified end-to-end against a real suite run: the spec emits `a11y.json` (15 flagged page/theme entries) and the report renders badges/filter/list with the real axe contrast data.
- On merge: confirm **Update visual baselines** publishes to `visual-baselines` even while the suite is red, and ends red with the `::error::` annotation.
- Note: 13 dark/light-theme `color-contrast` violations remain (tracked in INSTUI-5137) — the suite stays red-but-visible until those are fixed by design/component work.

Fixes INSTUI-5137


regression-test/cypress/support/a11y-capture.ts
```diff
@@ -0,0 +1,210 @@
...
+  nodes: CapturedNode[]
+}
+
+export type CapturedPage = {
+  page: { w: number; h: number }
+  violations: CapturedViolation[]
+}
+
+// Longer than this and the label stops being a label.
+const MAX_LABEL_TEXT = 60
```
git-nandor, line 78: Great job! I only found one small thing.
This is a bit confusing. I don't think we need a comment here.

git-nandor (changes requested):

ToMESSKa (changes requested):
Image

I think the link is missing from here

git-nandor (approved):

ToMESSKa (approved):

PR #2670, opened 2026-07-30
fix(regression-test): pair the theme page surface with its text color

drakeaharper:
✨ _Authored with AI assistance._

## Why

The **Visual regression** job has been failing on every open PR since 2026-07-27 — not just mine. PRs merged Jul 22–24 passed it; everything since fails. `feat/ui-card`, `vitest_browser_convert` and every other open branch report the same 13 `color-contrast` (serious) violations.

## Commit 1 — pair the page surface with its text color

`a1e26e3a9` started painting each page from the active theme's `background.page`, but never set a matching foreground:

```jsx
const pageBackground = semantics?.color?.background?.page ?? 'transparent'
<html style={{ background: pageBackground }}>
```

Inherited text therefore stays at its default dark value while the surface goes near-black:

| Theme | `background.page` | `color.text.base` |
|---|---|---|
| canvas | `#ffffff` | `#273540` |
| light  | `#F2F4F5` | `#273540` |
| dark   | `#10141A` | `#F2F4F5` |

Roughly 1.2:1 instead of the ~15:1 the paired token gives. `99746bfa5` then made the a11y check run across canvas + light + dark rather than canvas only, which turned it into a suite failure.

Fixed by setting `color` from `text.base` next to the background, on both `<html>` and `<body>` (the body needs it for the same `capture: 'fullPage'` reason the existing comment describes for the background).

## Commit 2 — remove hardcoded colors from three demo pages

| Page | Was | Ratio |
|---|---|---|
| `custom-icons` | Tailwind `text-gray-500` / `text-gray-700` | 4.38 light / 1.79 dark |
| `small-components` | `Text color="inherit"` inside a hardcoded `#AA0000` span | 2.38 dark |
| `tooltip` | bare `<input>` with no color, inheriting light text on the UA's white background | 1.13 dark |

## Result

| | a11y failures |
|---|---|
| master | **13** — matches CI exactly |
| + commit 1 | 6 |
| + commit 2 | **4** |

Verified locally against the full spec, and commit 1's 13 → 6 was independently confirmed by this PR's own CI run.

## The remaining 4 are not harness bugs — they need your call

All four are InstUI components sitting on InstUI's own `View` background variants, using the idiomatic pattern your `heading` and `progressbar` pages use:

- `<span class="baseButton__children">primary-inverse</span>` — inverse Button
- `<a class="view-link">jumps</a>` — Link
- View page `alert` / `success` / `danger` / `warning` variants with default text
- inverse `Text`

Two distinct contrast pairs remain:

```
#ffffff on #f2f4f5 — 1.10  (x3)   white inverse text on the LIGHT theme's page surface
#1c222b on #2b7abc — 3.51  (x1)   dark text on brand blue, under the 4.5:1 AA floor
```

The 1.10:1 case looks like a genuine defect: `View background="primary-inverse"` appears not to resolve to a dark surface under the light theme while `color="primary-inverse"` still resolves to white, so the two stop agreeing. That would affect any consumer using inverse variants on the light theme, not just this suite. See the [detailed comment](https://github.com/instructure/instructure-ui/pull/2670#issuecomment-5136318956).

I deliberately have **not** forced these green. The only way to do that from the harness is to override the text colors on your components inside the demos, which would hide the behaviour rather than fix it. I tried the non-invasive route first — passing `style={{ background: ... }}` to the wrapping `View` — and it's a no-op, since `style` isn't in View's `allowedProps`.

Happy to go either way: tell me the inverse tokens are working as intended and I'll adjust the demos to suit, or leave these red until the tokens are addressed.

## Context

Found while chasing red CI on #2666 and #2667 (two unrelated 11.7.4 regression fixes). Both add zero violations of their own and are blocked behind this job.

### DIFF
--- a/regression-test/src/app/custom-icons/page.tsx
+++ b/regression-test/src/app/custom-icons/page.tsx
@@ -97,7 +97,7 @@ function IconGrid({
           <Icon size={size} color={color} title={name} />
           <span
             style={{ fontSize: '15px' }}
-            className="text-gray-500 text-center leading-tight"
+            className="text-center leading-tight"
           >
             {name.replace('InstUIIcon_', '')}
           </span>
@@ -121,7 +121,7 @@ function SizeRow({
       {SIZES.map((size) => (
         <div key={size} className="flex flex-col items-center gap-1 p-1">
           <Icon size={size} color={color} title={name} />
-          <span style={{ fontSize: '15px' }} className="text-gray-500">
+          <span style={{ fontSize: '15px' }} className="">
             {size}
           </span>
         </div>
@@ -132,7 +132,7 @@ function SizeRow({
 
 function SectionHeader({ children }: { children: React.ReactNode }) {
   return (
-    <p className="text-xs font-bold mt-7 mb-2 pb-1 border-b border-gray-300 text-gray-700">
+    <p className="text-xs font-bold mt-7 mb-2 pb-1 border-b border-gray-300">
       {children}
     </p>
   )
--- a/regression-test/src/app/layout.tsx
+++ b/regression-test/src/app/layout.tsx
@@ -87,6 +87,12 @@ export default function RootLayout({
       ? newTheme.semantics(newTheme.primitives)
       : newTheme?.semantics
   const pageBackground = semantics?.color?.background?.page ?? 'transparent'
+  // Pair the surface with the theme's own text color. Painting only the
+  // background leaves inherited text at its default dark value, which on the
+  // dark theme's near-black surface is a serious color-contrast violation for
+  // every plain label on these pages and for any component using
+  // `color="inherit"`.
+  const pageColor = semantics?.color?.text?.base ?? 'inherit'
 
   return (
     // we need to make a new Map to reset counting on the server side
@@ -94,7 +100,7 @@ export default function RootLayout({
     <html
       lang="en"
       data-theme={themeKey}
-      style={{ background: pageBackground }}
+      style={{ background: pageBackground, color: pageColor }}
     >
       <head>
         <title>Component visual and regression test suite</title>
@@ -111,7 +117,11 @@ export default function RootLayout({
             in the captured baselines. */}
         <body
           className={inter.className}
-          style={{ background: pageBackground, minHeight: '100vh' }}
+          style={{
+            background: pageBackground,
+            color: pageColor,
+            minHeight: '100vh'
+          }}
         >
           {children}
         </body>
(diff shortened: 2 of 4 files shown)


drakeaharper:
✨ _Authored with AI assistance._

## Contrast data for the 6 remaining failures — two look like real defects, not demo mistakes

I pulled the actual axe colour pairs. They split cleanly, and I want the second group on record even though I'm about to adjust the demos so the suite goes green.

### Demo-page hardcoded colours (my problem, fixing in this PR)

| fg | bg | ratio | where |
|---|---|---|---|
| `#aa0000` | `#10141a` | 2.38 | `small-components` wraps `<Text color="inherit">` in `<span style={{ color: '#AA0000' }}>` |
| `#6a7282` | `#f2f4f5` | 4.38 | `custom-icons` uses Tailwind `text-gray-500` |
| `#364153` | `#10141a` | 1.79 | `custom-icons` uses Tailwind `text-gray-700` |
| `#000000` | `#10141a` | 1.13 | `tooltip` renders a bare `<input>` with no colour set |

None of these are component bugs — `color="inherit"` in particular is behaving exactly as documented.

### Possible genuine dark/light theme defects (your call)

**1. Inverse text at 1.1:1 on the light theme — appears 3×**

```
foreground #ffffff, background #f2f4f5, ratio 1.10
```

`#f2f4f5` is the **light** theme's `background.page`. So white inverse text is landing on the light theme's page surface. The affected pages use the idiomatic pattern that your own `heading` and `progressbar` pages use:

```jsx
<View background="primary-inverse" as="div" padding="small medium">
  <Button color="primary-inverse">primary-inverse</Button>
</View>
```

That reads as `View background="primary-inverse"` not resolving to a dark surface under the light theme, while `color="primary-inverse"` still resolves to white — so the two stop agreeing. 1.1:1 is effectively invisible text. If that's right it affects any consumer using inverse variants on the light theme, not just this suite.

**2. Body text on brand blue at 3.51:1**

```
foreground #1c222b, background #2b7abc, ratio 3.51
```

Below the 4.5:1 AA threshold for normal-size text.

### Why I'm changing the demos anyway

We need CI green to land two unrelated regression fixes (#2666, #2667) that are blocked behind this job, so I'm adjusting the demo pages to pass. That will make these two items **stop being visible in CI**, which is exactly why I'm recording the numbers here. If you'd rather I revert the demo changes for the inverse and brand-blue cases and leave the suite red until the tokens are addressed, say so and I'll do that instead — happy either way.

balzss:
@drakeaharper thanks for the PR, this issue will be solved in https://github.com/instructure/instructure-ui/pull/2669, so I'm closing yours in favor of that

PR #2671, opened 2026-07-31
fix(ui-table): don't append a sort state to unsorted v1 captions

balzss:
## Summary
An unsorted v1 `Table` with a plain string `caption` renders ` Sorted by undefined (undefined)` into both its `<caption>` and its accessible name:

```
aria-label="Movies Sorted by undefined (undefined)"
```

`getCaptionText` lost its `if (!sortInfo) return caption` guard in aac2b1790f (#2574), which restructured it around the new `TableCaption` function form. Shipped in v11.7.4. v2 is unaffected — `caption` must be a function there.

Our own docs site is affected too: the `__docs__` Properties, Params, Returns and ComponentTheme tables all use v1 with string captions.

## Changes
- Restore the early return in `Table/v1/index.tsx`. The function-caption branch is untouched, and the `as string` cast matches SECURITY.3 — the value feeds `aria-label`, which needs a string.
- Add `Table/v1/__tests__/Table.test.tsx`. v1 had no test coverage at all, which is why this shipped: the existing suite imports `/latest` (v2), and #2574 converted its last string captions to functions.

## Test Plan
- The two unsorted tests fail on unpatched source and pass with the fix. The sorted test passes either way, guarding against over-correcting.
- `pnpm run test:vitest ui-table` — 25 passed. Full unit suite — 2218 passed. `build:types` clean.
- Docs app: Properties and Component-theme captions render clean with matching `aria-label`s, and the function-caption sort form still works.

Worth knowing for future tests: the new file imports relatively on purpose. `getWorkspaceAliases()` in `vitest.config.mts` is only wired into the browser project, so a `@instructure/ui-table/*` import in the `web` project resolves to built `es/` output rather than source.

Fixes INSTUI-5138

Reported by @drakeaharper in [#instui](https://instructure.slack.com/archives/C0JCJ63TR/p1785354459934709) while upgrading canvas-lms onto public 11.7.4, where it broke 10 vitest files and 6 Selenium specs. Supersedes #2666.

### DIFF
--- a/packages/ui-table/src/Table/v1/__tests__/Table.test.tsx
+++ b/packages/ui-table/src/Table/v1/__tests__/Table.test.tsx
@@ -1,0 +1,93 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015 - present Instructure, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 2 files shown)


packages/ui-table/src/Table/v1/__tests__/Table.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+import { render, screen } from '@testing-library/react'
+import { MockInstance, vi } from 'vitest'
+// v1 accepts a plain ReactNode caption as well as the caption function that v2
```
joyenjoyer: Overly verbose, messy comment, please remove

joyenjoyer (changes requested):
Please reduce the commit message, it is quite long and verbose

ToMESSKa (approved):

joyenjoyer (approved):

PR #2672, opened 2026-07-31
fix(ui-motion): stop leaking elementRef onto emotion-wrapped DOM children

balzss:
## Summary

Any `<Transition>` child written with a `css` prop logs:

```
Warning: React does not recognize the `elementRef` prop on a DOM element.
    at span
    at BaseTransition
    at Transition -> DrawerTray -> DrawerLayout
```

`BaseTransition` treated `typeof child.type === 'object'` as "withStyle-decorated InstUI component". That's true of every forwardRef wrapper, including the one emotion's jsx runtime puts around any element carrying a `css` prop — and emotion forwards unknown props straight to the DOM node. `Tray`, `DrawerTray`, `RatingIcon` v2 and `Modal` (`constrain="parent"`) each render such a child.

## Changes

Narrow the check to children that actually declare `elementRef` in `allowedProps`. Emotion's wrapper doesn't, so `<div css={...}>` takes the plain `ref` path; a withStyle child keeps the chained `elementRef`.

This is deliberately **not** a revert of #2618 — reverting `renderChildren` brings back the `ref is not a prop` warning that PR fixed, verified by trying it rather than assuming. Two tests bracket the fix so neither behaviour can regress into the other.

## Test Plan

- `ui-motion` passes, as do `ui-tray ui-drawer-layout ui-modal ui-rating ui-overlays ui-toggle-details ui-tabs ui-popover` and the browser project. `build:types` clean.
- Both failure modes checked by temporarily swapping the implementation back.
- Docs app with devtools open: Alert dismiss/re-add and all three DrawerLayout trays — clean console, no `elementref` attributes in the DOM.

Fixes INSTUI-5139

Supersedes #2667, which proposed the same `allowedProps` gate.

### DIFF
--- a/packages/ui-motion/src/Transition/BaseTransition/index.ts
+++ b/packages/ui-motion/src/Transition/BaseTransition/index.ts
@@ -341,32 +341,35 @@ class BaseTransition extends Component<
 
     const child = ensureSingleChild(this.props.children) as ReactElement
 
-    const elementOnlyRef = (el: ReactInstance | Element | null) => {
-      if (el instanceof Element) {
-        this.handleRef(el)
-      }
-    }
+    // Pass elementRef only to children that declare it. Emotion's wrapper
+    // around `<div css={...}>` accepts any prop and forwards it to the DOM
+    // node, so it must take the plain `ref` path.
+    const acceptsElementRef = (
+      child.type as { allowedProps?: readonly string[] }
+    )?.allowedProps?.includes('elementRef')
 
-    // `typeof type === 'object'` => forwardRef wrapper (withStyle-decorated InstUI components)
-    const refProps =
-      typeof child.type === 'object'
-        ? {
-            // chain so the child's own elementRef still fires instead of being overwritten
-            elementRef: createChainedFunction(
-              (child.props as { elementRef?: (el: Element | null) => void })
-                ?.elementRef,
-              this.handleRef
-            ),
-            // fallback for forwardRef children that expose their node via `ref`, not elementRef
-            ref: elementOnlyRef
+    const refProps = acceptsElementRef
+      ? {
+          // chain so the child's own elementRef still fires instead of being overwritten
+          elementRef: createChainedFunction(
+            (child.props as { elementRef?: (el: Element | null) => void })
+              ?.elementRef,
+            this.handleRef
+          ),
+          // fallback for forwardRef children that expose their node via `ref`, not elementRef
+          ref: (el: ReactInstance | Element | null) => {
+            if (el instanceof Element) {
+              this.handleRef(el)
+            }
           }
-        : {
-            // for host el / plain class|fn: findDOMNode is the fallback
-            ref: (el: ReactInstance | Element | null) =>
-              this.handleRef(
-                el instanceof Element ? el : (findDOMNode(el) as Element) ?? null
-              )
-          }
+        }
+      : {
+          // host el / plain class|fn: findDOMNode is the fallback
+          ref: (el: ReactInstance | Element | null) =>
+            this.handleRef(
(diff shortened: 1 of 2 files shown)


packages/ui-motion/src/Transition/__tests__/Transition.test.tsx
```diff
@@ -275,4 +295,75 @@ describe('<Transition />', () => {
...
+      )
+
+      // the node reached handleRef, so the transition classes could be applied
+      expect(getByText('hello')).toHaveClass(getClass('fade', 'entered'))
+      await waitFor(() => {
+        expect(elementRef).toHaveBeenCalledWith(expect.any(Element))
+      })
+    })
+
+    // LX-4014 (#2618): a withStyle child plus an actually-running transition
```
joyenjoyer: Remove this ticket reference from the comment

packages/ui-motion/src/Transition/BaseTransition/index.ts
```diff
@@ -341,32 +341,38 @@ class BaseTransition extends Component<
...
-            // fallback for forwardRef children that expose their node via `ref`, not elementRef
-            ref: elementOnlyRef
-          }
-        : {
-            // for host el / plain class|fn: findDOMNode is the fallback
-            ref: (el: ReactInstance | Element | null) =>
-              this.handleRef(
-                el instanceof Element ? el : (findDOMNode(el) as Element) ?? null
-              )
+    // InstUI components expose an `elementRef` prop, which hands back the DOM
```
joyenjoyer: Rewrite this comment without Claude please, it is quite messy

packages/ui-motion/src/Transition/__tests__/Transition.test.tsx
```diff
@@ -275,4 +295,75 @@ describe('<Transition />', () => {
...
+    const warningsMatching = (mock: MockInstance, pattern: RegExp) =>
+      mock.mock.calls.filter((args: unknown[]) => pattern.test(args.join(' ')))
+
+    const elementRefWarnings = (mock: MockInstance) =>
+      warningsMatching(mock, /elementRef/)
+
+    const refIsNotAPropWarnings = (mock: MockInstance) =>
+      warningsMatching(mock, /`?ref`? is not a prop/)
+
+    // The repo compiles JSX with `jsxImportSource: '@emotion/react'`, so a child
```
joyenjoyer: I don't understand how this comment helps understanding these tests

joyenjoyer (changes requested):

packages/ui-motion/src/Transition/BaseTransition/index.ts
```diff
@@ -341,32 +341,35 @@ class BaseTransition extends Component<
...
-          }
-        : {
-            // for host el / plain class|fn: findDOMNode is the fallback
-            ref: (el: ReactInstance | Element | null) =>
-              this.handleRef(
-                el instanceof Element ? el : (findDOMNode(el) as Element) ?? null
-              )
+    // Only children that declare `elementRef` get one. `typeof child.type ===
+    // 'object'` also matches emotion's wrapper around `<div css={...}>`, which
+    // forwards it to the DOM.
```
ToMESSKa: I think this comment might be confusing as it references the old check which is no longer here.

ToMESSKa (dismissed):

joyenjoyer (changes requested):
Can you simplify the commit message body? It is overly verbose and reference canvas-rce tests and LX-4014, it is too long and adds no value.

joyenjoyer (approved):

HerrTopi (approved):

PR #2673, opened 2026-07-31
fix(ui-checkbox): announce the toggle variant as a toggle button

ToMESSKa:
INSTUI-5023

**ISSUE:**
- Checkbox toggle variant looks and behaves like a toggle button but announced as checkbox so role and state (checkbox/checked) don't match the presented control (toggle). A suggested fix is adding role=”button” and aria-pressed=”false/true” to the `<input>`, see https://instructure.atlassian.net/browse/LRN-11564

**TEST PLAN:**
- check the Checkbox toggle variant with VoiceOver and NVDA, it should be announced as a "toggle button"


joyenjoyer (approved):
looks great

HerrTopi (approved):

PR #2675, opened 2026-08-04
fix(emotion): only warn about a missing theme when the fallback is actually used

matyasf:
## Summary
- Move the "No theme provided for [InstUISettingsProvider], using default `canvas` theme." warning below the `isBaseTheme` early return so it only fires when the `canvas` fallback is actually used — previously it was a false positive whenever a valid theme was passed via the `theme` prop with no ancestor theme.
- Collapse the `currentTheme` if/else into a `hasAncestorTheme` flag; make the `InstUIProviderProps` import type-only.

## Test Plan
- Render `<InstUISettingsProvider theme={canvas}>` (no ancestor provider) and confirm no console warning appears.
- Render `<InstUISettingsProvider themeOverride={{...}}>` with no `theme` and confirm the warning still appears.

Fixes INSTUI-5142

### DIFF
--- a/packages/emotion/src/getTheme.ts
+++ b/packages/emotion/src/getTheme.ts
@@ -31,7 +31,7 @@ import type {
   ThemeOrLegacyOverride,
   SpecificThemeOverride
 } from './EmotionTypes'
-import { InstUIProviderProps } from './InstUISettingsProvider/index.js'
+import type { InstUIProviderProps } from './InstUISettingsProvider'
 declare const process: Record<string, any> | undefined
 
 /**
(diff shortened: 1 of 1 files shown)


HerrTopi (approved):

joyenjoyer (changes requested):
Please look into it, warning is not shown even when I use this line locally
```globalThis.process = { env: { NODE_ENV: 'development' } }```

matyasf:
> Please look into it, warning is not shown even when I use this line locally `globalThis.process = { env: { NODE_ENV: 'development' } }`

We could reproduce it in the end locally

joyenjoyer (approved):

PR #2677, opened 2026-08-04
feat(docs-app): limit the version select to a component's own versions

HerrTopi:
## Summary
- Build a per-package map of the library versions each component appears in and emit it as
  `activeVersions` in `docs-versions.json`.
- Drive the docs component-version select from that map instead of the full library version list.
- Omit the select entirely for packages with no version data (ui-motion, ui-position, ui-dialog,
  ui-focusable, ui-i18n, ui-portal, ui-selectable, ui-svg-images, ui-truncate-list).

## Test Plan
- On a versioned component (e.g. Avatar, Checkbox), confirm the version dropdown lists only the
  versions that component exists in, and that switching still loads the right docs.
- On a Util Component from an unmapped package (Transition, Position, Portal, Focusable, InlineSVG),
  confirm no version dropdown renders and the theme select still does.
- Check a component that was introduced mid-range or removed — its earlier versions should survive a
  gap rather than being truncated.

Fixes INSTUI-5145

### DIFF
--- a/packages/__docs__/buildScripts/DataTypes.mts
+++ b/packages/__docs__/buildScripts/DataTypes.mts
@@ -32,12 +32,15 @@ import type {
   SharedTokens
 } from '@instructure/ui-themes'
 
-type ProcessedFile =
-  Documentation &
+type ProcessedFile = Documentation &
   YamlMetaInfo &
   JsDocResult &
-  PackagePathData &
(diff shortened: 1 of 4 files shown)


balzss (approved):
looks good, works as expected

joyenjoyer (approved):
looks good

PR #2679, opened 2026-08-05
chore: bump TypeScript to 7.0.2

joyenjoyer:
## Summary
- Bump `typescript` to `7.0.2` in the root `package.json`
- Fix two `@ts-expect-error` directives (in `packages/__docs__/globals.ts` and ui-table's `Table.test.tsx`) that TS7's stricter validation now flags as stale, since they no longer suppress a real error
- Restore `__docs__`'s JSDoc extraction (`getJSDoc.mts`), which also depends on the removed Compiler API, by importing it from `@typescript/typescript6` (Microsoft's official TS 6.0 compatibility shim), scoped to that package only, the rest of the repo's type-checking/build still runs on TypeScript 7

Fixes INSTUI-5147

### DIFF
--- a/packages/__docs__/package.json
+++ b/packages/__docs__/package.json
@@ -140,6 +140,7 @@
     "react-docgen": "8.0.3",
     "style-loader": "4.0.0",
     "thread-loader": "^4.0.4",
+    "typescript-compiler-api": "npm:@typescript/typescript6@^6.0.2",
     "webpack-bundle-analyzer": "5.3.0"
   },
   "//dependency-comments": {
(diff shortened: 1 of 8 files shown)


packages/__docs__/package.json
```diff
@@ -140,6 +140,7 @@
     "react-docgen": "8.0.3",
     "style-loader": "4.0.0",
     "thread-loader": "^4.0.4",
+    "typescript-compiler-api": "npm:@typescript/typescript6@^6.0.2",
```
joyenjoyer, line 143: TS7 has no compiler API (that is used in getJSDoc). They promise a new API in 7.1 and recommends adding in this compatibility packages until it is released.

HerrTopi (approved):

git-nandor (approved):

PR #2687, opened 2026-08-12
feat(ui-tag): new Tag v3 implementation

git-nandor:
## Summary
- Add Tag `v3` (library `v11_8`) with the new design; `v2` preserved and frozen to `/v11_7` imports.
- `renderIcon` leading icon that auto-sizes to the Tag `size`; `href`/`onClick` render a focusable link/button body (underlined text for links).
- `dismissible` now renders a separate close button controlled by the new `onDismiss` callback — only the close button dismisses.
- Focus ring moved from the whole tag to the individual focusable elements (link/button body and close button).
- New `src/exports/c.ts` and `package.json` `./v11_8` + `./latest` point to v3.

## Test Plan
- Tab through a link + dismissible Tag: the label/icon group and the close button receive focus rings independently; clicking only the X dismisses.
- Visual: leading icon sizing across `small`/`medium`/`large`; link underline; focus-ring radius/height.
- Note: umbrella `@instructure/ui` `v11_8` wiring and the `small-components` regression page update are deferred to the release `bump` (umbrella `/latest` still resolves Tag to v2 until then).

Fixes INSTUI-5130

### DIFF
--- a/packages/__docs__/versioned-components.ts
+++ b/packages/__docs__/versioned-components.ts
@@ -24,6 +24,7 @@
 
 import * as V11_6 from '@instructure/ui/v11_6'
 import * as V11_7 from '@instructure/ui/v11_7'
+import * as V11_8 from '@instructure/ui/v11_8'
 // eslint-disable-next-line no-restricted-imports
 import { Guidelines } from './src/Guidelines'
 // eslint-disable-next-line no-restricted-imports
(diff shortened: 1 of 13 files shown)


danegivens:
This looks PERFECT! the only thing that was missed is the color of the tag on the rebrand light theme...in that theme the tag should be the opacity navy that is used for our secondary buttons!

adamlobler:
> This looks PERFECT! the only thing that was missed is the color of the tag on the rebrand light theme...in that theme the tag should be the opacity navy that is used for our secondary buttons!

It’s because this branch is still using v1.5.0 of the tokens. We should update it to the latest version.

adamlobler (approved):

PR #2690, opened 2026-08-17
feat(ui-scripts,ui-themes): generate theme tokens as css custom properties

HerrTopi:
## Summary
- Add a `buildCSSVariables` step to `build-themes` that flattens each theme's shared tokens and emits one stylesheet per theme under `src/themes/newThemeTokens/themesAsCSSVariables/`.
- Emit `cssThemesWithMediaQueries.css`, mapping the light and dark themes to `prefers-color-scheme`.
- Register the malva CSS plugin in `dprint.json` so the generated stylesheets go through the existing dprint pass.
- Mark `*.css` as side-effectful in `ui-themes` so bundlers don't tree-shake stylesheet imports.

## Test Plan
- Run `pnpm run build:themes` and check the 5 files under `packages/ui-themes/src/themes/newThemeTokens/themesAsCSSVariables/` — the directory is gitignored, so none of the generated CSS appears in this diff.
- With `cssThemesWithMediaQueries.css` loaded, toggle the OS colour scheme and confirm the `:root` values and `color-scheme` switch.

## Open questions for the reviewer
- The 188 custom properties are unprefixed (`--background-*`, `--spacing-md`). The existing style-dictionary emitter uses `instui-<theme>`. Adding a prefix later is breaking, so this should be settled before release.

Fixes INSTUI-5153

### DIFF
--- a/packages/ui-scripts/lib/build/buildThemes/buildCSSVariables.ts
+++ b/packages/ui-scripts/lib/build/buildThemes/buildCSSVariables.ts
@@ -1,0 +1,81 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015 - present Instructure, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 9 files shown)


packages/ui-scripts/lib/build/buildThemes/buildCSSVariables.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  for (const theme of themeDirs) {
+    const resolved = await loadTheme(path.join(root, theme))
+
+    const flatResult = flattenObj(resolved)
+    const cssVariables = Object.keys(flatResult).reduce(
+      (res, key) => `${res}--${key}:${flatResult[key]};`,
+      ''
+    )
+    sharedTokensByThemes[theme] = cssVariables
```
matyasf, line 75: This makes a CSS class called `light`/`dark`/`legacyCanvas`/`legacyCanvasHighContrast`. Arent the `light` and `dark` keywords too generic that it could cause hard to track down bugs because something overrides it? I'd prefix these with something like `InstUITheme-{name}`

matyasf (approved):
While the CSS is exported, the users can only pull it in with an export statement from the `src` folder, which is not nice. 
When I import it this way:
`import "@instructure/ui-themes/src/themes/newThemeTokens/themesAsCSSVariables/cssThemesWithMediaQueries.css"`
I get an error `TS2882: Cannot find module or type declarations for side-effect import of
@instructure/ui-themes/src/themes/newThemeTokens/themesAsCSSVariables/cssThemesWithMediaQueries.css`

Can you please look into this? First I think we should add it to the `exports` in the package.json, then AI recommended me to add also a `d.ts` file.

I've released version `11.7.5-pr-snapshot-1787571141446` from this PR, so its easier to test

matyasf (changes requested):
see my comment

HouseOfHawks:
> Can you please look into this? First I think we should add it to the `exports` in the package.json, then AI recommended me to add also a `d.ts` file.

Agreed! If we can truncate to a simple `import '@instructure/ui-themes/cssThemesWithMediaQueries.css';` then the DX would be fantastic.

Outside of that, testing a Next.js app router with that import in `layout.tsx`, throttling to 3G and stopping the page load before hydration - dark theme is correctly active (I'm in dark mode OS preference). Note that none of this page is currently _using_ any of those tokens; just inspecting in dev-tools:

Screenshot 2026-08-24 at 11 49 26 AM


And light mode is correctly overridden:

Screenshot 2026-08-24 at 11 52 01 AM

Great stuff!

matyasf (approved):

HouseOfHawks (approved):

PR #2694, opened 2026-08-21
INSTUI-5152: fix SSR hydration with stable id generation

HouseOfHawks:
This PR:
- Fixes ID Generation in place in `ui-utils` package,
- Deprecates `deterministicIdContext` and `generateId` function in lieu of the new `useDeterministicId` functional hook or `withDeterministicId` decorator,
- Enables strict mode in the regression test app,
- Disables no SRR hook in the regression test app, and
- Updates documentation regarding SSR guidance and best practices for consumers.

Known Issue - 1 (not a regression and not fixed in this PR):
```
<div> aria-describedby="Selectable___Rqfeminebn3alb-description" — pre-existing since 2019 (de742b4623): Selectable.getRootProps/getTriggerProps always sets aria-describedby, but Drilldown deliberately does not render getDescriptionProps
```

Example SSR loading/hydration in Next.js app router (results: clean hydration by components with previous breakage)

https://github.com/user-attachments/assets/a18e1f5b-1234-4e01-929c-93fae42e73d7

### DIFF
--- a/docs/guides/server-side-rendering.md
+++ b/docs/guides/server-side-rendering.md
@@ -1,0 +1,143 @@
+---
+title: Server side rendering (SSR)
+category: Guides
+order: 8
+relevantForAI: true
+---
+
+# Server side rendering (SSR)
+
+**InstUI works under SSR with no SSR-specific setup.** Set up
+[InstUISettingsProvider](/#InstUISettingsProvider) as you would in any app and
+render — there is nothing extra to configure, and no per-component opt-in.
+
+The rest of this page is the small number of cases where you do need to act.
+
+## Why it works
+
+Many components need an id for an element you never name: the target of an
+`aria-describedby`, the `<title>` inside an SVG, the panel a tab controls.
+InstUI generates those with React's built-in
+[`useId`](https://react.dev/reference/react/useId), which produces the same
+value for the same position in the React tree on the server and on the client,
+so the server HTML and the hydrated tree agree.
+
+Generated ids look like `ComponentName___token`, e.g. `FormFieldLayout___r7`.
+The token is the `useId` value with React's delimiters (`:r0:` on React 18,
+`«r0»` on React 19) stripped, so the id is always valid in a CSS selector. If
+you pass your own `id` prop, yours is used and nothing is generated.
+
+## Remove `instanceCounterMap` if you have one
+
+Before ids came from `useId`, InstUI counted component instances in a shared map,
+and earlier versions of this guide asked you to pass an `instanceCounterMap` to
+`InstUISettingsProvider` to keep server and client ids aligned. **That map is no
+longer read.** If you still pass one it is ignored — delete it.
+
+| Deprecated                                                    | Replacement                                   |
+| ------------------------------------------------------------- | --------------------------------------------- |
+| `instanceCounterMap` prop on `DeterministicIdContextProvider` | none needed — remove the prop                 |
+| `DeterministicIdContext`                                      | none needed — ids no longer come from context |
+| `generateId` from `@instructure/ui-utils`                     | `useDeterministicId` / `withDeterministicId`  |
+
+These are still exported for backwards compatibility and will be removed in the
+next major version.
+
+## Multiple React roots on one page
+
+`useId` guarantees uniqueness **within a single React root**. If a page mounts
+two or more independent roots — a micro-frontend layout, a widget embedded in a
+legacy page, or a [module federation](/#module-federation) host and guest — each
+root numbers its ids from scratch, so the roots can collide.
+
+This is the one case that needs your action. Give each root a distinct
+`identifierPrefix`, and pass the matching prefix to the server renderer so both
+sides agree:
+
+```javascript
+---
+type: code
+---
+// server
+renderToPipeableStream(<GuestApp />, { identifierPrefix: 'guest-' })
+
+// client
+hydrateRoot(document.getElementById('guest'), <GuestApp />, {
+  identifierPrefix: 'guest-'
+})
+```
+
+The same option exists on `createRoot` for client-only roots.
+
+## Next.js App Router
+
+`InstUISettingsProvider` uses React context, so with the App Router it has to
+live in a client component. Mark the layout that renders it with `'use client'`:
+
+```javascript
+---
+type: code
+---
+// app/layout.tsx
+'use client'
+import { InstUISettingsProvider, canvas } from '@instructure/ui'
+
+export default function RootLayout({ children }) {
+  return (
+    <html lang="en">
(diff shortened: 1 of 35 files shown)


package.json
```diff
@@ -121,7 +121,7 @@
   },
   "lint-staged": {
     "*.{js,ts,tsx}": [
-      "oxlint -c .oxlintrc.json --fix",
+      "oxlint -c .oxlintrc.json --fix --no-error-on-unmatched-pattern",
```
HouseOfHawks, line 124: Note: This was throwing an error when a commit only changed the `/regression-test/cypress/e2e/spec.cy.ts` file - happy to pivot if necessary.

regression-test/cypress/e2e/spec.cy.ts
```diff
@@ -141,14 +177,32 @@ const PAGES: PageSpec[] = [
...
+      'Inverse Link renders white on the light #f2f4f5 surface (1.1:1); same ' +
+      'dark-theme inverse token bug as the Button page.'
+  },
   {
     slug: 'menu',
     title: 'Menu',
     wait: 300,
+    // The menu is open on load (`defaultShow`), so it is server rendered open
+    // but unfocused until hydration applies the initial highlight.
+    awaitFocused: '[role="menuitem"]',
```
HouseOfHawks, line 205: Problem: We were hitting a visual regression with this component regarding focus state.

Technical Explanation: For real users, a real but bounded consequence of SSR. In most cases, menus are tied to an anchor element which opens a menu on click.

Details: An SSR-rendered `defaultShow` menu is genuinely unfocused for one hydration window (~120ms locally, longer on slow devices or the CI pipeline). No library can avoid that: focus requires JS. It's the same class of tradeoff as any interactive state that can't be expressed in HTML. Worth knowing, not worth blocking on and note this only affects menus that start open on load; the ordinary click-to-open path is unaffected.

Solution Explanation: This focus gate is an assertion, not a sleep. It retries until the settled state is real.

HerrTopi (approved):

docs/guides/server-side-rendering.md
```diff
@@ -0,0 +1,143 @@
+---
+title: Server side rendering (SSR)
+category: Guides
+order: 8
+relevantForAI: true
+---
+
+# Server side rendering (SSR)
```
matyasf, line 8: I like it, to the point and easy to understand

packages/ui-react-utils/src/DeterministicIdContext/withDeterministicId.tsx
```diff
@@ -42,11 +41,19 @@ type WithDeterministicIdProps = {
...
- * for deterministic id generation.
+ * This decorator injects a `deterministicId` prop into the decorated class,
+ * used to generate SSR-safe, hydration-stable ids.
  *
- * The context is there for the users to pass an `instanceCounterMap` Map which is then used
- * in the child components to deterministically create ids for them based on the `instanceCounterMap`.
+ * It is backed by React's built-in {@link https://react.dev/reference/react/useId `useId`},
+ * which produces the same id on the server and the client for a given position
+ * in the React tree, so ids no longer rely on a global instance counter. The
+ * injected `deterministicId(instanceName?)` function may be called multiple
```
matyasf: I would rephrase this, we are no longer interested in the past, so I'd remove the " so ids no longer rely on a global instance counter." part

packages/ui-react-utils/src/DeterministicIdContext/useDeterministicId.tsx
```diff
@@ -22,43 +22,51 @@
...
- * A React hook that provides deterministic ID generation for functional components.
+ * A React hook that provides SSR-safe, hydration-stable ID generation for
+ * functional components.
  *
- * This hook is the functional component equivalent of the `withDeterministicId` decorator.
- * It uses the `DeterministicIdContext` which is needed for deterministic id generation.
+ * This hook is the functional component equivalent of the `withDeterministicId`
+ * decorator. It is backed by React's built-in {@link https://react.dev/reference/react/useId `useId`},
+ * which produces the same id on the server and the client for a given position
+ * in the React tree, so ids no longer rely on a global instance counter.
```
matyasf: I would rephrase this, we are no longer interested in the past, so I'd remove the " so ids no longer rely on a global instance counter." part

regression-test/src/app/menu/page.tsx
```diff
@@ -25,58 +25,55 @@
...
 export default function MenuPage() {
-  // Disabling SSR is needed here because Menu does not work when it starts in
-  // the open state
+  // This page used to be wrapped in <NoSSR> because of a `mountNode` prop that
+  // resolved a DOM node (`document.getElementById('main')`). That node does not
+  // exist while pre-rendering, so the open menu was rendered inline on the
+  // server but portalled on the client — a structural hydration mismatch
+  // (React #418). Without `mountNode` the menu portals to document.body on both
+  // sides and hydrates cleanly, so the page can be server rendered like the
+  // rest. Don't reintroduce a DOM-resolving `mountNode` here.
```
matyasf: No need to comment on a fixed bug

regression-test/next.config.mjs
```diff
@@ -46,10 +46,10 @@ const nextConfig = {
   // resolves /route without needing a .html fallback config.
   trailingSlash: true,
   images: { unoptimized: true },
-  // strict mode needs to be disabled, so deterministic ID generation
-  // works. If its enabled, client side double rendering causes IDs to
-  // come out of sync. TODO fix
-  reactStrictMode: false,
+  // Deterministic ids come from React's `useId`, which is stable across
+  // StrictMode's double render, so strict mode is safe to leave on here — and
+  // keeping it on means the app catches hydration and id regressions.
```
matyasf: again, I think this comment is not needed

regression-test/cypress/e2e/spec.cy.ts
```diff
@@ -122,7 +148,17 @@ const PAGES: PageSpec[] = [
...
+  {
+    slug: 'button',
+    title: 'Button and derivatives',
+    wait: 100,
+    contrastSkipThemes: ['dark'],
+    contrastSkipReason:
+      'The `primary-inverse` Button keeps its light surface (#f2f4f5, correct ' +
+      'for an inverse variant) but its label resolves to white instead of the ' +
+      'dark text its own wrapper uses — 1.1:1. Fix the dark theme inverse ' +
+      'button text token.'
```
matyasf, line 160: These are pretty important findings, I'll create tickets from these (as I see they either need to be fixed in the examples here or a design issue)

matyasf (changes requested):
nice work! I just found some small changes needed for some comments

matyasf (approved):
Nice work!

PR #2695, opened 2026-08-24
test(INSTUI-5155): SSR layout shift lab — INVESTIGATION ONLY, not for merge

git-nandor:
> **INVESTIGATION ONLY — please do not merge.** This is the deliverable for INSTUI-5155 (*Investigate — SSR layout shifting in components*). It adds a throwaway measurement app, no library code. Its purpose is to give the team numbers to scope the follow-up SSR tickets. Merge it only if we decide the lab is worth keeping.
>
> **Stacked on `INSTUI-5152`**, not on `master` — the baselines below were recorded on top of that fix.

## Summary

- Adds `ssr-lab/`, a standalone Next.js app that server-renders InstUI **v2** components for real and measures how much content moves during hydration.
- Two failure modes found in the code: class components that recompute styles from `componentDidMount` (both `withStyle` and `withStyleNew` seed the first render with **empty** extraArgs, so the server only sees the incomplete pass), and components that measure the DOM in an effect.

## How to run it

```sh
cd ssr-lab && npm install
npm run build && npm start     # http://localhost:3200
```

Use the production build, not `npm run dev`. Pick a component from the index page; the panel top-right reports **CLS** and the scenario height at three points (server HTML → after `document.fonts.ready` → after hydration). Both numbers are needed: the height split separates a web-font swap from what hydration did, and it catches jumps the browser scores as **zero** because they land before the first paint (TopNavBar is exactly that case). 

To see a jump rather than just read it, throttle CPU to 20× in DevTools → Performance and enable Rendering → *Layout Shift Regions*.

## Testing custom component combinations

Two ways, no build step needed for the first:

- **As a file** — edit `ssr-lab/src/scenarios/custom.tsx` and build it again !

## Measurements

Production build, headless Chrome, 1280×900, 4× CPU throttling. Height delta is measured from the post-font baseline, so it is hydration only. Zero console errors on all 24 pages. Absolute numbers depend on viewport, machine and the scenario's own text content — the sign and the order of magnitude are what to compare.

| Scenario | CLS | Height after hydration | What moves |
|---|---|---|---|
| **TextArea** | **0.103** | **+532px** | autogrow height only settles after mount; the only scenario past the 0.1 "good" threshold |
| **TruncateText** | 0.037 | **−289px** | server sends the full text, the client truncates after hydration |
| **TopNavBar** | 0 | **0px → 164px** | renders **nothing** server-side; `Responsive` emits an empty div until it has a DOM. CLS is 0 only because it lands before first paint |
| DrawerLayout | 0.044 | 0px | measures content width to place the tray; moves horizontally |
| Table | 0.020 | +64px | `Table.Row/v2` two-pass styles |
| suite-app-shell | 0.017 | +64px | TopNavBar + Tabs on a realistic page |
| Breadcrumb | 0.004 | −84px | `TruncateList` collapses crumbs after measuring |
| Link / suite-dashboard | 0.001 / 0.002 | −16px / −19px | re-wrapping |
| Button, Tabs, TreeBrowser, SideNavBar, ToggleGroup, FileDrop, ProgressCircle, Rating, Calendar, ColorPicker, Drilldown, Pill, Select, **TextInput**, **suite-form** | ≤0.0002 | **0px** | no measurable shift |

**On the original FormFieldLayout hypothesis:** the grid does collapse to a single `"controls"` row in **v1** (label and messages land in implicit tracks). In **v2** it does not — all 6 instances produce identical `grid-template-areas` before and after hydration, because `useStyleNew` gets its params during render. The v2 shift on that page was the NumberInput, not the grid.

## Test Plan

- From the repo root: `cd ssr-lab && npm install && npm run build && npm start`, open http://localhost:3200 and confirm the panel reports numbers close to the table above.
- Check `/s/text-area`, `/s/truncate-text` and `/s/top-nav-bar` with DevTools → Rendering → *Layout Shift Regions* and CPU 20× to see the three worst cases.

Relates to INSTUI-5155

### DIFF
--- a/ssr-lab/src/scenarios/link.tsx
+++ b/ssr-lab/src/scenarios/link.tsx
@@ -1,0 +1,80 @@
+'use client'
+
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015 - present Instructure, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+import {
+  Link as lk,
+  Text as tx,
+  View as vw,
+  IconUserLine as iul
+} from '@instructure/ui/latest'
+
+const Link = lk as any
+const Text = tx as any
+const View = vw as any
+const IconUserLine = iul as any
+
+export default function Scenario() {
+  return (
+    <View as="div" maxWidth="34rem">
+      <Text as="p">
+        Egy bekezdés, benne{' '}
+        <Link href="https://instructure.design/">egy linkkel</Link>, hogy
+        látszódjon, elmozdul-e a körülötte lévő szöveg.
+      </Text>
+
+      <Text as="p">
+        Ikonos link:{' '}
+        <Link href="https://instructure.design/" renderIcon={<IconUserLine />}>
+          profil
+        </Link>{' '}
+        — az ikon szélessége befolyásolja a sortörést.
+      </Text>
+
+      <View as="div" margin="medium 0">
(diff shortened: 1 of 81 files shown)


matyasf:
Thanks, this tool is very useful. I've also tried it with the large `forms` example at https://instructure.design/forms and no layout shift :)

ssr-lab/src/scenarios/link.tsx
```diff
@@ -0,0 +1,80 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+import {
+  Link as lk,
```
HouseOfHawks, line 28: Just curious about all the type casting with every component implementation. Is it necessary to declare `any` for every import?

matyasf: its not necessary, just the AI was copying the regression test code.

There it has a purpose: the regression test reads from the compiled output (in the `/es/` folder) too add another testing layer here that tests our compiled code. but when reading from here somehow the type information is wrong (might be a bug?), so I've added this ugly workaround that is being copied by Claude. But it should have no effect

HouseOfHawks:
Higher level, I'm curious about how we are considering implementing this?

CI pipeline as a required pass? Could add recurring blockage, start to require delicate threshold management, etc., since some components are always going to shift to some degree. The promise for SSR is to create Skeleton loaders that are an approximation to the final rendered state. In some cases it's just unavoidable since we're limited to pure CSS (no JS measurements).

Dev tool to run and evaluate? Promising and probably where we achieve the most value, IMO.

Outside of that I enjoyed running this tool locally. Thanks for the work, and thanks for sharing @matyasf!

PR #2697, opened 2026-08-25
fix(ui-dialog): cancel the pending focus region activation on close

matyasf:
## Summary
- `Dialog.close()` cancels any still-scheduled `requestAnimationFrame` region activation and clears
  `_focusRegion` after blurring. Without it, a Dialog closed before that frame ran (e.g. a Tray
  opened and closed within one frame, which happens when rAF callbacks land late under load)
  activated a region nothing ever blurred — its document `keydown` listener then ran `scopeTab` on
  an unrendered element, `preventDefault`ing every later Tab press.
- Adds a Dialog regression test that flushes the activation frame manually after the close and
  asserts Tab still moves focus outside the Dialog.

## Test Plan
- This was the cause of the flaky Tray `should handle focus properly in complex cases` browser
  test. To reproduce the old failure, stub `window.requestAnimationFrame` to fire ~150ms late and
  run `packages/ui-tray/src/Tray/__tests__/Tray.test.tsx` — it fails on the first `userEvent.tab()`
  without this fix and passes with it.
- Worth a manual keyboard pass on Modal/Popover/Tray/Menu: open, close, then Tab around to confirm
  focus return and tab order are unchanged.

### DIFF
--- a/packages/ui-dialog/src/Dialog/__tests__/Dialog.test.tsx
+++ b/packages/ui-dialog/src/Dialog/__tests__/Dialog.test.tsx
@@ -272,6 +272,53 @@ describe('<Dialog />', () => {
       })
     })
 
+    it('should not leak a focus region when it is closed before its activation frame runs', async () => {
+      // The Dialog activates its FocusRegion in a requestAnimationFrame
+      // callback. On a busy machine that callback can land after the Dialog
+      // has already been closed. Activating a region for a closed Dialog
+      // leaked it: nothing blurred it afterwards, so its document `keydown`
+      // listener kept scoping every later tab press to an element that is not
+      // rendered anymore.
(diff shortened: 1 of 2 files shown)


joyenjoyer (approved):

balzss (commented):
although the code looks good, the 2 visual regression test changes look interesting. the table one seems unrelated which is even weirder, but the menu one might has something to do with the changes. it looks like the dark theme now not highlights the first menu option? also the other themes are not consistent with highlighting/not-highlighting the first item. do you think it's related to this change?

matyasf:
> @balzss the 2 visual regression test changes look interesting

I've re-ran the VRT and the changes are gone..

matyasf:
VRT is flaky, now I run it again, and now it shows a change in `Badge`

balzss (approved):

PR #2698, opened 2026-08-26
test: filter noisy default-theme console warning in browser tests

joyenjoyer:
This PR bypasses the noise in test logs caused by adding: ```No theme provided for [InstUISettingsProvider], using default <canvas> theme.``` multiple times.

INSTUI-5163

### DIFF
--- a/vitest.setup.browser.ts
+++ b/vitest.setup.browser.ts
@@ -39,3 +39,17 @@ if (typeof globalThis.process === 'undefined') {
 afterEach(() => {
   vi.restoreAllMocks()
 })
+
+//Do not let test components spam the logs because they are not wrapped in InstUISettingsProvider.
+const originalConsoleWarn = console.warn
+console.warn = (...args: Parameters<typeof console.warn>) => {
+  if (
+    typeof args[0] === 'string' &&
+    (args[0] as string).includes(
+      'No theme provided for [InstUISettingsProvider], using default'
+    )
+  ) {
+    return
+  }
+  originalConsoleWarn(...args)
+}


vitest.setup.browser.ts
```diff
@@ -39,3 +39,24 @@ if (typeof globalThis.process === 'undefined') {
...
   vi.restoreAllMocks()
 })
+
+// Components are rendered directly in tests, without wrapping them in an
+// <InstUISettingsProvider>, so `useTheme()`/`getTheme()` always fall back to
+// the default `canvas` theme and log this warning on every render. It's
+// expected in this context and just drowns out real warnings in CI, so we
+// filter it out here. This is a plain reassignment (not `vi.spyOn`), so it
+// survives the `vi.restoreAllMocks()` above and stays in effect for every
+// test, including ones that install their own `console.warn` spy on top of it.
```
matyasf: this is way too verbose. I'd just write:

Do not let test components spam the logs because they are not wrapped in `InstUISettingsProvider`.

vitest.setup.browser.ts
```diff
@@ -39,3 +39,24 @@ if (typeof globalThis.process === 'undefined') {
...
+const NOISY_WARNINGS = [
+  'No theme provided for [InstUISettingsProvider], using default'
+]
+const originalConsoleWarn = console.warn
+console.warn = (...args: Parameters<typeof console.warn>) => {
+  if (
+    typeof args[0] === 'string' &&
+    NOISY_WARNINGS.some((message) => (args[0] as string).includes(message))
```
matyasf: I would simplify this, no need for the array. Just do 
```
'No theme provided for [InstUISettingsProvider], using default'.includes(message)
```
no need for the generic filter code

matyasf (changes requested):
works well, just some stylistic changes

HerrTopi (commented):
Should we print a warning at all? That's why we have a default, don't we? In this case, canvas (who is hyper sensitive to warnings) can't use this with the default option

joyenjoyer:
> Should we print a warning at all? That's why we have a default, don't we? In this case, canvas (who is hyper sensitive to warnings) can't use this with the default option

This is out of scope for this PR and possibly a breaking change. It focuses on bypassing this warning in our tests. Let's talk about it elsewhere.

vitest.setup.browser.ts
joyenjoyer: Comment is changed

vitest.setup.browser.ts
joyenjoyer: Array removed

HerrTopi (approved):

matyasf (approved):

PR #2701, opened 2026-08-28
fix(ui-modal): avoid redundant Modal.Body re-renders from its observers

balzss:
## Summary

`Modal.Body` becomes a tab stop when it can be scrolled but has no focusable
child inside it. It read both of those from the DOM during `render()`, and its
resize and mutation observers called `forceUpdate()` on every change they saw.

So any DOM change inside the body caused a re-render, even when the tab stop
stayed the same. In jsdom tests those updates land outside `act()`, which is the
warning consumers hit on 11.7.4.

Now the value lives in state:

- `syncTabIndex()` reads the DOM, compares the result with the last value, and
  calls `setState` only if it changed. The compare has to come before
  `setState`: a class component schedules the update (and warns) before the
  updater runs, so returning early inside the updater is too late.
- `render()` reads state instead of the DOM.
- `componentDidUpdate()` calls `syncTabIndex()` too, because the observers don't
  catch every change. If only the text changes, the body can start to overflow
  without changing its own size, and neither observer fires. Thanks @matyasf for
  catching this and writing the test for it.

Same change in v1 and v2. No prop, theme, or export changes.

Measured on the branch: 20 no-op mutations inside the body went from 20
re-renders to zero in Chromium, and 10 mutations went from 10 `act()` warnings
to zero in jsdom, in both versions.

## Test Plan

- Keyboard only: open a modal with a scrollable body that has nothing focusable
  in it. Tab to the body, check that it takes focus and that Up/Down scroll it.
- With the modal open, add a focusable child to the body. The body should stop
  being a tab stop, and focus should go to the child. Remove the child and the
  body should be focusable again.
- With the modal open, make the body long enough to scroll and then short again.
  The tab stop should follow.
- Grow the body text only, so its own size doesn't change. The body should still
  become a tab stop.
- Worth a screen reader pass (VO/NVDA/JAWS) on the scrollable body's
  `aria-label`, since this touches the same code path as INSTUI-5046.

Fixes INSTUI-5166

### DIFF
--- a/packages/ui-modal/src/Modal/v2/ModalBody/index.tsx
+++ b/packages/ui-modal/src/Modal/v2/ModalBody/index.tsx
@@ -32,7 +32,7 @@ import { withStyleNew } from '@instructure/emotion'
 import generateStyle from './styles.js'
 
 import { allowedProps } from './props.js'
-import type { ModalBodyProps } from './props'
+import type { ModalBodyProps, ModalBodyState } from './props'
 import { UIElement } from '@instructure/shared-types'
 import ModalContext from '../ModalContext.js'
 
@@ -43,7 +43,7 @@ id: Modal.Body
 ---
 **/
 @withStyleNew(generateStyle, 'ModalBody')
-class ModalBody extends Component<ModalBodyProps> {
+class ModalBody extends Component<ModalBodyProps, ModalBodyState> {
   static displayName = 'ModalBody'
   static readonly componentId = 'Modal.Body'
   static readonly themeId = 'ModalBody'
@@ -54,6 +54,11 @@ class ModalBody extends Component<ModalBodyProps> {
     as: 'div',
     variant: 'default'
   }
+
+  state: ModalBodyState = { isFirefox: false, needsTabIndex: false }
+  // mirrors state.needsTabIndex, but readable synchronously in the observer
+  // callbacks below, where a pending setState would make this.state stale
+  private lastNeedsTabIndex = false
 
   ref: UIElement | null = null
   private resizeObserver?: ResizeObserver
@@ -92,9 +97,10 @@ class ModalBody extends Component<ModalBodyProps> {
 
     const finalRef = this.getFinalRef(this.ref)
     if (finalRef && typeof ResizeObserver !== 'undefined') {
-      this.resizeObserver = new ResizeObserver(() => this.forceUpdate())
+      this.syncTabIndex()
+      this.resizeObserver = new ResizeObserver(this.syncTabIndex)
       this.resizeObserver.observe(finalRef)
-      this.mutationObserver = new MutationObserver(() => this.forceUpdate())
+      this.mutationObserver = new MutationObserver(this.syncTabIndex)
       this.mutationObserver.observe(finalRef, {
         childList: true,
         subtree: true,
@@ -122,6 +128,32 @@ class ModalBody extends Component<ModalBodyProps> {
     this.mutationObserver?.disconnect()
   }
 
+  // The body is a tab stop only while it can be scrolled but holds nothing
+  // focusable. Both inputs come from the DOM, so the observers recompute them
+  // on resize and on subtree changes — which is most changes inside the body,
+  // the vast majority of them leaving the result identical. Comparing against
+  // the last computed value before calling setState keeps those callbacks from
+  // scheduling an update at all, rather than scheduling one React later
+  // discards: `setState` warns about updates outside `act()` in tests as soon
+  // as it schedules, so bailing out inside the updater would be too late.
+  syncTabIndex = () => {
+    const finalRef = this.getFinalRef(this.ref)
+    const hasScrollbar =
+      !!finalRef &&
+      Math.abs(
+        (finalRef.scrollHeight ?? 0) -
+          (finalRef.getBoundingClientRect()?.height ?? 0)
+      ) > 1
+    const needsTabIndex = hasScrollbar && findTabbable(finalRef).length === 0
+
+    if (needsTabIndex === this.lastNeedsTabIndex) return
+    this.lastNeedsTabIndex = needsTabIndex
+    this.setState({ needsTabIndex })
+  }
+
+  // this recursive function is needed because `ref` can be a React component.
+  // TODO rethink, the 'as' prop, likely its not a good idea to allow React
+  // components. See INSTUI-4674
   getFinalRef(el: UIElement): Element | undefined {
     if (!el) {
       return undefined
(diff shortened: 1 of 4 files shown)


git-nandor (approved):

packages/ui-modal/src/Modal/v2/ModalBody/index.tsx
```diff
@@ -122,6 +128,32 @@ class ModalBody extends Component<ModalBodyProps> {
...
   }
 
+  // The body is a tab stop only while it can be scrolled but holds nothing
+  // focusable. Both inputs come from the DOM, so the observers recompute them
+  // on resize and on subtree changes — which is most changes inside the body,
+  // the vast majority of them leaving the result identical. Comparing against
+  // the last computed value before calling setState keeps those callbacks from
+  // scheduling an update at all, rather than scheduling one React later
+  // discards: `setState` warns about updates outside `act()` in tests as soon
+  // as it schedules, so bailing out inside the updater would be too late.
```
matyasf, line 138: please simplify to 

```
  // The body is a tab stop only while it can be scrolled but holds nothing
  // focusable.
```

matyasf (changes requested):
Note: Please rephrase Claudespeak, there are some parts that are really hard to understand, e.g.

```
...so any DOM change inside the body re-rendered it even when the derived tabIndex was identical
```
what is a "derived tabIndex"??

This way a change might not be noticed in some cases, add `this.syncTabIndex()` to `componentDidUpdate()`. A failing case:

```

```

a failing test (just use this with less comments :) ):

```
    it('becomes a tab stop when the content grows without a DOM structure change', async () => {
      mockScrollable(false)
      const { rerender } = await render(<ModalBody>{'short body'}</ModalBody>)
      const body = page.getByText('short body').element()

      expect(body).not.toHaveAttribute('tabindex')

      // The body is already at its max height, so growing the content changes
      // `scrollHeight` but not the body's own box: the ResizeObserver stays
      // silent. React updates a lone text child by writing `nodeValue`, which
      // is a `characterData` mutation the MutationObserver does not subscribe
      // to, so it stays silent too. A late webfont swap or an image finishing
      // load looks the same from here.
      scrollHeightSpy.mockReturnValue(500)
      rerender(<ModalBody>{'a much longer body that now overflows'}</ModalBody>)

      // Nothing recomputes `needsTabIndex` any more: `componentDidUpdate` only
      // calls `makeStyles`, so the body is scrollable with no focusable child
      // and no way for a keyboard user to scroll it.
      await vi.waitFor(() => expect(body).toHaveAttribute('tabindex', '0'))
    })
```

and an example to repro:

```
import React, { useEffect, useState } from 'react'
import { Button, Checkbox, Flex, Modal, Text, View } from '@instructure/ui/latest'

const SHORT_TEXT =
  'This body fits inside the fullscreen modal, so it does not scroll and it ' +
  'correctly has no tabindex.'

// One long string, so React updates it in place as a single text node.
const LONG_TEXT = Array.from(
  { length: 120 },
  (_, i) =>
    `Paragraph ${i + 1}. The body now overflows, so a keyboard user needs the ` +
    'body itself to be focusable in order to scroll it.'
).join(' ')

const OVERLAY_SCROLLBAR_CSS = `
  [data-cid="ModalBody"] {
    scrollbar-width: none;
  }
  [data-cid="ModalBody"]::-webkit-scrollbar {
    width: 0;
    height: 0;
  }
`

type BodyInfo = {
  scrollHeight: number
  clientHeight: number
  clientWidth: number
  tabIndex: string | null
  focusableChildren: number
}

export function ModalBodyTabIndexPage() {
  const [open, setOpen] = useState(true)
  const [grown, setGrown] = useState(false)
  const [overlayScrollbars, setOverlayScrollbars] = useState(true)
  const [info, setInfo] = useState<BodyInfo | null>(null)

  // Poll the live DOM instead of the component's state, so the readout shows
  // exactly what a Tab press or a screen reader would see. The poll re-renders
  // this page (and with it Modal.Body) five times a second, which is the point:
  // even that does not bring the tabindex back, because `componentDidUpdate`
  // no longer recomputes it.
  useEffect(() => {
    const read = () => {
      const body = document.querySelector<HTMLElement>('[data-cid="ModalBody"]')
      if (!body) {
        setInfo(null)
        return
      }
      setInfo({
        scrollHeight: body.scrollHeight,
        clientHeight: body.clientHeight,
        clientWidth: body.clientWidth,
        tabIndex: body.getAttribute('tabindex'),
        focusableChildren: body.querySelectorAll(
          'a[href], button, input, select, textarea, [tabindex]'
        ).length
      })
    }
    read()
    const id = window.setInterval(read, 200)
    return () => window.clearInterval(id)
  }, [])

  const scrollable = !!info && info.scrollHeight - info.clientHeight > 1
  const isBug = scrollable && info?.focusableChildren === 0 && !info?.tabIndex

  return (
    <View as="div" padding="medium">
      {overlayScrollbars ? <style>{OVERLAY_SCROLLBAR_CSS}</style> : null}

      <Button onClick={() => setOpen(true)}>Open the modal</Button>

      <Modal
        open={open}
        onDismiss={() => setOpen(false)}
        label="Modal.Body tab stop repro"
        size="fullscreen"
        shouldReturnFocus={false}
      >
        <Modal.Header>Modal.Body tab stop repro</Modal.Header>

        {/* Text only: no focusable children, and a lone text node that React
            updates through `nodeValue`. */}
        <Modal.Body>{grown ? LONG_TEXT : SHORT_TEXT}</Modal.Body>

        <Modal.Footer>
          <Flex direction="column" gap="small" alignItems="start">
            <Flex gap="small">
              <Button color="primary" onClick={() => setGrown(!grown)}>
                {grown ? 'Shrink the body text' : 'Grow the body text'}
              </Button>
              <Button onClick={() => setOpen(false)}>Close</Button>
            </Flex>

            {/* Shrink the text back before flipping this: turning the gutter on
                or off while the body already scrolls changes its width, which
                fires the ResizeObserver and recomputes the tab stop. */}
            <Checkbox
              label="Overlay scrollbars (macOS default) — uncheck for classic scrollbars"
              variant="toggle"
              size="small"
              checked={overlayScrollbars}
              onChange={() => setOverlayScrollbars(!overlayScrollbars)}
            />

            <Text size="small">
              {info
                ? `scrollHeight ${info.scrollHeight} · clientHeight ${info.clientHeight} · ` +
                `clientWidth ${info.clientWidth} · scrollable: ${scrollable} · ` +
                `focusable children: ${info.focusableChildren} · ` +
                `tabindex: ${info.tabIndex ?? 'absent'}`
                : 'body not mounted'}
            </Text>

            <Text
              size="small"
              color={isBug ? 'danger' : 'success'}
              weight="bold"
            >
              {isBug
                ? 'BUG: the body scrolls, holds nothing focusable, and has no tabindex. ' +
                'Press Tab — focus skips the body, so its content cannot be reached by keyboard.'
                : scrollable
                  ? 'The scrollable body is a tab stop. With classic scrollbars this is ' +
                  'luck: the scrollbar shrank clientWidth, which woke the ResizeObserver.'
                  : 'Body is not scrollable yet — press "Grow the body text".'}
            </Text>
          </Flex>
        </Modal.Footer>
      </Modal>
    </View>
  )
}
```

balzss:
Thanks for the catch, you were right. `componentDidUpdate` only called `makeStyles`, so a change that made the body overflow without resizing it never recomputed the tab stop. Fixed in v1 and v2, and I added your test: it fails without the fix and passes with it.

Also done:
- Cut the comment above `syncTabIndex` to the two lines you asked for.
- Rewrote the PR description in plainer language. "Derived tabIndex" is gone; it now says what the component does before what changed.

matyasf (approved):

PR #2703, opened 2026-08-31
perf(ui-scripts,ui-icons): generate tree-shakeable named imports for lucide icons

joyenjoyer:
## Summary
- Generate the lucide icon barrel with named imports from `lucide-react` instead of a namespace
  import (`import * as Lucide from 'lucide-react'`)
- Annotate each generated `wrapLucideIcon(...)` call with `/*#__PURE__*/` so minifiers can drop
  unused icons
- Update the generator's unit tests to match the new output shape

## Test plan
- Create an isolated mini project and add v2 Button to it with a renderIcon prop. The prop should pick up a Lucide icon. Check if the final bundle contains other icons.

Fixes INSTUI-5171

### DIFF
--- a/packages/ui-scripts/lib/icons/generate-lucide-index.ts
+++ b/packages/ui-scripts/lib/icons/generate-lucide-index.ts
@@ -89,14 +89,17 @@ export default function generateLucideIndex() {
   const iconExports = allLucideIcons
     .map(
       (iconName) =>
-        `export const ${iconName}InstUIIcon = wrapLucideIcon(Lucide.${iconName})`
+        `export const ${iconName}InstUIIcon = /*#__PURE__*/ wrapLucideIcon(${iconName})`
     )
     .join('\n')
 
+  // Named imports (rather than `import * as Lucide`) let bundlers tree-shake
+  // lucide-react itself, and the /*#__PURE__*/ annotation above lets minifiers
+  // drop unused wrapLucideIcon(...) calls.
   // Output goes to src/generated/lucide/ so the import is relative to that:
   // ../../lucide/wrapLucideIcon resolves to src/lucide/wrapLucideIcon
   const content = `${HEADER}
-import * as Lucide from 'lucide-react'
+import { ${allLucideIcons.join(', ')} } from 'lucide-react'
 import { wrapLucideIcon } from '../../lucide/wrapLucideIcon'
 
 ${iconExports}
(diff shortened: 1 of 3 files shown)


balzss (changes requested):
I tested this with a small app: a v2 Button with `renderIcon`. I built it twice, once with vite 7 and once with webpack 5.

It works. Only the icon I imported ends up in the bundle.

- vite: 886.1 KB → 395.6 KB minified (241.1 KB → 112.8 KB gzipped)
- webpack: 861.0 KB → 389.8 KB minified
- Lucide SVG path nodes in the bundle: 6,967 → 4

Both parts of the change are needed. Rollup only needs `/*#__PURE__*/`. Webpack needs the named imports too. I tried the namespace import together with `/*#__PURE__*/`, and webpack still kept every icon.

One thing to fix: `generate-custom-index.ts:117` has the same problem. All 87 custom icons live in one module, and none of them have `/*#__PURE__*/`. So if you import one custom icon, you get all 87. That adds 26 KB gzipped.

Adding `/*#__PURE__*/` on line 117 fixes it in both bundlers. You don't need named imports there, because the `Paths` components are already in the same module. `wrapCustomIcon` has no side effects, just like `wrapLucideIcon`, so the annotation is safe. The current tests still pass. It's a one-line change, so I'd add it to this PR.

balzss (approved):

matyasf (approved):
Code looks good, tested with Claude, and it had only some very small nitpicks, none worth another round:

" The core of the change checks out: both wrapper functions are side-effect free, Babel keeps the annotation in all generated output, tsc and generator tests pass,
  and a measured webpack 5 single-icon consumer drops from about 489 KB to 8 KB."

packages/ui-scripts/lib/icons/generate-lucide-index.ts
```diff
@@ -89,14 +89,17 @@ export default function generateLucideIndex() {
...
     .join('\n')
 
+  // Named imports (rather than `import * as Lucide`) let bundlers tree-shake
+  // lucide-react itself, and the /*#__PURE__*/ annotation above lets minifiers
+  // drop unused wrapLucideIcon(...) calls.
   // Output goes to src/generated/lucide/ so the import is relative to that:
   // ../../lucide/wrapLucideIcon resolves to src/lucide/wrapLucideIcon
   const content = `${HEADER}
-import * as Lucide from 'lucide-react'
+import { ${allLucideIcons.join(', ')} } from 'lucide-react'
```
miskr-instructure, line 102: Human-friendlier generated code:
```js
`${HEADER}
import {
  ${allLucideIcons.join(',\n  ')}
} from 'lucide-react'
`

```

PR #2704, opened 2026-08-31
INSTUI-5169: prebundle react-dom to keep React versions in sync

git-nandor:
## Summary

- `react-dom/server` and `react-dom/client` were already in `optimizeDeps.include`, but Vite resolves those entries from the repo root, where `package.json` had `react` without `react-dom`. Both failed to resolve on every run, so Vite skipped prebundling them and discovered them mid-run instead — which can put React and react-dom in different optimizer generations and fail SSR tests with `Cannot read properties of null (reading 'useId')`. Adding `react-dom` to the root devDependencies makes the existing entries resolve.
- Adds `resolve.dedupe: ['react', 'react-dom']`, which was absent.
- Warns when an `optimizeDeps.include` entry cannot be resolved. Vite's own warning does not say what it costs, which is why this went unnoticed. Deliberately non-fatal.

## Test Plan

- Delete `node_modules/.vite/vitest` and run the full suite. The two `Failed to resolve dependency: react-dom/...` lines should be gone.
- Then check `node_modules/.vite/vitest/*/deps/_metadata.json`: `react`, `react-dom`, `react-dom/server` and `react-dom/client` should all be under `optimized` with `discovered` empty — nothing picked up mid-run.
- To see the new warning, temporarily add a bogus specifier to `PREBUNDLED_MODULES`; the run should print the warning and still complete.

Fixes INSTUI-5169

### DIFF
--- a/vitest.config.mts
+++ b/vitest.config.mts
@@ -55,6 +55,34 @@ PREBUNDLED_PACKAGES.filter((pkgName) => {
   }
 })
 
+// Prebundled alongside PREBUNDLED_PACKAGES, but resolved from node_modules.
+// SSR/hydration tests are the only that import these. Without prebundling,
+// Vite may re-optimize mid-run, breaking dynamic imports and
+// causing React and react-dom/server to use different optimizer generations.
+const PREBUNDLED_MODULES = ['react-dom/server', 'react-dom/client']
+
+const OPTIMIZED_DEPS = [...PREBUNDLED_PACKAGES, ...PREBUNDLED_MODULES]
+
+// Warn when an optimizeDeps entry cannot be resolved. This can cause Vite to
+// discover it mid-run, leading to different optimizer generations and SSR
+// failures.
+//
+// Uses CJS resolution; ESM-only entries may need a different check.
+OPTIMIZED_DEPS.forEach((entry) => {
+  try {
+    require.resolve(entry, { paths: [__dirname] })
+  } catch {
+    console.warn(
+      `[vitest.config] "${entry}" is in optimizeDeps.include but cannot be ` +
+        `resolved from the repo root, so Vite will skip prebundling it. ` +
+        `Expect it to be discovered mid-run instead, which can put React and ` +
+        `react-dom in different optimizer generations and fail SSR tests with ` +
+        `"Cannot read properties of null (reading 'useId')". Add it to the ` +
+        `root package.json devDependencies.`
+    )
+  }
+})
+
 // Create a hash of every file's updated time. This hash changes if any file's
 // updated time changes
 function getPrebundleStamp() {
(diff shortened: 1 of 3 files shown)


vitest.config.mts
```diff
@@ -55,6 +55,34 @@ PREBUNDLED_PACKAGES.filter((pkgName) => {
...
+  try {
+    require.resolve(entry, { paths: [__dirname] })
+  } catch {
+    console.warn(
+      `[vitest.config] "${entry}" is in optimizeDeps.include but cannot be ` +
+        `resolved from the repo root, so Vite will skip prebundling it. ` +
+        `Expect it to be discovered mid-run instead, which can put React and ` +
+        `react-dom in different optimizer generations and fail SSR tests with ` +
+        `"Cannot read properties of null (reading 'useId')". Add it to the ` +
+        `root package.json devDependencies.`
```
matyasf: IMO this is way too verbose. Just have it say
```
`[vitest.config] "${entry}" is in optimizeDeps.include but cannot be resolved from the repo root, Add it to the root package.json devDependencies.
```

And make it throw an error, we should correct this if happens

vitest.config.mts
```diff
@@ -55,6 +55,34 @@ PREBUNDLED_PACKAGES.filter((pkgName) => {
...
+// Vite may re-optimize mid-run, breaking dynamic imports and
+// causing React and react-dom/server to use different optimizer generations.
+const PREBUNDLED_MODULES = ['react-dom/server', 'react-dom/client']
+
+const OPTIMIZED_DEPS = [...PREBUNDLED_PACKAGES, ...PREBUNDLED_MODULES]
+
+// Warn when an optimizeDeps entry cannot be resolved. This can cause Vite to
+// discover it mid-run, leading to different optimizer generations and SSR
+// failures.
+//
```
matyasf: non need for this comment

vitest.config.mts
```diff
@@ -176,20 +204,14 @@ export default defineConfig({
...
-            'react-dom/server',
-            'react-dom/client'
-          ]
+          include: OPTIMIZED_DEPS
         },
         resolve: {
+          // Keep a single copy of React across the prebundled chunks and the
+          // on-demand transformed sources. Two copies mean react-dom installs
+          // its hooks dispatcher on one of them while components read the
+          // other, which throws on the first hook call.
```
matyasf: simplify to 

"Keep a single copy of React. Prevents errors when a dependency pulls in a different version."

matyasf (changes requested):
See my comments. As I see the real fix is adding `react-dom` to `package.json` as a dependency, the others are just safeguards.

Please always rephrase Claude's comments, they are way too verbose 95% of the time

matyasf (approved):

HerrTopi (approved):

PR #2706, opened 2026-09-03
chore: document the comment and commit message rules

balzss:
## Summary

- Add a Code comments section to `CLAUDE.md`, with a human-facing version in the
  contributor docs: explain why not what, one line, never reference the change itself.
- Fix `commitlint.config.js`'s `parserOpts.headerPattern`, which couldn't match a scoped
  subject like `fix(ui-button): ...`, so the conventional parser never ran and
  `subject-max-length` applied to nothing. Cap the subject at 100.
- Tighten `/commit`'s guidance on commit bodies, and drop the robot attribution line
  from commit messages.
- Remove ToMESSKa from the `/pr` reviewer roster — no longer on the team.

## Test Plan

- `subject-max-length: 100` checked against history: 78 of the last 2000 subjects exceed
  it, and they're the ones enumerating every package touched instead of naming the change.
- Verified the remaining rules still fire — a 120-char subject, a 120-char body line (from
  the preset's `body-max-line-length`) — and that comma scopes like
  `fix(ui-drawer-layout,ui-a11y-utils):` still parse.

Fixes INSTUI-5174


.claude/commands/commit.md
```diff
@@ -13,16 +13,37 @@ type(scope): imperative subject
...
+- Add .npmrc with hoisted node linker
+Build Tooling:
+- Update scripts/bootstrap.js
+
+✅ the reason the diff cannot show
+regression-test stays on npm so it keeps installing @instructure/ui
+the way an external consumer would.
+```
+
+Writing about _before_ and _after_ is encouraged — "Previously the placeholder only showed on hover" is exactly right in a commit message, which is permanently anchored to its own diff. (Code comments are different: see CLAUDE.md.)
```
matyasf, line 46: no need for this part:
"(Code comments are different: see CLAUDE.md.)"

matyasf (changes requested):
What about using the rules here for both commit and PR messages? https://github.com/ayghri/i-have-adhd/blob/main/skills/i-have-adhd/SKILL.md#rules (do not copy it literally, some rules are irrelevant here)

CLAUDE.md
```diff
@@ -17,6 +17,31 @@ External docs (preferred over guessing component APIs): https://instructure.desi
...
+
+**A comment must make sense to someone reading the file in a year who never saw the change that introduced it** — no access to the PR, the ticket, or the conversation. The same standard applies to commit messages; `/commit` has the specifics.
+
+- Explain **why**, never **what**. If the code already says it, delete the comment.
+- **One line.** Two or three only when the reason genuinely needs them.
+- **Never reference the change itself.** `now`, `new`, `previously`, `used to`, `this change`, `the fix`, `as discussed`, `per review`, `we decided`, `recently` — these only mean something next to the diff. Name the constraint instead.
+- **Never narrate the diff** (`// added onKeyDown handler`, `// updated to support X`) — that's what `git log` is for.
+- No commented-out code, no banner or separator comments, and **don't add comments to code you didn't change**.
+- Lowercase `//` on its own line above what it explains, never trailing. Ticket ids only on a real external blocker: `// TODO INSTUI-1234: <what unblocks it>`.
+- Leave the MIT license header alone — `notice/notice` in `eslint.config.mjs` enforces it.
```
git-nandor, line 30: eslint.config.mjs does not exist, the repo uses .oxlintrc.json

CLAUDE.md
```diff
@@ -17,6 +17,31 @@ External docs (preferred over guessing component APIs): https://instructure.desi
...
+// where the component re-rendered too often. Previously computed inline.
+const styles = useMemo(...)
+
+// ✅ names the constraint, reads standalone
+// getCSSStyleDeclaration costs ~100ms per call
+const styles = useMemo(...)
+```
+
+Prop docs are a JSDoc block with **one prose sentence** and no `@param`/`@type` — types come from TypeScript and `react-docgen`. See `packages/ui-alerts/src/Alert/props.ts`.
+
```
git-nandor, line 44: Alert/props.ts — broken reference, missing v1/v2

git-nandor (changes requested):

CLAUDE.md
balzss: Correct. `deed11cc24` replaced ESLint with oxlint. The rule name is the same, only the file changed. Fixed to `.oxlintrc.json`.

CLAUDE.md
balzss: The real paths are `Alert/v1/props.ts` and `Alert/v2/props.ts`, but neither fits. Several v2 props have multi-paragraph JSDoc. I removed the example and kept the rule.

balzss:
@matyasf That skill is about chat responses. Four of its 10 rules are already here, five do not apply to commits, and "number multi-step tasks" conflicts — that is the changelog body this PR removes.

PR #2707, opened 2026-09-03
fix(ui-select,ui-text-input): show messages for non-editable selects

git-nandor:
## Problem
A Select sets its TextInput to `readonly` to hide the keyboard cursor,
but the form field layer hides messages for readonly fields. Non-editable
Selects and SimpleSelects therefore rendered the error border without the message text.

## Fix
- `TextInput` v2 gets an optional `forceMessages` prop that keeps messages visible in the readonly state; `Select` v2 sets it where it forces readonly itself.

## Test Plan
- Docs app: a `SimpleSelect` with error message and no `onInputChange` should show the message text under the input.

````
<div>
  <TextInput
    renderLabel="TextInput readonly with forceMessages"
    interaction="readonly"
    defaultValue="Alaska"
    messages={[{ type: 'error', text: 'TextInput error message' }]}
    forceMessages
  />
  <SimpleSelect
    renderLabel="SimpleSelect with an error message"
    messages={[{ type: 'error', text: 'SimpleSelect error message' }]}
  >
    <SimpleSelect.Option id="opt-1" value="1">value 1</SimpleSelect.Option>
    <SimpleSelect.Option id="opt-2" value="2">value 2</SimpleSelect.Option>
  </SimpleSelect>

  <Select
    renderLabel="Select with an error message"
    inputValue="Alaska"
    messages={[{ type: 'error', text: 'Select error message' }]}
  >
    <Select.Option id="opt-1">Alaska</Select.Option>
    <Select.Option id="opt-2">Arizona</Select.Option>
  </Select>
</div>
````
- Visual regression: the `select` page has a new `MessagesExample` (Select + SimpleSelect with an error message) 

Fixes INSTUI-5172

### DIFF
--- a/packages/ui-select/src/Select/v2/index.tsx
+++ b/packages/ui-select/src/Select/v2/index.tsx
@@ -719,6 +719,7 @@ class Select extends Component<SelectProps> {
     const passthroughProps = omitProps(rest, Select.allowedProps)
     const { ref, ...triggerProps } = getTriggerProps({ ...passthroughProps })
     const isEditable = typeof onInputChange !== 'undefined'
+    const isForcedReadOnly = interaction === 'enabled' && !isEditable
 
     // props to ensure screen readers treat uneditable selects as accessible
     // popup buttons rather than comboboxes.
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Review by balzss — APPROVED** [review]

File: packages/ui-select/src/Select/v2/index.tsx [hunk 1]
```diff
@@ -719,6 +719,7 @@ class Select extends Component<SelectProps> {
     const passthroughProps = omitProps(rest, Select.allowedProps)
     const { ref, ...triggerProps } = getTriggerProps({ ...passthroughProps })
     const isEditable = typeof onInputChange !== 'undefined'
+    const isForcedReadOnly = interaction === 'enabled' && !isEditable
```
matyasf, line 722: I think this name is confusing, what is "forced read-only"? Maybe name it "isNonEditableSelect"?

matyasf (changes requested):
looks good, just a small naming change

matyasf (approved):

PR #2708, opened 2026-09-04
generate frozen themes for a component version

HerrTopi:
INSTUI-5188

Adds a generator that freezes theme tokens for one component version, so an older
version keeps rendering the same way while the shared `semantics` and
`sharedTokens` trees keep changing.

Run `pnpm run build:frozen-themes`. It prompts for the themes, package, component,
and version, then writes a `frozenThemes` folder into that component version.
Each file is narrowed to what the component actually reads — for Alert v2 that's
82 lines of semantics instead of 1600+.

What's in here.

- `ui-scripts`: the `build-frozen-themes` command, the prompt, and the extraction
  itself — a recording proxy for the component's own token reads, plus a
  TypeScript AST pass for the `sharedTokens` reads in `styles.ts`.
- `emotion`: `useStyleNew` and `withStyleNew` map a frozen theme's `component`
  onto `newTheme.components[componentId]`, so a frozen component resolves its own
  tokens.
- `ui-avatar`, `ui-badge`: wire Avatar v2 and Badge v2 to their frozen themes.
- Docs: `docs/contributing/frozen-themes.md`.

Reads that can't be resolved at build time, such as `sharedTokens[variant]`, throw
with the file, line, and column rather than silently dropping tokens.

Note: seTokensFrom is not handled currently.


packages/emotion/src/useStyleNew.ts
```diff
@@ -96,7 +96,12 @@ const useStyleNew = <
...
   // object passed to it. We use that instead of the current theme so backward compatibility stays intact
   const theme = frozenTheme
-    ? { newTheme: frozenTheme[themeKey] }
+    ? {
+        newTheme: {
+          ...frozenTheme[themeKey],
+          components: { [componentId]: frozenTheme[themeKey].component }
+        }
+      }
     : themeInContext
```
matyasf, line 105: I would throw an error if `useTokensFrom` is used with `frozenTheme`, so we dont accidentally use it when its not implemented

packages/emotion/src/withStyleNew.tsx
```diff
@@ -162,7 +162,12 @@ const withStyleNew = decorator(
       // if this component is an older component version but still uses the new theming system, it'll have a frozenTheme
       // object passed to it. We use that instead of the current theme so backward compatibility stays intact
       const theme = frozenTheme
-        ? { newTheme: frozenTheme[themeKey] }
+        ? {
+            newTheme: {
+              ...frozenTheme[themeKey],
+              components: { [componentId]: frozenTheme[themeKey].component }
+            }
+          }
```
matyasf, line 170: I would throw an error if `useTokensFrom` is used with `frozenTheme`, so we dont accidentally use it when its not implemented

packages/ui-scripts/lib/commands/build-frozen-themes.ts
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+import promptThemeInput from './promptThemeInput.ts'
```
matyasf, line 24: I think this should be merged to this file, we're not extracting code in other commands either

packages/ui-scripts/lib/build/buildThemes/extractThemeComponentData.ts
```diff
@@ -0,0 +1,566 @@
...
+    primitives: usedPrimitives,
+    semantics: usedSemantics,
+    sharedTokens: usedSharedTokens,
+    component: componentFn
+  }
+}
+
+const buildFrozenThemes = async (data: any) => {
+  for (let i = 0; i < data.themes.length; i++) {
+    const frozen = await resolveUsedThemeElements(
```
matyasf, line 549: As I see this code only extracts the top component from a multi-component package, isnt this an issue? e.g. I extracted `Menu`, and it failed to add `semantic.color.background.interactive.action.tertiary.hover` from `MenuItem`

packages/ui-scripts/lib/build/buildThemes/extractThemeComponentData.ts
```diff
@@ -0,0 +1,566 @@
...
+    join(
+      repoRoot,
+      'packages',
+      componentPackageName,
+      'src',
+      componentName,
+      version,
+      'styles.ts'
+    ),
+    2
```
matyasf, line 507: this is just called with "2", lets not make it a parameter

packages/ui-scripts/lib/build/buildThemes/extractThemeComponentData.ts
```diff
@@ -0,0 +1,566 @@
...
+  }
+
+  let param: ParameterDeclaration | undefined
+  const findParam = (node: Node) => {
+    if (
+      !param &&
+      (ts.isArrowFunction(node) || ts.isFunctionDeclaration(node)) &&
+      node.parameters.length > paramIndex
+    ) {
+      param = node.parameters[paramIndex]
```
matyasf, line 200: As I see this finds the first function with >2 parameters. it should find the default export, now it craches e.g. in `FormFieldLayout`

packages/ui-scripts/lib/build/buildThemes/extractThemeComponentData.ts
```diff
@@ -0,0 +1,566 @@
...
+ * access as "no tokens used" would strip tokens the component needs at runtime.
+ */
+class UnresolvableTokenUsageError extends Error {
+  constructor(message: string) {
+    super(message)
+    this.name = 'UnresolvableTokenUsageError'
+  }
+}
+
+const extractPaths = (filePath: string, paramIndex: number) => {
```
matyasf, line 174: This fails to extract symbols from `View`'s helper functions like `getBorderStyle`, `getSpacingStyle`` `getOffsetStyle`

matyasf (changes requested):
- for theme lookup `theme.key` is used in `withStyleNew` and `useStyleNew`, but this writes `canvasHighContrast`, when it needs` canvas-high-contrast`

See my other comments

PR #2709, opened 2026-09-05
INSTUI-4850: update dependencies and remove unused ones

matyasf:
Bump lots of smaller packages.

Update `vitest` to `5.0.0` (this speeds up our suite by ~20%!) and adjust for the breaking changes:
- locators match now text exactly, so `getByText` calls pass `{ exact: false }` to keep 4.x functionality
- Deleted `ui-codemods`'s `test:vitest` because it stopped working (config needs to be stored in the same place)

Remove `@testing-library/react`, `@testing-library/user-event`, `@testing-library/jest-dom`, the functionality of these are provided by `@vitest/browser` already. Jest-DOM's `toHaveTextContent` is `toMatchTextContent` in `vitest-browser`, rename this in tests

Remove root devDependencies nothing references: `chai`, `@types/jest`, `jsdom`,
`@emotion/cache`, `@vitejs/plugin-react`, `esbuild-loader`, `cross-spawn` and `read-package-up`.

`eslint-plugin-react` requires `eslint` at load time, so pnpm auto-installed its deprecated `eslint@9` peer. Declare
`eslint@10` at the root and allow it as the plugin's peer in `pnpm-workspace.yaml`.

Drop the `auto-install-peers` line from `.npmrc`, which pnpm 11 ignored.

Update the docs about vitest.


packages/ui-a11y-content/src/PresentationContent/__tests__/PresentationContent.test.tsx
```diff
@@ -49,7 +49,7 @@ describe('<PresentationContent />', () => {
     )
     const content = container.querySelector('[aria-hidden="true"]')
 
-    expect(content).toHaveTextContent('Hello World')
+    expect(content).toMatchTextContent('Hello World')
```
matyasf, line 52: this is because we are using now `@vitest/browser` instead of `jest-dom` for matchers

packages/ui-drawer-layout/src/DrawerLayout/__tests__/DrawerLayout.test.tsx
```diff
@@ -130,15 +130,17 @@ describe('<DrawerLayout />', () => {
       </View>
     )
     const { rerender } = await render(<TestComponent />)
-    expect(page.getByText(msg).element()).toBeInTheDocument()
+    expect(page.getByText(msg, { exact: false }).element()).toBeInTheDocument()
```
matyasf, line 133: vitest 4 -> 5 change https://vitest.dev/guide/migration/#locators-are-strict-by-default

packages/ui-table/src/Table/v1/__tests__/Table.test.tsx
```diff
@@ -65,7 +65,7 @@ describe('<Table /> (v1)', () => {
     it('uses the caption verbatim when no column is sorted', async () => {
       const { container } = await renderTable()
 
-      expect(container.querySelector('caption')).toHaveTextContent(/^Movies$/)
+      expect(container.querySelector('caption')?.textContent).toBe('Movies')
```
matyasf, line 68: `toHaveTextContent` no longer supports Regexp, so we assert here `textContent` equality. https://vitest.dev/guide/migration/#tohavetextcontent-now-performs-strict-equality

balzss (approved):
looks good, great work!

git-nandor (approved):

PR #2710, opened 2026-09-07
chore(ui-scripts,regression-test): ignore whole-pixel shifts in visual regression diffs

balzss:
## Summary
- `visual-diff` retries the comparison at small integer offsets before calling a screenshot changed, behind `--max-shift` (default 1, `0` restores exact matching). Scoring over the shared region also covers a screenshot one pixel taller with identical content.
- `viewportHeight` 800 → 2000, cutting the `fullPage` stitching that produces those shifts: 58 stitch operations → 37, and 17/32 → 28/32 pages captured in a single pass (measured against the `visual-baselines` branch).

A pixel-count tolerance can't do this job. On a real baseline a 1px translation produces **2825** differing pixels and a genuine 40×40 recolour produces **1479** — less than a factor of two apart, so any threshold loose enough to absorb the first would hide the second.

Verified against four real CI baselines — identical, shifted 1px, 1px taller, and a genuine recolour:

| | `--max-shift 0` (current) | default |
|---|---|---|
| identical | ok | ok |
| shifted 1px | **changed** (2825 px) | ok |
| 1px taller | **changed** (1280 px) | ok |
| real recolour | changed (1479 px) | **changed** |

## Test Plan
- **Expect a cascade of `changed` rows on this PR** — the viewport change alters layout, so it can't demonstrate its own success. Merging refreshes the baselines.
- Spot-check a few `changed` rows in the report to confirm they're the viewport reflow and not something unexpected.

Fixes INSTUI-5180

### DIFF
--- a/packages/ui-scripts/lib/commands/visual-diff.ts
+++ b/packages/ui-scripts/lib/commands/visual-diff.ts
@@ -48,6 +48,7 @@ type Args = {
   baselineDir: string
   outputDir: string
   threshold: number
+  maxShift: number
   failOnMissingBaseline: boolean
   prNumber?: string
   prUrl?: string
@@ -223,6 +224,64 @@ function diffMask(baseline: PNG, actual: PNG, threshold: number) {
     sizeMismatch: bw !== aw || bh !== ah,
     actual: a
   }
+}
+
+// Copy a w*h window out of `src` starting at (sx, sy). Callers clamp the window
+// to the source bounds first.
+function crop(src: PNG, sx: number, sy: number, w: number, h: number): PNG {
+  const out = new PNG({ width: w, height: h })
+  PNG.bitblt(src, out, sx, sy, w, h, 0, 0)
+  return out
+}
+
+// Is `actual` pixel-identical to `baseline` once shifted by up to `maxShift`?
+//
+// When a layout box rounds one device pixel differently, the whole painted
+// subtree moves — glyphs, borders, all of it. The image is unchanged, just
+// translated, but a direct comparison lights up every edge in it: a one-pixel
+// shift routinely produces several thousand differing pixels. A component that
+// moved a pixel and is otherwise identical is not a visual regression, so this
+// is the check that says so.
+//
+// A pixel-count tolerance cannot do this job. The shift above and a genuine
+// 40x40 recolor land in the same order of magnitude, so any threshold loose
+// enough to absorb the first also hides the second.
+//
+// Each offset is scored over the region the two images share, so the band that
+// shifts in from outside is never counted. (0, 0) is included deliberately: the
+// caller's comparison *pads* mismatched sizes, while this one *crops* to the
+// overlap, which is what lets "one pixel taller, same content" pass.
+/** @internal — exported only for tests; not part of the package's public API. */
+export function matchesWhenShifted(
+  baseline: PNG,
+  actual: PNG,
+  threshold: number,
+  maxShift: number
+): boolean {
+  for (let dy = -maxShift; dy <= maxShift; dy++) {
+    for (let dx = -maxShift; dx <= maxShift; dx++) {
+      // A positive dx means the content sits that many pixels further right
+      // than in the baseline, so actual (x, y) lines up with baseline
+      // (x - dx, y - dy).
+      const x0 = Math.max(0, dx)
+      const y0 = Math.max(0, dy)
+      const w = Math.min(actual.width, baseline.width + dx) - x0
+      const h = Math.min(actual.height, baseline.height + dy) - y0
+      if (w <= 0 || h <= 0) continue
+
+      // No output buffer — only the count matters here.
+      const numDiff = pixelmatch(
+        crop(actual, x0, y0, w, h).data,
+        crop(baseline, x0 - dx, y0 - dy, w, h).data,
+        undefined,
+        w,
+        h,
+        { threshold, includeAA: false }
+      )
+      if (numDiff === 0) return true
+    }
+  }
+  return false
 }
 
 // How much unchanged pixels are dimmed in the diff image so the changed pixels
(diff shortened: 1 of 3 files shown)


.github/workflows/visual-regression-selfcheck.yml
```diff
@@ -0,0 +1,144 @@
...
+      # both Cypress runs.
+      - name: Serve the app
+        run: nohup npm start > /tmp/http-server.log 2>&1 &
+        working-directory: regression-test
+
+      - name: Wait for the app
+        run: npx wait-on http://localhost:3000 --timeout 120000
+        working-directory: regression-test
+
+      # `trashAssetsBeforeRuns` wipes cypress/screenshots at the start of every
```
matyasf: trashAssetsBeforeRuns -- what is this referencing?

regression-test/cypress.config.ts
```diff
@@ -40,7 +40,14 @@ export default defineConfig({
...
   e2e: {
     viewportWidth: 1280,
-    viewportHeight: 800,
+    // Deliberately taller than a real browser window. `capture: 'fullPage'`
+    // scrolls the viewport down the document and stitches the slices together,
+    // and the stitch is the least reproducible part of the capture. Measured
+    // against the baselines on the `visual-baselines` branch, raising this from
+    // 800 takes the suite from 58 stitch operations to 37, and from 17 of 32
+    // pages captured in one pass to 28. Only `min-height: 100vh` on the body
+    // depends on this value.
```
matyasf, line 49: unnecessary comment IMO. Especially parts that talk about the past value. We should not keep history in comments.

Also why is this needed? Whats the issue with lots of stitch operations? If they are bad why not set this to 9999999?

packages/ui-scripts/lib/commands/visual-diff.ts
```diff
@@ -1580,6 +1653,12 @@ export default {
       describe: 'pixelmatch color threshold (0-1)',
       default: 0.1
     },
+    'max-shift': {
+      type: 'number',
+      describe:
+        'Treat a screenshot as unchanged when it matches its baseline exactly after being shifted by up to this many pixels. Absorbs whole-pixel layout rounding, which moves a component without altering it. 0 requires an exact match.',
```
matyasf, line 1659: "Absorbs whole-pixel layout rounding, which moves a component without altering it." is not needed

packages/ui-scripts/lib/commands/visual-diff.ts
```diff
@@ -1471,8 +1531,21 @@ function run(args: Args): number {
       actual: padded
     } = diffMask(baseline, actual, threshold)
 
-    const status: Status =
+    // Straight comparison first, so identical screenshots cost nothing extra;
+    // the realignment check below only runs on one that already failed it. The
+    // size guard keeps a real layout change from being shifted away — only a
+    // delta within the shift budget is a rounding artifact.
```
matyasf, line 1537: A better explanation:

Check if the change is just a layout shift

packages/ui-scripts/lib/commands/visual-diff.ts
```diff
@@ -225,6 +226,64 @@ function diffMask(baseline: PNG, actual: PNG, threshold: number) {
...
 
+// Copy a w*h window out of `src` starting at (sx, sy). Callers clamp the window
+// to the source bounds first.
+function crop(src: PNG, sx: number, sy: number, w: number, h: number): PNG {
+  const out = new PNG({ width: w, height: h })
+  PNG.bitblt(src, out, sx, sy, w, h, 0, 0)
+  return out
+}
+
+// Is `actual` pixel-identical to `baseline` once shifted by up to `maxShift`?
```
matyasf, line 237: As I see from the code, it takes into account shifting into any direction, not just up?

packages/ui-scripts/lib/commands/visual-diff.ts
```diff
@@ -225,6 +226,64 @@ function diffMask(baseline: PNG, actual: PNG, threshold: number) {
...
+// is the check that says so.
+//
+// A pixel-count tolerance cannot do this job. The shift above and a genuine
+// 40x40 recolor land in the same order of magnitude, so any threshold loose
+// enough to absorb the first also hides the second.
+//
+// Each offset is scored over the region the two images share, so the band that
+// shifts in from outside is never counted. (0, 0) is included deliberately: the
+// caller's comparison *pads* mismatched sizes, while this one *crops* to the
+// overlap, which is what lets "one pixel taller, same content" pass.
```
matyasf, line 253: way way too verbose. Please simplify this to 1-2 lines

packages/ui-scripts/lib/commands/visual-diff.ts
matyasf: it matches shift to ANY direction within the tolerance (including e.g. diagonal shifts)

packages/ui-scripts/lib/commands/visual-diff.ts
```diff
@@ -1471,8 +1531,21 @@ function run(args: Args): number {
...
+    // size guard keeps a real layout change from being shifted away — only a
+    // delta within the shift budget is a rounding artifact.
+    let status: Status =
       numDiff === 0 && !sizeMismatch ? 'unchanged' : 'changed'
+    if (
+      status === 'changed' &&
+      maxShift > 0 &&
+      Math.abs(baseline.width - actual.width) <= maxShift &&
+      Math.abs(baseline.height - actual.height) <= maxShift &&
+      matchesWhenShifted(baseline, actual, threshold, maxShift)
```
matyasf, line 1545: maybe add a `layout shifted: true` to the summary in this case?

matyasf (changes requested):
see my comments. Code looks OK, but I'd change the comments

PR #2712, opened 2026-09-08
feat(ui-tabs): add roving tabindex and manual activation mode

aaronshaf:
## Summary
- Give the Tabs tablist a proper roving tabindex: exactly one tab has `tabindex="0"` at a time, tracked per tab node instead of via a `Focusable` wrapper.
- Add `activationMode="manual"`, so arrow keys move focus without changing selection until Enter or Space is pressed (for screen readers that reserve arrow keys for their own cursor).
- Add Home/End support to jump to the first/last enabled tab, and set `aria-orientation` and `aria-selected="false"` on unselected tabs to match the WAI-ARIA tabs pattern.

## Test plan
- [x] `vitest run --project browser --project node packages/ui-tabs` — 61 passed
- [x] New `TabsKeyboard.test.tsx` covers arrow/Home/End navigation, manual activation, and roving tabindex placement for both v1 and v2 themes

INSTUI-5181

### DIFF
--- a/packages/ui-tabs/package.json
+++ b/packages/ui-tabs/package.json
@@ -29,7 +29,6 @@
     "@instructure/emotion": "workspace:*",
     "@instructure/shared-types": "workspace:*",
     "@instructure/ui-dom-utils": "workspace:*",
-    "@instructure/ui-focusable": "workspace:*",
     "@instructure/ui-i18n": "workspace:*",
     "@instructure/ui-motion": "workspace:*",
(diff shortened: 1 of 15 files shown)


git-nandor:
Thanks for the contribution! We’ve moved the fix to a new PR where we’re continuing the work. I’ll close this one in favor of the new PR so the fix can be tracked there.

[New PR](https://github.com/instructure/instructure-ui/pull/2723)

PR #2715, opened 2026-09-10
chore: add /release-notify skill to credit reporters after a release

git-nandor:
## Summary
- New `/release-notify` command: works out who reported the issues fixed in a release and drafts a Slack message crediting them by name. **Nothing is posted automatically** — the block is copy-pasted by a human.
- One `gather.mjs` script collects the release's Jira tickets and the `#instui` threads and narrows them down; the command itself decides who the reporter is.
- Reporters are found from two directions, because either can be missing: the Slack link in a Jira description, and channel threads quoting the issue key.
- Adds step 11 to the release process and documents the two read-only credentials (Slack bot token, Jira API token).

## Test Plan
- Run `/release-notify` after a release and check the draft credits the right people. The key thing to eyeball: the reporter must be the **thread's root author**, not the developer who pasted the Jira link into the thread.
- Needs `SLACK_BOT_TOKEN` and the three `JIRA_*` keys in the root `.env`, and the bot must be a member of `#instui` — see `/slack-setup`.
- Already verified against `v11.7.4..v11.7.5`: of a hand-written reporter list, 4 of 6 matched with no false positives. The other two are unreachable in principle — their report thread never quoted the issue key and their ticket never linked the thread, so nothing connects them.

Fixes INSTUI-5133

### DIFF
--- a/.claude/commands/release-notify.md
+++ b/.claude/commands/release-notify.md
@@ -1,0 +1,96 @@
+---
+description: Draft a Slack message that credits the people whose reported issues got fixed in the latest release — gathers the release's Jira tickets and the channel's threads, works out who reported what, and hands you a copy-pasteable block
+---
+
+Work out who reported the issues fixed in the **latest release**, and produce a **copy-pasteable Slack message** that credits them by name.
+
+Usage: `/release-notify` — no arguments. The range is always the two most recent `v*` tags.
+
+## Behavior contract (do not deviate)
+
+- **Never post anything to Slack.** You produce a draft; the user posts it. The token is read-only by design — don't look for a post tool.
+- **The reporter is the author of the thread's root message.** Whoever pasted the Jira link into the thread is the _developer_. Tagging them instead is this skill's worst failure mode: it credits our own team and misses the actual reporter.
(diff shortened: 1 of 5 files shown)


balzss (changes requested):
Works, but I think it solves the wrong problem.

We already know the reporter at triage time. `/slack-triage` resolves the user IDs in Step 1, and puts the thread link in the ticket in Step 5. Then we drop both, and this PR spends a six-month channel scan and 2.5 minutes rebuilding them by guesswork. Your own number: six of 22 tickets had a usable link.

Better: make `/slack-triage` store the reporter and the thread link as structured data (Jira Reporter field or a `reported-by-<slack-id>` label), and make it required. Then `/release-notify` just reads that field and writes the draft.

That means `gather.mjs` goes away completely, and the Jira credentials in `.env.example` with it — the `atlassian` MCP server already gives us Jira. Tickets triaged before this change won't have a reporter, and I'd accept that rather than keep a scanner for them.

git-nandor:
> Works, but I think it solves the wrong problem.
> 
> We already know the reporter at triage time...

I reworked it so the triage is now responsible for storing the Slack thread and reporter information in the ticket description. This allows gather.mjs to trace the merged ticket numbers back to the tickets included in a release and easily extract the reporter and thread details from them.

PR #2716, opened 2026-09-11
fix(ui-radio-input): set aria-required on RadioInputGroup when isRequired

balzss:
## Summary
- `RadioInputGroup` v2 sets `aria-required` on the `role="radiogroup"` element when `isRequired` is true, so screen readers announce the group as required.
- Updated the `isRequired` prop doc and the README "Required Fields" section.
- v2 only. Setting the native `required` attribute is not part of this change, it is tracked in INSTUI-5187.

## Test Plan
- With a screen reader, move focus into a `RadioInputGroup` that has `isRequired` and confirm the group is announced as required.
- Confirm form submission is still not blocked when no option is selected.

Fixes INSTUI-5186

### DIFF
--- a/packages/ui-radio-input/src/RadioInputGroup/v2/README.md
+++ b/packages/ui-radio-input/src/RadioInputGroup/v2/README.md
@@ -224,6 +224,8 @@ type: example
 
 Whenever a `RadioInputGroup` contains required fields, you must include a note explaining what the asterisk means — typically "Fields marked with an asterisk (\*) are required."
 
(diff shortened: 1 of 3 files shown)


matyasf (approved):
Works well in Vo and NVDA, but I did not get the 'required' announcement in JAWS + Chrome. I think its OK?

balzss:
> Works well in Vo and NVDA, but I did not get the 'required' announcement in JAWS + Chrome. I think its OK?

considering that the a11y audit didn't even flag the previous version, i think it's ok and still a good upgrade

git-nandor (approved):

PR #2720, opened 2026-09-16
feat(INSTUI-5156): new skeleton loader

HouseOfHawks:
This PR:
- Creates a new `useIsHydrated` context provider in `ui-react-utils`
- Creates a new `<SkeletonLoader>` component and `<SkeletonShape>` sub-component
- Updates the demo app with detailed instructions on implementations and use-case types


packages/__docs__/src/App/index.tsx
```diff
@@ -1178,8 +1178,11 @@ class App extends Component<AppProps, AppState> {
             hasShadow={false}
             renderCloseButtonLabel="Close"
             themeOverride={{
-              background: '#BF32A4',
-              color: 'white',
+              // v1.10.0 replaced Alert's single `background` token with
```
matyasf, line 1181: What is v1.10.0? And what has this to do with skeleton loaders?

HouseOfHawks: `instructure-design-tokens` update with the [Skeleton tokens](https://github.com/instructure/instructure-design-tokens/pull/18)

HouseOfHawks: That should be `v1.9.0`, though: https://github.com/instructure/instructure-design-tokens/pull/17/changes

HouseOfHawks: Either way, this is more of a placeholder until the tokens work is shipped. Will be removed.

packages/ui-pill/src/Pill/v2/styles.ts
```diff
@@ -43,23 +43,28 @@ const generateStyle = (
...
     warning: {
       color: componentTheme.warningTextColor,
-      borderColor: componentTheme.warningBorderColor
+      borderColor: componentTheme.warningBorderColor,
+      background: componentTheme.warningBackgroundColor
     },
     error: {
       color: componentTheme.errorTextColor,
-      borderColor: componentTheme.errorBorderColor
+      borderColor: componentTheme.errorBorderColor,
```
matyasf, line 66: I guess these (and the other theme changes) will be deleted once we can upgrade the design tokens version?

packages/ui-react-utils/src/HydrationContext/HydrationProvider.tsx
```diff
@@ -0,0 +1,77 @@
...
+  children?: ReactNode
+}
+
+/**
+ * Reads the hydration state **once per page** and hands it to every descendant
+ * through context.
+ *
+ * Doing it here rather than in each component matters for three reasons:
+ *  - class components can consume context but cannot call hooks, so this is what
+ *    lets `View` participate without being rewritten
```
matyasf, line 51: just FYI we have lots of class-based React components as of now, just just View

packages/ui-react-utils/src/useIsHydrated.ts
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * This is the only correct way to branch on hydration. A `typeof document !==
+ * 'undefined'` check does not work: `document` is already defined while
+ * `hydrateRoot` is running, so the server would render one tree and the first
+ * client render another. React treats that as a hydration mismatch and throws
+ * away the server HTML for that subtree.
+ *
+ * `useSyncExternalStore` avoids the mismatch by construction — React is
+ * required to use `getServerSnapshot` for the hydrating render and
+ * `getSnapshot` only afterwards.
```
matyasf, line 49: I think most of this comment is not needed. Also "This is the only correct way to branch on hydration" -- what about class components? shoudnt they use `HydrationGate`?

packages/ui-react-utils/src/useIsHydrated.ts
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * Prefer `useIsHydratedContext` from `HydrationContext` inside InstUI, so the
+ * whole page flips in a single commit instead of once per component.
+ *
```
matyasf, line 53: Either make this  a private API or omit comments talking about how InstUI should work internally, this is just confusing for users

packages/ui-skeleton/src/SkeletonLoader/README.md
```diff
@@ -0,0 +1,314 @@
...
+
+### Controlled
+
+Pass `isLoading` when the wait is a data fetch and you know when it ends.
+
+```js
+---
+type: example
+---
+class Example extends React.Component {
```
matyasf, line 20: Please always make functional React examples (unless it demonstrates some class based functionality)

packages/ui-skeleton/src/SkeletonLoader/README.md
```diff
@@ -0,0 +1,314 @@
+---
+describes: SkeletonLoader
+---
+
+`SkeletonLoader` wraps a loading region and owns everything assistive technology
+needs to know about it: the busy state, the polite announcement, and the error
+interruption. The shapes themselves are decorative and announce nothing.
+
+Render **one of these per region**, not one per card or row. A page that
+announces once per skeleton row is unusable with a screen reader.
```
matyasf, line 10: This is not needed; the docs app renders the comment above the component too, so this becomes a repetition 

Image

packages/ui-skeleton/src/SkeletonLoader/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+import type { SkeletonLoaderProps } from './props'
+import { SkeletonShape } from '../SkeletonShape'
+import { Text, Circle, Rectangle } from './shapes.js'
+
+/**
+---
+category: components
+---
+Wraps a loading region and owns everything a screen reader needs to know about
+it. The skeleton shapes themselves are decorative; this is what announces.
```
matyasf, line 43: Can you please rephrase this to less Claude speak? "owns everything a screen reader needs to know about" What does this mean? What does a screen reader need to know about? Does it simply mean that the loading is announced by screen readers, e.g. it has a live region?

> the skeleton shapes themselves are decorative; this is what announces.

So are they purely decorative or they serve some other purpose too?

packages/ui-skeleton/src/SkeletonLoader/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+    isLoading,
+    loadedLabel,
+    loadingLabel,
+    skeleton,
+    ...rest
+  } = props
+
+  const isHydrated = useIsHydratedContext()
+
+  // Uncontrolled: loading until hydration commits. Controlled: the caller wins.
```
matyasf, line 65: Does this mean:

"Uncontrolled: loading is `true` until the component is hydrated. Controlled: controlled by the `isLoading` prop"

? What does the caller win?

packages/ui-skeleton/src/SkeletonLoader/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const styles = generateStyle()
+
+  useEffect(() => {
+    if (isError) {
+      setStatus('')
+      return
+    }
+
+    if (loading) {
+      // Sub-second loads never announce — the content beats the timer.
```
matyasf, line 89: unnecessary, also untrue comment. what is announced depends on the delay value

packages/ui-skeleton/src/SkeletonLoader/props.ts
```diff
@@ -0,0 +1,116 @@
+/*
```
matyasf, line 1: 2 licenses here

packages/ui-skeleton/src/SkeletonLoader/README.md
```diff
@@ -0,0 +1,314 @@
...
+### Motion
+
+A highlight sweeps from the leading edge toward the trailing one, and keeps
+looping for as long as the skeleton is shown. The gradient is sized to twice the
+shape's width, so the highlight stays the same size relative to whatever it sits
+on — the same on a 32px avatar as on a 900px bar.
+
+The loop is open-ended on purpose. A shimmer that stops while the request is
+still in flight says the opposite of what it means: the page looks dead exactly
+when the user most needs reassurance that it isn't.
```
matyasf, line 280: I dont think this is needed

packages/ui-skeleton/src/SkeletonLoader/README.md
```diff
@@ -0,0 +1,314 @@
...
+
+**Motion respects `prefers-reduced-motion`,** and with an endless loop that is
+load-bearing rather than a nicety — it is the mechanism by which a user stops
```
matyasf, line 283: load-bearing 😢  please less Claudeisms

packages/ui-skeleton/src/SkeletonLoader/README.md
```diff
@@ -0,0 +1,314 @@
...
+<div>
+  <SkeletonLoader.Shape shape="text" size="xl" />
+  <br />
+  <SkeletonLoader.Shape shape="text" size="md" lines={3} />
+</div>
+```
+
+### Leading
+
+Pick the ramp the replaced text actually uses. InstUI headings are `1.25` and
```
matyasf, line 222: "Pick the ramp the replaced text actually uses" what does this mean??

packages/ui-skeleton/src/SkeletonLoader/README.md
```diff
@@ -0,0 +1,314 @@
...
+  }
+}
+
+render(<Example />)
+```
+
+#### When the request fails
+
+Catch the failure, **clear `isLoading`**, and render your own error content as
+children. The order matters: a skeleton still shimmering next to an error
```
matyasf, line 55: why does the order matter? Doesnt this happen in 1 frame?

packages/ui-skeleton/src/SkeletonShape/props.ts
```diff
@@ -0,0 +1,133 @@
...
+   * Fixed height for `shape="rectangle"`. Prefer `aspectRatio` for media, so the
+   * box scales with its column. Text height comes from `size`, not from this.
+   */
+  height?: string | number
+
+  /**
+   * Which line-height ramp the replaced text uses. InstUI headings are 1.25 and
+   * body text is 1.5; picking the wrong one shifts the layout on swap.
+   */
+  leading?: SkeletonLeading
```
matyasf, line 68: Please call this `lineHeight`, its much more common wording in CSS

packages/ui-skeleton/src/SkeletonShape/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+        shape,
+        size,
+        textDirection,
+        width
+      }
+    })
+
+    return (
+      <span
+        {...passthroughProps(rest)}
```
matyasf, line 94: please use omitProps here. `allowedProps` exists, so we the user can only use props that are allowed. e.g. `    const props = omitProps(this.props, Checkbox.allowedProps)`

packages/ui-skeleton/src/SkeletonLoader/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+        role="status"
+        aria-live="polite"
+        aria-atomic="true"
+        css={styles.srOnly}
+      >
+        {status}
+      </span>
+
+      <div
+        {...passthroughProps(rest)}
```
matyasf, line 113: please use omitProps here. allowedProps exists, so we the user can only use props that are allowed. e.g.     const props = omitProps(this.props, Checkbox.allowedProps)

packages/ui-skeleton/src/SkeletonShape/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+import generateStyle from './styles.js'
+import type { SkeletonShapeProps } from './props'
+
+/**
+---
+parent: SkeletonLoader
+id: SkeletonLoader.Shape
+---
+**/
+const SkeletonShape = forwardRef<HTMLSpanElement, SkeletonShapeProps>(
```
matyasf, line 41: This component has now a very hard to interpret API because most props are valid only in one combination, e.g. `diameter` has only meaning if its a circle, `aspectRatio` only if its a rectangle,... 
I would split this into 3 components, this makes the code much easier to read and removed the magic combinations, and makes `shapes.tsx` unnecessary

packages/ui-skeleton/src/SkeletonLoader/shapes.tsx
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * The three primitives from the design spec, as named shortcuts over
+ * `SkeletonLoader.Shape`. Composing pages from these rather than hand-rolling
+ * markup per feature is what keeps sizing and motion consistent.
+ *
+ * These deliberately carry no docs metadata of their own — they are thin
+ * aliases, documented alongside the shape they bind rather than as separate
+ * pages in the nav.
+ */
+const bindShape = (shape: SkeletonShapeProps['shape'], displayName: string) => {
```
matyasf, line 41: This would not be needed if the shapes would be different components (this would also simplify other things.)

packages/ui-view/src/View/v2/props.ts
```diff
@@ -226,6 +227,41 @@ type ViewOwnProps = {
...
+   * Which skeleton primitive to draw when `isLoading`. Defaults to `rectangle`,
+   * which fills the View's box.
+   */
+  skeletonShape?: SkeletonShapeType
+  /**
+   * Number of text rows to draw when `skeletonShape="text"`.
+   */
+  skeletonLines?: number
+  /**
+   * Type ramp step for `skeletonShape="text"`.
```
matyasf, line 256: What is a "Type ramp step"?

packages/ui-view/src/View/v2/README.md
```diff
@@ -888,3 +888,42 @@ type: example
...
+```
+
+Passing `isLoading="untilHydrated"` puts the skeleton in the server HTML and
+keeps it through the hydrating render, handing over to the children once
+hydration commits. It is opt-in rather than the default for `isLoading`, because
+defaulting to it would turn every existing View on a page into a skeleton.
+
+View only draws the shapes. The loading state still has to be announced, by a
+single [SkeletonLoader](#SkeletonLoader) region wrapping the whole area — one
+per region, never one per View.
```
matyasf, line 929: "the whole area — one per region, never one per View."

this is a bit hard to read here, I would remove it. It should be explained in SkeletonLoader

packages/ui-view/src/View/v2/index.tsx
```diff
@@ -211,21 +213,50 @@ class View extends Component<ViewProps> {
...
-        ref={this.handleElementRef}
-      >
-        {children}
-      </ElementType>
-    )
+    const renderContent = (loading: boolean) => {
+      return (
+        <ElementType
+          {...passthroughProps(props)}
+          aria-busy={loading || undefined}
```
matyasf, line 230: aria-busy should be set before passtroughProps, so users can override it if needed

packages/ui-react-utils/src/HydrationContext/HydrationProvider.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const isHydrated = useIsHydrated()
+
+  useEffect(() => {
+    // Written straight to the DOM rather than through React. This element is
+    // outside the React tree, so the write never participates in reconciliation
+    // and cannot cause a hydration mismatch.
+    const root = document.documentElement
+    root.setAttribute(HYDRATED_ATTRIBUTE, '')
+    return () => root.removeAttribute(HYDRATED_ATTRIBUTE)
+  }, [])
```
matyasf, line 66: HYDRATED_ATTRIBUTE is removed if this is unmounted. There can be multiple InstUISettingsProvider in a DOM tree, wont this be an issue if one is removed?

matyasf (changes requested):
See my comments. 
I think the code is mostly OK (except for the skeleton shapes and HYDRATED_ATTRIBUTE removed, see my comments), and please rephrase the comments/documentation made by Claude, its so convoluted and hard to read.

packages/ui-skeleton/src/SkeletonLoader/index.tsx
HouseOfHawks: The caller (parent) wins in priority. I'll rephrase.

PR #2721, opened 2026-09-16
feat(ui-spinner,ui-icons): add ai and ai-inverse spinner variants

balzss:
## Summary
- Add `ai` and `ai-inverse` variants to Spinner v2. They render the Ignite AI star instead of the circle.
- Export `wrapCustomIcon` from ui-icons so a component can build a themed icon without extending the shared icon set.
- Add a regression-test page for Spinner, covering all four variants at all four sizes.

## Test Plan
- Check the `ai` gradient in the canvas, light, dark, and canvas-high-contrast themes. The regression suite only captures the first three, so high contrast needs a manual look.
- Check `ai-inverse` on a dark background.
- With a screen reader, confirm `renderTitle` is announced once for the ai variants. They label the wrapper with `role="img"`, which differs from the circle's `<title>` element.
- The three new Spinner screenshots will show as "New" on the first run and become baselines on merge.

Fixes INSTUI-5177


packages/ui-spinner/src/Spinner/v2/index.tsx
```diff
@@ -78,6 +80,48 @@ const Spinner = forwardRef<HTMLDivElement, SpinnerProps>((props, ref) => {
...
+    </svg>
+  )
+
+  // The icon accepts only a string `title`, but `renderTitle` can be any node.
+  // So the label goes on the wrapping element instead.
+  const renderAIIcon = () => (
+    <>
+      <ScreenReaderContent id={titleId}>
+        {callRenderProp(renderTitle)}
+      </ScreenReaderContent>
```
matyasf, line 118: <title> would be nicer here, but for some reason if I change it to that, VO wont read it

packages/ui-spinner/src/Spinner/v2/README.md
```diff
@@ -61,3 +61,36 @@ type: example
...
+<div style={{ display: 'flex', alignItems: 'center', gap: '1rem' }}>
+  <Spinner renderTitle="Loading" variant="ai" size="x-small" />
+  <Spinner renderTitle="Loading" variant="ai" size="small" />
+  <Spinner renderTitle="Loading" variant="ai" />
+  <Spinner renderTitle="Loading" variant="ai" size="large" />
+</div>
+```
+
+Use `ai-inverse` on dark backgrounds. It draws the same icon in a single color
+instead of the gradient.
```
matyasf: Bit overexplained IMO. (also factually false: Use `ai-inverse` on dark backgrounds.")

I would just keep "Use the ai variants to show that an AI feature is working. Use `ai-inverse` on inverted color backgrounds." And merge the 2 examples into one

regression-test/src/app/spinner/page.tsx
```diff
@@ -0,0 +1,61 @@
+'use client'
```
matyasf, line 24: license is missing, this causes CI to fail

matyasf (changes requested):
the inverse variant looks bad in the `dark` theme

Image

HerrTopi (commented):
Is the color of the icons correct? Did design visually validate it?

packages/ui-spinner/src/Spinner/v2/README.md
```diff
@@ -61,3 +61,36 @@ type: example
...
+
+```js
+---
+type: example
+---
+<View background="primary-inverse" as="div">
```
HerrTopi: This should work properly on every theme. Let's find a way to have dark bg when the theme is dark

packages/ui-spinner/src/Spinner/v2/README.md
balzss: simplified

regression-test/src/app/spinner/page.tsx
balzss: added

packages/ui-spinner/src/Spinner/v2/README.md
balzss: after consulting with design, this was changed to `background="info"`

matyasf (approved):
Looks good! (But the inverse variant with dark theme is still broken, I guess that could be a separate ticket)

HerrTopi (approved):
Nice work, ready for merge

PR #2723, opened 2026-09-17
feat(ui-tabs,ui-view): add roving tabindex and manual activation mode

git-nandor:
## Summary

- Give the Tabs tablist a proper roving tabindex: exactly one tab has `tabindex="0"` at a time, tracked per tab node instead of via a `Focusable` wrapper.
- Add `activationMode="manual"`, so arrow keys move focus without changing selection until Enter or Space is pressed (for screen readers that reserve arrow keys for their own cursor).
- Add Home/End support to jump to the first/last enabled tab, and set `aria-orientation` and `aria-selected="false"` on unselected tabs to match the WAI-ARIA tabs pattern.
- Fix RTL-mirrored left/right arrows
- New `shouldUseFocusVisible` on `View` (passes `:focus-visible` through to the already-existing `customCSSSelector`), so the roving tabindex doesn't make a mouse click paint a focus ring. No-op for every existing `View` consumer.

**DOM output changes:** unselected tabs render `tabindex="-1"` and `aria-selected="false"` instead of omitting both attributes. Selectors asserting their absence need updating.

## Test Plan

- **VoiceOver + `activationMode="manual"`** — the acceptance criterion for the ticket and not coverable by CI: arrow through the strip, confirm Enter and Space select.
- **Focus ring** — clicking a tab must show no ring. Tabbing in or arrowing must show one. 
- **`tabOverflow="scroll"` in manual mode** — arrow to an off-screen tab and confirm it scrolls out from under the fade overlay.
- Check new manual activation section in the docs app.

Fixes INSTUI-5181


git-nandor:
I reverted the focus-ring behavior changes at the request of the Design and Accessibility teams. This will be fixed in the new v3 version.

